PR #24, opened 2016-05-13
Add asciinema.org as a dark site

nikolay:


demostanis-worlds:
Seems like it's not anymore

PR #50, opened 2016-12-11
Add selector for FB video progress thumbnail

jordanpurinton:
I've added a small fix for the Facebook video progress thumbnail being inverted as outlined in [this issue ](https://github.com/alexanderby/darkreader/issues/49).


alexanderby:
Hi. Thanks for your help. Unfortunately it will not work. There is already a facebook.com record (the list is sorted by URL), so please move your fix there (as a new line or inside existing string).

jordanpurinton:
@alexanderby Done.

alexanderby:
Great.

PR #63, opened 2017-04-23
Fix yandex.music concerts

BANOnotIT:


src/config/sites_fixes_v2.json
```diff
@@ -247,6 +247,14 @@
             "rules": [
                 "#watch-appbar-playlist * { -webkit-filter: none !important; }"
             ]
+        },
+        {
+            "url": [
```
alexanderby: This array can be replaced with "music.yandex.*". Also please keep an alphabetical order by domain name.

BANOnotIT: It cannot, 'case there're only 3 domains available for this service.

alexanderby: Dark Reader support wildcards like it is made for [Google](https://github.com/BANOnotIT/darkreader/blob/2d5d38bf172ea405c4437f92a96f10325cde5db5/src/config/sites_fixes_v2.json#L52). There's also "music.yandex.kz" and there may be more domains.

Please, place "music.yandex" config after "mail.live.com" so it will be easier to maintain the growing site list.

BANOnotIT: I know, but yandex have only 3 supported domains for yandex.music service.

alexanderby: But you can ensure there is at least one more zone https://music.yandex.kz/feed

PR #64, opened 2017-04-23
Add dark scroll bar

BANOnotIT:


alexanderby:
Great idea. It seems that only `<html>` element scrollbar is not inverted, but others scrollbars are OK.
Maybe it is possible to just invert "html::-webkit-scrollbar" (somehow pass it to inverted selectors, or copy a filter value for separate rule) and preserve default scrollbar appearence (["auto"](http://stackoverflow.com/a/9060446/4137472) value for scrollbar properties should do the job).

risualiser:
Maybe, it can be fixed in Dynamic theme? Now they are all white, even in sub-elements.

alexanderby:
I will try to move it here https://github.com/alexanderby/darkreader/blob/master/src/inject/dynamic-theme/modify-css.ts#L47
Not inverted colors should be picked to make them modifiable by filter settings.

alexanderby:
So, dark scrollbars are implemented in Dynamic mode.

PR #66, opened 2017-04-23
Add react website nav

BANOnotIT:


src/config/sites_fixes_v2.json
```diff
@@ -44,6 +44,10 @@
             ]
         },
         {
+            "url": "facebook.github.io/react/*",
+            "selectors":".hero, .nav-main, .nav-home"
```
alexanderby: Seems I should write a Wiki for editing this config and provide some panel for testing selectors to prevent unexpected results.
You should be attentive. There is a `{common}` keyword, which inserts Common Selectors (img, iframe etc.). If it will not be specified, the images and other content specified in `commonSelectors` will become inverted.

BANOnotIT: I know what `{common}` does, but it isn't work on react's site. So I tried to fix it like that. But idea about panel for testing is very cool! It will be useful!

alexanderby:
Please, update the fix for new config version https://github.com/alexanderby/darkreader#how-to-contribute

BANOnotIT:
[dub](https://github.com/alexanderby/darkreader/pull/226)

PR #67, opened 2017-04-23
Add nuclide.io

BANOnotIT:


src/config/sites_fixes_v2.json
```diff
@@ -104,6 +104,10 @@
             "rules": ".ComposeContent { background-color: white; }"
         },
         {
+            "url": "",
```
alexanderby: Maybe URL is missing.

BANOnotIT:
I don't think someone use nuclide.io...

PR #85, opened 2017-08-14
Make it run in firefox

nemanjan00:
There are still problems, but, now, it at least starts...


src/background/extension.ts
```diff
@@ -410,11 +410,15 @@ style && style.parentElement.removeChild(style);
          * installed in system.
          */
         protected getFontList(onReturned: (fonts: string[]) => void) {
-            chrome.fontSettings.getFontList((res) => {
-                // id or name?
-                var fonts = res.map((r) => r.fontId);
-                onReturned(fonts);
-            });
+            if(chrome.fontSettings != undefined){
```
alexanderby: Replace if/else with if/return, so that we can just simply remove this when fonts will be supported and leave a Todo comment.

src/manifest.json
```diff
@@ -1,4 +1,10 @@
 ﻿{
+    "applications": {
```
alexanderby: I would place such a strange FF-specific thing to the end of file.

src/manifest.json
```diff
@@ -1,4 +1,10 @@
...
+        "gecko": {
+            "id": "addon@darkreader.com",
```
alexanderby: I own darkreader.org domain, darkreader.com is under some chemistry manufacturer.

src/manifest.json
```diff
@@ -1,4 +1,10 @@
...
+            "strict_min_version": "42.0"
```
alexanderby: Did you try it on FF42? Currently I have FireFox 54 installed, let's stick to version that we can check.

yarn.lock
```diff
@@ -0,0 +1,1659 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
alexanderby: Remove this file. Let's use Yarn when there will be a problem with NPM.

alexanderby:
I've also added background fix for FireFox.

src/background/extension.ts
```diff
@@ -410,11 +410,15 @@ style && style.parentElement.removeChild(style);
...
+                chrome.fontSettings.getFontList(function (res) {
+                    // id or name?
+                    var fonts = res.map(function (r) { return r.fontId; });
+                    onReturned(fonts);
+                });
+            } else {
+                onReturned([]);
```
alexanderby: Let's provide at least generic font names instead of empty list https://developer.mozilla.org/en/docs/Web/CSS/font-family#<generic-name>

Currently it is not possible to enter font name but it should be picked from list.

nemanjan00:
I do have a lot work to do couple next days so I will take a look at it after that.

nemanjan00:
Or you can add list... :)

alexanderby:
Everybody is busy. There is only a small amount of work to make it run in Firefox. Then I will prepare it for the store, there is much harder work with all that icons and screenshots.

alexanderby:
I've fixed this extension work in Firefox, so closing your PR without merge.

nemanjan00:
When can we expect it published on addons store?

alexanderby:
It's under review.
Queue Position: 139 of 145

PR #98, opened 2017-09-25
Fix text color in reply composer at inbox.google.com

bstrdsmkr:
Both the inline reply and reply sub windows are black text on black backgrounds

### DIFF
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -76,7 +76,8 @@
             ],
             "rules": [
                 ".pI { background: rgba(255,255,255,.7) !important; }",
-                ".eB { background: rgba(255,255,255,.4) !important; }"
+                ".eB { background: rgba(255,255,255,.4) !important; }",
+                ".aR { color: rgba(255,255,255,.7) !important; }"
             ]
         },
         {


alexanderby:
I didn't notice the problem. Could you send a screenshot?

bstrdsmkr:
Sorry for the delay!
It's worth noting that this problem only manifests while the reply textbox is focused. If you select anything else (subject line for example), the reply text shows in 100% white and is perfectly legible. I've attached a screen cap of the text not visible, you can see Chrome's autocorrect underline on the "invisible" reply text
screen shot 2017-10-02 at 9 31 09 am

alexanderby:
Unfortunately it looks OK for me.

What happens if you past an image next to text, is it inverted? If so, maybe you have some custom theme. then you should walk up a tree a find an element causing an inversion.

For example try executing this script to find such elements containing `.aR`:
```javascript
document.querySelectorAll(".gbii .aR, .gbip .aR, .azp .aR, .aLF-aPX-KP .aR, .irc_bg .aR, .RY3tic .aR, [style*='url'] .aR");
```

bstrdsmkr:
Aha! Looks like this was caused by the Lazarus extension:
https://chrome.google.com/webstore/detail/lazarus-form-recovery/loljledaigphbcpfhfmgopdkppkifgno?hl=en

That being said, I'd still like to see this change integrated since Lazarus has saved me more than once =)

alexanderby:
Lazarus adds an icon as a background image, so Dark Reader inverts whole editor, but background image is mostly transparent. Currently I cannot suggest a selector that will work on all sites. Maybe there will be some image analyzer later.

Google Inbox, GitHub and many other modern websites save your draft as soon as you type, so you can disable Lazarus for them.

bstrdsmkr:
Lazarus does indeed allow blacklisting sites, and I had installed it an never looked back so after testing Inbox's form saving, I've blacklisted it in Lazarus and all is well.

Thanks again for all that you do for the internet. You're truly doing god's work.

PR #100, opened 2017-09-28
youtube subscription email preview images get inverted in gmail

Theyrealone:
fix youtube subscription emails incorrectly being inverted.
inverts table[background] the table[background] inversion must be before all the other inversions else it breaks the profile icon.

### DIFF
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -68,7 +68,7 @@
                 "google.*.*"
             ],
             "invert": [
-                ".gbii, .gbip, .azp, .aLF-aPX-KP, .irc_bg, .RY3tic, canvas.circle"
+                "table[background], .gbii, .gbip, .azp, .aLF-aPX-KP, .irc_bg, .RY3tic, canvas.circle"
             ],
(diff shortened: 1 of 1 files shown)


alexanderby:
That sounds strange, "background" attribute is deprecated and was used in ancient browsers. Can I have a look on a page where wrong inversion it happens?

Theyrealone:
image

alexanderby:
I mean a hyperlink.

Theyrealone:
I can forward you an email

alexanderby:
You can forward it to darkreaderapp@gmail.com

Theyrealone:
the image is a part of a table[class$=video-spotlight-width] that has the image as background

alexanderby:
Great. That looks like a compatibility with some old mail agents. This problem affects not only GMail.

Lets add `[background]` (without table) to `invert` at `common` section (at new line after `[style*=background...]`) and `[background] *` to `noinvert` at `common` so that images inside will not be inverted twice.

alexanderby:
Updating common fixes is always nervous :)

PR #106, opened 2017-10-11
text not visible

codGmer:
text not visible anymore

### DIFF
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -305,6 +305,12 @@
             "rules": [
                 "#scrim.app-drawer { background: rgba(255, 255, 255, 0.5); }"
             ]
+        },
+        {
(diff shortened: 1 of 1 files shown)


alexanderby:
Hi.
1. Please, sort fixes in alphabetical order by URL.
2. `"removebg": "*"` will do exactly the same, `"rules"` should be used as a last resort.
3. If possible, try to remove background for exact elements, not the whole page. Your rule can remove background color and images where you do not expect.

alexanderby:
The config file was simplified, now it is here https://github.com/alexanderby/darkreader/blob/master/src/config/inversion-fixes.config

You should use `REMOVE BG` rule in this case.

alexanderby:
A year has passed, fronter.com looks OK.

PR #111, opened 2017-10-18
Stopped invertion of the blue parts of the facebook header + profile pic

ITR13:
Couldn't figure out how to invert the part with the profile pic without inverting the name, so the name is still inverted.
First time doing this, so I may have specified some selectors wrong (I'm especially worried about the two long selectors I added), or put it the wrong place. 

I wanted to stop inverting dark-blue areas as the inverted color doesn't mix well with a dark background (The original color fits slightly more), but couldn't figure out how to do so for the main page elements without specifying the "#js_[somestring]" for each post

### DIFF
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -63,7 +63,12 @@
             "invert": [
                 "._3ixn, ._2teu, .uiStreamStory",
                 ".userContentWrapper canvas, ._4lqu, ._4lqt",
-                "._24ws, ._1z0-"
+                "._24ws, ._1z0-",
+                "._2s1x ._2s1y",
+                ".__tw",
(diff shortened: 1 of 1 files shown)


alexanderby:
Thanks for the try. There used to be bugs, when I tried to not invert the headers. The pop-ups became hidden and text became blurred. I will check out your fix, but maybe it would be better to define inverted background and text colors.

alexanderby:
I've just fixed (I hope so) the header. Your selectors for profile pics seem a bit complex. Could you send a screenshot or a direct link for a wrongly inverted page/

ITR13:
It seems to work now, so I assume it probably was something with how I inverted it in my fix. 

Not sure what I did, but it's probably wrong due to not really knowing what I was doing, as I said, first time :-P

PR #117, opened 2017-11-04
Fix Я.Радио track menu

BANOnotIT:


BANOnotIT:
Data rules are 'case they're background for links, under inversion. I tried to use `filter: none`, but it doesn't work, so i just change color of buttons with their background.

alexanderby:
Maybe you have to use `invert` then, I've added an explanation in ReadMe:
> When Dark mode is on, the whole page (root `<html>` element) is inverted by filter. To revert the images, videos etc. `invert` selectors are used, so the inversion will be applied to these elements twice. If inverted elements contain other elements that match the `invert` selectors, then these elements will be inverted 3 or more times. To prevent it `noinvert` selectors are used.

Some other notes:
- You can merge multiple URLs into `radio.yandex.*`
- `body` seems to be explicit, `.icon_share-white` should be enough. Or you can use `.track__menu` selector in your case to merge multiple selectors into one. If there are images inside `.track_menu` use `noinvert`:
```json
{
    "url": "radio.yandex.*",
    "invert": ".track__menu",
    "noinvert": ".track_menu *"
}
```
So, the list of fixes will remain compact.

BANOnotIT:
In case of yandex developed their own dark theme this PR should be closed

PR #138, opened 2017-12-04
Inversion fixes for mathprofi.ru and my.mail.ru/music, added 3 sites to dark list

Rult:

PR #140, opened 2017-12-04
Inversion fixes for mathprofi.ru and my.mail.ru/music, added 3 sites to dark list

Rult:
#138 and #139 are failed. This is ready version.
(You can also add "my.mail.ru/bk", "my.mail.ru/list", "my.mail.ru/ok" in array, I forgot it)


alexanderby:
Great work, thanks. Not inverting `html` element is an interesting solution. Unfortunately the sepia, grayscale etc. will not work in this case, but you gave an idea to be able to change the main inverting root or the ability to specify multiple root, so I will think on adding a `"root"` property into site fix config.

PR #142, opened 2017-12-05
Update fix_inversion.json

GCascaes:
Added non-inversion to some already dark website headers.

### DIFF
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -65,6 +65,10 @@
             "invert": ".footer-distributed, .navbar-inverse"
         },
         {
+            "url": "docs.microsoft.com",
+            "invert": ".theme-dark"
+        },
+        {
             "url": "ebay.*",
             "invert": [
                 "body[style*='background-image: url'] img",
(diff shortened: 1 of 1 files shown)


src/config/fix_inversion.json
```diff
@@ -207,6 +215,10 @@
             "url": "reddit.com",
             "invert": "#header-img"
         },
+        {
+            "url": "redmine",
```
alexanderby: Did you mean `redmine.org`?

GCascaes: Hi,

Actually no, redmine is often (and also in my case) used in the internal network of companies, so their address is usually just redmine/
Does it make it too open? If this is bad for merging, is there a way for me to apply that only to me, directly in the extension? I have it in the Inversion Fix Editor, but every now and then I have to open it and apply again, because it returns to default.

Thanks for the great extension you made ;)

alexanderby: I plan to improve the Editor during Christmas holidays, so that you will be able to use fixes for local sites and prevent them from being overriden by GitHub config. Currently I will only merge fixes for global web sites.

GCascaes: Ok, cool, thank you!

PR #148, opened 2017-12-14
partial fix for dedication.pl

senpl:
at least main paragraphs are visible.


alexanderby:
I've removed `*` from the end of URL. It should be used in cases when some website has multiple domain names (e.g. `google.com`, `google.pl` can be replaced with `google.*`).

PR #157, opened 2017-12-16
Fixed imdb.com

Mav3r1x:
Previous pull request was cancelled by Travis. I've fixed the navbar, titlebar, summary and poster. It's all black now.


alexanderby:
If there is an error, there is no need to open new pull request, but to edit an existing one.

PR #160, opened 2017-12-17
Fixed deviantart mobile and github unlogged page

Rult:
Can't get the Github header dark, because the drop-down list about profile starts to hide behind other elements


alexanderby:
No need to open a new PR, you can edit an existing one.

PR #164, opened 2017-12-27
fix "#discourse-comments" rule

senpl:
for pages like https://babeljs.io/docs/usage/polyfill/ element "#discourse-comments" should be inverted.


alexanderby:
I'll try to fix it by injecting CSS into iframe itself, it should be possible if Iframe is in the same domain.

PR #165, opened 2017-12-27
complete fix for div with discurrse comment

senpl:
for pages like https://babeljs.io/docs/usage/polyfill/
this one will work. all div with selected id should be inverted.

### DIFF
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -1,6 +1,7 @@
 {
     "common": {
         "invert": [
+            "div#discourse-comments",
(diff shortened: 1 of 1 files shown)


alexanderby:
1. `div` is not necessary selector. 
2. This will invert user pics.
3. As far as iframe domain name looks the same, it should be possible to inject stylesheet inside an iframe. This is a planned fix.

alexanderby:
I think proper IFrame handling will be fixed soon https://github.com/alexanderby/darkreader/issues/17

alexanderby:
IFrame handling was fixed, site looks OK.

PR #167, opened 2018-01-02
Add initial url section for ruby.sketchup.com and rubydoc.info

DanRathbun:
This is a inversion fix section for sites that are generated by the YARD Ruby documentation generator.
Most Ruby language Gems have their source on GitHub and documentation on www.RubyDoc.info/gems.
Many other Ruby projects that are also on GitHub have documentation generated (automatically by GitHub) and published to www.RubyDoc.info/github.
Trimble SketchUp (formerly Google SketchUp,) uses YARD to publish their Ruby APIs on their own server.
(The "header#navbar" selector is unique to SketchUp's YARD template, but the standard template does not have this body child element so no harm for RubyDoc.info hosted pages.)
-
This file change only does a few glaring fixes such as inverting the left nav iframe, and reverting the header bar.
There are some other things that need fixing inside the iframe#nav, but I do not know yet how to fix these.
-
* References for YARD :
https://yardoc.org/ 
http://www.rubydoc.info/gems/yard/file/README.md

### DIFF
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -337,6 +337,19 @@
             "invert": "#header-img"
         },
         {
+            "url": [
+                "ruby.sketchup.com",
+                "rubydoc.info"
+            ],
+            "invert": [
+                "header#navbar"
+            ],
+            "noinvert": [
+                "header#navbar a > img",
+                "iframe#nav"
+            ]
+        },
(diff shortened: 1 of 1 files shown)


alexanderby:
I plan to fix inversion inside of Iframes (only same domain can be fixed).

Issue #92, opened 2017-09-20
Maybe move devtools to a separate page?


BANOnotIT:
It isn't very comfort to use devtools in "modal".

alexanderby:
I agree. In next version a planned to make an editor with settings groups, theme editor, e.t.c for advanced users on separate page. Current solution is the simplest one, but as a workaround you can copy text from the text area, paste it into JSON editor and then past it back.

J-e-r-e-m-i-a-h:
is there anyway to make it... commit forever? every 10ish minutes i have to reopen the panel and click "apply" again

RalphLesch:
The extension updates (downloads) the current [fix_inversion.json](/alexanderby/darkreader/blob/master/src/config/fix_inversion.json) from github every 15 minutes, which "overwrites" the local custom config, although the changes made in the inversion fix editor are not changed.
It would be probably best to separate the local from the remote version and to overwrite the synced / downloaded remote version by the local one. Or in other words: merge local and remote versions by favoring local over remote.
When simply overwriting the remote site specific configs with the local ones, merge errors should be minimal.

alexanderby:
This will appear the next release. Also text format will be simplified.

PR #171, opened 2018-01-03
[WITHDRAWN] fixes for SketchUp Free (app.sketchup.com)

DanRathbun:
Here is the URL for the new SketchUp Free online cloud 3D modeling app:
https://app.sketchup.com/app?hl=en

### DIFF
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -31,7 +31,7 @@
             "noinvert": "#footer *, #photo-modal *"
         },
         {
-            "url": "app.sketchup.com",
+            "url": [ "app.sketchup.com", "edu.sketchup.com" ],
             "invert": [
                 "#canvas",
                 "#layers-container-overlay .sp-picker-container",
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -31,6 +31,48 @@
             "noinvert": "#footer *, #photo-modal *"
         },
         {
+            "url": "app.sketchup.com",
+            "invert": [
+                "#canvas",
+                "#layers-container-overlay .sp-picker-container",
+                ".about-tab img",
+                ".bottom-bar-brand-logo",
+                ".download-card-image",
+                ".category-thumbnail > img",
+                ".component-thumbnail > img",
+                ".dimension-details-section-header",
+                ".display-secondary-detail",
+                ".entity-info-label",
+                ".entity-info-layer-material-icon > img",
+                ".files-tab-header img",
+                ".layer-thumbnail > img",
+                ".login-email",
+                ".login-picture",
+                ".material-instance > img",
+                ".options-label:not(.option-toggle)",
+                ".option-toggle-label",
+                ".scene-attributes-title",
+                ".scene-properties-title",
+                ".scene-details .labeled-checkbox > div:first-child",
+                ".scene-thumbnail > img",
+                ".sp-input-container",
+                ".thumbnail img"
+            ],
+            "noinvert": [
+                "#instructorIframe",
+                "img",
+                ".checkbox-label .checkbox-checkmark",
+                ".entity-info-input",
+                ".entity-info-back-material-container > .inline-div",
+                ".entity-info-layer-material-container > .inline-div",
+                ".entity-info-layer-material-icon[style*='background-image: url']",
+                ".scene-description"
+            ],
+            "rules": [
+                ".entity-info-label, .options-label, .dimension-details-section-header, .scene-details, .scene-attributes-title, .scene-properties-title, .scene-description { color: #C9C9C9 !important; }"
+            ]
+        },
+        {
             "url": "atlassian.net",
             "invert": [
                 ".EditPage_withPanels_3Hr, #toolbar, #savebar-container"
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -31,7 +31,10 @@
             "noinvert": "#footer *, #photo-modal *"
         },
         {
-            "url": [ "app.sketchup.com", "edu.sketchup.com" ],
+            "url": [
+                "app.sketchup.com",
+                "edu.sketchup.com" 
+            ],
             "invert": [
                 "#canvas",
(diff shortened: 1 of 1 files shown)


DanRathbun:
Fixes became too complex to implement with Dark Reader.

Will implement a standalone Chrome extension instead.
This way I can decide which elements get inverted and leave the rest alone.

CLOSED THIS PULL REQUEST.

alexanderby:
I'm going to implement `"root"` selector (currently root inversion is hardcoded for root `<html>` element), maybe it will simplify your fix.

DanRathbun:
**I don't think so.** I'm up over **12,700** characters and **340** lines and still climbing.
It just got to be a real pain in the butt to keep pasting all that into yourlittle "developer tools" box.

This is all I **really** needed to invert:
```css
    .left-sidebar-container .file-view .actions-bar img,
    .about-team-tab-pane img, .browser-bottom-bar-action > img, .home-bar img,
    .js-bottom-bar-menus-container img, .js-bottom-bar-menus-container svg, 
    
    .layer-delete > img, .layer-visible > img, .left-tab:not(.about-tab) img,
    .loading-gif > img,
    .standard-views-container .standard-views-tabs > img,
    .tool-group-active-tool > img, .tool-flyout.expanded-base > img,
    .undo-redo-button img {
        filter: invert(100%);
    }
```

I also have some SVGs that are better adjusted changing the `fill` and `stroke` because they have color hover behaviors as well as the normal "lightened" tint. (They were dark gray originally.)

And then all the rest of the backgrounds and text colors are not all the same. 
With my own extension I can control them much better just by injecting a style sheet.

RalphLesch:
If most of the page should not be inverted, only certain elements, you can just undo the common page inverts, including `html` in your site specific rule (under `"noinvert"`) and then `"invert"` the elements you want to be dark.

For example to invert only the side bars:
```json
        {
            "url": "app.sketchup.com",
            "noinvert": "html, img",
            "invert": [
                ".home-bar",
                ".left-toolbar",
                ".right-sidebar-tabs-container",
                ".bottom-bar-container"
            ]
        }
```

But I agree that fiddling in the editor (developer tools) is annoying. A separate window would be great (#92).

DanRathbun:
> If most of the page should not be inverted, only certain elements, you can just ...

If it only were that easy. It's not. The devs hardcoded color and background-color into low level classes assigned to each low level visual HTML element, instead of letting styles cascade down. So you'd think that changing containers (as shown above,) might work, but it doesn't.

I really need to invert only half the **`img`** elements, and then adjust the background colors depending upon what they are. So a lot needs to be darkened, .. but inverting does not give me the control I need because the OEM interface is shades of gray that do not invert very well.

And again I also need to adjust fill and stroke colors, and add things that are not currently there, such as colored outline hover effects.

A standalone extension is best. (It has been awhile but I've been there before, both for Chrome and FireFox.)

> But I agree that fiddling in the editor (developer tools) is annoying. A separate window would be great 

Actually just reloading from my local JSON file would be best, that way I can use my own editor.
This is what I do when working my own extension. Save in the editor, SHIFT+TAB to to Chrome window with the Extensions page, and click the Reload for my extension, reload the app page I'm injecting into.

RalphLesch:
Ok, I see. Then, before you create a new extension, here are some alternatives:

- [Stylish](https://chrome.google.com/webstore/detail/stylish-custom-themes-for/fjnbnpbmkenffdnngjfgmeleoegfcffe): custom style sheet injection.
Probably exactly what you need? Has an integrated editor, which updates the style on save.
On [userstyles.org](https://userstyles.org) you can share your result if you want to. ;)
- [Tampermonkey](https://tampermonkey.net) ([chrome webstore](https://chrome.google.com/webstore/detail/tampermonkey/dhdgffkkebhmkfjojejmpbldmpobfkfo)): JavaScript injection.
If you need more then style injection, for dynamic results, like removing style sheets, customizing classes, replacing elements etc., this would be my first choice.

> Actually just reloading from my local JSON file would be best, that way I can use my own editor.
This is what I do when working my own extension. Save in the editor, SHIFT+TAB to to Chrome window with the Extensions page, and click the Reload for my extension, reload the app page I'm injecting into.

For development purposes, you could also disable the remote style loading of DarkReader temporary, with a litte "hack" in [config_management.ts](/alexanderby/darkreader/blob/7b525bb000f0555da0e2990dd0dcd858024b3c26/src/background/config_management.ts#L15) or the corresponding compiled config_management.js in the extension folder (`[chrome profile folder]/Extensions/[extension ID]/[version]/background/`):
Replace the remote URL for `fix_inversion.json` in [line 15](/alexanderby/darkreader/blob/7b525bb000f0555da0e2990dd0dcd858024b3c26/src/background/config_management.ts#L15) by the local URL below, so the local file gets loaded on updates (extension (re)loads and every 15min).

Happy styling and page manipulation!

DanRathbun:
> @RalphLesch  For development purposes, you could also disable the remote style loading of DarkReader temporary, ...

Way ahead of you. I've permanently disabled Dark Reader for app.sketchup.com by adding it to the exclusion list on DR's "**Site list**" tab. I don't need to be fartin' with the config files. (Better things to do.)

> Then, before you create a new extension, here are some alternatives ...

Thanks, but those are not really alternatives. This is not for **me**. This is for the SketchUp community.

I can't be promoting non-technical users to go install some 3rd party extension and then follow complicated instructions to install or enable a stylesheet or config files, etc., or making tweaks in a GUI editor.
(1) It's all much too complex for most people.
(2) The app is still undergoing fundamental builds, often several release per week. It really needs it's own extension to push automatic updates.

And, also I am known in the community (~ 10 years now,) and a registered SketchUp developer. Pointing towards some non-community authored extension may not be accepted well. (And I don't know as I could publicly "endorse" such a solution. Liability, etc.)

Besides I'm a programmer. Fluent in (among other coding languages,) Ruby, HTML, CSS and Javascript. I've created Chrome and Grease Monkey extensions in the past. And I'm 95% done with this one's Alpha phase.

**Not a big deal.**

... regards.

PR #172, opened 2018-01-04
Fixed some already added sites, supported wikia

Rult:
DA: fixed mini-avatars on previews
MathProfi: fixed imgs
VK: fixed text on album previews
Wikia: setting black background and white text

### DIFF
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -502,12 +502,15 @@
             "url": "wikia.com",
             "invert": [
                 ".wds-global-footer",
+                ".wds-global-footer__header",
                 "#WikiaBarWrapper"
             ],
             "rules": [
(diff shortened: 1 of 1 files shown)


src/config/fix_inversion.json
```diff
@@ -498,6 +498,21 @@
             "invert": ".qrcode .icon-logo",
             "rules": ".qrcode { border: 8px solid black; }"
         },
```
RalphLesch: - Light themed *.wikia.com pages are fine, although the removal of the background image is not really necessary in my opinion.
- Dark themed *.wikia.com pages are falsely inverted, should be excluded. To take an example from the main page: http://memory-alpha.wikia.com
- wikia.com and fandom.wikia.com pages are not covered, images and dark parts are inverted.

Checkout my fix (#175)  as an alternative, which should cover (hopefully) all subpages.

Rult: Wow, that's really good and compact version.
I'll create new request without wikia.

PR #175, opened 2018-01-07
Fix for Wikia.com

RalphLesch:
Full coverage for all *.wikia.com pages, with support for both light and dark-themed sites, latter are excluded / not changed, background images are generally preserved, to keep the page style intact.

Main pages ([www.wikia.com](http://www.wikia.com), [fandom.wikia.com](http://fandom.wikia.com)), (non-wiki pages) are also covered.

### Notes about possible check fails
The order of the site fixes are important, to prioritize for e.g. the fix for `www.wikia.com` over `*.wikia.com` . This conflicts (possibly) with the alphabetic order of site fixes, but is currently the only solution, as far as I know.

### DIFF
--- a/src/config/dark_sites.json
+++ b/src/config/dark_sites.json
@@ -17,6 +17,7 @@
     "destinytracker.com",
     "discordapp.com",
     "dotabuff.com",
+    "fandom.wikia.com",
     "ffmpeg.org",
     "filterblade.xyz",
     "frootvpn.com",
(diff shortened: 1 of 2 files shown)


Rult:
Good for PC, but on mobile version (change user-agent to check) it's falsely inverted.

RalphLesch:
Mobile sites should be fixed now. Let me know, if I missed anything else.

alexanderby:
I need to think how to resolve these URLs.

alexanderby:
Works well in Dynamic mode.

Issue #119, opened 2017-11-07
wildcard in site list doesn't always work


jmona789:
I use google photos for pictures, I like it inverted while browsing but not when viewing a picture because of he overlay adding "photos.google.com/photo" works but if I search for something and then view a picture the URL is "https://photos.google.com/search/query/photo/" so I tried adding 
"https://photos.google.com/search/*/photo or even "https://photos.google.com/*/photo/" to the not invert list but none of that works.  I tried serval different variations none of which work

RalphLesch:
The star works only in the domain part, for the rest you can use some RegEx: `^photos.google.com/.*/photo/`.
See for e.g. [this comment](/alexanderby/darkreader/issues/86#issuecomment-355656408) for more details.

alexanderby:
Unfortunately wildcards don't work well and need to be reworked.

alexanderby:
It should work better with Dynamic Theme mode, but here is what is planned to do with list https://github.com/alexanderby/darkreader/issues/373

Issue #28, opened 2016-06-12
Feature request: separate black/white lists


ershov:
Hi,
I love this extension and I have a proposal for improvement.
Today, there's only one list of sites which acts as 'inverted list' or 'no touch list' depending on the mode.
This doesn't make much sense because the sites that are chosen to be inverted should always be inverted, not depending on the default mode. And vice versa.
The suggestion is: there should be two separate lists for sites to be inverted and for sites to stay untouched. And the 'default' mode will do the default to the sites that are not in any of the lists.

alexanderby:
@ershov This figure explains why one Site List covers all use cases.
![dark-reader_site-list_diagram](https://cloud.githubusercontent.com/assets/9638588/16033692/1fd826e8-3216-11e6-9f8b-6c7101d77a19.png)

ershov:
I'm not quite sure what it explains.

The behavior that I see today is as follows:
1. Add an URL (e.g. google.com) to the 'sites list'
2. If 'not invert listed' selected, the site is not inverted.
3. If 'invert listed only' selected, the site is inverted.

Once I start using 'sites list', I can't switch the default mode.
The site 'google.com' is always white => I always want it inverted.
The site 'store.steampowered.com' is always dark => I never want it inverted.
But there are lots of unlisted sites which I want to see in the 'default' mode.

alexanderby:
There are two use cases:
1. You want most of the sites to be dark. Then use "Not invert listed" for ignoring sites which are inverted incorrectly (e.g. already dark, but inverts into white).
2. You want few sites to be dark. Then use "Invert listed only" to specify what to invert.
These two strategies are illustrated at the bottom of the figure.

ershov:
Right.
Here's the problem: I have to choose the mode from the very beginning and can't change the mode once I have any sites in the list.

alexanderby:
I find this case very specific, and adding additional list will make UI more complex. So, I think it's more simple and universal to add a JSON config editor.

RalphLesch:
Related: #147 
For e.g. subdomains cannot be fully controlled by the current include/exclude mechanics and URL syntax.
If not using both include and exclude lists, I would suggest changing the URL syntax to allow better control of the includes/excludes.

alexanderby:
I was thinking of adding unlimited count of settings groups. Currently I'm refactoring the UI, so will work on it later.

RalphLesch:
Sub groups for domains should be enough to solve most URL match problems.
[...]

PR #176, opened 2018-01-07
Fixed *.gamepedia.com

RalphLesch:
Fix for *.gamepedia.com sites, with support for both light and dark themes.


alexanderby:
I need to think how to make it work. Currently try to merge your fixes into one domain.

RalphLesch:
Oh boy, I can try, but it will not be easy.
If you are reworking the UI, with subgroups for domains (#28,), and changing the URL regexp parser (#119), the URL match problems should be much easier to handle.

Issue #143, opened 2017-12-06
Rules created with the developer tool are not properly reapplied.


milochristiansen:
I have created a few custom rules via the developer tool, but they are only applied to the site when you open the tool and click "apply". Refreshing the page works, but navigating to a new page on the same site does not apply the custom rules until you open the editor and click apply again.

It seems anything entered in the developer tool only affects the page that was open when "apply" was clicked.

(By "rules" I mean "invert" and "noinvert" entries, as well as "rules" entries.)

Note: Sometimes, very rarely, it works. I have no idea what causes it to work or not work at a particular time. Furthermore, if you go back to a page you visited before (via a link, not the browser back button) it often works fine on that page (well, maybe not *often*, but more often than it does normally it seems).

fwpowell:
I can confirm this behavior on Chrome 64.0.3282.24 (Linux) / Chrome 63.0.3239.84 (Windows).
When you create the rules using the developer tool, initially, they're applied and function for a period of time. After a number of automatic refreshes (or duration), the rules are seemingly 'forgotten' and require opening the developer tool and clicking 'Apply'.

To recreate this you can use the Observium demo site: demo.observium.org (U: demo / P: demo) as an example (set the refresh period to ~1 min, top right drop-down menu).

Same as @milochristiansen, I have little idea as to what's causing this. Doesn't seem to be very common either as I've been looking for around a month and this is the first other mention I've seen. 

Truly an amazing extension, the absolute best I've tried in the category - thank you for all the effort that went into it.
Please let me know if there's any additional info I can provide

alexanderby:
Every 15 minutes the config is loaded from GitHub and overrides your editor values. Currently I'm doing UI refactoring, editor will be placed into separate page, then I will think about merging GitHub config values with user's values. Currently you can submit a PR for GitHub config.

jordancrane:
The ability to merge GitHub and user config would be useful in some cases. My specific example is that I have some work-specific internal sites that I have fixed with the developer tool, but since they are internal sites it would be useless for everyone else to have them in their configuration files as well. It's not a huge annoyance to just press the "apply" button again when I visit the sites in question, but it would be a nice feature to have local configuration changes persist.

Mav3r1x:
Glad to know that you're already working on this issue. Maybe create a section called " local rules" in the developer tools, completely different from the inversion json file that is pulled from github.

Gusted:
@milochristiansen check with latest darkreader version.

PR #181, opened 2018-01-12
Github header fix

unstope:
Added Github .Header element into invert list so that it inverts once more and becomes dark

### DIFF
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -151,7 +151,8 @@
                 ".bg-gray-light",
                 ".bg-gray-dark",
                 ".bg-gray-light .btn-block",
-                ".bg-gray-dark .form-group"
+                ".bg-gray-dark .form-group",
+                ".Header"
             ]
         },
         {


alexanderby:
Have a look at my latest fix for Facebook header. Chrome usually fails to render dynamic content that has filter applied multiple times (that was the case with Facebook, Linkedin etc). Maybe try to set Light color and dark text for the header via `"rules"`, invert (or not) the logo.

unstope:
It works but after Chrome is restarted custom rules are not injected into the page. If I click Apply button they become injected till the next Chrome restart.

image

RalphLesch:
That is already a known bug: #143

unstope:
I have checked the behavior of first variant and it works the same way as the custom rules (until I restart Chrome). I guess these bugs are related to each other.

You can see how Github looks like before restart.

image

alexanderby:
Profile icon get inverted. When you click on "+" near profile icon, the context menu gets partially disappeared. So, i would suggest to use background and color rules fix.

alexanderby:
I've fixed the header by https://github.com/alexanderby/darkreader/commit/7a76fdc2cbade046c39c1917c8f6116cac25159b#diff-7d816c61670a7107b3850c13d1fa2861

PR #185, opened 2018-01-18
Fixed Evernote web

vitaly-zdanevich:
Text editor was white.


vitaly-zdanevich:
Sorry incorrect place of the edit.

PR #186, opened 2018-01-18
Evernote web client - fixed color of editor

vitaly-zdanevich:


spthorn:
This can be closed... the Evernote editor is darkened just fine already.

PR #188, opened 2018-01-19
partial fix for reading livescience.com bacground

senpl:
still not whole page inverted but at least it is now readable.

### DIFF
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -226,6 +226,12 @@
             ]
         },
         {
+            "url": "livescience.com",
+            "invert": [
+                "background-color"
+            ]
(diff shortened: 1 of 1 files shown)


alexanderby:
Looks strange. Your current selector refers to an element `<background-color>`, but I doubt that this element can exist. Please, check.

- `background-color` selector will refer to elements like `<background-color>`.
- `.background-color` selector will refer to elements like `<div class="background-color">`.
- `#background-color` selector will refer to elements like `<div id="background-color">`.
- `[style*='background-color']` selector will refer to elements like `<div style="background-color: black;">`.

senpl:
in new version no change required

PR #191, opened 2018-01-22
Fix math equations on wikipedia.org

alexalekseyenko:
Ran across some math using the `mwe-math-fallback-image-display` class. This makes it possible to read on a dark background.


alexanderby:
There were 2 whitespaced instead of 4, but I fixed it myself.

PR #193, opened 2018-01-24
Add electronics-tutorials.ws support

vital-net:
"removebg": "img" because Gif are transparents and look is bad.


alexanderby:
Nice solution. There was an extra white space and I removed it https://travis-ci.org/alexanderby/darkreader/builds/332851479#L1031

PR #195, opened 2018-01-27
Fixed electron and tproger

Rult:
About commentary section on TProger articles - I don't know how to fix it correctly.


alexanderby:
I think it is not possible to fix, there is an IFrame, and it's "src" points to another domain. Browser will not allow to inject a script there.

PR #199, opened 2018-01-31
Fix vk.com: article previews and image viewer

risualiser:


alexanderby:
Let me know when I can merge it :)

risualiser:
Thx, but you can merge it right now, if I find new fixes, I'll add it in new PR.
I had problems with checking on mobile version.

alexanderby:
Thanks. What do you mean by mobile version? As far as I know it is not possible to install extensions on mobile Chrome.  There was an idea for separate mobile browser, but it seems to be much work. I haven't used mobile Firefox, but as far as I know it would be possible to install extensions there. Currently porting to Firefox is in progress.

risualiser:
Try mobile Yandex.Browser, it let you use extensions.
Just go to Chrome Web Store and install what you want (some extensions are disable, but your is available).

alexanderby:
This is very cool.

PR #200, opened 2018-01-31
Global rule for 'Atlassian Confluence' background

sylverpyro:
Atlassian Confluence deployments all contain a ' div id="main-bg" class="main-bg" ' tag at the top of the generated pages that are not matched by any of the global invert rules.  

Site specific invert rules in the form of 
"url": "site.com", 
"invert:" ".main-bg"
work, however adding a rule per site seemed ridiculous as there is an unlimited number of these around the internet (I interface with two on a daily basis myself and they do not share a common URL element to glob).  

This added to the common -> invert section however fixes it for all Confluence sites that I've tried it on.

### DIFF
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -7,7 +7,8 @@
             "[style*='background:url'], [style*='background-image:url']",
             "[style*='background: url'], [style*='background-image: url']",
             "[background]",
-            "twitterwidget"
+            "twitterwidget",
+            "[id='main-bg'], [class='main-bg']"
         ],
         "noinvert": [
             "[style*='background:url'] *, [style*='background-image:url'] *",


alexanderby:
1. Use `#main-bg` instead of `[id='main-bg']`, `.main-bg` instead of `[class='main-bg']`.
2. There is a chance that `.main-bg` element can be found on some non-Confluence site. Maybe there is some Confluence-specific classname for some parent element, so you will be able to use a nested selector like `.atlassian .main-bg`.
3. Current list of fixes in Dev Tools is overriden after every update. Having a personal list of fixes is planned to help making fixes for Local Area Network sites.

alexanderby:
I hope Dynamic mode works well with Atlassian.

PR #209, opened 2018-02-04
fix for tasks in new google calendar view

senpl:
In current google calendar there is tasks that could be viewed. I added inversion for that element.

### DIFF
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -69,6 +69,10 @@
             "invert": ".footer-distributed, .navbar-inverse"
         },
         {
+            "url": "calendar.google.com",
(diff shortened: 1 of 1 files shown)


alexanderby:
Did it work? I see that tasks are inside an IFrame, we cannot fix IFrame content at the moment.

senpl:
> Did it work? I see that tasks are inside an IFrame, we cannot fix IFrame content at the moment.

It works after page is loaded then clicking apply. So maybe refresh just after all other loads would make it work. For me even manual clicking apply is better then nothing.

alexanderby:
I've merged it. I'm not sure, if `calendar.google.com` will conflict with `google.*`, there is already an issue for similar cases.

PR #215, opened 2018-02-07
build: add package-lock.json

thetric:
this is recommended by npm (see https://docs.npmjs.com/files/package-lock.json)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,3159 @@
+{
+  "name": "darkreader",
+  "version": "3.5.2",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "abbrev": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
+      "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==",
+      "dev": true
+    },
+    "ajv": {
+      "version": "4.11.8",
+      "resolved": "https://registry.npmjs.org/ajv/-/ajv-4.11.8.tgz",
+      "integrity": "sha1-gv+wKynmYq5TvcIK8VlHcGc5xTY=",
+      "dev": true,
+      "optional": true,
+      "requires": {
+        "co": "4.6.0",
+        "json-stable-stringify": "1.0.1"
+      }
+    },
+    "ansi-regex": {
+      "version": "2.1.1",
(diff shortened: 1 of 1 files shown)


alexanderby:
The package list will look absolutely different soon, so this file will be rewritten almost completely.

alexanderby:
I haven't chosen yet. It seems suitable for Inker and Taucharts projects and I wanted to try it on Dark Reader to find possible issues and disadvantages.

There are some pros and cons:

### Pros
- Simplicity, just about 400 lines of code to debug. I tried to write it without overengineering. The complexity is moved into plugins and you can enable them if you need them.
- Usage of native API for handling events instead of huge React event system.
- It is possible to get DOM node before rendering a child node. In React (and similar APIs) it requires additional `componentDidMount() -> setState()` cycle.
- Small bundle size (7KB vs 108KB in React).
- Control over everything (from author's point of view).
- Simple API for animations.
- Native-like API for HTML attributes and CSS properties.
- `class` instead of `className`. This is no doubt a **killer feature** :)

### Cons
- No community.
- No best practices.
- No ready for use components.
- Requires a deep JS and DOM API knowledge.
- No dev tools (like React Dev Tools extension).
- It was not used much, who knows what are the other downsides.

So, Malevic is more suitable for standalone widgets, charts, small apps with dynamic UI and for writing a web app alone, but I'm not sure if it is suitable for a team work and long time perspective.

thetric:
[Preact](https://preactjs.com/) might be an alternative, too. Some points you mentioned are met:
* small code base (<1k loc, estimated)
* no synthetic events
* small bundle size (3kb preact, 7kb animations)
* `class` instead of `className`
* component libraries
* devtools

See also https://preactjs.com/guide/differences-to-react. Might be worth a look :)

alexanderby:
Then I would choose Inferno :) I heard that the developer was hired by Facebook.

Everything suits well for a simple UI, so for me this is a choice between mainstream and own solution.
- Mainstream gives you ready for use infrastructure but less flexibility. It's complexity usually grows as far as developers are trying to cover all the use cases and move far away from initial goals. I'm already experienced with React and know what to expect. It would be a great choice for Dark Reader: there is no network limit, multi-touch or animations.
- Own solution gives you fun :) and flexibility. But you can see that I've already used own MVVM framework 4 years ago and today it looks ugly (but maybe Angular, Backbone or Knockout look ugly today too).
- Non-mainstream solutions usually give you problems in a year or two when they are no more maintained. You have to rewrite your app for another framework or learn the source code and fork it.

I'm so much tired with these choices: ES6 vs. TypeScript, Webpack vs. Gulp+Rollup, React+JSX vs. everything else, CSS Grid vs. Flexbox. Nothing is ideal. From a product point of view you should just code and don't waste time on thinking :)

PR #221, opened 2018-02-08
Improve compatibility with Lazarus extension

phoenixeliot:
Lazarus is (well, was) an extension for saving text field inputs, that I still have installed. It adds a background image with a small icon to text fields that's clickable for its menu, but that background image was causing the text to be black-on-black. This line prevents the text from being flipped if the background image is the Lazarus icon (or Lazarus's saving icon).

Lazarus is no longer published on the chrome store, so I understand if you don't particularly want to bother with being compatible with it officially. But a lot of people probably still have it.

I believe these images are under the GPL. https://github.com/lazarus-recovery/chrome_patched/tree/master/images


alexanderby:
I'm thinking on implementing personal or optional fixes, so that performance will not be affected for the majority of users. Currently I'm making some refactoring to be able to easily add more features, so currently let's put your fix on hold.

alexanderby:
Sorry, but more than a year has passed. Many things changed since that. This file no longer accepts PRs. Also changes in Dev Tools are saved persistently.

PR #232, opened 2018-02-18
Add gfycat.com

opssemnik:
gfycat is already a dark site (dark purple towards navy blue), however its footer is white.
Add the whole body to exclusion zone so footer gets inverted, also add rule to not invert the video frame

### DIFF
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -190,6 +190,11 @@
             "noinvert": "img"
         },
         {
+            "url": "gfycat.com",
+            "invert": "body",
+            "noinvert": "video"
+        },
+        {
             "url": "gigaom.com",
             "invert": ".bg"
         },


alexanderby:
This site is displayed white for me
image

risualiser:
@opssemnik Can you visit gfycat again? 
It's white by default for me and now also can be perfectly inverted by new Dynamic theme (from "More" tab).

XanderLuciano:
gfycat's homepage and detail pages are all white themed, but the individual "gif" pages have a dark background. This should be a CSS selector fix and not domain level.

Example page: https://gfycat.com/damploathsomecow

alexanderby:
The config file format has changed. If you are not using Dynamic mode, but using Filter mode, please, submit changes to `inversion-fixes.config`.

PR #236, opened 2018-02-19
Fix xda-developers

opssemnik:
Fix white header (div) 
Fix White header buttons ( i )
Fix White header text ( span ) 
Fix White floating text on images (thumb_meta)

### DIFF
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -732,6 +732,10 @@
             ]
         },
         {
+            "url": "xda-developers.com",
+            "invert": "div.thumb_meta, i, div#header.header-logged-out.portal-header,div#header.header-logged-in.portal-header, span"
+        },
+        {
             "url": "yandex.*",
             "invert": [
                 ".afisha, #wd-_afisha",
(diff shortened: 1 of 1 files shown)


src/config/fix_inversion.json
```diff
@@ -731,6 +731,10 @@
                 ".numberedequation, .displayformula"
             ]
         },
+        {
+            "url": "xda-developers.com",
+            "invert": "div.thumb_meta, i, div#header.header-logged-out.portal-header,div#header.header-logged-in.portal-header,div#header.header-logged-out,div#header.header-logged-in, span"
```
alexanderby: 1. There is a rule called "CSS Specificity". To describe it in simple words: `#id` gives you 100 points, `.class` gives 10 points, `div` gives 1 point. `div#header` will give you 101 points. So, you can just leave two selectors `.thumb-meta, #header` in this case.
2. `span` selector is very generic, the rule will be applied to elements you don't expect. Try to find a more specific one.
3. If the "invert" value is too long, you can split it into multiple lines like
```json
{
    "invert": [
        "selector1",
        "selector2"
    ]
}

alexanderby:
Dynamic mode works well with xda-developers.com

PR #237, opened 2018-02-19
Fix premierefc

opssemnik:
Fix the already dark main game showcase (div), its text (h1).
Do not invert game images

### DIFF
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -225,6 +225,11 @@
             ]
         },
         {
+            "url": "globosatplay.globo.com/premierefc/ao-vivo",
+            "invert": "h1, div#ao-vivo.palco.pre-evento.pfc",
+            "noinvert": "img"
+        },
+        {
             "url": [
                 "google.*",
                 "google.*.*"
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -230,6 +230,10 @@
             "noinvert": "img"
         },
         {
+            "url": "globosatplay.globo.com/universal/",
+            "invert": "a, header.rmq-e9c50d2d, div.rmq-26e388b"
+        },
+        {
             "url": [
                 "google.*",
                 "google.*.*"


opssemnik:
Oh and by the way, the URL has to be like that, the main domain (globosatplay.globo.com) is used for all ondemand services that "GLOBO" licenses here in brazil (which includes both international channels like fox and universal services) and local channels (in this case ,premierefc is for Brazillian football with some LATAM games)

opssemnik:
I also added universal play to this patch, to keep alphabetical order

src/config/fix_inversion.json
```diff
@@ -224,6 +224,15 @@
                 ".ui_light .navbar"
             ]
         },
+        {
+            "url": "globosatplay.globo.com/premierefc/ao-vivo",
```
alexanderby: I think you can merge these 2 fixes into 1 `globosatplay.globo.com`

src/config/fix_inversion.json
```diff
@@ -224,6 +224,15 @@
...
+            "invert": "h1, div#ao-vivo.palco.pre-evento.pfc",
```
alexanderby: `#ao-vivo.palco` should be enough. Have a look if `h1` element has some class.

src/config/fix_inversion.json
```diff
@@ -224,6 +224,15 @@
...
+            "noinvert": "img"
+        },
+        {
+            "url": "globosatplay.globo.com/universal/",
+            "invert": "a, header.rmq-e9c50d2d, div.rmq-26e388b"
```
alexanderby: `a` selector is too generic. `header` and `div` can be removed

alexanderby:
For making changes you can edit existing Pull Request

src/config/fix_inversion.json
alexanderby: I've re-read about the URL case, let it be 2 separate.

alexanderby:
Many changes and improvements were made since the PR was opened, please, submit new PR if changes are still needed.

PR #240, opened 2018-02-19
Add CS:GO Subreddit

opssemnik:
do not invert banner.
Also do not invert sidebar, on csgo`s subreddit it is already dark (navy blue)


alexanderby:
I've applied your fix to all Subreddits. There seems to be a Chrome bug with mail popup, so I've hidden it.

PR #247, opened 2018-02-20
Fix inversion for central.bitdefender.com

SteveJamesM:


src/config/fix_inversion.json
```diff
@@ -76,6 +76,11 @@
             "url": "calendar.google.com",
             "invert": "#:4.fc"
         },
+        {
+            "url": "central.bitdefender.com",
+            "invert": "div#left_nav.left_nav, nav.top_nav, div#user_avatar",
```
alexanderby: There is a rule called "CSS Specificity". To describe it in simple words: `#id` gives you 100 points, `.class` gives 10 points, `div` gives 1 point. `div#left_nav.left_nav` will give you 111 points, but a shorter `#left_nav` gives 100.

alexanderby:
There is a rule called "CSS Specificity". To describe it in simple words: `#id` gives you 100 points, `.class` gives 10 points, `div` gives 1 point. `div#left_nav.left_nav` will give you 111 points, but a shorter `#left_nav` gives 100.

PR #248, opened 2018-02-21
Fix LaTeX images on terrytao.wordpress.com

alexalekseyenko:
The equations on this site are rendered as black symbols on a white background. Inverting these images makes the equations much easier to read.


alexalekseyenko:
Replaced by #249

alexanderby:
You should be able to edit an open PR, no need to create a new one.

PR #252, opened 2018-02-23
Do not invert country flags in /r/europe on Reddit

pkrakowiak:
Inverted colors make the country flags look wrong (for example white becomes black, etc.). Flags are important :) and should be displayed as is, so that national colors are left intact.


pkrakowiak:
So, according to Travis reddit.com/r/europe does not belong alphabetically between react.js and reddit.com/r/GlobalOffensive...

alexanderby:
The reason was in case sensitive sort.

PR #253, opened 2018-02-24
Fix color inversion for habitica.com

Yutsuten:
WIthout these changes, the site becomes very ugly...


alexanderby:
Great, thanks.

Yutsuten:
Thank you!

PR #264, opened 2018-03-04
Add /inbox to mail.ru music

risualiser:
Also shorten it and remove useless code (.profile__cover_empty)


alexanderby:
I,m not sure how regexp will work currently, I will change the behavior to make it work soon.

risualiser:
Ok, so I'll change it to more shorter variant.
`my.mail.ru/(.*/.*/)?music(/|$)`

alexanderby:
Regexps are not intended to work. I would like it to be similar to Glob patterns, so it will be more readable and there will be a chance to use ignore patterns like `google.*` `!google.com/maps`.

PR #266, opened 2018-03-05
Fix r/Firefox header

MindlessCirno:
Don't invert text on header


risualiser:
I think there are should be one header fix for all subreddits.

alexanderby:
Unfortunately they all are styled differently

PR #267, opened 2018-03-05
r/Paranormal

MindlessCirno:
Don't invert r/Paranormal header image


alexanderby:
We have case sensitive sort at the moment ("P" is upper than "e")

PR #281, opened 2018-03-15
flooxer.com is already dark

ivanzorrilla:


risualiser:
@alexanderby I think you can fix it by yourself and add in new list.

alexanderby:
@ivanzorrilla Site list was moved to `dark-sites.config`

PR #291, opened 2018-03-18
Update fix_inversion.json to fix white background in Google Hangouts …

michaelm2:
…sidebar in Gmail

When logged into gmail the google hangouts sidebar widget is loaded from an iframe that renders with a white background because of line 19.  I'm not sure what downstream impacts removing this line will have so do with as you will.

### DIFF
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -15,9 +15,6 @@
             "input",
             "[background] *"
         ],
-        "removebg": [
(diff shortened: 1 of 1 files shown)


alexanderby:
Inverting content inside IFrame is planned. Inverting the whole IFrame leads to inverting images and videos inside it. So currently it is better to use white background, because text usually becomes black on black.

PR #305, opened 2018-03-23
Better fix for imdb

risualiser:
Still bad z-index on header, but now only on video/photo wrappers, if you didn't scroll it before.

### DIFF
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -336,7 +336,18 @@
         },
         {
             "url": "imdb.com",
-            "invert": "#nb20, .vital, .poster, .slate, .slate_fade_top, .slate_fade_bottom"
+            "invert": [
+                "#nb20 > div, #navbar-suggestionsearch",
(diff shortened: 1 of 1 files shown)


risualiser:
@alexanderby Great! Can I now add fix for Instagram in old one?

risualiser:
@alexanderby There are "Dynamic theme generator" and "Simple dark theme" in your plans, so, will we get three modes ("classic" old with users' fixes, "simple" and "smart") in new release?

alexanderby:
I've decided to try some more methods. In the nearest release there will be possibility to choose between default CSS filters, SVG filter and static theme.

CSS filters `hue` function makes colors less saturate, so custom SVG color matrix should fix it.

Static theme is much more performant and doesn't produce rendering bugs, but often removes colors from items that should stay colored, or makes icons black on black. It's OK for some users but not OK for others.

Then I will start to work on dynamic theme generator.

risualiser:
@alexanderby How many features are left to add before the release?

PR #326, opened 2018-03-28
Better fix for yandex

risualiser:
Moved mobile fixes to special html tag to avoid inversion conflicts

Also, with the "Filter+" mode on Mobile Firefox there is a black screen on yandex.ru.
Even without any code from devtools for this site.

### DIFF
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -1564,49 +1564,36 @@ yandex.*
 
 INVERT
 body.b-page_theme_pure_night
-.afisha
-#wd-_afisha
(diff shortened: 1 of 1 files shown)


alexanderby:
SVG filters work poorly on Firefox. I hope that dynamic theme approach will work, so no fixes from Firefox developers will be required.

risualiser:
There is a similar bug, but only in Chromium browsers:
1) Go to https://www.yandex.ru/themes/night
2) Type some site address, for example w3.org
3) When you type dot symbol (and link appears in results), page become black

PR #331, opened 2018-03-29
Code Combat Addition to Fix Inversion

Seerean:
Initial Inversion
badinversion

After my fixes
seereanadjustedinversion

Please consider my inversion fixes for codecombat.com

### DIFF
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -94,6 +94,10 @@
             ]
         },
         {
+            "url": "codecombat.com",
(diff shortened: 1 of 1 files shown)


alexanderby:
Since March 27 we are using simplified config version https://github.com/alexanderby/darkreader/blob/master/src/config/inversion-fixes.config

Please, move your fix there.

The old config is used by users, who haven't received extension update yet or who cannot update Chrome.

chrisnnchris:
I have added my combat changes to inversion-fixes and reverted any changes to the original inversion_fixes. Please review when you are able to.

PR #345, opened 2018-04-03
inversion-fixes - Fix QR codes on @trezor wallet (wallet.trezor.io)

ondrejsika:


ondrejsika:
I thank you for great plugin. If you want help with maintenance, let me know :) And add somewhere Bitcoin or Litecoin address for donations ;)

PR #368, opened 2018-04-15
Dynamic Theme mode

alexanderby:
- Analyze existing CSS
- Analyze background images
- Enable by default for Firefox and Mobile


risualiser:
That's really great work. Now I'm going to use Firefox on mobile instead of Chromiums.

Issue #373, opened 2018-04-17
URL match patterns


alexanderby:
Implement user-friendly URL glob patterns with negation ability and some special behavior. Similar to [globby](https://www.npmjs.com/package/globby), [extension match patterns](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Match_patterns) etc.

### How it should work:
- `*` matches everything
- `*/*.pdf` matches PDF file extension
- `google.com` matches www.google.com, mail.google.com, google.com/search etc.
- `google.*` matches google.com, google.by etc.
- `*.google.com` matches mail.google.com, inbox.google.com etc.
- `google.com/mail/*` matches google.com/mail/inbox etc.
- `google.com/*.pdf` matches google.com PDF files.
- `localhost:*`should match localhost and any port
- `ftp://*` should match FTP protocol only
- `/^.google\.com\/mail/` should behave as a regular expression` *(but should not be implemented yet, maybe there is no need for it)*.

`*` can only be surrounded by dots and first `/` in host part and by last `/` and file extension dot at path part. `*` corresponds to one host or one path part or to many parts if placed at start or end. Queries should not be allowed.
- `*.google.com` OK
- `goo*.com` error
- `google.com/blog/*` OK
- `google.com/blog*` error
- `google.com/*/blog` OK 
- `google.com/*.pdf` OK
- `google.com/2018-07-*.jpg` error
- `google.com/search?q=cat&p=dog` error, we won't investigate if it should match search?p=dog&p=cat, use regular expressions for that

### URL lists
URL lists are used in user's Site List config and fixes configurations (each record can have multiple URLs). `!` should reverse pattern result.
- `google.com, !*.google.com` should match google.com except it's subdomains.
- `google.com, !mail.google.com` should match everything that matches `google.com` except everything that matches `mail.google.com`
- `google.com, !mail.google.com, mail.google.com/compose` should not match everything that matches `mail.google.com` except everything that matches `mail.google.com/compose`

### Pattern specificity
Pattern specificity is used to determine which exact match from config file to use. It should behave similar to CSS specificity.
- `*` 0.1.0.0
- `google.*` 1.1.0.0
- `google.com` 2.0.0.0
- `*.google.com` 2.1.0.0
- `mail.google.com` 3.0.0.0
- `mail.google.com/*` 3.0.0.1
- `mail.google.com/mail` 3.0.1.0
- `mail.google.com/mail/*` 3.0.1.1
[...]

PR #393, opened 2018-04-24
Remove simulator.io

risualiser:
No need this anymore.


risualiser:
I think there should be a feature which allow user to toggle global Dark List.

alexanderby:
@risualiser This should be fixed by #373 , clicking on Toggle button will produce `!example.com` pattern.

PR #407, opened 2018-04-25
Add dynamic theme rules for Google Calendar

iki:


alexanderby:
I've updated ReadMe regarding Dynamic Theme config. I'm not sure if improving image analysis will not impact the performance and the need of `INVERT` disappears. When in dynamic mode, `INVERT` usage is not recommended. You can use it for transparent dark images or SVG elements, that are not visible on dark backgrounds. The problem in this case is inline header color, so try using `INLINE` rule. I haven't decided, if all inline styles should be automatically processed, it may impact the performance.

PR #419, opened 2018-04-26
checkmark when adding a video to a playlist

fubrick:


fubrick:
This change was meant for the Dynamic mode.

PR #421, opened 2018-04-26
Fix white background once signed into elearning

prstorero:
Since the developers may not have access to the eLearning portal, I have provided screenshots that demonstrate the fix.

image

image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5,6 +5,15 @@ svg
 
 ================================
 
+elearning.utdallas.edu/webapps/portal/execute/tabs/tabAction?tab_tab_group_id=_1_1
+NO
+
+INVERT
+.contentBox.contentBox-edit
+.content.clearfix
+
+================================
+
 facebook.com
 
 INLINE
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -5,6 +5,15 @@ svg
 
 ================================
 
+elearning.utdallas.edu/webapps/portal/execute/tabs/tabAction?tab_tab_group_id=_1_1
```
alexanderby: Let's shorten the URL

src/config/dynamic-theme-fixes.config
```diff
@@ -5,6 +5,15 @@ svg
...
+NO
```
alexanderby: We have `INVERT` and `NO INVERT` commands, `NO` does nothing

prstorero: That's what I kept trying, but every time I would hit apply, the developer tool would move the NO up there for some reason. I'll move it back down like it's supposed to be.

prstorero:
I tried editing the original commit, but I guess it made a new one with the requested changes. This is my first pull request, so I'm unfamiliar with how to make changes once requested.

PR #430, opened 2018-04-29
Update dynamic-theme-fixes for engadget

prstorero:
Inverts the engadget website. Note that the dividers on the main page can't seem to be made darker using the invert command. Attempting it reversed the inversion on the rest of the page. Looks fine on article pages.


alexanderby:
The is a bug, there is no way to fix it with config.

PR #447, opened 2018-05-04
Dynamic theme fix for news.google.com

cderk6:
Before:
image

After:
image


alexanderby:
Fixed in version 4.3.1, all inline styles (except inline background images) are handled now.

PR #450, opened 2018-05-04
Add fixes for animecharactersdatabase

peteradnan:


alexanderby:
I've just added inline style handling, there will be no more need in INLINE rule, wait for update.

Issue #404, opened 2018-04-25
Dynamic Mode - Blender wiki not Inverted


sonicdee:
Blender Wiki is not converted at all with dynamic mode

see:
https://wiki.blender.org/index.php
or deeplink:
https://wiki.blender.org/index.php/Extensions:2.6/Py/Scripts/Nodes/Nodes_Efficiency_Tools

alexanderby:
Looks fine now (v4.5.6). Soon it will be possible to remove strange backgrounds using Dev Tools.

PR #451, opened 2018-05-04
Dynamic theme fix for wiki.blender.org

cderk6:
There may be a larger issue at hand here, but here's a temp fix at the very least. #404 

Before:
image

After:
image


alexanderby:
When fixing Dynamic Theme mode, you should use `INVERT` only for images, canvases and SVGs.

In this case it is a bug, CSS `@import` support is missing.

PR #459, opened 2018-05-05
Make Facebook icons more legible

prstorero:
Makes the Facebook icons in the top bar more legible for the dynamic theme engine.

Before: 
image

After: 
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -26,7 +26,16 @@ facebook.com
 INVERT
 ._2n_9
 ._59fc
-._5lxr
+._5lxs._3qct._p
+.uiToggle._4962._1z4y._330i._4kgv.openToggler
+.uiToggle._4962._4xi2._z9r.openToggler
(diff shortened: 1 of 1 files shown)


alexanderby:
They become light, but when you click on them they become dark.

prstorero:
I made changes with a new commit that correct the dark images for the clicked version as well.

alexanderby:
There is a problem with earth icon, when you click it the content in popup becomes inverted and it seems like filters are applied to other popups twice.

Overall you should only apply CSS filters to void elements (icons, schemes, canvases, SVGs) and avoid applying filters to text blocks and photos. This way Filter and Filter+ work, but we are trying to avoid it in Dynamic mode.

If it's hard to do it for Facebook, leave it as is. Our image analysis will be improved to be able to handle sprite images like this https://www.facebook.com/rsrc.php/v3/y4/r/Sfw9IoCZwGA.png

prstorero:
Strange, it worked for me when I opened it

image

But I understand you would rather approach this with the image analysis instead.

PR #468, opened 2018-05-08
added gidonline.in - dark already

krupenja:
added gidonline.in - website is dark already


risualiser:
It should be before giphy.

risualiser:
I'm not the dev, but I think that there are should be only popular sites in list.
There are tons of sites like your one, I guess.

alexanderby:
@risualiser I think we will make a clean-up when it will hit the performance, currently there are just 91 sites in list.

PR #481, opened 2018-05-09
Fix reddit signup banner

risualiser:

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -54,6 +54,14 @@ INVERT
 
 ================================
 
+reddit.com
+
+INVERT
+.listingsignupbar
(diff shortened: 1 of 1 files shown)


alexanderby:
I think we should avoid inverting elements twice in Dynamic mode, this is a reason why Filter often cause rendering failures.

In this case there are two images: picture with mascot and planet at top-left corner.

Picture was detected as light background image and was inverted, but I just tried to reduce it's brightness instead of inverting:
image

The second picture was detected as a light icon, and I don't see the way to detect it other way. I was thinking to add `REMOVE BG` rule to dynamic mode config to simply remove wrong backgrounds, maybe we can declare image URL and instruct what to do with image.

risualiser:
Ok, got it.

PR #486, opened 2018-05-10
Add Discordapp.com

Fenris95:
DarkReader makes the Discord WebApp kinda blue, and discolours role colours. It's already a dark theme by default.


risualiser:
I think it can be just "discordapp.com/channels".
Because other pages are inverted correctly.

Fenris95:
Changed to "discordapp.com/channels".

PR #497, opened 2018-05-13
Invert Reddit logo text on Reddit redesign.

rhencke:
On Reddit's new site, the logo currently does not invert correctly with Dark Reader and the Dynamic Theme engine.

This change updates it so it does.  Screenshots below.

Before:

screen shot 2018-05-13 at 9 21 52 am

After:

screen shot 2018-05-13 at 9 22 34 am

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -54,6 +54,13 @@ INVERT
 
 ================================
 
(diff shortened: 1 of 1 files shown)


rhencke:
@risualiser Gosh darn it.  Thanks.  Let me see if I can find a more stable selector to use.

PR #511, opened 2018-05-16
Add NO INVERT rule for Wikipedia

aidalgol:
Don't invert the images in the mouseover popups on internal page links.

### DIFF
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -1775,7 +1775,7 @@ INVERT
 NO INVERT
 .mwe-math-fallback-image-inline
 .mwe-math-fallback-image-display
-.mwe-popups * image
+.mwe-popups image
 
 ================================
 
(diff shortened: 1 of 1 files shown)


src/config/inversion-fixes.config
```diff
@@ -1775,6 +1775,7 @@ INVERT
 NO INVERT
 .mwe-math-fallback-image-inline
 .mwe-math-fallback-image-display
+.mwe-popups * image
```
alexanderby: This selector looks strange. Your selector is read as `SVG image inside any element that is inside element with class="mwe-popups"`. Maybe you can remove `*`, it will  make it more performant.

aidalgol: For some reason I thought that without the `star` the selector would match only `image` elements that are _immediate_ descendants of elements with the class `mwe-popups`.

alexanderby: `.mwe-popup > image` means immediate descendant.

PR #519, opened 2018-05-17
Invert Reddit logo text on Reddit redesign (again)

rhencke:
The CSS classes the Reddit redesign uses are apparently not very stable.

As per @risualiser 's comment in https://github.com/darkreader/darkreader/pull/497#issuecomment-389833203, the CSS class changed to `.gPxcKZ`, then as of this moment is yet again changed to `.eeHZFh`.

This attempt takes a (hopefully) more robust approach.  The top navigation bar has the element ID of `header`.  Within it is a link to the main page (`a[href="/"]`), and within this, the SVG logo is contained.  It is split into two elements.  The first, the logo circle does not need inverting.  The second - the text (`svg:nth-child(2)`), does.

For reference, the element we are targeting (and surrounding DOM) looks like such:

screen shot 2018-05-17 at 5 58 27 pm

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -57,7 +57,7 @@ INVERT
 reddit.com
 
 INVERT
-.iAsPXl
+#header a[href="/"] svg:nth-child(2)
 
 ================================
 


alexanderby:
The new reddit uses these fancy styled components. Soon I will add support for URLs in INVERT section.

PR #527, opened 2018-05-23
Add op.gg fixes

odCat:
Fixed the display of Champion Icons/Images in various places on the page, so they are not inverted


alexanderby:
`op.gg` should work as `*.op.gg` too. In future it is planned that `*` will be not considered for alphabetical order, currently it is better not to use it in config.

PR #561, opened 2018-06-13
Indonesian Language for Dark Reader

phenomx8:
Indonesian Language for Dark Reader, sorry my upload request didnt work, it says that its need Push Request so, I use create file instead


alexanderby:
That's great, thanks. It will be published after some bug fixes.

I've updated the doc, maybe you have no permission to upload files, looks strange.

PR #513, opened 2018-05-17
Improve Simplified Chinese (zh_CN) translation quality

WalterMa:
I love the extension！
As a Chinese user, I'd like to improve simplified chinese translation quality.
Hope this helpful.

PR #572, opened 2018-06-18
update zh_TW.config

lcynot:
apply some translation that make more sense


alexanderby:
Thank you. My translation was made via Google Translate from Chinese Simplified https://github.com/darkreader/darkreader/pull/513

PR #584, opened 2018-06-23
Dynamic theme fix for goodreads.com

abdullahau:
Logo of the web-page was wrongly inverted.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -37,6 +37,13 @@ img
 
 ================================
 
+goodreads.com
+
+INVERT
+[class="siteHeader__logo"]
+
+================================
+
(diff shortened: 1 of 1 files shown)


alexanderby:
I have some comments:
1. `[class="siteHeader__logo"]` can be replaced with `.siteHeader__logo`, prefer shorter selector. Sometimes you can see `[class]` is used in config, this trick is used when selector should match `class="a"` but not `class="a b"`. Possible solutions in this case are `[class="a"]` or `.a:not(.b)`.
2. The logo was automatically inverted, but maybe the delay was too long for you. I will add an ability to add image URL instead of CSS selector, so that there will be no need in image analysis.
image

alexanderby:
Looks like not relevant anymore.

PR #592, opened 2018-06-26

checktravis:
top nav bar was not inverting at all, added #gb for support.google.com


alexanderby:
There seems to be some bug with Dynamic mode. I will have to fix it without the need of a config change.

alexanderby:

PR #603, opened 2018-07-05
Fix LaTeX images everywhere

alexalekseyenko:
I realize it's probably best to avoid adding things to the "*" section, but I keep enabling this rule locally using the dev tools and I'm not sure what else to do. I don't want to spam PRs or fill up `inversion-fixes.config` with this same rule for a dozen different sites. This appears to be a very common way to embed LaTeX on a page and I can't imagine inverting `img.latex` is going to break very much (how many non-LaTeX images have a `.latex` class?).

### DIFF
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -20,7 +20,7 @@ NO INVERT
 [style*="background-image: url"] *
 input
 [background] *
-img.latex
+img[src^="https://s0.wp.com/latex.php"]
 
 REMOVE BG
 .compatibility-with-darkreader-below-4-3-3
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -20,6 +20,7 @@ NO INVERT
 [style*="background-image: url"] *
 input
 [background] *
+img.latex
 
 REMOVE BG
 .compatibility-with-darkreader-below-4-3-3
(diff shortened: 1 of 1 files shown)


alexanderby:
I'm not familiar with LaTeX. Maybe we can use `img[src^="https://s0.wp.com/latex.php"]` selector, if websites utilize it in the same way?

alexalekseyenko:
Good thinking---just pushed that change.

CJShearer:
Do changes in inversion-fixes.config, specifically the ones made in af0dde3, affect the dynamic theme engine? If so, this fix does not appear to work on [this page](https://artofproblemsolving.com/wiki/index.php/Binet%27s_Formula).
___
image
___

I have tried placing the following in dev tools (dynamic-theme-fixes.config), but it has no effect.
```

artofproblemsovling.com

CSS
img.latex, img.latexcenter {
    filter: invert(1);
}
```
Could something be overriding my rules? Adding such an invert tag manually with the page inspector works. Perhaps this is related to [#996](https://github.com/darkreader/darkreader/pull/996#issuecomment-492873740)?

alexalekseyenko:
I only use Filter+, so I'm not familiar with the intended Dynamic behavior. @alexanderby thoughts?

Regarding the question of fixing LaTeX in general, it might not be a bad idea to just add
```
img.latex
img.latexcenter
````
to `inversion-fixes.config`. I added those fixes to my local dev tools a long time ago and it hasn't caused any problems for me. The https://artofproblemsolving.com link is yet another example of pages with that pattern.

PR #614, opened 2018-07-10
Fix hackmd.io editor

quilicicf:
Hi, 

I've just discovered hackmd.io and it rocks so I figured it deserves to be beautiful when using dark reader!
I've tested this fix in `Click on dark reader icon > Dev tools` and it looks nice now.

## Before

selection_007

## After

selection_006

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -48,6 +48,13 @@ INVERT
 
 ================================
 
+hackmd.io
+
+INVERT
+.ui-edit-area
+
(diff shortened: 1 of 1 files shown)


alexanderby:
Seems like you are using Filter mode, then you should add this fix into `inversion-fixes.config` file.

Here is how it currently looks in Dynamic mode
image

quilicicf:
I just discovered dynamic mode and it rocks!
I guess this PR has no reason to be anymore, if it's alright with you I'll close it.

PR #624, opened 2018-07-16
Update inversion-fixes.config for codecademy.com

locofocos:
codecademy has a partially dark UI. This makes it entirely dark, at least during the lessons. (There are other areas of the site, like documentation, that I haven't touched)

These are the page elements I made sure to cover:
- Beginning of a lesson before the editor loads, the editor pane on the right
- Code elements (short ones inline with instructions)
- Code elements (larger boxes)
- Side navigation drawer (choose different exercises in the current lesson)
- Top and bottom nav bars
- codecademy logo in the top nav bar

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -7,6 +7,14 @@ INVERT
 
 ================================
 
+codecademy.com
+
+INVERT
+.CodeMirror-cursors
+span[class^="burger"]
+
+================================
+
 docs.google.com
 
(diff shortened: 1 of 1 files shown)


locofocos:
Screenshot:
image

locofocos:
Thinking about this more, the reason I made fixes to the Filter mode (instead of Dynamic mode) is that the cursor didn't look right in Dynamic mode. I'll update this to change Dynamic mode instead. Should be much simpler.

PR #643, opened 2018-07-28
Fix to make Google Drive titlebar dark

krupenja:


alexanderby:
I will have a look into this bug, it is common for some Google sites.

alexanderby:

PR #645, opened 2018-07-28
Added not inverted elements on Google Keep

SciFreakholic:
A couple elements in the sidebar where not inverted correctly in the dynamic mode, so I just added them in the config file.


alexanderby:
There seems to be a bug. I will have to add a `CSS` section into this config, to make it able fix such issues using some static CSS in case if some complex changes are needed.

alexanderby:

PR #653, opened 2018-08-01
Old Reddit title color

sumiflow:
Reddit redesigned their site to look different a while ago but the old design is still an option under a user's preferences.  This inverts the color for post titles when a user has the old design set in their preferences.  It has no effect on the new design.  Without this, post titles are hard to read since they are a very dark blue on a dark background.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -95,6 +95,7 @@ INVERT
 reddit.com
 
 INVERT
+.title
 #header a[href="/"] svg:nth-child(2)
 
 ================================
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -95,7 +95,7 @@ INVERT
 reddit.com
 
 INVERT
-.title
+a.title
 #header a[href="/"] svg:nth-child(2)
 
 ================================


risualiser:
For me, dynamic theme on Reddit [works ok now](https://imgur.com/a/PDWIeRQ), and bad [after your change](https://imgur.com/a/wueYret).

risualiser:
I think that current "dark" blue color of links is better than what are you suggesting (it's a bit sharp also).
But anyway, this decision is better:
`body.with-listing-chooser a.title`
It works only on common pages, not subreddits.

sumiflow:
I see what you mean.  Although my suggestion works for some, you are right that it is worse for many  subreddits.  I didn't notice this because I had "custom themes" in reddit preferences turned off.  For me, every subreddit has that dark blue color on dark background.  Your suggestion of "body.with-listing-chooser a.title" seems to do the same as mine but applied only to the main reddit site.

I wish I knew a CSS selector that could tell whether or not custom themes were turned on but I couldn't find any.  I might be stuck with the dark blue on every subreddit.

risualiser:
Now I can only recommend you to use [this theme](https://chrome.google.com/webstore/detail/material-dark/npadhaijchjemiifipabpmeebeelbmpd) for Chrome.
I think, if you'll make your whole browser dark, these links won't be so dark for you.

sumiflow:
That's a nice theme and I do appreciate the suggestion but unfortunately it doesn't help me any with the font color.
Here it is: [https://imgur.com/a/Ikeu3JW](https://imgur.com/a/Ikeu3JW)

"Filter" mode fixes the readability problem but it doesn't have different colors for visited vs unvisited links.  I guess I'll have to pick my poison.  Thanks for your help.

risualiser:
I think that "Filter+" will be good for you.
filter plus

sumiflow:
Yeah that one is better.  I set it to Filter+ for reddit.com only.

PR #655, opened 2018-08-03
eBay/itm background fix

kaovilai:
This is one way to fix go around background image issue on /itm pages on dynamic theme such as  https://www.ebay.com/itm/ASRock-H270M-ITX-ac-Mini-Itx-MotherBoard/263839765302

An alternative would be to fix add CSS capability to dynamic and hide/add styles such as hiding the background from https://ir.ebaystatic.com/pictures/aw/pics/cmp/ds3/imgbg.jpg

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -24,6 +24,16 @@ INVERT
 
 ================================
 
+ebay.com/itm
+
+INVERT
+#Body
+#gh-gb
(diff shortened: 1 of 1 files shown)


alexanderby:
Sometimes bugs like this are caused by uBlock Origin, it inserts it's own CSS for hiding some ads, but the other extensions cannot see it. Maybe making able to add some static CSS into config will fix the issue.

PR #659, opened 2018-08-06
Czech translation

jy-r:
translation to Czech
with help of @multiflexi


alexanderby:
Great, thanks. It will be published soon.

PR #662, opened 2018-08-08
Add a finished localization file for Romanian.

odCat:
I wan't able to find the context in which @custom_browser_theme_on/off
is used. So I used the masculine in translation, which should be fine.

Also it would be nice if someone else would take a second look. 😄


alexanderby:
Custom browser theme is a new Firefox feature. I'm thinking of removing this feature, or moving this to some advances settings page.
image

alexanderby:
I'm not sure about "@invert_listed_only" and "@not_invert_listed" phrases if they mean that only listed websites will be inverted or will be ignored.

PR #664, opened 2018-08-08
Update dark-sites.config

PizzaLovingNerd:


risualiser:
I don't think that adding personal blogs is a good idea. 
It'd be better for you to just simplify html elements on your site.
But anyway, you can shorten the code:
```

pizzalovingnerd.com
pizzalovingnerd.gitlab.io

INVERT
i

```

PizzaLovingNerd:
Alright I'll remove this and add instructions on how to use darkreader on the site

risualiser:
There are should not be any instructions for any specific sites. It makes extension experience more complicated. You can just make your site dark, so users will be motivated to use Dynamic or Static Theme instead of old Filter (and yes, you can warn your users about this). And maybe developer will be more motivated to make detector of already dark pages.

PizzaLovingNerd:
I'm not going to add the instructions on the site, not as part of the extension. I would make it dark, but it's a static site in hugo, and I couldn't find any css files to edit. My old site hosted on github instead of gitlab was using jekyll, which has css files, and I edited them to be dark.

PR #667, opened 2018-08-11
Update inversion-fixes.config for swappa.com

erockce:
An initial attempt to eliminate default glaring white background color when presenting images by the Light Gallery CSS JQuery library used on the [Swappa](swappa.com) web site.  This does the trick because Darkreader is using default inversion rules for white.  However the side affect makes the navigation controls disappear on the hover event.  I'm struggling finding the correct CSS selector foreground color to modify so open to suggestions.

Test using one of the Swappa listings, [YBB041](https://swappa.com/listing/YBB041/view).

Source CSS [lightgallery.min.css](https://static.swappa.com/static/jquery/lightgallery/css/lightgallery.min.css)


alexanderby:
The domain in config should be the domain name of the website, that is actually displayed, not the domain of the CSS file. So maybe you can just move `CSS` from `static.swappa.com` to `swappa.com`.

erockce:
I admit that I haven't done a lot of front-end web development and having difficulty getting this finished specifically having the right defined color for navigation controls when  browsing pictures.  Continued testing and research indicates that "last defined style wins"[1][2].  What rules does Dark Reader follow to determine how its defined styles are inserted into a page?

References:
* [1][CSS Styles: Inline vs. Internal vs. External](https://cssreset.com/css-styles-inline-vs-internal-vs-external/)
* [2][Does embedded css always override external css?](https://stackoverflow.com/questions/31369919/does-embedded-css-always-override-external-css)

alexanderby:
Sorry, but more than a year has passed. Please submit a new PR if still relevant. Have a look at updated docs.

PR #677, opened 2018-08-14
Show Dark Sky element for current condition

apaulo11:
Added darksky.net exception to show animated element to the left of current conditions.


alexanderby:
`input` is a too generic selector. Maybe there is a bug with Dynamic mode, so you can open an issue or wait until I add a `CSS` property into config soon.

PR #683, opened 2018-08-16
Created it.config

giorgiogamberini:
Italian translation for the extension up for review.
Already built with gulp and tested, seems to fit well along the layouts.


alexanderby:
Grazie!

PR #685, opened 2018-08-16
Add finishing touches to the Ro translation

odCat:


odCat:
A summary of the changes:
* **Fix the invert_listed strings.** Changed the words a little, but the meaning remains; considering the space constrains.
* **Change the @setup_add_site_hotkey string to make it sound better.**
* **@custom_browser_theme strings changed to feminin.** Checked on Firefox and there should be space for one more character

PR #735, opened 2018-09-19
Create `ar.config` for Arabic translation

atefBB:
Work in progress !


alexanderby:
Sorry, but about a year has passed. Please open a new PR if you would like to create a translation. Many new labels where added. And also app redesign is planned, many labels are going to be changed.

PR #755, opened 2018-10-07
Updated dark-sites to add draftkings.com

nitrocode:


nitrocode:
@alexanderby does https://github.com/darkreader/darkreader/blob/master/src/config/dark_sites.json also need to be updated? why is there a `dark_sites.json` and `dark_sites.config` ?

alexanderby:
@nitrocode This files are there for historical reasons. One day I decided that storing site list in a plain file is simpler than JSON. These files are still used by [Dark Retired](https://chrome.google.com/webstore/detail/dark-retired/oibheihomapbjogmoabgfbkchjchpdfp) extension, but will be removed soon.

nitrocode:
@alexanderby I see where you're coming from. It would be nice to note this as a comment in the json so others aren't confused.

```
    "_CURRENTLY_DEPRECATED_AND_ONLY_USED_FOR_DARK_RETIRED.com",
    "_UPDATE dark_sites.json INSTEAD",
```

PR #762, opened 2018-10-11
Translate "dark" to "escuro" in pt_br

joaocv3:


joaocv3:
It also adds consistency, since we call it "modo escuro" in the `@store_listing` session, which is correct.

PR #800, opened 2018-10-29
improve dynamic theme for postgresql.org

splanquart:
Postgresql is a popular database system.


alexanderby:
I was in hurry, I thought this is a Filter config. For Dynamic mode `INVERT` should be used only for small elements. Here is a bug, some style element was not analyzed for some reason.

alexanderby:
This issue was already fixed yesterday, same problem happened for Google Drive and many other websites. I will publish the fix tomorrow.

PR #814, opened 2018-11-03
Fix for logo and More arrow icon for BBC

krupenja:
Note that using `bbc.com/*` is not posssible as it breaks the logo at bbc.com/home and bbc.com/reel

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1,3 +1,22 @@
+bbc.com/news
+bbc.com/sport
+bbc.com/weather
+bbc.com/travel
+bbc.com/capital
+bbc.com/autos
+bbc.com/culture
+bbc.com/future
+bbc.com/sounds
+bbc.com/food
+bbc.com/bitesize
+bbc.com/earth
+
+INVERT
+.orb-nav-section.orb-nav-blocks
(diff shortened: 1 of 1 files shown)


alexanderby:
OK, let it be. I think it will be improved when we finally rework URL templates, so writing `bbc.com !bbc.com/reel` will make it shorter.

rolagg:
Suddenly, this fix inverts logo incorrectly and the site itself automatically gets successful dark theme without any config.
My settings: Debian 9.6/Firefox 64.0b14

With fix:

image

Without:

image

But if it's just on my side, I'd recommend to save some space in config file by using regexp:
```bbc.com/(news|sport|weather|travel|capital|autos|culture|future|sounds|food|bitesize|earth)```

krupenja:
Thanks, did not know that Darkr Reader Dev tools support regex!

This looks like something browser-specific. The BBC logo looks OK in Chrome with this fix but I have now tried in Firefox and there indeed it looks OK without the fix (and broken with the fix).

I guess it would be better for the maintainer to revert the fix unless there is a way to implement browser-specific rules.

alexanderby:
The dev tools are not supposed to support RegExps, it is a side effect. Maybe patterns like this will be allowed, but still the work on it is not started.

It looks like some difference in image analysis between Chrome and Firefox.

PR #816, opened 2018-11-04
double-invert images in docs.google.com

stylehouse:
there may be a non-two-step way to do it.

see https://docs.google.com/document/d/1mPFvwqeS6NK2hb_7JKNusttIO5C9MlGyBntMA4qawRs/edit?usp=sharing for an image.

### DIFF
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -302,6 +302,14 @@ INVERT
 [aria-label^="Google Account"]
 div[title="Profile"]
 
+CSS
+.kix-embeddedobject-view svg {
+filter: invert(100%) hue-rotate(180deg) !important
+}
+.kix-embeddedobject-view svg image {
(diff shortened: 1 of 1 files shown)


src/config/inversion-fixes.config
```diff
@@ -302,6 +302,14 @@ INVERT
 [aria-label^="Google Account"]
 div[title="Profile"]
 
+CSS
+.kix-embeddedobject-view svg {
+filter: invert(100%) hue-rotate(180deg) !important
```
alexanderby: Please add these selectors under `INVERT` rule, it will produce the same CSS code.

stylehouse:
hmm indeed. that's better, but both seem to work only once you've recently applied the rules (even without change) via the INVERSION FIX EDITOR, something slightly beyond the page's lifetime. Festival time, back in a week or so.

alexanderby:
Hi, please open a new PR if still relevant.

PR #821, opened 2018-11-05
Update for hangouts.google.com background

vaindil:


src/config/inversion-fixes.config
```diff
@@ -696,6 +696,9 @@ CSS
 .g-Qx-r4m2rf-wZVHld {
     display: none;
 }
+.g-Qx-eb {
+    background: rgb(255, 255, 255) !important;
```
alexanderby: There is `REMOVE BG` rule for this, it will produce the equivalent CSS code.

vaindil: Good point, my bad. Fixed, thank you!

PR #822, opened 2018-11-05
fixed wrong inversions on last.fm

Cerbereon:


alexanderby:
Type selectors (like `div` or `p`) are not necessary when there are class selectors (like `.masterhead-logo`) due to [CSS specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) rule.

Anyway it's a minor issue, so I'm merging your work, thanks.

Issue #660, opened 2018-08-07
Subdomain of sites in global dark list may not be dark


yodahuang:
We need a more powerful syntax to specify what sites are truly "dark", just a domain names isn't enough.

An example would be [ffmpeg.org](https://ffmpeg.org). Its main site is dark, but [its Wiki (trac.ffmpeg.org/wiki)](https://trac.ffmpeg.org/wiki) is bright.

To make things worse, this setting can not be toggled by user since it's hard-coded in the global dark list. Hope there's an elegant solution for these situation.

afarah1:
I'm having the same issue with trac.ffmpeg.org. I came here looking for a way to remove a website from this "global dark list" or at least understand what this is but found no documentation, just this issue.

Oussebon:
The Humble store and Fanatical are both on the global dark list for me - despite the Humble store always having a bright background, and Fanatical's blog having a white background (unlike the main store page which is naturally dark grey) https://www.fanatical.com/en/blog

Hubro:
I just Googled a question on https://faq.i3wm.org and the website is *insanely* bright, but it's in the dark list so I can't enable dark mode. This should be overridable by the user.

val-litvak:
I am having the same issue with pro.coinbase.com and docs.pro.coinbase.com 

pro.coinbase.com is dark normally, so I added it to the dark list, but I wanted to make docs.pro.coinbase.com dark(it is white and there is no way to change the theme) but it seems like it is not possible with the current implementation.

val-litvak:
Thanks @alexanderby ! Your solution would normally work, but because `pro.coinbase.com` is in global dark list, it did not work for me, please see my PR which fixes the problem(tested locally): https://github.com/darkreader/darkreader/pull/830

metalim:
Huge Ping on this one!
It's awful to see Dark Reader being disabled on many subdomains, because you "tried hard" to make it work on main domains of those sites. https://support.discordapp.com/ burns my eyes.

camdenorrb:
I'm also experiencing this with https://trac.ffmpeg.org/wiki

Feez:
https://www.epicgames.com/fortnite/en-US/patch-notes/v8-10
dark btw

Programatic:
[...]

PR #830, opened 2018-11-09
fix-pro-coinbase-com-allow-docs-to-be-dark references #660

val-litvak:
Please see my comment in https://github.com/darkreader/darkreader/issues/660 for more info on why I am submitting this PR.

It turns out that in the current implementation it is not possible to set `docs.pro.coinbase.com` to be dark, because `pro.coinbase.com` is in "global dark list".

@alexanderby suggested that I try `^pro.coinbase.com` which did not work for me when I added it to my dark list via the extension UI.

However, I did some digging and realized that in the actual codebase there is a list of sites that are marked as dark by default, and it had `pro.coinbase.com`. Changing it to `^pro.coinbase.com` fixed it.

If someone wishes, they would still be able to add `docs.pro.coinbase.com` to dark list (if for example they preferred the docs to be in light rather than dark).

I am also wondering whether it would be good to add a way to override (in this case I needed a way to remove) items from the dark list.

Also, I had to put the ^pro.coinbase.com in the very beginning of the config file because the test ensures that it's alphabetized.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -1,3 +1,4 @@
+^pro.coinbase.com
 alienwarearena.com
 animeonline.su
 app.getpocket.com
@@ -89,7 +90,6 @@ poeapp.com
 poelab.com
 pr0gramm.com
 primevideo.com
-pro.coinbase.com
 raidtime.net
 realgamernewz.com
 runicgames.com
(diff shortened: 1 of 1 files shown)


alexanderby:
I decided to not use special symbols in this list, maybe sorting of such symbols will be improved in future.

AlexByte:
Add option to disable this list/

alexanderby:
It is planned.

PR #839, opened 2018-11-13
added duckduckgo

Highom:
duckduckgo has it's own dark theme


rolagg:
But not by default.

alexanderby:
You are right, we only add fully dark websites to this list.

PR #856, opened 2018-11-21
Added inversion for charts in Jira

maik-kulbe:


alexanderby:
Hello, could you please explain, why `jira.*`? Does it match some local address?

maik-kulbe:
@alexanderby IME most non-cloud Jira-Server/-DC instances use the domain-scheme jira.example.org or jira.tld.

alexanderby:
OK, I planned to add some local fixes tool, but seems like there are a lot of important work to do at the moment, so it will not be added soon. Did your fix work in Dev Tools? I'm afraid it can work for `jira.company`, but will break on `jira.company.com`. Even if it works now, it is unintentionally, and it could be broken after URL templates refactoring.

So, I'm adding `jira.*.*` too.

PR #863, opened 2018-11-27
Downgrade event-stream to 3.3.4 and remove flatmap-stream@0.1.1 dependency

cfal:
flatmap-stream@0.1.1 steals crypto coins and is a dependency of event-stream 3.3.6: https://www.theregister.co.uk/2018/11/26/npm_repo_bitcoin_stealer/

One possible course of action here is to downgrade to 3.3.4 which does not have the flatmap-stream as a dependency: https://github.com/dominictarr/event-stream/issues/116

Tested that 'gulp release' still builds.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1764,19 +1764,17 @@
       }
     },
     "event-stream": {
-      "version": "3.3.6",
-      "resolved": "https://registry.npmjs.org/event-stream/-/event-stream-3.3.6.tgz",
-      "integrity": "sha512-dGXNg4F/FgVzlApjzItL+7naHutA3fDqbV/zAZqDDlXTjiMnQmZKu+prImWKszeBM5UQeGvAl3u1wBiKeDh61g==",
-      "dev": true,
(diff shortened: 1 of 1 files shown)


alexanderby:
Thank you for the info. I used `gulp-connect` package (sponsored by Jet Brains), which had `event-stream` as a dependency, for live-reloading of a separate static page during HTML & CSS coding. Looks like it stopped working since locales were added in June, so I decided to simply [remove](https://github.com/darkreader/darkreader/commit/106a796c2e3acf5d37b782fe35d77c35726010eb) the `gulp-connect`.

If you run `npm audit` you will see some high vulnerabilities, but they are not actually vulnerabilities here. Anyway I will remove `gulp` or at least update it up to `4.0.0`.

alexanderby:
So, I [replaced](https://github.com/darkreader/darkreader/commit/5bd0cf2e168af7d871875a7e89c8ab9369ef9c4e) Gulp with Vanilla JS using as few packages as possible.

PR #869, opened 2018-11-29
Added messages.android.com, broadened discordapp.com in dark-sites.config

adinklotz:


rolagg:
I'd not recommend to ignore the whole Discord site.
It's better to use this filter 
`discordapp.com/(activity|library|store|channels)`
to not affect pages like this:
https://support.discordapp.com/hc/en-us
https://discordapp.com/jobs
https://discordapp.com/open-source

adinklotz:
Makes sense -- I didn't realize those other pages were still white by default.

Issue #873, opened 2018-12-01
issues in the display of GitHub profile 


gilbertohasnofb:
In dynamic mode, the contributions grid in an user profile becomes impossible to read. The colours also do not correspond to the scale on the bottom right:

![image](https://user-images.githubusercontent.com/5700795/49329662-5d019480-f57a-11e8-8cb7-572592cda517.png)

Static mode does a bit better with the grid, but makes the rest of the page looking buggy:

![image](https://user-images.githubusercontent.com/5700795/49329683-84f0f800-f57a-11e8-9498-f58aab99ff10.png)

![image](https://user-images.githubusercontent.com/5700795/49329686-9508d780-f57a-11e8-9f27-2180e9a78bfd.png)

Filter mode makes the grid look quite better, except that the scale becomes inverted, the lighter the green the more contributions an user made in a day:

![image](https://user-images.githubusercontent.com/5700795/49329696-c2ee1c00-f57a-11e8-86a0-108f4b6ef96d.png)

gilbertohasnofb:
@krupenja you are right, my apologies. I am closing this and adding the extra info to that opened issue.

PR #875, opened 2018-12-02
[Dynamic] Make Github contribution grid dark

rolagg:
Partial fix for #873.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -104,6 +104,13 @@ img
 
 ================================
 
+github.com
+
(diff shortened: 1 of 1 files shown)


alexanderby:
I think I should treat SVG as a single image, instead of analyzing separate elements. Current algorithm thinks that every cell is a separate icon, so it can cause issues like this:
screen shot 2018-12-06 at 11 29 46

alexanderby:
A better fix was suggested https://github.com/darkreader/darkreader/pull/1701

Analyzing SVG as a whole image is still undone.

PR #878, opened 2018-12-03
Time settings

### LINKED ISSUE #10: Time settings
I recently switched to Dark Reader because I've found that it's generally better about inverting the colors consistently.

One feature I miss from the extension I had been using ([Night Reading Mode](https://chrome.google.com/webstore/detail/night-reading-mode/chcciiimjmlgcoabgfdhkfjdcgfpgndi)) was the ability to have dark mode "On", "Off", or "Auto", the last of which would turn dark mode on during a range of times (i.e., at night). It also let you override the above setting per-domain.

As an example, you could have it Night Reading Mode to "Auto" (so it'd become dark from maybe 7:00 p.m. to 7:00 a.m.), but have something like hslpicker.com be always regardless "Off".

alexanderby:
- Ability to set Dark Reader active hours.
- Fixes #10.
image

### DIFF
--- a/src/_locales/be.config
+++ b/src/_locales/be.config
@@ -4,6 +4,7 @@
 @loading_please_wait
 Ідзе загрузка, пачакайце
 
+
 #==== Top section ====
 
 @on
@@ -28,6 +29,12 @@
 @setup_hotkey_toggle_extension
 Уст. спалучэнне
 клавіш
+
+@time_settings
+Налады часу
+
+@set_active_hours
+Усталюйце перыяд актыўнасці
 
 @page_protected
 Гэтая старонка ахоўваецца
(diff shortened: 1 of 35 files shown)


sypl:
Why is this feature hidden behind a button that only appears on hover? Nobody would ever find that unless they read the blog. Seems an on/off/timer option set would be ideal?

alexanderby:
@sypl You are right, this is implemented in Safari version.
screen shot 2019-01-22 at 16 22 44

But this requires redesign of the app, there are already too much settings to show. It is planned to do, but currently I work on other problems.

sypl:
Ah I see, understood. Thanks. Great extension!

PR #881, opened 2018-12-04
Add rules for feedly.com

cewilliams:
Invert article listing and opened article


alexanderby:
I don't see any problem with Feedly. Could you please add a screenshot or steps to reproduce the problem?

PR #902, opened 2018-12-10
Update dark-sites.config

wedemmoez:
added imgur since it is already dark


alexanderby:
There were many discussions, Imgur sometimes can be white.

PR #910, opened 2018-12-12
Update dark-sites.config

Baine:
* added 9gag.com, since it offers a dark mode switch now


alexanderby:
This website is white by default, but Dark Sites list should include websites that are always dark.

Baine:
alright, my mistake then. will close this request in that case

PR #949, opened 2018-12-24
Invert facebook.com chat bar when collapsed

Tansc161:
Before: 
before

After: 
after


alexanderby:
They are using a 1px sprite as a background image in this area, so I will add support for custom CSS in config soon to remove it.

PR #951, opened 2018-12-25
Create some custom inversion for search results in bing.com

Aaron19960821:
bing.com creates some useful widget for better search results. Currently darkreader is not supporting them well. See the following screenshot:
image
We provide some fix for these widgets:
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -26,6 +26,19 @@ img[src^="/graphs/"]
 
 ================================
 
+bing.*
+bing.*.*
+
+INVERT
+.b_cnvsug
+.b_focusLabel
+.b_richcard
+.cbtn
+.trans_button
+.nosmab
+
+================================
+
 chtoes.li
 
 INVERT


alexanderby:
Could you please check, if Bing works well without your fix? There could be some issue with not fully loading the stylesheet for analysis. Currently everything works well for me.

Aaron19960821:
> Could you please check, if Bing works well without your fix? There could be some issue with not fully loading the stylesheet for analysis. Currently everything works well for me.

The Bing has 2 versions, local and international version. I have checked and international version have some problems, see the first screenshot I attached. And you can try the following link: 
https://www.bing.com/search?q=dns&qs=n&FORM=BESBTB&sp=-1&pq=&sc=0-0&sk=&cvid=9A996AF2B79A479B96ACEE0232D19AF5&ensearch=1

PunyFlash:
The video and image search seems to be broken too
image
image

alexanderby:
OK, I see the bug with image results. Such large areas should not be inverted by filter, so I think I should have a look what's wrong.

PunyFlash:
Also found casual search to be broken (like on Aaron's screenshots) on MacOS. On Windows it works well

Aaron19960821:
> OK, I see the bug with image results. Such large areas should not be inverted by filter, so I think I should have a look what's wrong.

Ok, I will also take a look. Please tell me if there's something I can help.

kaovilai:
Reporting additinal link with widget issues.
Only text in the widget is black excluding the link right below widget.
https://www.bing.com/search?q=best+pdf+editor&PC=U316&FORM=CHROMN

specifically class df_c rwrl has css applied to it not being inverted.
Chrome dev tool found this. Unticking the color under here fixed the widget of text extracted from techradar.com

.df_c .rwrl, #b_results .df_c .b_entityTitle {
    color: #111;
}

PR #959, opened 2018-12-28
White background for google search shopping items

JoelBalmer:
Heya! Just noticed that the images from google search shopping results weren't showing up well because of the black background.

I tried to fix with the dev tools but wasn't able to apply custom css (or REMOVE BG or NO INVERT features either, for that matter). I just played around in chrome dev tools and was able to get the desired result with the css in this commit. See an example of what I'm trying to fix in the below image.

Cheers!

dark_reader_google_img

### DIFF
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -627,6 +627,9 @@ CSS
 .eB {
     background: rgba(255, 255, 255, 0.4) !important;
 }
+.pla-unit img {
+    background: rgba(255, 255, 255, 0.7) !important;
+}
 
 ================================
 


alexanderby:
`inversion-fixes.config` is used by Filter mode. For Dynamic mode `dynamic-theme-fixes.config` file should be edited. Currently it doesn't support custom CSS, but it will be added in the meantime.

JoelBalmer:
Ah cool, so if i've understood right, the fact that custom css isn't supported is only for testing purposes (i.e. my changes on `inversion-fixes.config` will still be applied correctly if my PR is merged and deployed)? Thanks!

alexanderby:
@JoelBalmer I see you are using Dynamic mode on your screenshot, so changes for it should be made in `dynamic-theme-fixes.config`. If you submitted these changes for Filter mode, then I will merge them (`inversion-fixes.config` supports custom CSS).

JoelBalmer:
Brilliant thanks, yes they were made on inversion fixes config

alexanderby:
Hi, a long time has passed. Please submit a new PR if still relevant.

PR #971, opened 2019-01-04
Add Dutch translation

Galaxyzd:
Translated fully by me, not perfect but it should work.


alexanderby:
Thank you, I had to shorten some phrases, maybe after redesign there will be more space for them.

PR #978, opened 2019-01-06
Remove humblebundle.com from dark sites

kraxarn:
Humble Bundle uses a lot of light themes, for example in game info and profile view.

### DIFF
--- a/src/config/dark_sites.json
+++ b/src/config/dark_sites.json
@@ -35,7 +35,6 @@
     "hangouts.google.com",
     "hardcoregaming101.net",
     "heavybit.com",
-    "humblebundle.com",
     "hyper.is",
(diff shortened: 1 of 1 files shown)


alexanderby:
@kraxarn Looks like it was removed from a wrong file. `dark_sites.json` is still used by [Dark Retired](https://chrome.google.com/webstore/detail/dark-retired/oibheihomapbjogmoabgfbkchjchpdfp) extension for older browsers, some day I will remove it or move to another folder, but code changes for old version are needed. I've just added a warning into the JSON file, to reduce the chance of mistakes. Please, use `dark-sites.config` file, it has a plain text format.

alexanderby:
Another user just fixed it https://github.com/darkreader/darkreader/pull/1028

PR #990, opened 2019-01-09
Add Scratch 3.0 Editor styles

TheV360:
This fixes the block selection pane, making the text readable and making the entire UI look really nice.


alexanderby:
One note: you could use `${color}` template so that brightness and contrast could be applied to color. See more info here https://github.com/darkreader/darkreader#using-dev-tools

PR #996, opened 2019-01-11
Fix Latex elements on wikibooks

adrian-valente:
For example here: https://en.wikibooks.org/wiki/LaTeX/Mathematics


CJShearer:
It appears this patch is broken, at least in Firefox.
___
 Annotation 2019-05-15 200752
___
Reversing changes in commit bb21559 appears to leave the page in the state originally desired.
___
1t4ydccz bmp
___
Would you check [that wikibooks ](https://en.wikibooks.org/wiki/LaTeX/Mathematics)page again and see if it still looks right? I think recent changes to the dynamic theme engine may have made bb21559 obsolete.

PR #1012, opened 2019-01-18
Add anilist.co to dynamic-theme-fixes.config

mgziminsky:
Use the sites dark theme as default

Copied the variable values for the sites native dark theme, which isn't the default theme for the site.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -7,6 +7,23 @@ CSS
 
 ================================
 
+anilist.co
+
+CSS
+:root {
+    --color-background: 39,44,56 !important;
+    --color-foreground: 31,35,45 !important;
+    --color-foreground-grey: 25,29,38 !important;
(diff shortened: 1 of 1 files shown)


alexanderby:
Did you try it in Dev Tools? `39,44,56` is not a valid CSS color, but `rgb(39, 44, 56)` is. Also you could use `${LightColor}` template so that brightness and contrast would be applied to color and it will be bright in light mode. But if you find this case rare and too tricky, you can leave dark colors.

mgziminsky:
I did try it. The site uses variables to define its theme colors, and those are valid variable definitions. I didn't come up with these rules, they are directly from the sites own stylesheet. The site just doesn't use the dark theme by default

alexanderby:
Great, I'm not sure if `!important` could be used with CSS variables, but if you say it works, then let it be.

mgziminsky:
Yeah, I don't actually know for sure about the !important, but it doesn't break anything so I thought I might as well add it.

PR #1015, opened 2019-01-18
DDG has a built in dark mode

theLMGN:


alexanderby:
By default Duck Duck Go is bright, but this list should contain fully dark websites. It is important for Filter mode, cause it inverts dark websites into white, which can cause much frustration.

PR #1019, opened 2019-01-20
Update dynamic-theme-fixes.config. Fix for vk.com.

swit33:
Added a fix to vk.com's online icon.

I don't really shure is it necessary to add this, because it's a Russian social network, and not so manu users of this extencion use it. But i made this one for myself and maybe it can help.


swit33:
I'll close reques for now
I'm only getting used to GitHub, so i reqested too early.
I'll find all problems that needed to be fixed and then i'll make a pull request/

PR #1021, opened 2019-01-21
added indiexpo.net

Leone25:
added indiexpo.net and youtube.com because they are already dark or they have a dark mode


alexanderby:
YouTube is white by default, so I removed it from the list.

PR #1023, opened 2019-01-21
Added reddit.com 

f1nnl3strange:
reddit has a dark mode so I added it


alexanderby:
By default Reddit is white, but this list should include only fully dark websites. YouTube is also white by default and Creators Studio has only white theme.

PR #1026, opened 2019-01-21
removed discordapp.com

Galaxyzd:


alexanderby:
I'm pretty sure somebody will add it again today :)

PR #1039, opened 2019-01-24

leizhuchem:
default chrome highlight color is difficult to recognize under dark theme. changed color to orange background with white font color.


alexanderby:
I think I understand what's the problem with selection colors. Maybe I could get `::setection` part from the selector and increase lightness for such backgrounds.

Gusted:
#2470

PR #1042, opened 2019-01-25
Adding imgur to dark sites

koyuawsmbrtn:


alexanderby:
Currently there is a problem, when people open images or gifs in new tab: they become having white background.

koyuawsmbrtn:
Oh that's bad. I'll close this then.

PR #1051, opened 2019-01-27

waterdrinkerAP:
Some elements on this site were displaying white text on bright yellow background, making the text unreadable.
I tried fixing it in various other ways, but this it the only thing that worked.
It is not perfect, but at least the text is now readable.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -370,6 +370,15 @@ INVERT
 
 ================================
 
+online.rbb.bg
+
+INVERT
+.has-context:hover
+.active
+.btn
+
+================================
+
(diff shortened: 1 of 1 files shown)


alexanderby:
While I merged it, there could be better ways like using recently added custom CSS support like:
```
CSS
.light-background-color {
    background-color: ${black} !important;
}
.light-background-image {
}
```

waterdrinkerAP:
CSS was the first thing I tried. But even with the !important the site was overwriting it somehow. To be honest, this site is doing some stuff way beyond my understanding. The elements I wanted to change were not handled by a CSS file. I think they were styled with javascript somehow. Anyway it is just a dirty fix to at least make the stuff readable. Thanks for adding it.

PR #1057, opened 2019-01-29
Should be a standard img darkerner & of course hover to see original …

ZNosX:
Should be a standard img darkerner & of course hover to see original img function~

Simple yet I think could help ppl who always got blinded by all those shining imgs @ nite.


alexanderby:
Nice idea, but needs some time and thoughts before being merged. The latest release will make dev tools changes be applied permanently, so you will be able to use it.

Issue #630, opened 2018-07-21
Vim-Vixen no invert on hit boxes


wviana:
Hello, I'm a darkreader user and also a [vim-vixen](https://github.com/ueokande/vim-vixen/)

When I press `f` on Vixen I get a some hits boxes.

So I got the Vixen Source code and found this class `.vimvixen-hint`. I tried to place a `NO INVERT` in the *dev tools* with an `*` , but I think there is no way to apply a `NO INVERT` to all sites. Is there a URL for extensions ?

Could you help me to place this rule?

robsmith11:
I am also looking for a solution to this.

On a related note, have you gotten "static" mode to work with Vim Vixen?  For me it just blacks out the entire screen.

wviana:
Does someone could give us some tip of where does the config file is interpreted. Looked around but couldn't figure out.

nonetrix:
I am still having this issue

PR #1058, opened 2019-01-29
Compatibility with link hinting of VimVixen add-on

Vis5:
In dynamic night mode, inversion of VimVixen's link hinting is not recognizable. This change overrides the inversion to default look on all sites.
Issue here: #630.


Vis5:
I see. Problem with color template was that hint-boxes looked completely unrecognizable in dark mode. But, to get in line with the coding style and various brightness contrasts, I changed the default color codes (yellow, black, gold) to something more distinguishable. I think it should be okay now.

PR #1066, opened 2019-01-30
Fix content in 01net.com

adrian-valente:
Fix text in titles and content in 01net.com (popular french language tech website).
Better implementation of #1064 

Before:
01net1

After:
01net2

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -9,11 +9,10 @@ CSS
 
 01net.com
 
-INVERT
-.titre-article
-.blocx3
-.bloc p
-.display-inline-block
+CSS
+html, body {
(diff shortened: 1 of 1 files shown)


alexanderby:
Strange bug happens on this website. White your fix works, it is not recommended to invert such large areas. Maybe some solution like this can help?
```
CSS
.bloc p, .titre-article {
    color: ${black} !important; // will become `white` in dark mode
}
```

alexanderby:
I think yes, this is a bug. Dark Reader style becomes empty for some reason.

alexanderby:
The reason was that many texts have `color: inherit`, but `<html>` and `<body>` elements also have their color `inherited`, but they are root elements and cannot inherit a color.

Issue #1056, opened 2019-01-28
support Static filter syntax for dark-sites.config


wis:
[Static filter syntax](https://github.com/gorhill/uBlock/wiki/Static-filter-syntax)
[dark-sites.config](https://github.com/darkreader/darkreader/blob/d9e9be642be13aa913974573b16bb899c79c7226/src/config/dark-sites.config#L68) has `i3wm.org`, there needs to be a way to exclude sub domains. `faq.i3wm.org` should be excluded.

wis:
I think you're trying to reinvent the wheel with the making of a new spec, library and syntax instead of using [Static-filter-syntax's static-extended-filtering](https://github.com/gorhill/uBlock/wiki/Static-filter-syntax#static-extended-filtering) which's used by the most popular ad blocking extensions including ublock origin which has a very useful element picker feature that you can borrow.

wis:
I think you're trying to reinvent the wheel with the making of a new spec, library and syntax instead of using the established [o](https://adblockplus.org/en/filters) [n](https://github.com/gorhill/uBlock/blob/1de821d99b4d2fa9756fdd451823e0886cef3534/src/js/static-ext-filtering.js) [e](https://adblockplus.org/filter-cheatsheet). 
[Static-filter-syntax's static-extended-filtering](https://github.com/gorhill/uBlock/wiki/Static-filter-syntax#static-extended-filtering) are be used for hiding elements by the most popular ad blocking extensions including ublock origin which has a very useful element picker feature that you can borrow.

wis:
I think you're trying to reinvent the wheel with the making of a new spec, library and syntax instead of using the established [o](https://adblockplus.org/en/filters) [n](https://github.com/search?q=Filtering+repo%3Agorhill%2FuBlock+path%3A%2Fsrc%2Fjs+language%3AJavaScript&type=Code) [e](https://adblockplus.org/filter-cheatsheet). 
[Static-filter-syntax's static-extended-filtering](https://github.com/gorhill/uBlock/wiki/Static-filter-syntax#static-extended-filtering) are used for selecting elements to hide by the most popular ad blocking extensions including ublock origin which has a very useful element picker feature that you can borrow.

alexanderby:
Thank you for the samples. The main goal is to make it simple to use.

alexanderby:
Subdomains should not be affected now (only `www` will be affected).

PR #1080, opened 2019-02-05
add 'imgur.com' to dark-sites.config

wis:


wis:
oh sorry. 
seems to be Blocked on: #373 and #1056

Issue #1007, opened 2019-01-16
Github inline code background


ricab:
Inline code in github markdown renderings originally has light grey background. With dark reader, that background is lost: `some inline code`.

Chekote:
Here's a couple of screenshots illustrating the problem:

<img width="799" alt="screen shot 2019-02-27 at 10 25 40 am" src="https://user-images.githubusercontent.com/550486/53483793-2f918d80-3a7a-11e9-80b7-0dc9076a4aa5.png">

<img width="800" alt="screen shot 2019-02-27 at 10 25 55 am" src="https://user-images.githubusercontent.com/550486/53483799-33bdab00-3a7a-11e9-835b-b7e6a5a6d867.png">

PR #1081, opened 2019-02-06
Fix for GitHub Markdown inline code (Issue #1007)

italodirenzo93:
Proposed fix for #1007

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -235,8 +235,10 @@ INVERT
 CSS
 .markdown-body code,
 .markdown-body pre {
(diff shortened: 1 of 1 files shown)


alexanderby:
I'm not sure if it's OK to force light background and dark text in dark mode. I tried this solution so that code blocks look similar in dark and light mode:
```
CSS
.markdown-body pre,
.markdown-body code {
    background-color: ${rgba(27, 31, 35, 0.1)} !important;
}
```

italodirenzo93:
I wasn't very keen on it either. Going by the issue description, I thought that was what was desired. I decided to use your solution but added a rule to preserve the transparent background of code blocks inside of <pre> elements.

PR #1086, opened 2019-02-07
Update dark-sites.config epicgames.com is dark.

m4gnify:
https://www.epicgames.com


Logerfo:
The site is not not completely dark, see it for yourself: https://www.epicgames.com/fortnite/en-US/patch-notes/v8-00. This PR should be reverted.

PR #1111, opened 2019-02-17
Add discordapp.com to dark sites

alexalekseyenko:
Looks like the Discord web app has a dark theme by default


alexanderby:
There were many Pull Requests, that were adding and removing Discord. Some web pages are white, so this website cannot be included into this list. Maybe some day an option will be added to ignore the Dark Sites list.

PR #1112, opened 2019-02-17
Update dynamic-theme-fixes.config for web.archive.org

LentUsername:
Added entry for web.archive.org
Made timeline readable in dark mode
Made circles on the homepage calendar visible
Inverted logo

Timeline
before:
from

after:
after2

Homepage
before:
hpb

after:
hpa

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -573,7 +573,9 @@ INVERT
 #wm-logo
 
 CSS
-.measure{z-index:0!important}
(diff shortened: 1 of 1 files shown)


LentUsername:
I also could not determine why some of the circles are invisible. Perhaps there's a bug? It appears to be affected by page size.
Page 570x937, extension Off:
off
extension On:
on
This is different from larger page size on screenshots in the first comment.

I resorted to overlaying circles on top of text which slightly tinted it.

PR #1134, opened 2019-02-28

SirLT:
Fix alphabetical order of previous pull request.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -569,6 +569,15 @@ INVERT
 
 ================================
 
+webhallen.com
+
+CSS
+.drop-countdown-time {
+    COLOR: rgb(23, 30, 54);
+}
+
+================================
+
 what-if.xkcd.com
 
 INVERT


alexanderby:
Hi. You could edit your previous Pull Request, if you found an error.

Please, pay attention, that the fix should work well in Dark and Light mode. So maybe you should replace `COLOR: rgb(23, 30, 54);` with the initial color, placed in special template `${}`, like `color: ${rgb(23, 30, 54)} !important;`. `!important` ensures that this rule will not be overriden by other rules.

src/config/dynamic-theme-fixes.config
```diff
@@ -569,6 +569,15 @@ INVERT
 
 ================================
 
+webhallen.com
+
+CSS
+.drop-countdown-time {
+    COLOR: rgb(23, 30, 54);
```
alexanderby: Please use `color` in lower case, use `${rgb(x, x, x)}` to ensure fix works well in light and dark mode, and add `!important` keyword.

alexanderby:
Hi, a long time has passed. Please submit a new PR with requested changes if still relevant.

PR #1157, opened 2019-03-08
Youtube video title color change

Abragus:
Fixed issue where the video title in Youtube would be yellow.


alexanderby:
Please, pay attention that `${black}` color should be used, so that right color is shown in light mode and brightness and contrast are applied properly. But the problem is that YouTube still look black in light mode due to some bug.

PR #1163, opened 2019-03-12
Fix FFMPEG's wiki being bright?

camdenorrb:
I haven't tested this at all, but shouldn't this allow that specific sub domain to be darkified via Dark Reader? If not please tell me what to change in order for this to work.


alexanderby:
It is planned to do this by `!` symbol, but not supported yet.

PR #1166, opened 2019-03-13

coch110149:
medium.freecodecamp.org highlighting was unreadable. Fixed using one of their specified colors that highlights to a darker color that is easier to read with dark theme


alexanderby:
Please check how fix works in Light mode, maybe `${color}` template should be used.

PR #1177, opened 2019-03-16
viduro.xyz has a dark mode.

Troplo:


alexanderby:
Hi, the website in this list should be fully dark by default. Also alphabetical order is wrong.

PR #1180, opened 2019-03-17
Fix background and message icon for soundcloud.com

nasso:


alexanderby:
Hi, please pay attention that fix should work in light mode too, so `${color}` template are used. Also `!important` keyword should be used to prevent overrides.

PR #1188, opened 2019-03-19
Updated dynamic-theme-fixes.config YouTube's Copyright page

TheAlienDrew:
Fixed the background SVG (changed color to black and converted it to URI to use)
Fixed the logo that would appear above this svg by inverting it.


alexanderby:
Is it possible to add this fix to `youtube.com` and replace base64 image with `INVERT` rule?

TheAlienDrew:
Try the invert yourself... It inverts the entire element, which means
everything in the body. That's why I did it the base64 way.

alexanderby:
Then let's use `background-image: none !important;`.

TheAlienDrew:
Going to close this and do another pull request in the right area

PR #1189, opened 2019-03-19
Updated dynamic-theme-fixes.config Youtube Policies page

TheAlienDrew:
Fixed the background SVG (changed color to black and converted it to URI to use)
Fixed the logo that would appear above this svg by inverting it.


alexanderby:
I would also put this one into single `youtube.com` section, embedded images make this file grow too much.

PR #1190, opened 2019-03-19
Updated dark-sites..config Youtube Music website

TheAlienDrew:
Check for yourself, music.youtube.com is already dark mode by default.


alexanderby:
Not available in my region :) I will merge this fix, but if you see any white page on this website or website is not dark by default, the whole site should be excluded from list.

PR #1191, opened 2019-03-19
Updated dynamic-theme-fixes.config GitHub

TheAlienDrew:
Subtle fix for off color separation when looking at some pull requests.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -269,6 +269,9 @@ CSS
 .markdown-body pre code {
     background-color: transparent !important;
 }
+.new-discussion-timeline .closed-banner {
+    border-bottom-color: rgb(23, 24, 28);
+}
 
 ================================
 


src/config/dynamic-theme-fixes.config
```diff
@@ -269,6 +269,9 @@ CSS
 .markdown-body pre code {
     background-color: transparent !important;
 }
+.new-discussion-timeline .closed-banner {
+    border-bottom-color: rgb(23, 24, 28);
```
alexanderby: The fix should look well in Light and Dark mode and filter values should be applied, for this purpose `${color}` templates are used, where `color` is the source color, usually white or black. Also `!important` should be added to prevent override by another rule.

TheAlienDrew:
I understand the !important, but what about this $[color] thing? I understand the purpose for the Light theme, but does this change the color automatically to it's darker shade when dark mode is turned on?

alexanderby:
The value in `${color}` template (pay attention on curly braces) is inverted in dark mode.

TheAlienDrew:
> The value in `${color}` template (pay attention on curly braces) is inverted in dark mode.

Thanks, that is very useful.

TheAlienDrew:
Going to fix with a different pull.

PR #1193, opened 2019-03-20
Updated dynamic-theme-fixes.config Google profile fix

TheAlienDrew:
The profile picture in the top right corner was inverted, now it's fixed by inverting it again.

Should work for all sites that have the small profile picture.


alexanderby:
Thanks for noticing. There could be some change on the website. But the reason is in another selector. For some reason I cannot edit this PR, so I will close it and remove unnecessary line, that causes the issue.

alexanderby:
Here are my changes https://github.com/darkreader/darkreader/commit/d7827340da8f343c3df2c275b8d888ec28094ee2

PR #1194, opened 2019-03-20
Viduro.xyz and www.Viduro.xyz is now fully dark by default.

Troplo:


alexanderby:
`www` is not necessary in this case.

PR #1225, opened 2019-04-07
Update dynamic-theme-fixes.config - datadog

sloncho:
When the log detail window is opened (by clicking on a log row), the selected row is not visible. This makes it work.

Originally in the config editor I found

```

app.datadoghq.com

INVERT
.HostMap-canvas

```

Adding the CSS section worked.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -57,7 +57,7 @@ INVERT
 
 CSS
 .log-dt-event.active, .log-dt-event.active:hover, .log-dt-event:hover {
-    background-color: rgb(37, 45, 58);
+    background-color: rgb(37, 45, 58) !important;
 }
 
 ================================
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -151,6 +151,18 @@ span[class^="skycon swip"]
 
 ================================
 
+app.datadoghq.com
+
+INVERT
+.HostMap-canvas
+
+CSS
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -151,6 +151,18 @@ span[class^="skycon swip"]
 
 ================================
 
+app.datadoghq.com
```
alexanderby: Please, preserve alphabetical order by URL. By the way, there is already `app.datadoghq.com` in the list.

src/config/dynamic-theme-fixes.config
```diff
@@ -151,6 +151,18 @@ span[class^="skycon swip"]
...
+
+INVERT
+.HostMap-canvas
+
+CSS
+.log-dt-event.active, .log-dt-event.active:hover, .log-dt-event:hover {
+    background-color: rgb(37, 45, 58);
```
alexanderby: Please, check, how it looks in light mode. Most likely the fix should look like `background-color: ${white} !important;` where `${white}` is a pattern containing source color value, and `!important` prevents overrides by other rules.

src/config/dynamic-theme-fixes.config
sloncho: done

src/config/dynamic-theme-fixes.config
sloncho: Unfortunately the proposed {white} made it unreadable in light mode, and did not change the original behavior in dark.. Added "!important"

Issue #1232, opened 2019-04-09
Google maps does not invert canvas


kiralex:
Hi, 

I have an issue with google maps : the canvas does not invert correctly. I have done a fix and I will open a pull request. 

Brefore: 
![Before](https://user-images.githubusercontent.com/3044771/55825273-f8e17500-5b05-11e9-9048-9e07c63a5784.png)

After:
![After](https://user-images.githubusercontent.com/3044771/55825302-0434a080-5b06-11e9-9271-e0eb3387a1e8.png)

Hope this will helps :smile:

kiralex:
The CI seems to not pass. I'll fix that :wink

alexanderby:
Hi, do you think that markup differs on Linux from other operating systems?

There is already `#app-container.vasquette:not(.app-imagery-mode):not(.app-globe-mode) .widget-scene-canvas` selector that does the job, inverts scheme and doesn't invert globe or street views.

StraggleCraft:
Any update on this issue?

Gusted:
@StraggleCraft If you have darkreader 4.9.9 It should be fixed.

Gusted:
#2497

PR #1233, opened 2019-04-09
Fix google maps on linux

kiralex:
closes https://github.com/darkreader/darkreader/issues/1232


alexanderby:
Hi. Sorry, I haven't tried it on Linux. Do you think this problem still happens?

inthreedee:
I'm on Arch Linux with Firefox v69.0 and google maps is correctly inverted for me using a fresh install of Dark Reader, v4.8.0.  No modifications were needed.

PR #1242, opened 2019-04-13
Fixes for tjournal.ru

jtraub:
Another popular site covering wide-range of Internet-related topics.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -728,6 +728,18 @@ INVERT
 
 ================================
 
+tjournal.ru
+
+INVERT
+mark
+
+CSS
+#page_wrapper {
+  background-color: ${#fff4e0} !important;
+}
+
+================================
+
 translate.google.*
 translate.google.*.*
 


jtraub:
I should note that there are 3 popular content sites (dtf.ru, tjournal.ru and vc.ru) belonging to the same company in the Russian part of the Internet. They use the same engine to power these sites; hence, having the same layout.

Maybe rules for all 3 of them should be merged into a single record? On the other hand, I can see that layout changes are boing rolled out with delay. First, they've updated vc.ru and then 2 days later - tjournal.ru. dtf.ru seems to have old layout.

Neurotoxin001:
They change design so it's broken again

image

Neurotoxin001:
So now I must wait update? Also they have https://dtf.ru site too, and there is all is fine even after redesign. They have same engine.

jtraub:
@Neurotoxin001 you can manually add changes to the config of your DarkReader.

Just click 'Dev tools' button. You will see a popup with textarea in it. Locate tjournal.ru entry there and replace entry with what you see [here](https://github.com/darkreader/darkreader/blob/b9d5b05b86b2ca4a0caf7958082448f855a55e37/src/config/dynamic-theme-fixes.config#L1006)

Or you can wait until @alexanderby reviews my changes and merges them.

Regarding DTF.ru. Yes, they share the engine between vc.ru, tjournal.ru, and dtf.ru. However, the markup on dtf.ru is slightly different. I didn't have a lot of time to understand what actually causes DarkReader misbehaviour on tjournal.ru.

Anyway, you can ping me if there will be more breaking changes on tjournal.ru / dtf.ru / vc.ru. I will try to fix them.

jtraub:
@Neurotoxin001 do you use an adblocker with [RU AdList](https://forums.lanik.us/viewtopic.php?f=102&t=22512) filter enabled?

Neurotoxin001:
Yes.

jtraub:
@Neurotoxin001 well, I do believe that the problem is caused by [this rule](https://hg.adblockplus.org/ruadlist/file/58bfe2d4a629/css-fixes-experimental.txt#l819) in the RU AdList filter.

This explains why we don't observe the problem on dtf.ru.

If you haven't updated your DarkReader configs you can confirm by disabling you adblocker or RU AdList.

jtraub:
@Neurotoxin001 I closed my PR as I confirmed that the real culprit is a CSS fix from RU AdList filter.

You can either add your own user style or ask RU AdList mainainer to remove/modify the rule.

PR #1251, opened 2019-04-19
GitHub: Zero Contribution Day Fix

harrisbegca:
The GitHub contribution section shows the number of contributions per day. However, it looks like this currently:
Light
With the proposed change, the section would look like this:
Dark


alexanderby:
Nice solution, thanks.

PR #1259, opened 2019-04-21
discordapp.com is already dark

DooML:
It already offers a dark theme by default ( + option to switch to light theme ) and Dark Reader only ruins it


alexanderby:
Hi, there are some white pages on Discord, so it currently cannot be added into this list. When option to ignore Dark List will be added, it will be possible to add Discord there.

PR #1269, opened 2019-04-26
Update ko.config

little7279:
Korean Translation fix


alexanderby:
Thank you. I used Google Translate and Chinese translation as the source. Will be published soon.

PR #1294, opened 2019-05-05
Fix hover inversion for Slack when using Refined's Visible Threads

g3rv4:
[Refined](https://refined.chat/) is an extension for slack that makes a bunch of things better.

One of the things it does is making threads visible (aka: [Visible Threads](https://refined.chat/visible-threads)).

When visible threads are enabled, hovering on a conversation message highlights all the messages in the same conversation. But with Dark Reader, the hover color can't be really appreciated.

In this PR, that's fixed. Steps to see the feature in action:

1. Install Refined
2. Enable visible threads
3. Hover over a reply

Expected behavior: You should see the original message highlighted.
Current behavior: The color change is so subtle it's hard to see.

Here you can see how it looks with this change:

visibleThreadsOnDark

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -700,6 +700,16 @@ slack.com
 INVERT
 .slack_logo > img
 
+CSS
+.refined-conversation-hover,
+.refined-conversation-hover .c-message--hover {
+    background: #595959 !important;
+}
+.refined-conversation-hover .c-message__reply_bar:hover,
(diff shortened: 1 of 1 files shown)


g3rv4:
I was able to mess with how my extension looks with Dark Reader in this commit https://github.com/g3rv4/Refined/commit/81a7b2fbb8b2d207570a092942e0d5d869a67ea0 which IMO is a cleanest solution (since almost none of Dark Reader users are Refined users).

PR #1311, opened 2019-05-10
Most math elements on Wikipedia are double inverted

scholtes:
Most math elements contain both the `.mwe-math-element` and the `.mwe-math-fallback-image-inline` class, and therefore were being double-inverted (i.e., undoing itself).

`.mwe-math-element` was added as part of #382.  `.mwe-math-fallback-image-inline` was added as part of #1261, likely because there were some elements which contained `.mwe-math-fallback-image-inline` but not `.mwe-math-element`.

However, as far as I can tell, most math elements contain both `.mwe-math-element` or `.mwe-math-fallback-image-inline` and therefore get inverted twice.

Tested in Chrome

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -866,7 +866,6 @@ INVERT
 .mw-wiki-logo
 .central-textlogo__image
 .svg-Wikimedia-logo_black
-.mwe-math-fallback-image-inline
 
 ================================
 


scholtes:
A smarter solution might be to check if any of the child element's parents have one of the classes to be inverted, but I'm guessing that DarkReader does this all via CSS and that this would be a much heavier change.  Suggesting this to revert #1261 for now as a workaround.

alexanderby:
It is possible to add CSS rule like this:
```
INVERT
.selector1

CSS
.selector2 .selector1 {
    filter: none !important;
}
```
Also could you please give some sample pages?

TallGuysFreeADMIN:
Example: https://en.wikipedia.org/wiki/Pythagorean_theorem

HyperionFalls:
Another example:
https://en.m.wikipedia.org/wiki/Bra%E2%80%93ket_notation

Issue #1330, opened 2019-05-19
How to detect if darkreader is active?


Richienb:
Is there a way to detect if **darkreader** is active so my website can adjust accordingly?

If there isn't such a way, I suggest that a variable should be defined and an event like `themechanged` on the document element should be triggered when the theme changes.

alexanderby:
My opinion is that website should not rely on extensions. I'm not sure if the extension can store it's own global objects and trigger it's own events. The content script runs in a sandbox.

Gusted:
```javascript
document.querySelector('meta[name="darkreader"]');
```

milahu:
see also #4342

PR #1331, opened 2019-05-19
Update rules for richie-bendall.ml

Richienb:
`richie-bendall.ml` now supports dark mode automatically. This PR updates the config files to address that.

The only issue is that there is no way to detect if `darkreader` is requesting dark mode so the user will have to manually activate it. #1330

Depends on #1342.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -121,6 +121,7 @@ ptt.cc
 raidtime.net
 rawg.io
 realgamernewz.com
+richie-bendall.ml
 runicgames.com
 sanctum.geek.nz/arabesque
 sandervanderburg.blogspot.com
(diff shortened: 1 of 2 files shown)


alexanderby:
I see that Dark Reader works as intended, always making website dark when user wants it. Only websites that are always dark, should be placed in Dark Sites list, this is important for those who use Filter mode.

While this is great, that you provided a dark theme, I don't think that website should be tied on extension.

Maybe a better and simpler solution for you would be using [`prefers-color-theme` media rule](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme), which is already supported by Firefox and Safari and will be supported in Chrome 76 (Canary at the moment). This way users, that have system dark mode enabled, will see a dark theme on your website.

alexanderby:
Hi. This PR is marked as a draft. Please close it if no longer needed, or mark it as ready so that I can merge it.

PR #1335, opened 2019-05-22
Update nba.com and mcdonalds.com

OscarGodson:
nba.com and mcdonalds.com both had backgrounds that weren't properly changing the background color from white.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -457,6 +457,15 @@ INVERT
 
 ================================
 
+mcdonalds.com
+
+CSS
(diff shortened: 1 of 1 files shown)


alexanderby:
Unfortunately I was not able to check your changes. I didn't find an element with specified selector on Mcdonalds website, and `nba.com` redirects me to `tribuna.com/nba`, maybe this is some redirect based on my location.

Please, check that your fixes work well in Light mode too. To achieve this `${default_color}` template is used. Also `!important` keyword should be used.

alexanderby:
Hi. A long time has passed. Please submit a new PR (or multiple for each site) if still relevant.

PR #1341, opened 2019-05-24
Update rollup-plugin-typescript2 to fix security issues

Richienb:
This PR resolves the following NPM Security Advisories:

- https://www.npmjs.com/advisories/786
- https://www.npmjs.com/advisories/786
- https://www.npmjs.com/advisories/813
- https://www.npmjs.com/advisories/755
- https://www.npmjs.com/advisories/788
- https://www.npmjs.com/advisories/786

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "rollup-plugin-commonjs": "9.2.1",
     "rollup-plugin-node-resolve": "4.0.1",
     "rollup-plugin-replace": "2.1.0",
-    "rollup-plugin-typescript2": "^0.21.1",
+    "rollup-plugin-typescript2": "0.21.1",
     "ts-jest": "24.0.0",
     "ts-node": "8.0.3",
     "typescript": "3.3.3333",
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "rollup-plugin-commonjs": "9.2.1",
     "rollup-plugin-node-resolve": "4.0.1",
     "rollup-plugin-replace": "2.1.0",
-    "rollup-plugin-typescript2": "0.19.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -34,7 +34,7 @@
     "rollup-plugin-commonjs": "9.2.1",
     "rollup-plugin-node-resolve": "4.0.1",
     "rollup-plugin-replace": "2.1.0",
-    "rollup-plugin-typescript2": "0.19.3",
+    "rollup-plugin-typescript2": "^0.21.1",
```
alexanderby: Please, remove `^`, it can break a build when future versions of this package will be released, especially that according to SemVer versions starting with `0` can change their API.

alexanderby:
By the way, what is your OS, Node and NPM versions? There are so many changes in `package-lock`.

Richienb:
@alexanderby I'm using:
Node 11.15.0
NPM 6.7.0
Windows 10

package.json
Richienb: Done in https://github.com/darkreader/darkreader/pull/1341/commits/44b9ec68343ced120a97b9f10dad3f38fec1cf4b

Richienb:
@alexanderby Just wondering, is there a reason why you stick to NPM and not another package manager like Yarn which works faster?

alexanderby:
Hi. Thanks for noticing the vulnerabilities. The configuration was updated to use as few dependencies as possible. This project uses Node LTS. The reason to stick to NPM is that there is no benefit of Yarn for this small project. At the same time using it requires an extra installation step. The lock file—the main feature of Yarn—was already implemented in NPM some years ago, so I see no benefit in it.

PR #1342, opened 2019-05-24
Allow detection and listening for dark mode changes

Richienb:
This PR allows darkreader to set a Global Window Variable that contains information about the theme and if dark mode is enabled. It also adds an event to the document element.

TODO:
- [ ] Finish implementation
- [ ] Document usage in README
- [ ] Test for all settings.

Closes https://github.com/darkreader/darkreader/issues/1330.

### DIFF
--- a/src/background/extension.ts
+++ b/src/background/extension.ts
@@ -96,7 +96,6 @@ export class Extension {
                 }
                 const url = await this.tabs.getActiveTabURL();
                 const info = await this.getURLInfo(url)
-                handleManual(!info.isProtected && info.isInDarkList)
                 return info;
             },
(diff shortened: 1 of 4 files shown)


alexanderby:
Sorry, I think we should discuss the feature before submitting the code.

Some code review:
- Use self-descriptive file names, variables and function names.
- Don't change code that is not related to PR, all the refactoring and formatting changes should be made in a separate PR.
- Try matching the code style (`npm run code-style` command was added, that uses ESLint).
- Avoid using `innerHTML` (use `textContent` instead).
- Be accurate when injecting scripts with dynamic content.
- Don't leave `console.log`.

PR #1348, opened 2019-05-27
Change some things to French translation.

locness3:
This PR changes some things to the French translation to make it more "Frenchy" and correct.

### DIFF
--- a/src/_locales/fr.config
+++ b/src/_locales/fr.config
@@ -1,5 +1,5 @@
 @extension_description
-Thème sombre pour chaque site Web. Prenez soin de vos yeux, utilisez Dark Reader pour votre navigation nocturne et quotidienne.
+Un thème sombre pour chaque site Web. Prenez soin de vos yeux, utilisez Dark Reader pour votre navigation nocturne et quotidienne.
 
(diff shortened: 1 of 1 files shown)


src/_locales/fr.config
```diff
@@ -120,7 +120,7 @@ Filtre+
 Statique
 
 @engine_dynamic
-Dynamiq.
+Dynamique
```
alexanderby: This word is too long for the small button. In future versions "Filter" and "Filter+" will be merged, so there would be more space.

src/_locales/fr.config
```diff
@@ -14,14 +14,14 @@ Marche
...
-Basculer le site actuel
+Activer/Désactiver le site actuel
 
 @setup_hotkey_toggle_site
 Configurer la touche de raccourci
 du basculement du site actuel
 
 @toggle_extension
-l'extension bascule
+Activer:Désactiver l'extension
```
alexanderby: Maybe "/" should be used same as on line 17.

locness3: That seems to be a typo.

PR #1375, opened 2019-06-14
Create ukrainian localization file

lytvoles:


alexanderby:
Hi, thanks for the translation! Are you sure about "мережевий осідок"?

alexanderby:
OK @lytvoles, I will merge it. My opinion is that "сайт" looks more commonly used, but you are Ukrainian, so you decide.

PR #1376, opened 2019-06-14
Invert YouTube like bar in dynamic theme

luuknieuwdorp:
Without this, likes and dislikes are almost the same color making it hard to differentiate between the two.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1043,6 +1043,9 @@ CSS
 
 youtube.com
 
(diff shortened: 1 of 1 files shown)


alexanderby:
Hi, are you using Firefox? YouTube serves different site versions for Chrome and Firefox, so I will have a look more precisely there.

luuknieuwdorp:
Ah, yes I am. I assumed it would be the same. I only tested this on Firefox.

alexanderby:
I think it looks OK now.

PR #1393, opened 2019-06-25
Update dark-sites.config

garrtt:
Adding google.com to Site List will add messages.google.com (which has a dark mode)


alexanderby:
Hi, this list should contain websites that are always dark by default.

PR #1419, opened 2019-07-09
Add overdodactyl.github.io/ShadowFox to dark sites

alexalekseyenko:
The ShadowFox page is [as one might expect] already dark-themed. Welcome to the dark side, ShadowFox.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -109,6 +109,7 @@ open.spotify.com
 opencritic.com
 opendota.com
 orteil.dashnet.org/cookieclicker
+overdodactyl.github.io/ShadowFox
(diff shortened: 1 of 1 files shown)


rolagg:
I think it actually works good with extension.

Without extension:



With extension:

alexalekseyenko:
That's interesting. This is what it looks like for me with the extension:
Screen Shot 2019-07-10 at 09 32 23
Screen Shot 2019-07-10 at 09 32 34

I'm using:
Brave Version 0.66.99 Chromium: 75.0.3770.100 (Official Build) (64-bit)
Dark Reader 4.7.13

rolagg:
Oh, right, I forgot about classic Filter mode.

PR #1420, opened 2019-07-10
remove steamdb.info from global dark list

xPixv:
image

image

image

image

It worked fine in dark mode before.

Thank you for your consideration.


camobiwon:
Saw this as well this morning, hoping they remove it soon

PR #1423, opened 2019-07-12
Patch 1

dimuththarindu:
Added the Sinhala Language.
Not fully completed.


alexanderby:
Hi @dimuththarindu! Please, check the lines count and new lines, it should be the same for all locales. Then tests will pass.

alexanderby:
Sad news, I don't see Sinhala language in this list https://developer.chrome.com/webstore/i18n

Maybe I will add an option to choose preferred locale. But it seems that Chrome doesn't support `si` locale.

PR #1425, opened 2019-07-12
Added webtoons.com text fix to dynamic theme config file

jebjordan:
Changes the webtoon tile text so that it can actually be seen in dark mode (and light mode) with DR on. 
(before it was kinda really light grey and hard to see)

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -962,6 +962,25 @@ INVERT
 
 ================================
 
+webtoons.com
+
+CSS
+.card_item .info {
+    color: black;
+}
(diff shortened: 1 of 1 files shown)


alexanderby:
Hi, please pay attention, that `!important` keyword should be used and fix should work in light mode too (try using `${source_color_here}` pattern so that the value will be substituted later.

jebjordan:
I tested it in dark and light mode and it worked. 

My goal was just to make it look like the text normally does before DR but with a few small changes, since that's basically the way the site devs intended it. 

Looked fine on my end. Anyways thanks for accepting the pill request.

alexanderby:
Great, thank you.

PR #1434, opened 2019-07-16
Added CSS Fix for technews.tw

nick322:
Added CSS Fix for technews.tw

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -837,6 +837,33 @@ INVERT
 
 ================================
 
+technews.tw
+
+CSS
+div#main.wrapper {
+    background-color: #000;
+}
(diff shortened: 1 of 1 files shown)


alexanderby:
Hi, please pay attention that `Important` keyword should be used. Also please pay attention, that fix should work in light mode too (use `${source_color_here}` pattern so that inverted value will be substituted when necessary).

nick322:
I tested it in dark and light mode and it worked.

My goal was just to make it look like the text normally does before DR but with a small change (white shadows on the text), since that's basically the way the site devs intended it, and gives best read ability in both modes.

not fix is look like 
圖片 20190720 11

fix 
圖片 20190720 12


Looked fine on my end. Anyways thanks for accepting the pull request.

Issue #1093, opened 2019-02-10
Cannot see black chess pieces on lichess.org


nitrocode:
I noticed the black pieces were viewable when I disabled darkreader.

https://lichess.org/training

kristian-94:
Would like to use this extension on lichess, but of course it makes the board look all wrong and inverted. Would be perfect if it inverted the rest of the page but not chess boards.

trentks:
Can confirm this issue exists... luckily lichess offers a dark-mode, and this site can just be excluded from darkreader. It does mean having to allow cookies (whitelisting lichess in cookie autodelete) for this selection to persist though.

pfgithub:
Lichess has these errors in the console with darkreader enabled

```
Content Security Policy: This page's settings blocked the loading of a resource at blob:https://lichess.org/...
Content Security Policy: This page's settings blocked the loading of a resource at blob:https://lichess.org/...
Content Security Policy: This page's settings blocked the loading of a resource at blob:https://lichess.org/...
Content Security Policy: This page's settings blocked the loading of a resource at blob:https://lichess.org/...
Content Security Policy: This page's settings blocked the loading of a resource at blob:https://lichess.org/...
```

there are 5 errors and 5 missing piece types (all black pieces except the king disappear)

lopnax:
If you add this lines in the extension settings you are gonna see the black pieces correctly
```
lichess.org

CSS
[...]

PR #1436, opened 2019-07-17
Add lichess.org to dark-sites.config

trentks:
Lichess is currently broken with Dark Reader--see issue #1093. Black chess pieces (other than the king) are not drawn. Lichess has its own dark theme, so it's best to disable dark reader globally and have users select the native dark theme that lichess offers.


alexanderby:
Hi, this lists holds websites that are always dark. It is important for primitive Filter mode and prevents white flashes there. I think the issue should be fixed using the Dev Tools.

Issue #1407, opened 2019-07-03
Todoist Subdomain Issue


Rakleed:
Dark Reader is automatically disabled on todoist.com/app, since you can enable the dark theme for it - this is correct.

But in the settings (todoist.com/prefs) with the dark Todoist theme turned on, the background remains white, as on the Templates page (todoist.com/templates). Screenshots: https://tlgur.com/d/g31DN3O8, https://tlgur.com/d/GVQV9pa8

So please make Dark Reader disable only to todoist.com/app.

insipidlight:
This seems to still be an issue? I had to manually add the `app` after creating a new exclusion in the sites list.

PR #1437, opened 2019-07-17
Add todoist.com/app to dark-sites.config

Rakleed:
Fix https://github.com/darkreader/darkreader/issues/1407

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -151,6 +151,7 @@ terminal.sexy
 textfiles.com
 tf2mart.net
 thealiendrew.github.io
+todoist.com/app
 toneden.io
 totalwarwarhammer.gamepedia.com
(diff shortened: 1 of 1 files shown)


alexanderby:
Hi, I cannot login to see it myself? Is the dark mode enabled on this website by default? Only websites that are always dark should be added into Dark Sites list.

Rakleed:
Hello! So the Dark Sites list includes only sites that are always with a dark theme or which use a dark theme by default, but in the settings you can change it to a light one?

If this is so, then not only Todoist, but also some sites already added to the list are not suitable. For example, [opencritic.com](opencritic.com) and [app.getpocket.com](app.getpocket.com).

alexanderby:
Feel free to edit this list if you see mistakes. From my side I'm going to add ability to override global dark list.

PR #1438, opened 2019-07-17
Fix mark tag on Medium

secf4ult:


alexanderby:
Hi, why did you make it green? Please note that fix should work well in light mode too (use `${source_color}` template to properly invert or not invert color). By the way, I see this bug is unstable,  maybe there is some issue with CSS priority.

alexanderby:
Hi, why did you make it green? Please note that fix should work well in light mode too (use `${source_color}` template to properly invert or not invert color). By the way, I see this bug is unstable,  maybe there is some issue with CSS priority.

secf4ult:
@alexanderby The mark tag didn't render properly when I created this PR, now it does, I think this PR isn't relevant anymore.

PR #1450, opened 2019-07-20
Update Sinhala Lan

dimuththarindu:


alexanderby:
I've merged it, but this language seems to be not supported (like my native Belarusian too).

I suggest you to submit a Chromium and Firefox bugs.

By the way, when you run `navigator.languages` in Chrome and Firefox console, what you get?

alexanderby:
I've just checked, running Chrome with `lang=si` flag should work for the extension.

PR #1459, opened 2019-07-25
Update dark-sites.config

FReeshabh:
Got blinded reading Plural sight article at night.
Added Discord to the list, as the default dark mode is alright, adding the dark reader makes it font for some fonts to be readable.
Removed PluralSlight from the slight cause blinding people is mean

Ideally there should be a useful defined list for this, but for now this will do since this issue has multiple complaints and many people have complained asking for not to be blinded.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -50,6 +50,7 @@ dotapicker.com
 dpreview.com
 draculatheme.com
 draftkings.com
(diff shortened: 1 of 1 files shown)


alexanderby:
Hi. It is planned to add ability to override global dark list. So currently, since Discord has some white pages, it is better to not store it in this list.

PR #1480, opened 2019-08-04
Fix dumb ukrainian translations

33kk:
When i saw "Цідило" i just wanted to die...... "Осідки" is gorgeous.


alexanderby:
Hi, I was guessing there's something wrong, but left the decision for the author of initial translation. Anyway good work https://github.com/darkreader/darkreader/pull/1375

PR #1502, opened 2019-08-16
[Do not merge] Fix for tjournal.ru

jtraub:
As @Neurotoxin001 pointed out in https://github.com/darkreader/darkreader/pull/1242#issuecomment-521118613 tjournal.ru has been updated with the new design which doesn't work with current DarkReader configs.

This PR resolves the problem

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1009,8 +1009,8 @@ INVERT
 mark
 
 CSS
-#page_wrapper {
-  background-color: ${#fff4e0} !important;
+div.main.layout {
+  background-color: ${#F4F4F4} !important;
 }
 
 ================================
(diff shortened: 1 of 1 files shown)


jtraub:
This is how it looks for me after the fix has been applied.
image

jtraub:
Latest commit  makes links inside <mark> blocks slightly more readable

Before
image

After
image

jtraub:
It seems that the initial issue with the site background has been caused by adblocker. I will investigate this tomorrow.

P.S. DarkReader works just fine with tjournal.ru if I use Firefox.

jtraub:
I have confirmed that DarkReader itself [works just fine](https://github.com/darkreader/darkreader/pull/1242#issuecomment-523529942). Hence, I close the PR.

PR #1506, opened 2019-08-19
fix wikipedia.org li::marker

sizur:


gguman:
Chrome (Version 70.0.3538.77 (Official Build) (64-bit)) 
problems with formulas using li::marker on wikipedia.org

alexanderby:
@gguman So sad I didn't notice your comment. Unfortunately this works in Firefox only. Maybe I will add some template like `$ff{li::marker}` for Firefox-only fixes.

PR #1522, opened 2019-08-25
feat: toggle with system dark mode media query change

### LINKED ISSUE #745: Auto toggle based on macOS Mojave Dark Mode status
As you might know, the newly released macOS Mojave has a great Dark Mode implementation where all supported apps automatically change color theme to Dark when Dark Mode is toggled.

It would be great if Dark Reader could be compatible with Dark Mode and automatically be toggled based on the Dark Mode status.

ColCh:
closes #745

demo https://gph.is/g/aXmLjmR

<details><summary>GIF</summary>

</details>


src/_locales/en.config
```diff
@@ -41,6 +41,8 @@ by browser
 This site is in global
 Dark List
 
+@system_dark_theme
+System dark theme
```
ColCh: do I need to create other translations as well?

src/background/extension.ts
```diff
@@ -190,19 +195,33 @@ export class Extension {
 
     private startAutoTimeCheck() {
         setInterval(() => {
-            if (!this.ready || this.user.settings.automation !== 'time') {
+            if (this.user.settings.automation !== 'time' && this.user.settings.automation !== 'system') {
```
ColCh: we still need to check by interval as a fallback solution

src/background/extension.ts
```diff
@@ -190,19 +195,33 @@ export class Extension {
...
                 return;
             }
-            const isEnabled = this.isEnabled();
-            if (this.wasEnabledOnLastCheck !== isEnabled) {
```
ColCh: refactored out into handleAutoCheck method

src/background/tab-manager.ts
```diff
@@ -48,6 +49,12 @@ export default class TabManager {
                 } else if (message) {
                     port.postMessage(message);
                 }
+
+                port.onMessage.addListener(({ type }: Message) => {
+                    if (type === 'color-theme-media-query-change') {
```
ColCh: is this ok to communicate with content scripts like that?

src/definitions.d.ts
```diff
@@ -63,7 +63,7 @@ export interface UserSettings {
     changeBrowserTheme: boolean;
     notifyOfNews: boolean;
     syncSettings: boolean;
-    automation: string;
+    automation: '' | 'time' | 'system';
```
ColCh: just replaced typing to something more exact

src/inject/index.ts
```diff
@@ -40,3 +41,8 @@ port.onDisconnect.addListener(() => {
     logWarn('disconnect');
     cleanDynamicThemeCache();
 });
+
+port.onDisconnect.addListener(watchForMediaQueryChange(() => {
```
ColCh: **we need** to listen to page event because Chrome not fires media query events for extension's `window` object

src/ui/popup/components/header/style.less
```diff
@@ -45,7 +45,7 @@
             display: inline-block;
             height: @size-control-inner + 2 * @size-border;
             left: 100%;
-            opacity: 0;
+            opacity: 0.7;
```
ColCh: made this one to be visible a bit because it's very hard to notice with zero opacity

alexanderby:
Thanks for the Pull Request, nice work! Currently I'm in a business trip and will be able to review it on Monday.

src/inject/index.ts
alexanderby: Did you see some Chrome bug submitted?

src/background/extension.ts
ColCh: I suppose interval code is not needed

That was a workaround for non-firing event for media query change

src/inject/index.ts
ColCh: Yes. This code may be removed after Chrome bug would be fixed

ColCh:
Thank you for fixes in the branch!

Issue #1498, opened 2019-08-14
Background texture of askubuntu.com too bright


cole8888:
The background texture of askubuntu.com is too bright and makes viewing the page quite difficult. I've attached an image of the page.

![image](https://user-images.githubusercontent.com/32819560/63063183-f9ce4d00-bed1-11e9-9382-a31e4f4b857e.png)

The texture is loaded with a URL which references an image, so I was unable change it's colour using developer tools from the extension or the browser. Removing this element does not cause any issues, so blocking it may be the most simple solution.

This is probably the same as #1476, except they only mention the left sidebar in their issue. 
I am using the dynamic theme generator.

FranklinYu:
One more site: https://dba.stackexchange.com

cole8888:
Fixed :+1:

jtraub:
@FranklinYu well, the issue is related to askubuntu. I haven't noticed your report until today.

Anyway, I sent a PR which removes the background on all StackExchange sites.

PR #1533, opened 2019-08-28
Remove body background image on StackExchange network

jtraub:
I've started to toggle DarkReader more frequently and immediately noticed that many of StackExchange sites have body background images which don't look nice in the dark mode. See https://ell.stackexchange.com or https://emacs.stackexchange.com/

Other users have noticed it. See PR #1527, https://github.com/darkreader/darkreader/issues/1498#issuecomment-523196238

While many of StackExchange sites do not have any background image I don't think it is worth it to remove the image by separate fixes. Hence, I've added such a broad fix.

P.S. It also makes PR #1527 obsolete.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -87,6 +87,7 @@ CSS
 ================================
 
 askubuntu.com
+*.stackexchange.com
 
 CSS
 body {


jtraub:
@alexanderby maybe 3c6d32f8e88089184f6 should be reverted. See https://github.com/darkreader/darkreader/blob/master/src/config/dynamic-theme-fixes.config#L90 and https://github.com/darkreader/darkreader/blob/master/src/config/dynamic-theme-fixes.config#L1227

PR #1540, opened 2019-08-30
Remove max-width on body of DevTools

jtraub:
Sometimes I have to look into really long lines in dynamic-fixes.config when I try to understand what goes wrong on a site and fix it. It becomes really tedious due to long lines and expanding DevTools popup doesn't help since its style bears the `max-width: 60rem` property

image

I propose removal of `max-width` as I believe it basically does nothing in regards to the UI.


alexanderby:
Thanks for the Pull Request, I will review it on Monday.

PR #1543, opened 2019-09-03
Add outlook.office.com to the list

tvalentius:
Added outlook.office.com, since they already have a built-in dark theme.


alexanderby:
Hi, this list should contain website, that are dark by default. This is important for Filter mode.

Issue #1266, opened 2019-04-24
Dark mode doesn't work on YouTube


xin12pzx24:
Today, i found darkreader doesn't work on YouTube. i don't know why.
But i fix it by replacing the default settings in darkreader.
If you don't want to fix it yourself, you can use my solution.
Or maybe you can take some time to fix it.
It may be a bit lengthy, but it's a temporary solution.

**Code below:**
```
youtube.com

INVERT
#checkmark
.ytp-volume-slider-handle
.toggle-bar.style-scope.paper-toggle-button
.align-by-text.style-scope.ytd-comment-replies-renderer
g#youtube-red-paths
#like-bar
.yt-simple-endpoint.style-scope.ytd-grid-video-renderer
.title.style-scope.ytd-guide-entry-renderer
.style-scope.ytd-guide-signin-promo-renderer
#video-title.style-scope.ytd-compact-video-renderer
.more-button.style-scope.ytd-comment-renderer
#submenu.menu-container.style-scope.ytd-multi-page-menu-renderer
#content.style-scope.ytd-expander
#author-text.yt-simple-endpoint.style-scope.ytd-comment-renderer
#simplebox-placeholder
#upload-info

CSS
#textarea, paper-button {
    color: ${black} !important;
}

.style-scope.ytd-page-manager, 
.style-scope.ytd-masthead, 
.style-scope.ytd-guide-renderer, 
#header.style-scope.ytd-app, 
#columns.style-scope.ytd-watch-flexy, 
.style-scope.ytd-popup-container,
.style-scope.ytd-searchbox,
#label.style-scope.ytd-toggle-theme-compact-link-renderer,
#label.style-scope.ytd-compact-link-renderer,
#subtitle.style-scope.ytd-compact-link-renderer,
#video-title.style-scope.ytd-compact-playlist-renderer,
#video-title.style-scope.ytd-compact-radio-renderer{
    background-color: black !important;   
    color: ${black} !important;   
}
```

alexanderby:
Hi, please update Dark Reader up to 4.7.11, if you already have the latest version, reset your changes in Dev Tools.

MobiuSRIT:
Hail 4.7.11 -- I was going mad after youtube pushed it's new light/dark theming and dark reader stopped.  I'm so happy this is working again in this version.  Thank you.

alexanderby:
@MobiuSRIT Maybe you have to reset changes in Dev Tools. By the way, it seems that today they reverted the change.

[...]

PR #1569, opened 2019-09-15
Fixes illegible volume slider on YouTube videos.

mvastola:
See #1266

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1614,6 +1614,12 @@ html:not(.style-scope) {
     --yt-spec-general-background-b: ${#f1f1f1} !important;
     --yt-spec-general-background-c: ${#e9e9e9} !important;
 }
+.ytp-volume-slider-handle {
+    background-color: rgba(255, 255, 255, 0.9);
+}
+.ytp-volume-slider-handle::before, .ytp-volume-slider-handle::after {{
+    background-color: rgba(255, 255, 255, 0.5);
+}
 
 ================================
 


src/config/dynamic-theme-fixes.config
```diff
@@ -1614,6 +1614,12 @@ html:not(.style-scope) {
     --yt-spec-general-background-b: ${#f1f1f1} !important;
     --yt-spec-general-background-c: ${#e9e9e9} !important;
 }
+.ytp-volume-slider-handle {
+    background-color: rgba(255, 255, 255, 0.9);
+}
+.ytp-volume-slider-handle::before, .ytp-volume-slider-handle::after {{
```
alexanderby: `{{` should be a single `{`

src/config/dynamic-theme-fixes.config
```diff
@@ -1614,6 +1614,12 @@ html:not(.style-scope) {
     --yt-spec-general-background-b: ${#f1f1f1} !important;
     --yt-spec-general-background-c: ${#e9e9e9} !important;
 }
+.ytp-volume-slider-handle {
+    background-color: rgba(255, 255, 255, 0.9);
```
alexanderby: Please check how it looks in Light mode, if it is not good, try using `${source-color}` pattern, that will be inverted and brightness/contrast will be applied. Also you should use `!important` keyword to prevent overrides by other style sheets.

So, it could look like:
`background-color: ${white} !important;`

BEEDELLROKEJULIANLOCKHART: @alexanderby, is "255,255,255" not superior to "white"? It is more consistent, and must be converted to RGB anyway, so it requires less interpretation.

Issue #1430, opened 2019-07-15
Dark Reader overwrites Firefox Themes


Metazolid:
Whenever I enable and disable Dark Reader, it overwrites the function of the previously enabled Firefox theme. I have to manually disable and reenable it for it to work, OR activate Dark Reader before the theme, which I then have to reenable again.
It would be great if the Addon wouldn't interfer with the installed themes.

zacharytyhacz:
What version of firefox are you using? When you say theme, are you talking about the top area ( URL, search, and tabs area ) ?

gibfahn:
(hopefully the same issue as above)

If I disable Dark Reader on Firefox Nightly (macOS version `70.0a1 (2019-07-23) (64-bit)`) and restart Firefox, this page looks like this:

![image](https://user-images.githubusercontent.com/15943089/61779245-40abb400-adf8-11e9-90f2-aabe4638374d.png)

If I then enable Dark Reader (in `about:addons`) it looks like this:

<img width="973" alt="image" src="https://user-images.githubusercontent.com/15943089/61779267-4c977600-adf8-11e9-8681-bc3984752cad.png">

This persists even after I restart Firefox. I can fix it by disabling and then enabling the current theme (in my case Default as I have Dark Mode in my system preferences), but I have to do that on every restart.

gibfahn:
I worked around this by:

1. Running `shadowfox` and choosing `Uninstall Shadowfox` (I previously had it installed but it no longer seems necessary).
2. Going to about:profiles and clicking `Restart with Add-ons Disabled`
3. Then I clicked `Refresh Firefox` at the popup:

<img width="398" alt="image" src="https://user-images.githubusercontent.com/15943089/61780710-eb24d680-adfa-11e9-9261-325afed4bb7d.png">

gibfahn:
The quickest workaround I've found is to go to `about:addons`, click `Themes`, and either `Enable` or `Disable` the `Dark Theme`. Whether it's enabled or disabled doesn't matter if you have Dark Mode enabled systemwide, but the switch fixes the colour issues.

zacharytyhacz:
Interesting, I didn't think this extension would affect the actual browser theme. I'm using the default Mozilla Firefox theme ( version `68.0.1 (64-bit)` ) on macos high sierra `10.13.6`.

However, I'm confident this can be an easy fix if someone were to make create a pull request. I may take a look, though I'm busy.

Naheulf:
[...]

PR #1608, opened 2019-10-06
Fix browser theme setting in Firefox

gibfahn:
The new `frame` and `tab_background_text` properties must be set in
order to avoid setting to defaults in newer Firefox versions (e.g. on
Nightly).

At the moment the old `accentcolor` and `textcolor` properties return a
warning, so we can avoid removing them to work with older versions of
Firefox for as long as possible.

Fixes: https://github.com/darkreader/darkreader/issues/1430
Refs: https://github.com/darkreader/darkreader/commit/181ec067486ab4602d6d1c78ed5be9eac089139d

### DIFF
--- a/src/background/window-theme.ts
+++ b/src/background/window-theme.ts
@@ -13,6 +13,7 @@ const themeColorTypes = {
     accentcolor: 'bg',
     button_background_active: 'text',
     button_background_hover: 'text',
+    frame: 'bg',
     icons: 'text',
     icons_attention: 'text',
     popup: 'bg',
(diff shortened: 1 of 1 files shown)


alexanderby:
I'm surprised that somebody uses this feature. I was going to remove it, since Firefox has many bugs with it, like resetting the theme to default white when the extension is disabled. At the same time built-in dark Firefox theme looks rather good.

Also I remember last year it was not possible to change text selection color, maybe it's fixed now.

gibfahn:
>I'm surprised that somebody uses this feature

It's not so much that I use this feature, it's more that it's enabled by default and then started breaking :).

>I was going to remove it, since Firefox has many bugs with it, like resetting the theme to default white when the extension is disabled. At the same time built-in dark Firefox theme looks rather good.

Agree it's not necessary as Firefox (now) does this natively.

PR #1612, opened 2019-10-07
make youtube comments and links more clear and nice

RsPlayer:


RsPlayer:
changed to white color since it fits better than grey

RsPlayer:
I think the problem of white links and grey comments is only with firefox, closing

PR #1655, opened 2019-10-23
Remove twitch.tv

WoodenMage:
Theyre using black background and pure white, blinding font color. No reason not to be able to change that to more toned and easy on the eyes version like we could all along until recently.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -168,7 +168,6 @@ thealiendrew.github.io
 toneden.io
 totalwarwarhammer.gamepedia.com
 trustreaming.tv
-twitch.tv
 ufplanets.com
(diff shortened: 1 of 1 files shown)


IamGreen112:
Agreed. Made an account on here just to comment because when I went on twitch I was blinded by how bright it was and had to manually set it to dark mode on the site. Don't see why we're forced to  use their ugly option.

TG4LAaron:
There is the builds all i did was remove twitch from the list
[build.zip](https://github.com/darkreader/darkreader/files/3763827/build.zip)
[build-firefox.zip](https://github.com/darkreader/darkreader/files/3763828/build-firefox.zip)

IamGreen112:
Thanks!

Issue #1285, opened 2019-04-29
Support website's native dark mode


rugk:
Firefox 67 introduced the new [`prefers-color-scheme` CSS media feature](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme). (It's not currently available for any other browsers that support browser/WebExtensions.)

With that websites can detect a dark system mode and adjust their design.

It would obviously be great, if you could integrate that, somehow. And there are many possibilities.

---

That said, I've already tried that (to force a particular design) and it is really not a nice thing to implement. At least forcing the black design works, as this is how websites likely implement it, but anything else is likely not really possible…

So this is rather **a proof-of-concept** (although it works quite fine if you just want to force the websites to use their native black design.). 

**[Test the add-on here!](https://addons.mozilla.org/firefox/addon/dark-mode-website-switcher/?utm_source=github.com&utm_medium=issue&utm_content=https://github.com/darkreader/darkreader/issues/1285&campaign=external-github)** 

Source code: https://github.com/rugk/website-dark-mode-switcher/

What is easy is:
* get the system state and apply the add-on based on it. (just a `.matchMedia` query and you know it)

What is hard is:
* forcing websites to use a particular design

You may also just disable this add-on on sites, where you detect it used. It opens many possibilities/ideas…

**Edit:** Due to new Firefox APIs the [v2](https://github.com/rugk/website-dark-mode-switcher/releases/tag/v2.0) of the linked extension is much more powerful and now can easily and reliably force the dark or light mode on websites.

--- 

Test sites are linked in the add-on, but here again:
* https://webkit.org/blog/
* https://stuffandnonsense.co.uk/blog/redesigning-your-product-and-website-for-dark-mode

rugk:
BTW, as the current way is quite hard, I asked Mozilla to implement an API or so for this feature: https://bugzilla.mozilla.org/show_bug.cgi?id=1547818

Please upvote, if you want this, too.

awtrimpe:
Support for __prefers-color-scheme__ will now be available in Chrome 76

https://blog.chromium.org/2019/06/chrome-76-beta-dark-mode-payments-new.html

jacob-j:
Changing the media `screen` variable to `screen and (prefers-color-scheme: dark)`  in the files `src/inject/dynamic-theme/index.ts` and `src/inject/dynamic-theme/style-manager.ts` would solve this for dynamic theme.

[...]

PR #1667, opened 2019-10-28
Global Dark Sites: Added a few Apple news websites

lemeb:
A few Apple news sites that have been making the switch after the release of iOS 13.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -1,5 +1,6 @@
 2600.com
 5stardata.info
+9to5mac.com
 animeonline.su
 app.keeweb.info
 app.plex.tv
@@ -107,6 +108,8 @@ jsfiddle.net
 killtheradio.net
 kissmanga.com
 linux.org.ru
+macrumors.com
+macstories.net
 marte.dev
 mcrpw.github.io
 mcskinhistory.com
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -1,6 +1,8 @@
 2600.com
 5stardata.info
+9to5google.com
 9to5mac.com
+9to5toys.com
 animeonline.su
 app.keeweb.info
 app.plex.tv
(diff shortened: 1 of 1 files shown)


alexanderby:
Hi. These sites support native dark mode, but are not dark by default when system Light appearance is selected. Well, things got complex, I have to think about the purpose of this list, it is becoming useless. Maybe the option is needed to detect native dark mode support and prevent doing anything.

lemeb:
I think that's right, especially since some of the sites that are _currently_ in the list (like the Docker Docs) are not dark by default but offer the option...

I think it might be wise to create two lists:
- `dark_sites.config`: websites that are dark by default anywhere
- `prefers_color_scheme_sites.config`: websites that are dark if the browser communicates a dark mode, which is the case of these websites, [Bugzilla](https://hacks.mozilla.org/2019/05/firefox-67-dark-mode-css-webrender/), the Docker docs, etc. That's where — to take an example — macstories.net would lie.

Let's imagine a case where the user has Dark Reader turned ON globally, "use system color scheme" OFF, and wants to visit, say, macstories.net:
- If the system color scheme sent by the browser `== 'dark'`: don't use `prefers_color_scheme_sites.config`, since macstories.net already knows that the user prefers a dark website.
 - If the system color scheme sent by the browser is `!= 'dark'`: use `prefers_color_scheme_sites.config`. Instead of generating a dark CSS, Dark Reader injects `prefers-color-scheme: dark`, as proposed in #1285, and let macstories.net render its native theme.

I have to mention that I'm on Firefox on Catalina with Dark Mode enabled, and that Firefox is _not_ passing along that I'm on Dark Mode:
- With Chrome and Safari macstories.net is natively dark, 
- But with Firefox, is natively light with Dark Reader "OFF" and artificially dark with Dark Reader "ON".

FranklinYu:
Bugzilla implemented it in a simple way, but Docker docs site seems to have a UI switch for users to override this settings.

> * If the system color scheme sent by the browser `== 'dark'`:
> 
>     * don't use `prefers_color_scheme_sites.config`, since macstories.net already knows that the user prefers a dark website.

By "don't use it" did you mean "treat it the same as `dark-sites.config`"?

lemeb:
> By "don't use it" did you mean "treat it the same as `dark-sites.config`"?

Yep! When the user goes on the website, it is dark by default, so our job is done here to a degree. If the user prefers to activate light mode on the website in question, well, free to them!

PR #1673, opened 2019-10-29
Update for almost invisible highlights.

Myshor:
Polar Relieve shows some stats and it is almost invisible in default dark settings. This will make them black (like they are without dark mode). You can check https://flow.polar.com/training/relive/4002051395 to see the problem. ;)


Myshor:
This changes are included in my next request.

PR #1682, opened 2019-10-31
Fix github search bar background color

audip:
Github search bar color fixed from being blended with the navbar to have a black background that makes it stand out and add a rounded border to conform with GitHub design


audip:
Visual diff for the proposed changes, please take a look at before and after screenshots

Before:
image

After:
image

Issue #558, opened 2018-06-11
[Dynamic] Github contribution graph not very readable


krupenja:
See screenshot, Dynamic theme. Also note that the colors in the legend below the graph do not correspond to the colors on the graph. Looks better with Static, Filter and Filter+.

![image](https://user-images.githubusercontent.com/26354715/41255373-fccc0efc-6dce-11e8-9f98-0fd421453cd2.png)

gilbertohasnofb:
> Looks better with Static, Filter and Filter+.

It does look better, but still with a lot of problems: 

* Static mode does a bit better with the grid, but makes the rest of the page looking buggy:

![image](https://user-images.githubusercontent.com/5700795/49329683-84f0f800-f57a-11e8-9498-f58aab99ff10.png)

![image](https://user-images.githubusercontent.com/5700795/49329686-9508d780-f57a-11e8-9f27-2180e9a78bfd.png)

* Filter mode makes the grid look quite better, except that the scale becomes inverted, the lighter the green the more contributions an user made in a day:

![image](https://user-images.githubusercontent.com/5700795/49329696-c2ee1c00-f57a-11e8-86a0-108f4b6ef96d.png)

* Filter+ makes the page become fully black with no text nor images for me.

krupenja:
@gilbertohasnofb As a temporary workaround, you can use https://github.com/StylishThemes/GitHub-Dark to generate a custom dark theme. then, in Dark Reader, you can enable Static theme, open Dark Reader CSS Editor and paste your generated theme there. This will get you a decent-looking contribution graph:

![image](https://user-images.githubusercontent.com/26354715/49534249-bc88d880-f8c9-11e8-9652-9203479a5262.png)

gilbertohasnofb:
@krupenja Thanks for the tip, I will check that out!

Mrgove10:
Any news on when this will be fully implemented in darkreader ?

PR #1701, opened 2019-11-06
[Dynamic] Recolored github.com profiles' contrib grids.

TheDaemoness:
This PR also changes the legend at the bottom of the contribution grids to match. The means by which this is done is VERY hacky, but it preserves the use of Github's own logic to determine which calendar cells get which colors.

As for why grayscale,
* It looks better on dark mode than the original colors do.
* It translates to light mode very well.
* It's (arguably) less-opinionated about what the color scheme should be than most other choices.


alexanderby:
Fixed #558

PR #1708, opened 2019-11-08
feature: improve timely tracking overview

FinnBueno:
Removed timeline background image to get rid of ugly dotted white lines.
Inverted stopwatch icon back to white.


FinnBueno:
Here are before and after screenshots.
**Before**
image
**After**
image
I initially wanted to edit the dotted lines to be darker, but it turns out this entire background is an image, so I decided to remove them instead.

PR #1720, opened 2019-11-11
QUICKFIX!!! facebook.com (i) siteinfo icon

Myshor:
There was too many changes in my last fixes.
it seems ._34k2 was related somehow with something I removed earlier.


Myshor:
@alexanderby : merge before release... I found one thing I needed to remove in my last fixes. ;)

PR #1722, opened 2019-11-12
Add DuckDuckGo to the global dark sites list

yanisvieilly:


yanisvieilly:
Apologies, I added it as they now detect whether your OS is set to light or dark mode, and set the DDG page theme accordingly. But it won't be dark by default if your OS is set to light mode.

PR #1723, opened 2019-11-12
All ::placeholder brightness reduce.

Myshor:
As I found many pages with placeholder having color almost the same as entered text and invert can make some pages ugly then I propose brightness reduction.
It seems to be best solution. No color changes for dynamic theme and making it a little bit darker to see it's not text already entered in the text field.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -9,6 +9,9 @@ CSS
     border-color: ${#c59d00} !important;
     color: ${#302505} !important;
 }
+::placeholder {
+    filter: brightness(50%) !important;
+}
 
 ================================
 


alexanderby:
Maybe something like `opacity: 0.5` could be better.

Myshor:
Changing to your proposal in my dev tools in dark reader... Seems to work fine too. :)

alexanderby:
I think that using opacity should have less performance issues. Though I don't know how to measure it. But usually Chrome and especially Firefox have filter bugs. Also `75%` opacity could look better for some websites. Later I will try to fix gamma by using non-linear color scale.

Myshor:
I checked my cornercase site and opacity: 0.5 is tolerable. 0.75 looks a little bit to bright.
On other sites it seems to be better solution. It's not darkening it that much if background is not black in dark mode.

If you want to check my corner case: https://myrooms.ryanair.com/account/register-info/independent

I'll make change in pull request.

alexanderby:
Feel free to submit a new PR. Anyway I will release it in the morning.

Myshor:
Ok... Closing this PR.

PR #1726, opened 2019-11-13
facebook.com - text message with background

Myshor:
._4a6n
invert for text color for messages with background...
Now it ends almost always with white and sometimes it's hard to read.
Checked invert with some messages and seems to work nice.


Myshor:
Abandon fix - it inverts emot icons in post too.

PR #1732, opened 2019-11-14
Fix #1621: Fix the Duplication of text for all shortcuts

### LINKED ISSUE #1621: "Toggle current site" shortcut duplicates when clicking another button.
When using the extension on Firefox and interacting with the extension overlay there is a visual bug when clicking the "Toggle current site" shortcut which causes it to duplicate.

Peek 2019-10-10 07-17

RyanWils:
Steps to solving this issue 

When another button is clicked it 
In this function listed below it matches the parents/siblings 
image
It then calls the insertNode function and inserts the shortcut message again 
image
When its sets the shortcut it goes into this function
image

The code I added does the following
    //create variables for both the toggle switches
    const shortcutClass = document.getElementsByClassName("shortcut");
    const toggleCurrent = document.getElementsByClassName("shortcut").item(0);
    const toggleExtension = document.getElementsByClassName("shortcut").item(1);

    //If all the toggles come back in the class, it means that the shortcuts exist and its not the 
    creation
    if (shortcutClass.length == 3) {
        //Check to see if there is any repeats of Toggle 
        if (toggleCurrent.innerHTML.lastIndexOf("Toggle") > 5) {
            //if there is repeats then take the first instance of the string (Example of what the string 
            looks like "Toggle extension\nAlt+Shift+DToggle extension\nAlt+Shift+D")
            toggleCurrent.innerHTML = 
            toggleCurrent.innerHTML.slice(0,toggleCurrent.innerHTML.indexOf("Toggle",1));
        }
        //Same logic for the extension button 
        if (toggleExtension.innerHTML.lastIndexOf("Toggle") > 5) {
            toggleExtension.innerHTML = 
            toggleExtension.innerHTML.slice(0,toggleExtension.innerHTML.indexOf("Toggle",1));
        }
    }

### DIFF
--- a/src/ui/controls/shortcut/index.tsx
+++ b/src/ui/controls/shortcut/index.tsx
@@ -89,8 +89,8 @@ export default function ShortcutLink(props: ShortcutLinkProps) {
         });
     }
 
-    function onUpdate(props) {
-        props.textContent = shortcutMessage;
+    function onRender(node) {
+        node.textContent = shortcutMessage;
     }
 
     return (
@@ -98,7 +98,7 @@ export default function ShortcutLink(props: ShortcutLinkProps) {
             class={cls}
             href="#"
             onclick={onClick}
-            updated={onUpdate}
-        >{props.textTemplate(shortcut)}</a>
+            attached={onRender}
+        ></a>
     );
 }
--- a/src/ui/controls/shortcut/index.tsx
+++ b/src/ui/controls/shortcut/index.tsx
@@ -18,18 +18,7 @@ interface ShortcutLinkProps {
 export default function ShortcutLink(props: ShortcutLinkProps) {
     const cls = mergeClass('shortcut', props.class);
     const shortcut = props.shortcuts[props.commandName];
-    const shortcutClass = document.getElementsByClassName("shortcut");
-    const toggleCurrent = document.getElementsByClassName("shortcut").item(0);
-    const toggleExtension = document.getElementsByClassName("shortcut").item(1);
-
-    if (shortcutClass.length == 3) {
-        if (toggleCurrent.innerHTML.lastIndexOf("Toggle") > 5) {
-            toggleCurrent.innerHTML = toggleCurrent.innerHTML.slice(0,toggleCurrent.innerHTML.indexOf("Toggle",1));
-        }
-        if (toggleExtension.innerHTML.lastIndexOf("Toggle") > 5) {
-            toggleExtension.innerHTML = toggleExtension.innerHTML.slice(0,toggleExtension.innerHTML.indexOf("Toggle",1));
-        }
-    }
+    const shortcutMessage = props.textTemplate(shortcut);
 
     let enteringShortcutInProgress = false;
 
@@ -100,11 +89,16 @@ export default function ShortcutLink(props: ShortcutLinkProps) {
         });
     }
 
+    function onUpdate(props) {
+        props.textContent = shortcutMessage;
+    }
+
     return (
         <a
             class={cls}
             href="#"
             onclick={onClick}
+            updated={onUpdate}
         >{props.textTemplate(shortcut)}</a>
     );
 }
--- a/src/ui/controls/shortcut/index.tsx
+++ b/src/ui/controls/shortcut/index.tsx
@@ -18,6 +18,18 @@ interface ShortcutLinkProps {
 export default function ShortcutLink(props: ShortcutLinkProps) {
     const cls = mergeClass('shortcut', props.class);
     const shortcut = props.shortcuts[props.commandName];
+    const shortcutClass = document.getElementsByClassName("shortcut");
+    const toggleCurrent = document.getElementsByClassName("shortcut").item(0);
+    const toggleExtension = document.getElementsByClassName("shortcut").item(1);
+
+    if (shortcutClass.length == 3) {
+        if (toggleCurrent.innerHTML.lastIndexOf("Toggle") > 5) {
+            toggleCurrent.innerHTML = toggleCurrent.innerHTML.slice(0,toggleCurrent.innerHTML.indexOf("Toggle",1));
+        }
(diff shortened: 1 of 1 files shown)


src/ui/controls/shortcut/index.tsx
```diff
@@ -18,6 +18,18 @@ interface ShortcutLinkProps {
 export default function ShortcutLink(props: ShortcutLinkProps) {
     const cls = mergeClass('shortcut', props.class);
     const shortcut = props.shortcuts[props.commandName];
+    const shortcutClass = document.getElementsByClassName("shortcut");
```
alexanderby: Hi, thanks for the provided solution. However it looks like a hack. The `ShortcutLink` class should know nothing about its implementations.

src/ui/controls/shortcut/index.tsx
```diff
@@ -18,6 +18,18 @@ interface ShortcutLinkProps {
...
+    const toggleCurrent = document.getElementsByClassName("shortcut").item(0);
+    const toggleExtension = document.getElementsByClassName("shortcut").item(1);
+
+    if (shortcutClass.length == 3) {
+        if (toggleCurrent.innerHTML.lastIndexOf("Toggle") > 5) {
+            toggleCurrent.innerHTML = toggleCurrent.innerHTML.slice(0,toggleCurrent.innerHTML.indexOf("Toggle",1));
```
alexanderby: Use of `innerHTML` is discouraged.

RyanWils:
@alexanderby 
Hi Alexander, Thanks for reviewing this Pull Request. 
Steps to why this occurs,
1. Initial variable is set to node.textcontext
2. It creates the node with the text ("Toggle Extension\nCtrl+D") when you click on the button cause it re renders the elements. 
3. it sets it the node.textcontext to the initial in the OnBlur function.  

This OnBlur function was mainly used to reset the text when you click on the toggle and then click off it.
I will look deeper into this issue.

src/ui/controls/shortcut/index.tsx
RyanWils: My logic for this issue consists of the following
1. If node.content is equal to "..." then set it to the initial text
2. If node.content is equal to "Toggle" then do not set the node context node as this means that it was already created by the re render function.

But the issue that I ran into was that there was no actual way to get the node.content so I grabbed the InnerHTML.

src/ui/controls/shortcut/index.tsx
RyanWils: From my knowledge props.shortcuts just stored the shortcuts to toggle the site, not the node.content.
Notice this behaviour 
image
I added a test message to the end of the line
image

What I assume it did 
1.set the text of the node  "Toggle Extension\nCtrl+D"
2.append the initial "Toggle Extension\nCtrl+Dtest"

String is now - "Toggle Extension\nCtrl+DToggle Extension\nCtrl+Dtest"

1. inital = node.content 
1.set the text of the code "Toggle Extension\nCtrl+D"
2. append the initial "Toggle Extension\nCtrl+DToggle Extension\nCtrl+Dtest" + "test" 

String is now "Toggle Extension\nCtrl+DToggle Extension\nCtrl+DToggle Extension\nCtrl+Dtesttest"

Refer to my comment below

alexanderby:
[Here](https://github.com/alexanderby/malevic/blob/master/tests/dom.tests.tsx) are various DOM tests. But they are run in js-dom environment, so maybe the problem will not be reproducible. Also, maybe it is expected problem if you mix virtual DOM with real DOM manipulation.

So, maybe let's not use Virtual DOM for text content in this case and always stick to using `textContent` property. Malevič.js has ability to pass a function, that will be executed after parent DOM node was updated, like:
```jsx
function Component(props) {
    return <div>
        {(domNode) => domNode.textContent = props.text}
    </div>;
}
```

RyanWils:
@alexanderby Hi Alexander,
Sorry for the late reply. I have been exploring through the Malevič.js documentation and I am pretty new to using Malevič.js, so there's a lot of grey areas.
As you talked about not sure if its the same thing, but the after the parent nodes are updated it will pass a function. 
I have submitted another commit which fixes the duplicates text using the OnUpdate function.
Not sure if this is an appropriate fix, feedback would be appreciated.

src/ui/controls/shortcut/index.tsx
```diff
@@ -88,11 +89,16 @@ export default function ShortcutLink(props: ShortcutLinkProps) {
         });
     }
 
+    function onUpdate(props) {
```
alexanderby: Let's rename this function to `on Render` (see why later). `props` is an HTMLElement (HTMLAElement to be more precise), we can rename the argument and add a type to it (since we are using TypeScript).

src/ui/controls/shortcut/index.tsx
```diff
@@ -88,11 +89,16 @@ export default function ShortcutLink(props: ShortcutLinkProps) {
...
+        props.textContent = shortcutMessage;
+    }
+
     return (
         <a
             class={cls}
             href="#"
             onclick={onClick}
+            updated={onUpdate}
```
alexanderby: Let's add `attached={onRender}`. `attached` handler is invoked after DOM element is created and attached to DOM for the first time, `updated` is called after DOM was synchronized with virtual DOM. I had an idea to add `rendered` handler that would be called in both cases.

src/ui/controls/shortcut/index.tsx
```diff
@@ -88,11 +89,16 @@ export default function ShortcutLink(props: ShortcutLinkProps) {
...
         >{props.textTemplate(shortcut)}</a>
```
alexanderby: The problem was in this line, it creates a string and a virtual text node is created for it. But since we use real DOM manipulations, let's make text empty. We can just leave it as `<a ... />`.

RyanWils:
I have put in your requested changes. 
It seems that the attached function is used in other parts of the code and the parameter was not casted for example
image
To keep the code consistent I did not cast (HTMLAnchorElement) to the node.
image

PR #1738, opened 2019-11-15
pyszne.pl - circle with estimated time invert

Myshor:
When you wait for your order there was unreadable circle with estimated time.


Myshor:
Adde fix for dark background in logos in promotions. Most logos were black or very dark.
Invert was not good solution as it was breaking colors in pictures.

PR #1742, opened 2019-11-16
Update dark-sites.config

N190392:
Added:
sportsurge.net
summertimesaga.com


alexanderby:
Sorry, I cannot add some websites to the list. In my country it could be considered as a distribution of adult content and punished up to 3 years of prison.

Kurtoid:
Out of curiosity, what country?

PR #1744, opened 2019-11-17
linkedin.com - two logos invert

Myshor:
Linked text was black and unreadable.


Myshor:
Added fix for black company logos dark background using CSS.

PR #1755, opened 2019-11-18
Update dark-sites.config

merajmasuk:
Added typescript


alexanderby:
Hi, sorry, but this website is bright by default. But this list should include dark sites only.

merajmasuk:
Understood

PR #1762, opened 2019-11-20

cartond:
Adding https://www.bigocheatsheet.com/ but not for the red-outlined box


cartond:
Forgot an invert

PR #1771, opened 2019-11-22
::placeholder opacity change

Myshor:
Fix for almost white placeholder text when background is black in darkmode.

See https://myrooms.ryanair.com/account/register-info/independent as example.


alexanderby:
Good :)

Issue #490, opened 2018-05-11
Google Sheets is changing text color


jeff275:
![screen shot 2018-05-11 at 8 16 23 am](https://user-images.githubusercontent.com/39192561/39926048-a1408d72-54f3-11e8-9b7e-0dc527b2cb9b.png)

arthurblake:
What I've noticed happening is when you type text into a new cell in Google Sheets, the color comes out unreadable/invisible. Seems to be something caused by the new dynamic theme. Maybe the dynamic theme is not picking up a dynamic CSS color change. If you copy one cell to another or reset the color (with google sheets) it fixes it. But it's annoying if you are typing in a lot of data into the sheet.

jeff275:
Yes, exactly. It's a lot of work on highly used spreadsheets.

krupenja:
Quick workaround for this is to select all affected cells and set colour to black from Sheets menu bar. Obviously only works if you only use black colour for text.

jeff275:
It's less work to just disable Dark Reader on the Google Sheets.

rtman:
Just noticed this too, its super frustrating

electricityforprogress:
I believe this is due to the mode which you are using.  Change the settings to Filter+ under DarkReader --> more --> filter+, I only use this setting with docs.google ...  Same thing happens with some Gmail emails and Calendar which you may want to toggle settings for those sites too.

![darkreader_filterplus](https://user-images.githubusercontent.com/22551283/46906382-a0237d80-ced0-11e8-8025-d9ecdcfeadc5.png)

Fabian42:
Confirmed, switching to "Filter+" for docs.google.com works around the issue. But it's still a valid bug, isn't it? There's no reason why the "Dynamic" filter should change the text colour, so this issue should be reopened.

Fabian42:
Ah, ok. I just saw a red "Closed" label and thought it referred to this one.

MavropaliasG:
Can confirm, please fix!

ufo2mstar:
The "Filter +" theme engine works perfectly for sites like Google Sheets (and even github, it looks like)..
So would a feature request like defaulting the experimental theme engine to 'known-issue' sites be more apt for the fix? @Fabian42 @darkreaderdev

Fabian42:
Can you please reword that? I don't understand what you mean.

alexanderby:
It is planned to fix the issue this month.

[...]

PR #1772, opened 2019-11-22
🎨 Fixes Dynamic Google Sheets input color

favorable-mutation:
There is currently an issue with the Dynamic theme engine where, upon text entry to a cell in Google Sheets, the input text will become a very light color that is barely visible in either Dark or Light mode.

These changes prevent that from happening for the three available modes of text entry (click on cell and type, double-click on cell and type, click on cell then click on top bar and type), and are scoped to specifically affect a cell being edited in Google Sheets.

Likely, a change to the actual code for the Dynamic theme engine is in order, but it seems to me that a workaround is in order, given how long the bug has existed.

See https://github.com/darkreader/darkreader/issues/490 for more information.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -334,6 +334,17 @@ INVERT
 .docs-homescreen-icon
 .kix-equation-toolbar-icon
 .kix-equation-toolbar-palette-icon
+.cell-input
+
+CSS
+.cell-input {
+  background-color: ${black} !important;
+  color: ${white} !important;
+}
+
+.cell-input > span, .cell-input > font {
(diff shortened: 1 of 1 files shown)


alexanderby:
Glad to see a workaround. Did you check what happens when you open the document without Dark Reader after you've made some changes with it? Did you check if text color is editable?

favorable-mutation:
Yup! After editing with Dark Reader on, closing the Google Sheet, disabling the extension, and turning it back on, the text is black (on white background). Text colors are editable both with Dark Reader enabled and without.

favorable-mutation:
(Rebased off of master)

alexanderby:
One bug I've found was when you edit a colored text, it resets. But it's much better than it used to be before.

PR #1781, opened 2019-11-24
Add several sites to exclusion config

crse:
Imageboard
- Arisuchan
- Lainchan

Forum
- ~~Lewd.sx~~

Gaming
- Gamejolt

SNS
- Imgur

~~R18~~
- ~~Exhentai~~
- ~~Nhentai~~

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -79,7 +79,6 @@ enlightenment.org
 envs.net
 erai-raws.info
 eternal.abimon.org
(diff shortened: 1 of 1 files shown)


alexanderby:
Sorry, I cannot add adult websites to this list. It can be considered as a distribution and punished in my country.

crse:
Updated!

alexanderby:
Thanks, it can sound paranoidal, but there were real cases.

Issue #1785, opened 2019-11-24
Please do not change scroll bar because it makes scroll bar invisible. And that a big problem.


PranavBhattarai:
Take this GitHub page for example: 
![Screenshot was taken from Ubuntu 19.10 of Firefox browser](https://user-images.githubusercontent.com/32763112/69497916-08cd8e00-0f0a-11ea-8afb-c8571c796ea0.png)

What I did here is, I increased the **Contrast** by **50+** (which is the maximum allowed contrast).
Increasing the contrast by that much on OLED (or even in LCD) screen, make Letters shinner and more visible. It looks more vibrant and sexy.

But the downside of this is that it makes the scrollbar invisible. We a cool extension with cool options user can toggle but this is a really sad downside.

**My point is there should be a way around this to make the scroll not increase its blackness when a user increases the contrast of webpage from its default value to prevent going invisible.**

Issue #1592, opened 2019-09-30
Scrollbar is wrongfully inverted with a dark theme in XFCE4


Nicram01:
I use Firefox on a Linux XFCE4 Desktop environment. Some websites implement a scroll bar. The scroll bar on my XFCE4 theme (Blackbird) is already dark, and when I turn the dark reader on the scroll bar is wrongfully inverted. Is there a way to reverse the invert on all pages of the scroll bar.

![Wrongfull_Invert](https://user-images.githubusercontent.com/31846298/65882791-40dea780-e385-11e9-9bb8-c760107f8640.png)

Issue #1343, opened 2019-05-24
Scrollbar barely visible in Chrome


Raijin56:
Hi,

First of all, thank you very much for Dark Reader. It works awesome.

I installed the Chrome plugin to my father's laptop and yesterday he told me that he is not able to see the scrollbar. He has 63 years and can't see it. I have 27 and I also find very difficult to see it most times, because there is very little contrast between the background of the scrollbar and the bar itself.

Would you be so kind as to change the colors of the scroll bar to improve the contrast?

As an example, the scrollbar of the new dark mode of Windows 10 is completely visible and the colors look good. Maybe it can be used as a reference.

Thank you very much and kind regards!

Raijin56:
I taken the colors of the Windows' scrollbar and transferred to Dark Reader. I'm not a designer so the colors will not match perfectly with the ones of Dark Reader, but the scrollbar is completely visible now., so they are a good start. I also restored the original width of the scrollbar, that in stock Dark Reader is reduced for some pointless reason.

![image](https://user-images.githubusercontent.com/1304436/58768234-44f5e680-8598-11e9-8470-94a3cf1fcefe.png)

Here are so someone could test and improve them and/or submit to Dark Reader.

I import them opening the Dark Reader contextual menu > Tools. Then, at the top of the document, under * and CSS, there are two entries named ".jfk-bubble" and ".vimvixen-hint {", and a separator line "======". Put the following code just before the separator:

`/* bar background*/
::-webkit-scrollbar {
    background-color: #151515;
    /*color: #b4b1ab;*/
    width: 1rem;
}
/* bar */
::-webkit-scrollbar-thumb {
    background-color: #454545;
}
/* bar when mouse is hover it */
::-webkit-scrollbar-thumb:hover {
    background-color: #7a7a7a;
}
/* bar when pressed and moved */
::-webkit-scrollbar-thumb:active {;
    background-color: #a6a6a6;
}
/* nothing apparently
::-webkit-scrollbar-corner {
    background-color: #323334;
}
*/
/* nothing apparently
* {
    scrollbar-color: #3f4142 #353637;
}
*/`

Raijin56:
4,5 months ago I opened this issue and I event spent my time in providing a solution to make it easy for the developer to implement it. It was completely ignored by the developer and could've been resolved just copy pasting the solution. It was for the accessibility of an older person, not a whim. A whim was the stupid decision of dwarfing the scroll bar, in the first place.

[...]

Issue #1323, opened 2019-05-15
Scroll bar won't show with contrast all the way up


dwlsalmeida:
Scroll bar won't show with contrast all the way up, so it is impossible to know where we're at in a page.
![Here's an attached picture for reference.](https://user-images.githubusercontent.com/11738710/57781255-2d95ad00-7700-11e9-83db-755fb5486814.png)

hamdyl052:
The workaround is to set the theme engine non-Dynamic, e.g. Filter, Filter+, or Static. Go to tab More, then select a theme engine.
Another workaround is use Light instead of Dark. It's better than glaring white background.

Gusted:
#2402

PR #1793, opened 2019-11-26
Don't style the scrollbar

frozencemetery:
In many cases, applying our style to the scrollbar will override user
styles or OS theme settings, resulting in an unusuable scrollbar.
Investigated by ebbit1q.

Signed-off-by: Robbie Harwood <rharwood@redhat.com>
Resolves: #1785
Resolves: #1592
Resolves: #1343
Resolves: #1323
Resolves: #858


frozencemetery:
As a workaround until this merges, firefox users will want to look at https://github.com/darkreader/darkreader/issues/858#issuecomment-467411405 (webkit/chromium users should be able to adopt a similar fix looking at the source, I think).

alexanderby:
Hi. Some users want to style scrollbars, and some not. So, some option is needed. Currently I'm reworking the UI and have plans for changing the settings structure.

Issue #1273, opened 2019-04-27
Math formulas on a lot of pages are invisible


Fabian42:
Similar to #649. Might be caused by #356.

There seems to be a certain way of embedding Math formulas that many websites use that doesn't work well with Dark Reader. Often it's completely invisible (black font on black background), sometimes a slight shadow is visible, but usually not enough to properly read the formula.

Some examples

* https://en.wikipedia.org/wiki/Invertible_matrix
![image](https://user-images.githubusercontent.com/10481506/56844922-f2e6e480-68b8-11e9-84ff-60979dcdd61a.png)
Here the "Filter+" mode fixes it, but in the default it's invisible.

* https://elearning.physik.uni-frankfurt.de/data/FB13-PhysikOnline/lm_data/lm_281/modul_2/teil_8/node48.html
![image](https://user-images.githubusercontent.com/10481506/56844929-127e0d00-68b9-11e9-9234-995238e2de42.png)
Here all four modes look similarly bad. The only way to read the formula is to disable Dark Reader for this site.

There are also cases where all four modes have completely invisible formulas, I just didn't find any right now.

NewbieToEverything:
Similar issue encountered, I use Wikipedia as mathematical reading material quite often and this bugs me very much. Hope there exists a fix.

Just-Harry:
Perhaps a configuration setting could be put in place for this?
To not modify the `background-color` of images, unless the alpha value of the computed `background-color` of the image is below a certain value, in which case set the `background-color` of the image to the original (non-inverted) `background-color` of its parent-element.

---

As a personal ad-hoc solution, I'm using a bookmarklet which toggles the `background-color` of all images between white, and their original `background-color`:

`javascript:(/* Licence: CC0 */ function () {for (let img of document.body.getElementsByTagName('img')) {img.style.setProperty('background-color', img.style.getPropertyValue('background-color') === 'white' ? 'initial' : 'white', 'important');}})();`

Although, one could also use Dark Reader's dev tools to make the `background-color` of all images white by placing the following rule in the CSS section of the very first `*` URL selector.

```
img {
    background-color: white !important;
}
```

rkanati:
Bizarrely, removing the math-related classes under `INVERT` for `wikipedia.org` in the style editor fixes this for me - can someone else try?

Just-Harry:
> 
> 
[...]

PR #1805, opened 2019-12-03
🧮 Adds a CSS rule to invert Wikipedia math

favorable-mutation:
As described in https://github.com/darkreader/darkreader/issues/1273, there is currently an issue with the Dynamic theme that causes math rendered in a specific way (particularly on Wikipedia) to be near-invisible.

These changes invert such elements specifically for Wikipedia, though a larger fix addressing the root cause of the issue may eventually be necessary.

I've checked in Chrome that, after the change is applied, the math elements are visible (white-on-black) on the page [cited in the issue](https://en.wikipedia.org/wiki/Invertible_matrix) and [another page](https://en.wikipedia.org/wiki/Binary_search_algorithm) in Dark Mode, and checked that the elements are still visible (black-on-white) when switching to Light Mode.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1997,6 +1997,9 @@ div .thumbimage[src$=".png"],
 div .thumbimage img[src$=".png"] {
     background-color: white;
 }
+.mwe-math-element {
+    filter: invert(1);
+}
 
 ================================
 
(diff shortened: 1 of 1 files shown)


alexanderby:
I noticed a strange thing, that your selector in CSS section is exactly the same as in INVERT selection. But then I realized that all inversion selectors for Wikipedia were broken for 3 month due to Firefox-specific fix.

PR #1808, opened 2019-12-04
Fix filter for chrome extension "bookmark sidebar"

ars-gh:
Prevent invert the extension's iframe.
Target extension is https://chrome.google.com/webstore/detail/bookmark-sidebar/jdbnofccmhefkmjbkkdkfiicjkgofkdh

### DIFF
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -12,6 +12,7 @@ svg image
 [style*="background-image: url"]
 [background]
 twitterwidget
+iframe#blockbyte-bs-sidebar
 
 NO INVERT
 [style*="background:url"] *


alexanderby:
Hi, could you please explain, what are you trying to achieve? Do you want the whole page look black, and this menu white?

cyse-ars:
No, dark reader overwrite plugin's bg color which show in iframe.

[when disabled](https://imgur.com/r3EKBIN)
[when enabled](https://imgur.com/YwKPsQW)
[when enabled and patched](https://imgur.com/xTo3G7D)

This problem causes in specific users only, who using extensions, this and "bookmark sidebar".
So if project not support other extensions, please kick this request.

alexanderby:
When I checked, the background of the extension was white. Is it configurable?

cyse-ars:
Yes

[open setting](https://imgur.com/nMDK27l)
[toggle this opt](https://imgur.com/PgHszPa)

alexanderby:
Hi @ars-gh, sorry for the delay. I think that if the panel is white by default and Dark Reader works as expected, then no changes are necessary. In future I will add a CSS section into settings, so that users could make their fixes for rare cases.

ars-gh:
Thanks for considered, and suggesting future feature. 😄

PR #1813, opened 2019-12-05
gazetawroclawska.pl - .atomsArticleTile__tileImg

Myshor:
fix for .atomsArticleTile__tileImg
invert when in selected areas - this does not invert tiles in gallery mode.


Myshor:
Added invert for .atomsPromotionsQuiz__imageWrapper
to fix inverted colors of image.

PR #1814, opened 2019-12-05
naszemiasto.pl - .atomsArticleTile__tileImg

Myshor:
fixes for .atomsArticleTile__tileImg - no more inverted images in gallery


Myshor:
Added invert for:
.atomsPromotionsQuiz__imageWrapper
.atomsPromotionsLink__imageWrapper

Including new fixes from gazetawroclawska.pl
As they have exactly the same template and classes on their page.

PR #1819, opened 2019-12-07
Update theme for courses.fit.cvut.cz

janpicka:


alexanderby:
Hi, @janpicka, if you have time, please check how your fix works in light mode. Instead of specifying colors like `#abc !important` try using something like `${src_color} !important`, where `src_color` is a color, that element has without Dark Reader. This pattern will be substituted with changed color.

PR #1824, opened 2019-12-10
linkedin.com - fix invert for .mb3

Myshor:
Invert fix for .mb3 - Added :.not(.container-with-shadow)


Myshor:
Removed one dot that should not be included.

Myshor:
.org-jobs-recently-posted-jobs-module added to exclude in .mb3 invert

PR #1826, opened 2019-12-10
flow.polar.com - invert for icon in weekly sleep

Myshor:
added invert .sleep-chart-yaxis.end for icon in weekly sleep chart


Myshor:
Added CSS fix for group black icons with dark background.
.card__item-icon--rounded img

PR #1828, opened 2019-12-11
Fix yellowy background on Basecamp

deathau:
Dynamic theme on Basecamp chose an ugly, dusty, yellowy colour for the background by default: rgb(39, 29, 22)
Changed that to pure black, along with the navbar and fixing the shadow.


alexanderby:
Hi @deathau, are you sure that URL should be 3.basecamp.com?

deathau:
Pretty sure, yeah. This is specifically for "Basecamp 3" which runs of the "3.basecamp.com" domain.

alexanderby:
Great. I've made a change to make the fix look correct in light mode, but didn't visit the website. Please, let me know if there's an error.

PR #1829, opened 2019-12-12
Lighter background for devblogs.microsoft.com

yurikus:
Black is too black everywhere, set lighter background for devblogs.microsoft.com:

.postcontent, #single-wrapper{
    background-color: #282828 !important;
}

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -308,6 +308,15 @@ video#vjs_video_3_html5_api.vjs-tech {
 
 ================================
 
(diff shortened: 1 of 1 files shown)


alexanderby:
Hi, maybe let's try `transparent` color?

yurikus:
Is there a way configure white inversion -> specified color globally?

alexanderby:
Not yet, but it is planned to add ability setting colors for neutral background and text.

alexanderby:
I mean, that if you set `transparent` in this case, then the brightness, contrast and inversion could be configured.

PR #1846, opened 2019-12-22
Fix Yandex Translate

Rakleed:
I fixed many icons of Yandex Translate and its internal pages.

It turned out quite a few lines. If you can simplify this, let me know.

And I noticed a bug: if I open the translate.yandex.com/collections page, when I close the Collections, I automatically go to the translate.yandex.com main page, but at the same time my INVERTs do not apply to the main page (since the page does not reload). I also do not know how to fix this.


alexanderby:
Hi, thanks. There could be a problem analyzing icons, packed into singe file. It's quite complex problem to analyze such images.

PR #1851, opened 2019-12-23
Fix Google Apps for Google Search

Rakleed:
Fixed invert Google Apps to work for all languages.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -708,6 +708,29 @@ CSS
 
 ================================
 
+google.*
+google.*.*
+
+INVERT
+.gb_hc
+.gb_ec
+.gb_x.gb_Vb
+.gb_x.gb_Ub
+.gb_0b
+#dictionary-modules img
+a.gb_b > div
+.gb_D.gb_oc
+
+CSS
+.RNNXgb {
+    box-shadow: 0 0 2px 0 ${rgba(0,0,0,0.16)}, 0 0 0 1px ${rgba(0,0,0,0.08)} !important;
+}
+.Gor6zc {
+    background-color: white !important;
+}
+
+================================
+
 google.*/maps
 google.*.*/maps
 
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -708,6 +708,29 @@ CSS
 
 ================================
 
+google.*
```
alexanderby: Hi @Rakleed, please have a look at `www.google.*` and `www.google.*.*` and move your fixes there.

src/config/dynamic-theme-fixes.config
```diff
@@ -708,6 +708,29 @@ CSS
...
+#dictionary-modules img
+a.gb_b > div
+.gb_D.gb_oc
+
+CSS
+.RNNXgb {
+    box-shadow: 0 0 2px 0 ${rgba(0,0,0,0.16)}, 0 0 0 1px ${rgba(0,0,0,0.08)} !important;
+}
+.Gor6zc {
+    background-color: white !important;
```
alexanderby: Please double check if `${color}` pattern should be used to make it look good with light color scheme.

src/config/dynamic-theme-fixes.config
```diff
@@ -2092,29 +2115,6 @@ body, .BA-topbg {
 
 ================================
 
-www.google.*
```
alexanderby: I think that `www.` should be still there for Google Search, since there are absolute different products hosted on other subdomains.

Rakleed: Why add `www.` If everything works fine even without them?

src/config/dynamic-theme-fixes.config
Rakleed: I did not change this part of the code, and I don’t know how it works.

alexanderby: OK

src/config/dynamic-theme-fixes.config
alexanderby: It's because google.com can have many subdomains, and if they are not listed in config, then `google.*` fixes will be applied to them. But we don't know for sure if they are not breaking something on these subdomains.

Also I'm going to make more consistent behavior, treat `google.com` and `*.google.com` as different websites.

Rakleed: OK

PR #1859, opened 2019-12-25
Fix border color on www.baidu.com with a Tampermonkey Script

Scriptbeta:
(!)  Fixes has no effect on the webpages without the script.

The Tampermonkey script that conflict:
https://github.com/langren1353/GM_script/
https://github.com/langren1353/GM_script/issues/51

Fixed websites:
www.google.*     www.baidu.com     bing.com

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -222,6 +222,15 @@ CSS
 
 ================================
 
+bing.com
+
+CSS
+.na_cnt .nws_itm, .nws_itmb, #b_content #b_results li, body #b_header {
+    border-color: rgb(55,55,55) !important;
+}
+
+================================
+
 bitbucket.org
 
 INVERT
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -222,6 +222,15 @@ CSS
 
 ================================
 
+bing.com
+
+CSS
+.na_cnt .nws_itm, .nws_itmb, #b_content #b_results li, body #b_header {
+    border-color: rgb(55,55,55) !important;
```
alexanderby: Please use `${default-color}` pattern to make fix work in light mode.

src/config/dynamic-theme-fixes.config
```diff
@@ -2112,6 +2130,12 @@ CSS
 .Gor6zc {
     background-color: white !important;
 }
+#cnt #hdtbSum, #cnt>#appbar {
+    background: rgb(24,26,27);
```
alexanderby: Please use ${element-default-color} pattern to make fix work in light mode and make it configurable.

src/config/dynamic-theme-fixes.config
```diff
@@ -2112,6 +2130,12 @@ CSS
...
+}
+.srg .g{
+    border-color: rgb(53,57,66) !important;
```
alexanderby: Could you please explain what are you trying to achieve?

alexanderby:
Hi, I think we should not deliver fixes for 3rd party scripts to all users. This month I'm going to add a separate CSS section other than dev tools for rare cases important for some users.

PR #1861, opened 2019-12-26
Translation: Swedish

kartik918:
Add sv.config in _locales


kartik918:
My line count is preserved according to en.config, I don't get the reason for the failed check.

alexanderby:
Hi @kartik918, sorry for the delay.

The problem is in line endings, it is CRLF instead of LF. I don't know how it could happen, I force LF in `.gitattributes` config. But it seems like your local Git settings made the file to be uploaded with CRLF.

Please upload the file with LF line endings if possible.

kartik918:
@alexanderby thanks for helping me out with the failed check.
I've fixed it in the latest commit.

alexanderby:
Thanks, maybe `.gitattributes` is misconfigured. I tried forcing LF so that it stays the same on Windows, macOS and Linux no matter of user's Git configuration.

kartik918:
I checked .gitattributes, ``* text=auto`` should be enforcing LF line endings.

PR #1866, opened 2020-01-02
fix button color on tincercad.com

Rakasch:
The buttons are on a canvas which always has white background.
In dynamic mode, the buttons have grey background with grey icons. 
Set the buttons background to white to be unobtrusive. 
Now you can see the buttons borders and the icon.

image


Rakasch:
will remove https from url, since Travis CI build failed on that

PR #1879, opened 2020-01-05
Update dark-sites.config

Mizuchi:
These sites have native dark mode support

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -87,6 +87,7 @@ flooxer.com
 fortnite-api.com
 frankerfacez.com
 gamejolt.com
+gamersky.com
 geektyper.com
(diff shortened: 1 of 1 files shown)


Rakleed:
@Mizuchi https://github.com/darkreader/darkreader#fixing-wrong-inversion
> If some website is already dark (**_has all pages dark by default_**), you can add it to dark-sites.config file (please, preserve alphabetical order).

And the comment of the developer https://github.com/darkreader/darkreader/pull/1667#issuecomment-548699769

> Hi. These sites support native dark mode, but are not dark by default when system Light appearance is selected. Well, things got complex, I have to think about the purpose of this list, it is becoming useless. Maybe the option is needed to detect native dark mode support and prevent doing anything.

PR #1885, opened 2020-01-09
Update dark-sites.config

skyzuma:


alexanderby:
Hi @skyzuma and @yc5, only websites that are dark by default should be listed there.

skyzuma:
there is dark by default if i switch to dark ... dont need darkreader ... why u closed and delete my entries?

PR #1900, opened 2020-01-16
Fix for instructure.com

BananaManCJ:


BananaManCJ:
Before
image
After
image

PR #1904, opened 2020-01-17
Added Outlook.live.com

TheAlienDrew:
Outlook has a dark mode now


alexanderby:
Hi, this list should include sites that are always dark by default.

PR #1930, opened 2020-01-25

xlynx555:
Fix white background in dark theme for expanded article view on inoreader.com

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -936,8 +936,8 @@ CSS
 inoreader.com
 
 CSS
-div.article_expanded{
- background-color: rgb(31, 35, 38); !important;
+.article_expanded {
+    background-color: rgb(31, 35, 38) !important;
 }
 
 ================================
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -933,6 +933,15 @@ CSS
 
 ================================
 
+inoreader.com
+
+CSS
+div.article_expanded{
```
alexanderby: `div` selector part is unnecessary, `.article_expanded` should be enough. Please put ` ` (space) before `{`.

src/config/dynamic-theme-fixes.config
```diff
@@ -933,6 +933,15 @@ CSS
...
+ background-color: rgb(31, 35, 38); !important;
```
alexanderby: There should be only one `;` at the end of line. Please use 4 spaces for indentation. Check, how your fix works in Light mode, maybe you should replace `rgb(31, 35, 38)` with `${white}` template, so that `white` color (or any other value) could be configurable with app settings.

xlynx555: Formatting fixed.
As for color - it looks appropriate with rgb(31, 35, 38) in both Light and Dark modes. Changing to ${white} makes background white on a dark page (in Light mode) which is not looking good

PR #1969, opened 2020-02-07
Update dark-sites.config

sanjaylord:
Request removal from Global Dark list.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -96,7 +96,6 @@ femto.pw
 filterblade.xyz
 flooxer.com
 fortnite-api.com
-forum.snahp.it
 frankerfacez.com
 gamejolt.com
 geektyper.com


alexanderby:
Hi, this site looks dark, is there a reason for removal?

sanjaylord:
> Hi, this site looks dark, is there a reason for removal?

Thanks for the response! Yes, it does actually look dark since the forum is using a Dark theme by default. But once you log into the site, you're able to change the board theme to a "Light" version instead. Which then cannot be inverted at night because the site now happens to be in the global ignore list :(

Screenshot: https://i.postimg.cc/0Q7shhM5/screenshot-106.png

alexanderby:
By the way, you should be able to invert websites in this list, this feature was added several months ago. Simply click a checkbox with website name, it is not greyed out.

sanjaylord:
> By the way, you should be able to invert websites in this list, this feature was added several months ago. Simply click a checkbox with website name, it is not greyed out.

Wow, I can't believe I missed that toggle. It's working as intended now, the site can now be inverted. Thanks a lot for this wonderful extension btw.

PR #1973, opened 2020-02-08
Add Raindrop app to dark sites list

d1vanloon:
The web-app for Raindrop.io defaults to dark mode.


alexanderby:
Looks like it is not always dark.
Screen Shot 2020-02-18 at 12 35 59

d1vanloon:
Oops, forgot to check the login page. Once the user gets past that, the app's dark mode is nicer than what Dark Reader attempts.

PR #1979, opened 2020-02-09
Fixes "Trending today" bar on reddit

crwn1337:
Previous class was only the body, which didnt cover the tab.


alexanderby:
I hope I've found a solution prone to class name changes.

PR #1987, opened 2020-02-11
Subscene site logo on upper right becomes dark.

Mav3r1x:


Myshor:
You need to make changes in commit.

One more line is needed before INVERT.

PR #1999, opened 2020-02-13
Bring dark_sites.json (for older browsers) to parity with dark-sites.config.

DroneBetter:
It's currently at less than half the length.


alexanderby:
Hi @DroneBetter, thanks for reminding about this old version. I will move it to another repository.

PR #2002, opened 2020-02-14
Make checkboxes easier to read on GitHub

nasso:


alexanderby:
Hi @nasso, could you please share a link to an example?

nasso:
So I just realized when trying it out that this issue is only happening in Windows, because GitHub uses disabled check-boxes, which are barely visible with the Windows 10 theme:

### Live demo:
- [ ] Not checked
- [x] Checked

### Without the patch:
74784818-1300ed80-52a9-11ea-8753-3a1e2f78facc

### With the patch:
74784718-d7feba00-52a8-11ea-9229-cd74765c1cd3

A proper fix should somehow manage to only change the appearance of check-boxes in Windows, but I have no idea how that would work with just CSS.
Another option is to provide a custom check-box item to replace the native one but that's doesn't sound like a great idea; though it might be the best option we have..?

alexanderby:
If it happens for all checkboxes on Windows, maybe I could think of adding a default style for Windows only. Also it is planned to add a custom CSS section for user's style sheets.

PR #2037, opened 2020-02-24
Update INVERT for google spreadsheet's formula

sgavka:
Update INVERT for google spreadsheet's formula in cell


alexanderby:
Thanks for the fix. Weird issue with cells. I always suggest to use Filter mode for Google Sheets.

PR #2045, opened 2020-02-27
Fixes for Instagram.com

merajmasuk:
Comments are now visible when the dynamic-theme is activated on Instagram.

Before:
image

After:
image

Known Issue:
* The Instagram icon on the navbar gets inverted.
image


alexanderby:
Hi @merajmasuk 

Previous fixes were strange. Dark Reader seems to work fine in Instagram without fixes, only sprite icons should be fixed. Please, try inserting the fixes from this Pull Request.

PR #2052, opened 2020-02-28
tweakers.net - Fix Dynamic on this site

Gusted:
Weird repeat arrow? It's not even suspossed to view for desktop version...
Before
before
After
after

The top bar was? Drunk I guess
Before
before
After
after

More repeating arrows? and the arrow was placed in a sub section of the element instead of the parent element.... 
Before
before
After
after

Box doesn't matter when on darkmode
Before
before
After
after

Couldn't see the replycounts on banners articles
Before
before
After
after

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2350,6 +2350,10 @@ CSS
 tweakers.net
 
 CSS
+.ankeiler>a.commentCount {
+    background-image: url(../../../g/icons/commentcounter_small_bg.svg) !important;
+    text-decoration: none !important;
+}
 .thumb.category {
    border: none !important;
 }
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2350,7 +2350,6 @@ CSS
 tweakers.net
 
 CSS
-CSS
 .thumb.category {
    border: none;
 }
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2347,6 +2347,15 @@ CSS
 
 ================================
 
+tweakers.net
+
+CSS
+a.showMoreItems {
+   background-image: none !important;
+}
+
+================================
+
 underconsideration.com/brandnew
 
 CSS
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2350,8 +2350,108 @@ CSS
 tweakers.net
 
 CSS
+CSS
+.thumb.category {
+   border: none;
+}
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -2347,6 +2347,118 @@ CSS
...
+CSS
+.ankeiler>a.commentCount {
+    background-image: url(../../../g/icons/commentcounter_small_bg.svg) !important;
+    text-decoration: none !important;
+}
+.thumb.category {
+   border: none !important;
+}
+#categoryBrowser li a{
+     background: none !important;
```
alexanderby: Use 4 spaces

src/config/dynamic-theme-fixes.config
```diff
@@ -2347,6 +2347,118 @@ CSS
...
+}
+
+#categoryBrowser li.more {
+    background-image: url("../../../../g/if/categories/arrows.png") !important;
+    background-color:
```
alexanderby: Missing value

src/config/dynamic-theme-fixes.config
```diff
@@ -2347,6 +2347,118 @@ CSS
...
+#categoryBrowser.sidebar li.more {
+    background-position: 282px 10px;
+}
+
+a.showMoreItems {
+   background-image: none !important;
+}
+
+.corner::before,.corner::after {
+    border: 6px solid #333 !important;
```
alexanderby: Please, check how the fix looks in light mode. Use `${#333}` pattern putting a source color into `${}`, later the changed value will be substituted.

Gusted: It looks like this ![https://i.imgur.com/jt8OGPJ.png] ${#333}
And like this ![https://i.imgur.com/V2FMjyx.png] with #333 So I don't think it's handy to use ${}
It looks like this in light mode ![https://i.imgur.com/T8aDoYp.png]

src/config/dynamic-theme-fixes.config
Gusted: Fixed Dev tools Doesn't Support tab? (It just unfocus the page)

src/config/dynamic-theme-fixes.config
Gusted: Removed I don't know why it was there

alexanderby:
Great, thanks.

Vistaus:
Thanks a lot! I had the exact same issues (I'm an avid Tweakers member :) ). I tested your changes, but everything looks the same as before, i.e. still broken. So I think you need to tweak (no pun intended) your code some more.

Gusted:
Could you provide a screenshot? Because for me it's now just working as intended.

Did you make sure you hit Apply?

Issue #2049, opened 2020-02-28
Icons replaced on office.com (Office 365)


iSnackyCracky:
On the Office 365 Dashboard page (logged in to https://www.office.com with an Office 365 business account) the Excel and SharePoint icons are "replaced", but this isn't necessary.

darkreader enabled:
![image](https://user-images.githubusercontent.com/20481463/75537141-c76fe180-5a16-11ea-9ef4-92a8fb0538ea.png)

darkreader disabled:
![image](https://user-images.githubusercontent.com/20481463/75537231-ecfceb00-5a16-11ea-9ecc-551c755791d8.png)

darkreader enabled, but default icon restored:
![image](https://user-images.githubusercontent.com/20481463/75537297-0b62e680-5a17-11ea-9434-0478876338a3.png)

Seems like the background-image `of span.ms-ohp-svg-Icon.ms-ohp-Icon.ms-ohp-Icon--excelLogo.ms-ohp-Icon--excelLogoFill.ng-star-inserted::before` and `span.ms-ohp-svg-Icon.ms-ohp-Icon.ms-ohp-Icon--sharepointLogo.ms-ohp-Icon--sharepointLogoFill.ng-star-inserted` gets "replaced".

Unfortuantely I wasn't sure how to fix this, in chromium dev-tools it looks like the change comes from an empty style-object:
![image](https://user-images.githubusercontent.com/20481463/75539240-37339b80-5a1a-11ea-8a34-c8db9bd43d8d.png)
`<style class="darkreader darkreader--sync" media="screen"></style>`

And I didn't find the stated `blob:https://www.office.com/85fb` anywhere.

Gusted:
`office.com

CSS
.ms-ohp-svg-Icon.ms-ohp-Icon--excelLogo:before {
[...]

PR #2059, opened 2020-03-01
office.com - Icon fixs

Gusted:
#2049


alexanderby:
Until I understand what's wrong with the icon, lets not embed images into this file. My suggestion is a little bit odd, but works.

PR #2063, opened 2020-03-02
Fixed github.com fork tree

Mertyn:


alexanderby:
Hi @Mertyn, could you please share an example of such tree?

I see Canvas only, but no images with class "network-tree".

Mertyn:
Okay @alexanderby
You can see it [here](https://github.com/darkreader/darkreader/network/members) for yourself or in the screenshot here:
image

alexanderby:
Great, thanks. I used the `INVERT` section for network graph and fork tree. The background color is a little bit different, but it should be adjusted in JS code.

PR #2066, opened 2020-03-03
added startpage.com since it now has a native dark theme

geekabit2:


alexanderby:
Sorry, but this website is bright by default. But this list should include only websites, that are always dark. This is important for users of Filter mode. In future I think I will add a detection for website's own dark theme.

Issue #2067, opened 2020-03-04
Red Hat website issue


UchihaSR:
when i went to red hat officiak website to chck the ex-200 exam objective. i encountered this situation

 
![rhcsa_darkreader](https://user-images.githubusercontent.com/52357560/75856423-58c9c400-5e1a-11ea-86e3-c95f63eaa863.jpg)

Gusted:
https://www.redhat.com use for some elements like the panels #shadow-root (open) [More info](https://stackoverflow.com/questions/34119639/what-is-shadow-root) and that means the have their own DOM and I think when darkreaders gets all element. It won't detect whats in shadow root. I will take a look a make CSS fix for redhat what probaly gonna make it little bit better. A lot of the elements have same classes....

Gusted:
I have now got just this

`
================================

redhat.com

CSS
p, li {
   color: ${black}
}
.PFElement {
    background-color: ${white};
}

================================`

Gusted:
Hi their! 
Could you please try to get the latest version of Darkreader 4.9.4 and check if this issue still persist?

UchihaSR:
yes, it is solved now. Thanks

PR #2073, opened 2020-03-05
redhat.com - manual fix css for shadow-root DOM

Gusted:
#2067 Fix
Before
Before

After
After


alexanderby:
Thanks, looks strange, Shadow DOM should be observed when elements are in Open mode, maybe there's some bug.

PR #2086, opened 2020-03-07
Adjust issue with AWS CloudWatch Dashboards

urda:
## Problem

I spend a ton of time in my AWS console. When I'm viewing a dashboard on a page with a URL *similar* to:

```
https://AWS-REGION.console.aws.amazon.com/cloudwatch/home?region=AWS-REGION#dashboards:name=DASHBOARD-NAME-HERE
```

Dashboards ended up having this large white bar:

before_fix

**So I fixed it:**

after_fix

Let me know if you have any questions.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -227,6 +227,11 @@ img[alt^="WEB_FreeTier"]
 .lb-is-lazyloaded
 .lb-none-v-margin.lb-img
 
+CSS
+.header-background {
+  fill: none !important;
+}
+
 ================================
 
 baike.baidu.com


src/config/dynamic-theme-fixes.config
```diff
@@ -227,6 +227,11 @@ img[alt^="WEB_FreeTier"]
 .lb-is-lazyloaded
 .lb-none-v-margin.lb-img
 
+CSS
+.header-background {
+  fill: #181A1B !important;
```
alexanderby: Hi, adding static colors is not good, the fix will not work in Light mode or will look strange after filter value change.

Try using `fill: ${white} !important;` or `fill: none !important;`.

urda: Thanks for the tip, I'm making the update and testing now.

urda: I have made the change requested and rebased.

PR #2087, opened 2020-03-07
Remove trailing whitespace in dynamic-theme-fixes.config

urda:
During my efforts to make my AWS console prettier in #2086 , my editor discovered a chunk of trailing whitespace.

So I nuked it.


alexanderby:
Thanks, CSS text formatting is not validated.

PR #2090, opened 2020-03-07
facebook.com - fix for black promotion icon

Myshor:
change path for promotion icon to be more universal


Myshor:
Invert for  likes icons for recommended pages
image

Myshor:
Removed invert for ._4k6w
Forced png instead of blob in CSS.
image

PR #2099, opened 2020-03-10
github.com - fix color for gitako

diartyz:
fix background-color for [gitako](https://addons.mozilla.org/en-US/firefox/addon/gitako-github-file-tree/?src=search) in firefox

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -870,6 +870,7 @@ a[href="https://github.com"]
 github.com
 
 INVERT
+.gitako-position-wrapper
 .octotree-sidebar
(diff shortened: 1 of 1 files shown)


alexanderby:
There is a problem not with Dark Reader (at leas a latest build), but with Firefox. It returns a wrong MIME type for another extension's style sheets, and Dark Reader blocks a request for security reasons. I will try to think about workaround.

alexanderby:
Screen Shot 2020-03-12 at 13 07 50
Done.

alexanderby:
https://github.com/darkreader/darkreader/commit/511ba13c3d644552620c32f74a6815199886ea26

PR #2107, opened 2020-03-12
Improved Shadow DOM styling

alexanderby:
- Override Shadow DOM styles (open mode).
- Watch for Shadow DOM elements.
- Watch for Shadow DOM attribute styles.


Gusted:
I just checked the branch shadow-dom on redhat.com that has shadow-doms It's not working still...

alexanderby:
I hope it is fixed. There is a small flash, because Chrome developers to disable `customElements` API for extensions, so I had to use `setInterval`. Maybe I will use `requestAnimationFrame` for this, but I'm afraid about performance.

Gusted:
Just run some performance test, because just now make the code a lot faster and all new code optimizated for performance especially for firefox users if I see this issues

PR #2113, opened 2020-03-13
facebook.com - event question mark icon invert fix

Myshor:
Remove .sx_a1a9dd
Add #event_tabs #reaction_units span img
image


Myshor:
One more attempt for correct page promotion icon invert as it is again black and almost invisible.
image

PR #2114, opened 2020-03-13
COVID-19 Map is dark

Gusted:
The COVID-19 Incidents map is already dark by default


Gusted:
wasnt fully dark the loading screen isn't

PR #2129, opened 2020-03-15
Duolingo match the pairs fix

ashtonx:
During match the pairs quiz, the pairs that are are matched should be grayed out.


OwenAR:
This currently seems to be broken again.

See default:
1 duolingo matching normal|Screenshot_2021-09-15 Learn a language for free

vs same thing with darkreader:
2 duolingo matching darkreader|Screenshot_2021-09-15 Learn a language for free


If I squint, I think the text of the already-matched pair (wáng and 王) is *maybe* a *TINY* bit *brighter*...?

But yeah, already-matched entries are practically indistinguishable from still-to-match entries.

(btw, when I toggle darkreader on, it seems to apply the changes in steps, and oddly, it actually passes through a phase where the entries are already dark but *are* distinct.)

ashtonx:
I think duo did this before, didn't use it for a while so i didn't notice. I started using it again but didn't install dark reader on that webapp yet so I didn't notice. Submitted fix.

Issue #2130, opened 2020-03-16
Make devtools in a separate tab?


tmpm697:
I'm in situation that if I open "Dev tools" it will pop-up its own window that from window manager aspect cause some problem. As my window manager is not implemented ability to "button" to close the window, I have to use an X tool to kill "Dev tools" window that cause firefox to crash immediately.

Is there a way to use firefox tab for "Dev tools" instead of a separate window?

Gusted:
Is this a good idea? A close button.

![](https://i.imgur.com/umt0tdr.png)

tmpm697:
@Gusted Yes please a close button.

Gusted:
#2133 :)

alexanderby:
Hi @tmpm697. As a workaround press F12, type `location.href` in Console, open the URL you see, add it to bookmarks.

PR #2133, opened 2020-03-16
Devtools - Close button

Gusted:
Add a close button to the devtools UI
Related Issue: #2130

### DIFF
--- a/src/ui/devtools/components/body.tsx
+++ b/src/ui/devtools/components/body.tsx
@@ -52,6 +52,10 @@ function Body({data, actions}: BodyProps) {
         wrapper.reset();
         setState({errorText: null});
     }
+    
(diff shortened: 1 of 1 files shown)


alexanderby:
Sorry, but the UX decision look not good to me. Most of the users will not need this button. There should be an ability to open this page in a new tab or a new window. This option should not be visible, but rather hidden in some menu to not mess with main action buttons. Maybe `<a>` link should be used. Let me think about it for some time, or propose some other ideas.

Gusted:
So far I see this was a issue for Firefox Users so maybe isFirefox() and than for chrome a seperate window and for firefox users a new tab?

Gusted:
NEW UI

PR #2134, opened 2020-03-16
Add invert for easypost images

jasonleibowitz:
Add Invert rule for images on easypost. 

Before
Screen Shot 2020-03-16 at 11 04 43 AM

After
Screen Shot 2020-03-25 at 9 39 59 AM

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -587,6 +587,13 @@ INVERT
 
 ================================
 
+easypost.com
+
+INVERT
+img
+
+================================
+
 education.github.com
 
 INVERT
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -587,6 +587,13 @@ INVERT
 
 ================================
 
+easypost.com
+
+INVERT
+img
```
alexanderby: Please specify a more specific selector. For example if element or its parent has "class" attribute, you can use class selector, like `.class-value`. If inverted images have some specific path, you can use an attribute selector like `img[src="path.jpg"]` or `img[src*="common-pattern"]` if multiple images have some common name part.

PR #2154, opened 2020-03-22
facebook.com - new fix for memories inverted img

Myshor:
Before:
image

After:
image


Myshor:
Fix for likes icons for recommended pages as they sometimes are  but sometimes <i>.
Catched that there is always .img class used.
image

Myshor:
reworked memories picture fix to be without classes thst change all the time

PR #2182, opened 2020-03-27
 Added dot.kde.org body background

GabMus:
fixes original inverted pattern background that otherwise looks like static noise


Gusted:
# Before



# After

PR #2188, opened 2020-03-27
Handle Empty shadowObserver

Gusted:
Pretty annoying in console

### DIFF
--- a/src/inject/dynamic-theme/watch.ts
+++ b/src/inject/dynamic-theme/watch.ts
@@ -175,15 +175,13 @@ export function watchForStyleChanges(update: (styles: ChangedStyles) => void) {
     }
 
     function subscribeForShadowRootChanges(node: Element) {
-        if (nodesShadowObservers.has(node)) {
+        if (nodesShadowObservers.has(node) || node.shadowRoot != null) {
             return;
         }
         const shadowObserver = new MutationObserver(handleMutations);
-        if (shadowObserver.takeRecords().length != 0) { // Don't handle empty Shadowobservers
-            shadowObserver.observe(node.shadowRoot, mutationObserverOptions);
-            shadowObservers.add(shadowObserver);
-            nodesShadowObservers.set(node, shadowObserver);
-        }
+        shadowObserver.observe(node.shadowRoot, mutationObserverOptions);
+        shadowObservers.add(shadowObserver);
+        nodesShadowObservers.set(node, shadowObserver);
     }
 
     const mutationObserverOptions = {childList: true, subtree: true, attributes: true, attributeFilter: ['rel', 'disabled']};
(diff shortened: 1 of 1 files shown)


alexanderby:
Hi @Gusted, could you please share an example URL?

src/inject/dynamic-theme/watch.ts
```diff
@@ -179,9 +179,11 @@ export function watchForStyleChanges(update: (styles: ChangedStyles) => void) {
             return;
         }
         const shadowObserver = new MutationObserver(handleMutations);
-        shadowObserver.observe(node.shadowRoot, mutationObserverOptions);
-        shadowObservers.add(shadowObserver);
-        nodesShadowObservers.set(node, shadowObserver);
+        if (shadowObserver.takeRecords().length != 0) { // Don't handle empty Shadowobservers
+            shadowObserver.observe(node.shadowRoot, mutationObserverOptions);
```
alexanderby: As far as I understand, the error happens when shadowRoot is not an instance of Shadow Root.

Gusted:
open the console in this pull reqeust

src/inject/dynamic-theme/watch.ts
```diff
@@ -179,9 +179,11 @@ export function watchForStyleChanges(update: (styles: ChangedStyles) => void) {
             return;
         }
         const shadowObserver = new MutationObserver(handleMutations);
-        shadowObserver.observe(node.shadowRoot, mutationObserverOptions);
-        shadowObservers.add(shadowObserver);
-        nodesShadowObservers.set(node, shadowObserver);
+        if (shadowObserver.takeRecords().length != 0) { // Don't handle empty Shadowobservers
```
alexanderby: It looks like it will always return 0, since the observer is not observing anything.

Gusted: It's observing a empty {}

alexanderby:
So, it happens for custom elements, that have shadow root attached in close mode, so a simple `element.shadowRoot != null` check somewhere should work.

src/inject/dynamic-theme/watch.ts
```diff
@@ -179,9 +179,11 @@ export function watchForStyleChanges(update: (styles: ChangedStyles) => void) {
             return;
```
alexanderby: So, you can basically add a check for a shadow root in this `if` block.

Gusted:
>?

src/inject/dynamic-theme/watch.ts
```diff
@@ -175,7 +175,7 @@ export function watchForStyleChanges(update: (styles: ChangedStyles) => void) {
     }
 
     function subscribeForShadowRootChanges(node: Element) {
-        if (nodesShadowObservers.has(node)) {
+        if (nodesShadowObservers.has(node) || node.shadowRoot != null) {
```
alexanderby: It should exit, when shadow root is null.

alexanderby:
Thanks for noticing, Travis did break, so I switched to GitHub actions.

Gusted:
Yeah I saw Travis was quite broke has to commited twice to trigger Travis

PR #2191, opened 2020-03-29
acm.sdut.edu.cn/onlinejudge3 is dark

dreamerblue:
This site has its own dark theme settings.


dreamerblue:
Sorry for wrong committed file. I will fix it later.

PR #2196, opened 2020-03-30
Fix vote buttons in all stackexchange sites

Mukundan314:
Before Fix:
Screenshot_2020-04-01 Hell0 W0rld scored by ASCII values
After Fix:
Screenshot_2020-04-01 Hell0 W0rld scored by ASCII values(1)


Mukundan314:
Since the section I added matches all stackexchange sites the sections for specific stackexchange sites that already existed are not being applied. Should I merge all stackexchange site sections to a single section?

PR #2201, opened 2020-03-31
docs.google.com - Again invert fix

Gusted:
Welp Guess it or not more invert

# Before

# After


Gusted:
Already fixed with another fix

PR #2213, opened 2020-04-02
Fix 10fastfingers.com

Mukundan314:
Before Fix:
Screenshot_2020-04-02 Typing Test English - 10FastFingers com
After Fix:
Screenshot_2020-04-02 Typing Test English - 10FastFingers com(1)

**Note:** Time is supposed to be hidden

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -30,6 +30,9 @@ CSS
 body {
     background-image: none;
 }
+#speedtest-main .hide-time {
+    color: transparent !important;
+}
 #inputfield {
     color: #333 !important;
 }


Gusted:
Hmm Could you post a screenshot from devtool tools of that element. I find it weird Darkreader changed a transparent color.....

Mukundan314:
20200415_17h29m53s_grim
This happens because the site does not set the color to transparent but instead sets it to the same color as the background.

Gusted:
I've just checked with darkreader off  Seems like it displays the time

Mukundan314:
Clicking on the time should hide it.

Gusted:
That makes more sense clicking on it

PR #2215, opened 2020-04-02
Stackoverflow now has dark theme

AdrianMuntean:
[Stackoverflow](https://stackoverflow.blog/2020/03/30/introducing-dark-mode-for-stack-overflow/) now has the option to use the dark mode.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -211,6 +211,7 @@ snazzah.com
 spacebattles.com
 spacespacespaaace.space
 speedtest.net
+stackoverflow.com
 staszic.waw.pl
 statbot.net
 steam.tv


Gusted:
It's in beta and it's a opt-in so it's mean not by default so this will be probaly a closed Pull Request.

AdrianMuntean:
> It's in beta and it's a opt-in so it's mean not by default so this will be probaly a closed Pull Request.

Ok, then I'll close this PR and maybe re-open if/after the feature is stable.

gregk-git:
It's not in default dark theme and you can't use the dark theme without being signed in. It very likely won't ever be the default theme for StackOverflow, either. There's very little reason to add it to the list, if you want to use SO's theme, then just disable DR.

PR #2224, opened 2020-04-03
web.whatsapp.com - dropdown button

Gusted:
So it's make the background white when hovering.

# Darkreader Turned off

# Before

# After


Gusted:
Dynamic. Whatsapp updates too much smh.

Issue #2229, opened 2020-04-05
Activated like / dislike buttons on Youtube not changing to blue color anymore


Maxxiii:
When you click on like or dislike button on Youtube video, it's not changing color to blue anymore. Only text description near icons (liked/disliked) change color. Same thing for like / dislike for comments section, but it's even worse there, because there is no text for this buttons, so "activated" / (pressed) likes indistinguishable from "not activated"

Gusted:
Seems like youtube updated css selectors like usual :/ Will look into it and provide a fix.

Gusted:
Provided a fix #2230

Gusted:
https://github.com/darkreader/darkreader/pull/2230#issuecomment-613969547

PR #2230, opened 2020-04-05
youtube.com - like/dislike color

Gusted:
Youtube changed some CSS selectors.

Related Issue: #2229 
# Before

# After


Gusted:
Nope Isn't the same when Dakreader off Youtube changed :?

PR #2233, opened 2020-04-05
Fixes for duolingo

ashtonx:
Fix originally Introduced in #1490 updated values to current ones.

before:
 
after:


ashtonx:
Second is update values for the match the pairs quiz I also lowered opacity to make contrast better (i found previous value to be lacking after repeated use)

before:

after:

Issue #2236, opened 2020-04-06
QR code, inverted !


promeneur:
With your PC with Chrome go to web.whatsapp.com
get a QR code.

QR code is inverted then when you scan QR code you can't connect to your account !

You must disable darkreader to scan the good QR code.

Gusted:
To make a fix it will look like this..
![](https://i.imgur.com/ql67ZXd.jpg)

Gusted:
https://pastebin.com/Rgyz0gHU

Gusted:
I close this issue as #2209 has been merged and will be avaible in the next update of Darkreader. Thank you for reporting this issue!

PR #2209, opened 2020-04-01
Fix for web.whatsapp.com

HassanAbass:
When activating dark mode on https://web.whatsapp.com, my mobile whatsapp scanner cant read the code.
So i added a border to separate the qrcode from the dark background, in the web.whatsapp.com block

PR #2238, opened 2020-04-06
web.whatsapp.com - update class

Gusted:
QR code has new class
Related Issue #2236


yawnston:
might wanna squash those commits 😃 

I was the one who originally made the inversion for the QR code but I guess they changed the CSS class at some point, maybe relying on document structure (something like taking the `landing-main` div and going a few divs down to div[data-ref]) would keep this from breaking for longer.

Gusted:
Hmm good one! I will look into it and edit the css selector

Gusted:
#2209

PR #2239, opened 2020-04-06
Update for README.md

Gusted:
Giving the user a option to host a locally up to date api Because I just saw the CDN is running 4.9.2...


Gusted:
Second Opnion on this

alexanderby:
Maybe let's make it shorter, I think it will be easier to understand then.

PR #2241, opened 2020-04-07
youtube.com - Video Information Corrected Color

Gusted:
This code is for Views/Upload Date
```
#metadata-line.ytd-video-meta-block span.ytd-video-meta-block {
    color: var(--ytd-metadata-line-color, var(--yt-spec-text-secondary)) !important;
}
```
This code is for the youtube Channel name
```
yt-formatted-string[has-link-only_]:not([force-default-style]) a.yt-simple-endpoint.yt-formatted-string {
    color: var(--yt-endpoint-color, var(--yt-spec-text-primary)) !important;
}
```

# Darkreader Off so you see it has other colors

# Before

# After


Gusted:
https://github.com/darkreader/darkreader/commit/0a367acd3b11138eb5bc61bb704edd34562344f3

PR #2252, opened 2020-04-09
Fix for instructure.com

jake2918:
LATEX Equations were previously black text making it impossible to read with Dark Reader turned on.    I just added 1 line to invert the color to make it readable.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1376,6 +1376,7 @@ instructure.com
 
 INVERT
 .title_3bek::after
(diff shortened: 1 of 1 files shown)


Gusted:
Could you provide a screenshot? I can't acces any of the materials from instructure.com the CSS Selector seems good. Just need to know if the invert is good enough or any custom CSS Maybe.

jake2918:
Sure,  Here it is with the new code:


image


and without:


image


Thanks!

Gusted:
That's good with a INVERT!

PR #2261, opened 2020-04-10
Fixes for  journal.tinkoff.ru

Rakleed:
Fix pages like journal.tinkoff.ru/list/the-fastest-cars, journal.tinkoff.ru/about, journal.tinkoff.ru/user* (requires login to your account)

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1436,6 +1436,21 @@ h1.logo
 
 ================================
 
+journal.tinkoff.ru
+
+INVERT
+.g2tvv
+._3aEco
+._1Dhxk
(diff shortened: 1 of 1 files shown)


Gusted:
Seems good to me. But to see if the CSS Selectors for `journal.tinkoff.ru/user*` are right and don't have a account and don't have intention to make one. Could you make a screenshot of 2 seprate users with this fix so it's comfirmed it's not for each user another CSS selector.

Rakleed:
I was a little mistaken, you can see the changes without having to log in to your account. For example, [this user](https://journal.tinkoff.ru/user4783/).

Gusted:
Looks perfectly fine!

PR #2264, opened 2020-04-11
fix databricks code mirror colours

chinwobble:
Before:
image

After:
image


Gusted:
How does it look without Darkreader on?

Gusted:
@chinwobble Hi their! can you post an screenshot without darkreader enabled?

chinwobble:
@Gusted this is what the site looks like without dark reader.
image

PR #2268, opened 2020-04-11
sport.pl - articles background fix + logo invert

Myshor:
articles background fix + logo invert

Before:
image

After:
image


Myshor:
Added img to invert rule to fix main page

PR #2297, opened 2020-04-16
Fix for stackexchange sites

Mukundan314:
Before fix:
Before fix

After fix:
After fix

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -39,9 +39,8 @@ CSS
 body {
     background-image: none !important;
 }
-.profile-cards--graph .line {
-    stroke: Green;
-    fill: none;
+.profile-cards--graph {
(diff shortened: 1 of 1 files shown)


Gusted:
Could you give an example url where I can check this fix?

Kurtoid:
Here's one: https://askubuntu.com/users/359408/kurtoid?tab=topactivity

Looks good to me

Gusted:
well..

Gusted:
As I'm seeing it actually make it's more unvisble. I'm getting something wrong because the original images from OP are showing otherwise

Mukundan314:
I am not sure its seems to work for me on the link that @Kurtoid sent:
screenshot

I also diffed my local config with the one on the this branch there does not seem to be any changes

Gusted:
Are you using Firefox?

Mukundan314:
Yes
Firefox Version: 75.0
Dark Reader Version: 4.9.4

Mukundan314:
Seems to work on chromium too

Gusted:
IDK how but for me it's still ooks like that

Gusted:
Woops had a stylus theme on that I didn't know.

PR #2313, opened 2020-04-17
fix for facebook.com

Myshor:
1. change back to old class for next/prev icons on page panel
.sp_NfKX-XV69KF => .sp_hk4DJV_EEeW
image


Myshor:
Added alternative ID for pagelet to invert icon.

Issue #2315, opened 2020-04-17
SVG images too dark when inverted


jamy015:
A lot of websites now have their logos as SVG. These get inverted by Dark Reader, which is pretty ugly. As an example (not the only one), check out coolblue.nl:

![image](https://user-images.githubusercontent.com/3188721/79615633-8b4a2a80-8103-11ea-8769-d6f597d57677.png)

I tried to invert the logo back using the `INVERT` directive. However, this results in a logo that's darker than the real logo. Compare with Dark Reader switched off and on:

![image](https://user-images.githubusercontent.com/3188721/79615678-a4eb7200-8103-11ea-9e60-9a6a34cf575b.png)
![image](https://user-images.githubusercontent.com/3188721/79615690-acab1680-8103-11ea-9ce9-9c997f6a6eb7.png)

(All adjustments in Dark Reader are switched off, i.e. brightness, contrast, sepia and grayscale.)

How can I adjust the logo back to the real, non-darkened version?

Gusted:
Hello fellow Dutchman :P

I'm assuming you use dynamic theme as of that. 
Dynamic theme fills in the element another color --darkreader-in line-fill or something in that way, that's why inverts don't fixs it. I will take another closer look soon and provide a CSS fix

Gusted:
```
coolblue.nl

CSS
path[fill="#F60"] {
    --darkreader-inline-fill: #F60 !important
}
path[fill="#FFF"] {
    --darkreader-inline-fill: #FFF !important
}```

Gusted:
![](https://i.imgur.com/bC6gUgf.png)

Gusted:
When #2317 will be merged this issue will be closed.
Thank you for reporting this issue to darkeader.

PR #2317, opened 2020-04-18
Fix for coolblue related sites

Gusted:
- Fix SVG logo to be the original color
- Related issue #2315

# Before

# After


jamy015:
Fix should probably also be applied to coolblue.be and hotorangemedia.nl

Gusted:
> Fix should probably also be applied to coolblue.be and hotorangemedia.nl

Thank you for the reaction I've added them as well

Issue #1963, opened 2020-02-06
ngrok.com: SVG font remain dark on a dark background


GauthierPLM:
The SVG element on [ngrok.com](https://www.ngrok.com/) main page uses a dark background and a dark font, making the text unreadable.

I attached an image to illustate this.

I use Firefox Nightly and Dark Reader version 4.8.8 with default settings (Dynamique style and default font and color settings).

<img width="670" alt="Capture d’écran 2020-02-06 à 10 49 00 AM" src="https://user-images.githubusercontent.com/2579741/73898916-4b2e2680-48ce-11ea-85f5-8de37e9429e7.png">

Myshor:
Made some fixes which ends with:
![image](https://i.ibb.co/sK3Fj41/ngrok.png)

Fix is waiting for acceptance. ;)

MariusCC:
Wikipedia is using svg and many are rendered unreadable when maximized.
Looks like this in the page: 
![image](https://user-images.githubusercontent.com/1683264/79635727-5083d900-817b-11ea-8448-4cb51ca23dbf.png)
but unreadable when zoomed in:
![image](https://user-images.githubusercontent.com/1683264/79635762-75784c00-817b-11ea-9a11-5db1a6f66f26.png)

Maybe a good solution will be to render them as intended by the author - ie black text on white background - as trying to invert the colors in the image will lead to strange results.

Myshor:
First of all... Not nice to ressurect something that was already burried and fixed. 😄 
Second thing. As you show some examples in screen. Can you give exact URL? It will be easier to analyse what and how can be fixed. 😉

MariusCC:
Sorry - I saw the issue still open.
That's the page: 
`https://en.wikipedia.org/wiki/HIV`
and the image:
`https://en.wikipedia.org/wiki/HIV#/media/File:HI-virion-structure_en.svg`

I'm using Chromium Version 81.0.4044.92 (Official Build) Arch Linux (64-bit) if that makes a difference.

Myshor:
Will be fixed with #2321 ;)
You can check how the fix will look like - I shared screens after fixes in Pull Request.

MariusCC:
Tried your fix using the dev tools in the extension.
I think 0.40 for transparency is still not readable. 0.75  is much better even if probably at night will be a little discomfort.
`background-color: rgba(255, 255, 255, 0.75) !important;`

MariusCC:
[...]

PR #2321, opened 2020-04-18
fix for wikipedia.org

Myshor:
Added CSS fix for transparent svg files on zoom mentioned in #1963

URL to check:  https://en.wikipedia.org/wiki/HIV#

Before:
image

After:
image

Checked with other pics on this page and seems to work fine when checker is set as background. :)

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3228,7 +3228,7 @@ div .thumbimage[src$=".png"],
 div .thumbimage img[src$=".png"] {
     background-color: white;
 }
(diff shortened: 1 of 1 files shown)


Myshor:
First of all... we don't want all SVGs change... (I think Wiki logo is SVG too but nvm)
Second thing - in most places wiki makes PNG preview of SVG (it can be found it was svg but still final screen is PNG)

I added background to that kind of files when media player is disabled and it is shown like this:
https://en.wikipedia.org/wiki/File:HI-virion-structure_en.svg

Gusted:
`a[src$=".svg"] > img` This would resolve false positives and only target transparent svg that aren't viewable/

Issue #1077, opened 2019-02-05
Incorrect background-image generated


DianaNites:
The https://www.musl-libc.org/how.html page has, by default, a `background-image: url(puzzle-a2.svg);` for .listpanel

But enabling the dynamic theme overrides that with `background-image: url(https://www.musl-libc.org/how.html/puzzle-a2.svg);`, which is incorrect. It should be https://www.musl-libc.org/puzzle-a2.svg, or even better don't change the link at all?

Untested with any other theme option.

Resolved locally by overriding the CSS back.

Gusted:
Hey their!

I'm working on a fix for this, I only need to get working ../ for backwards folders(else it could going to break another urls) and than a PR will be made.
Thank you for reporting this issue!

Gusted:
#2323 A Pull request has been made and will be merged soon(Somewhere next week)

PR #2323, opened 2020-04-18
Fix resolving absolute paths for non-index pages

Gusted:
Related issue #1077

[Test-pages.zip](https://github.com/darkreader/darkreader/files/4497032/Test-pages.zip)
That's the test pages I've used.

As you know mostly if not all modern website don't have .html .htm .php after their pages as they use index.html or similar so when you request the url it was automaticly grab index.html for you etc.

But as for older websites they sometimes still use .html and .php for their sub domains sites. That can causes issues with loading images. As of like [this site](https://www.musl-libc.org/how.html) It's loads an image on  on the .listpanel (This screenshot was been made with this fix as you can see the background image).
As in the css of this sites it's sets the background-image to `    background-image: url(puzzle-a2.svg);` An incorrect way for modern browsers and HTML5 but back than an very normal way to load images. 

So darkreader will come to the image and spit back out that it absoluteURL is at `https://www.musl-libc.org/how.html` While this will be correct if how.html wasn't a file but a folder. 

With this Patch it will check if the last pathPart(not the image ofc) ends with .html .htm .php and splice this from the array as this isn't a real path/folder but a file. And it will check if the pottential .html .htm .php isn't a backwards indicator.

As of the test pages you will need to host it on a localhost 
That can be simply with python
`python -m http.server 1337` for Python 3 on windows
`python3 -m http.server 1337` for Python 3 on Linux distro's

You will find their some files and folders. I think you can figure yourself out to test this first without the fix and than with the fix and see that nonindex.html will as asked get the good url to load the image and in SomeFolder/FolderForBackwardsTest you will find the both files have ../ in the css to show it will detect automaticcly if the potetional is a backwards indicator.

Regards,
Your new member Gusted

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,8 @@
     "debug": "node tasks/debug.js",
     "prepublishOnly": "npm test && npm run api",
     "release": "npm test && node tasks/release.js",
-    "test": "jest --config=tests/jest.config.js"
+    "test": "jest --config=tests/jest.config.js",
+    "test:inject": "jest --config=tests/inject/jest.config.js"
   },
   "main": "darkreader.js",
   "repository": {
--- a/src/inject/dynamic-theme/url.ts
+++ b/src/inject/dynamic-theme/url.ts
@@ -17,6 +17,16 @@ export function getAbsoluteURL($base: string, $relative: string) {
         return u.href;
     }
     const pathParts = b.pathname.split('/').concat($relative.split('/')).filter((p) => p);
+    let lastRealURLPath: string;
+    if (pathParts[pathParts.length - 2] == "..") { // Check if the path isn't backward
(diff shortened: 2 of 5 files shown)


Gusted:
Approved the unit test and the Simplifer looks good

Gusted:
> I've simplified it a little bit and added some tests. Not sure if I understand how CSS files resolve paths correctly.

The current implemetation is Correct so if the test pass it should be a green light.

alexanderby:
Need to find some doc how CSS resolves files paths.

Gusted:
[CSS level-1](https://www.w3.org/TR/CSS1/#url)
[CSS level-2](https://www.w3.org/TR/CSS2/syndata.html#uri)
[CSS level-3](https://drafts.csswg.org/css-values-3/#urls)
[CSS level-4](https://drafts.csswg.org/css-values-4/#urls)

PR #2325, opened 2020-04-18
fix weather underground

blairun:


Gusted:
Example link? as I can't find any css class with `.bar-on`

blairun:
example link: https://www.wunderground.com/forecast/us/or/portland/KORPORTL1045?cm_ven=localwx_10day

PR #2327, opened 2020-04-19
Negative Pattern

Gusted:
#373 The negative things

You can see with this patch
support.discordapp.com 
blog.discordapp.com
Aren't in the darklist anymore but discordapp.com is still is.

Cons
example.com/this/to/be/tested/would/not/be/detected/correctly
directory's are buggy but this would work with subdomains where they are mainly if not all used for so far I know.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -1,3 +1,5 @@
+!blog.discordapp.com
+!support.discordapp.com
 2600.com
 5stardata.info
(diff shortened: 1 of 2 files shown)


BananaManCJ:
Once this is merged, I know that dlive.tv is dark and community.dlive.tv is not

Gusted:
Going to move this to a bigger problem and a New PR for it

Issue #2342, opened 2020-04-20
isbgpsafeyet.com


jamy015:
https://isbgpsafeyet.com/ is fairly broken in dynamic mode. It works pretty well in filter mode, though.

Gusted:
![](https://i.imgur.com/x1hyBSI.png)
They have like a whole list of this color so... ![](https://i.imgur.com/pogE0WB.png)

Myshor:
Quickfix done with removing problematic background:
![image](https://user-images.githubusercontent.com/56877029/79912191-ed67a000-8421-11ea-9b19-8ae3151ce7e5.png)

Will it be acceptable?

Gusted:
All the titles and h3 h4 etc are not visible

Myshor:
I'll do more fixes to the branch tomorrow. I turned off PC already. 😉

Myshor:
Added fixes for headers and button color.
![image](https://user-images.githubusercontent.com/56877029/79951129-6e518680-8478-11ea-8e07-774640c08078.png)

Gusted:
#2348 has been merged

PR #2348, opened 2020-04-21
fix for #2342

Myshor:
Quick and easy CSS fix for #2342


Myshor:
Quickfix done with removing problematic background:
image

Will it be acceptable?

Myshor:
Added fixes for headers and button color.
image

Issue #2361, opened 2020-04-23
Edge extension doesn't list extension store as protected


BananaManCJ:
In the Edge version of the extension, https://microsoftedge.microsoft.com/addons is not listed as "protected by browser." Usually it would say "This page is protected by browser."

System Information:
 - OS: Windows 10
 - Browser: Edge Chromium
 - Version: 80.0.361.69

You should also investigate if Edge protects https://chrome.google.com/webstore so it could possibly be themed.

Gusted:
Chromium so most likely webstore of google has also been flagged as protected page

PR #2362, opened 2020-04-23
List Edge Add-ons store as protected

BananaManCJ:
This will make it list https://microsoftedge.microsoft.com/addons as protected in Chrome as well, but most Chrome users wouldn't ever go to that page. To make it only protected in Edge would require app changes.


BananaManCJ:
Fixes #2361

Gusted:
Can't find any documentation for protected pages.

PR #2369, opened 2020-04-24
google apps icon invert fix

Myshor:
fix excluding classes and independent from language
image


Myshor:
Checked for PL and EN.
@Gusted, @alexanderby can you check if it works in your languages too? ;)

Gusted:
NL works

PR #2379, opened 2020-04-25
epicgames.com is dark

Kurtoid:


PeterMinin:
As previously noted (#1168), it's not all dark. From a quick look around, it seems like only `epicgames.com/store/*` is.

Gusted:
It's very hard to get all subdomains checked especially if you their isn't a global solution to see all avaible paths etc

PR #2389, opened 2020-04-26
fixes for facebook.com

Myshor:
Added fixes to have dark mode map even in new design.


Myshor:
Yup. Checked both designs.
This can be merged and will not break anything in old design.

PR #2390, opened 2020-04-26
Fix logo color on amazon.com when placing an Amazon Smile Order.

KyleGospo:
image

image


KyleGospo:
Also corrects gift orders
image

image

PR #2406, opened 2020-04-27
Fixed Jira issue headers to show correct background color

sdo91:


Gusted:
Hi @sdo91 
Could you give an example link?

Gusted:
@sdo91 ^

sdo91:
Hmm, not sure about an example link. It would look something like this: https://mycompanyname.atlassian.net/browse/theissuenumber

Gusted:
Could you give an screenshot Before and After, you can of course redact sential info I just need to see how it looks like after the fix.

PR #2410, opened 2020-04-28
Fix for repl.it editor cursor

### LINKED ISSUE #988: [Dynamic] Cursor in Repl.it, possibly other Monaco Editor sites
Cursor is invisible in Repl.it in the Monaco editor in both the (site's) dark and light theme. Note that unfortunately dark theme only pertains to the actual REPLs and editor, and the rest of the pages on the site  are still bright white (so disabling DarkReader for entire site is not really an option).

I assume this might be an issue on other sites using Microsoft Monaco Editor as well.

I tried using the Developer Tools to fix this myself but no luck. The Developer Tools for Dynamic mode don't allow CSS sections to be added (they revert the file).

Mukundan314:
Fixes #988

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2398,6 +2398,15 @@ p, li {
 
 ================================
 
+repl.it
+
+CSS
+.monaco-editor .cursor {
+    background-color: ${#000};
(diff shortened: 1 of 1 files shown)


Gusted:
Example link?

PR #2411, opened 2020-04-28
Devtools patch

### LINKED ISSUE #2158: Confirmation before resetting custom css
For Static mode, you can specify your own css. So for websites that I go to frequently, I designed my own css. But one misclick and it resets to default, no way to get it back. So could anyone consider adding a confirmation box before reset? 

I could probably use some other custom css injecting addon for my use case, but I don't trust it with the permission to access all my data.

Gusted:
Reset button confirmation.
resolved #2158

### DIFF
--- a/src/background/devtools.ts
+++ b/src/background/devtools.ts
@@ -81,6 +81,7 @@ export default class DevTools {
     private static KEY_DYNAMIC = 'dev_dynamic_theme_fixes';
     private static KEY_FILTER = 'dev_inversion_fixes';
     private static KEY_STATIC = 'dev_static_themes';
+    private static KEY_DONTASK = "dev_do_not_ask_again";
 
     private getSavedDynamicThemeFixes() {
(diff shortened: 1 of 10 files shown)


alexanderby:
As far as I understand, the issue #2158 refers to CSS editor too.

So, let's split this PR into parts:
1. Tab key for Dev Tools. See how to better handle it [here](https://github.com/alexanderby/mova/blob/master/src/editor/index.js#L4).
2. Conformation dialog for CSS editor. I think we will need it for many operations, like resetting settings, so I will create a <ModalDialog> element to use in many places. Also, I think we can just use page's own local storage to not deal with message passing and clattering the background page.
3. We will copy/paste this for Dev Tools and for other places too.

Gusted:
As of splitting this

Gusted:
DO NOT DELETE Devtools-Patch

PR #2438, opened 2020-05-03
fixes for facebook.com

Myshor:
Two icons fix for not using black inverted blobs.
.sp_Jc8OKpJq5NW .sp_C8jQ3Z1xUrS
image


Gusted:
10/10 Pull request, just because of the picture showing in the example

PR #2443, opened 2020-05-05
Remove rules for metacritic.com

Rakleed:
Since they changed the site design, this rule is obsolete.

[metacritic.com](https://www.metacritic.com/)


Gusted:
Thank you for Contributing 😄

PR #2448, opened 2020-05-05
fix for facebook.com

Myshor:
Lightening bar for stories.
image


Myshor:
Added two different fixes for Black "X" 'caused by forcing inverted blob as background:
image

PR #2461, opened 2020-05-07
Automation patch

Gusted:


src/ui/popup/main-page/sun-moon-icon.tsx
```diff
@@ -0,0 +1,55 @@
+import {m} from 'malevic';
```
alexanderby, line 1: If it is a full copy-paste, let's use `git mv SRC_PATH DST_PATH` command for `sun-moon-icon.tsx`, `system-icon.tsx` and `watch-icon.tsx` files. This way the history will be preserved. Note that file paths in JS `import` for old design should be corrected.

src/ui/popup/main-page/style.less
```diff
@@ -4,6 +4,25 @@
...
+    &__time {
+        background-color: @color-heading;
+        border-radius: 50%;
+        color: @color-control-fore;
+        cursor: pointer;
+        display: none;
+        height: @size-text-small-height;
+        pointer-events: none;
+        position: absolute;
+        right: 4.78125rem;
```
alexanderby, line 17: Maybe try some beautiful number, like `calc(100% - @size-text-small-height / 2)`.

src/ui/popup/automation-page/style.less
```diff
@@ -0,0 +1,111 @@
...
+        &:hover &__button,
+        &:hover &__checkbox {
+            background-color: @color-control-hover;
+        }
+    }
+    
+    .toggle {
+        width: 100%;
+    }
+}
```
alexanderby, line 111: Missing new line at the end of file

src/ui/popup/main-page/style.less
Gusted: I'm not quite sure if an beautiful number can come to this number as in the other original CSS for this icon the right was -0.2.....

alexanderby: The original was placed in the corner, but here the whole width is `100% - 2 * @size-border`, the right edge of the 2nd button is `67% - @size-border / 3`, so maybe the number `67% - @size-border / 3 - 2 * @size-text-small-height / 3` should do the trick.

alexanderby: Oh wait, it is `left` position, but for `right` try `33% + @size-border / 3`.

Issue #2467, opened 2020-05-08
[Broken Website] 


theairbend3r:
**Website**
https://tailwindcss.com/

**How to Reproduce**
An example for this is
> 1. Go to x site

**Expected behavior**
Navbar and sidebar, along with some other components do not invert colours.

**Actual behavior**
Navbar and sidebar should invert colours.

**Screenshots**
![image](https://user-images.githubusercontent.com/60727948/81374422-d660c780-911c-11ea-8980-32afa8d3afdd.png)

**System info:**
 - OS: Ubuntu 18
 - Browser: Firefox
 - Darkreader Version: 4.9.8

Myshor:
I made fixes for that white backgrounds + headers texts will be more readable.
![image](https://user-images.githubusercontent.com/56877029/81437718-ed7ad600-916b-11ea-9327-569e8bc25c97.png)

theairbend3r:
Amazing! Thanks for the speedy PR! :)

PR #2471, opened 2020-05-08
fix for tailwindcss.com

Myshor:
Fix for issue #2467
image


Myshor:
Main page after fixes:
image

PR #2475, opened 2020-05-09
Add admin.microsoft.com to dark-sites.config

bherila:
Dark Reader doesn't work properly with https://admin.microsoft.com/ 

However, https://admin.microsoft.com/ has its own Dark Mode setting which works great! 

The way to enable this is to click the "Dark Mode" button which results in a POST to https://admin.microsoft.com/admin/api/navigation/darkmode with payload 

```js
{IsDarkModeOn: true}
```

Does this qualify to be added to dark-sites.config?

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -1,6 +1,7 @@
 2600.com
 5stardata.info
 acm.sdut.edu.cn/onlinejudge3/
+admin.microsoft.com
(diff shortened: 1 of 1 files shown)


Gusted:
Sorry, no this doesn't qualify for dark site as it's native dark theme and not by default and no changing back to light theme...

PR #2481, opened 2020-05-10
fix for facebook.com

Myshor:
Added fix for black maximize icon in movies.


Myshor:
Removed ._3w97 invert (i) website information icon.
Added .3w97 to css brightness fix to have it white in both old and new design.

Myshor:
Final results can not be better until old design for facebook will be totally abandoned.
No pictures this time. 😜

PR #2482, opened 2020-05-10
Replace rollup packages with newer maintained

Gusted:
> rollup-plugin-commonjs: `10.1.0`  → @rollup/plugin-commonjs: `^11.1.0`
> rollup-plugin-node-resolve: `5.2.0` → @rollup/plugin-node-resolve: `^7.13`
> rollup-plugin-replace: `2.2.0` → @rollup/plugin-replace: `^2.3.2`
> Added @rollup/pluginutils: `^3.0.10`

> npm WARN @rollup/pluginutils@3.0.8 requires a peer of rollup@^1.20.0 but none is installed. You must install peer dependencies yourself. 

Adding @rollup/pluginutils as devedepency to newest version fixed this warning.

It's preserving rollup-plugin-typescript2 for the compilers errors


package.json
```diff
@@ -31,6 +31,10 @@
     "malevic": "0.18.2"
   },
   "devDependencies": {
+    "@rollup/plugin-commonjs": "^11.1.0",
```
alexanderby, line 34: Let's use strict packages' versions (without `^`). This way we will ensure that everyone (including reviewers from Mozilla) is using the same packages, and has the same build results and bugs. Also this can protect from vulnerabilities like [this](https://www.theregister.co.uk/2016/03/23/npm_left_pad_chaos/) or [this](https://snyk.io/blog/malicious-code-found-in-npm-package-event-stream/).

Gusted: Woops seems like --save-dev isn't the best option for automatic Install smh

alexanderby: You can add `--save-exact` or `-E`

Gusted: Noted

PR #2483, opened 2020-05-10
Fix for wikipedia.org

Rakleed:
Fix three dots on the main page


Myshor:
Some screen and/or example URL? I can not find this class using devtools in Chrome.

Rakleed:
https://ru.wikipedia.org/wiki/Заглавная_страница

Myshor:
Ok... I see it now. On en and pl mainpage I could not find that. 😃

PR #2493, opened 2020-05-11
Devtools tab

Gusted:


src/ui/devtools/components/body.tsx
```diff
@@ -30,11 +30,23 @@ function Body({data, actions}: BodyProps) {
...
         }
+        node.addEventListener('keydown', (e) => {
+            if (e.key === 'Tab') {
+                e.preventDefault();
+                const start = node.selectionStart;
+                const end = node.selectionEnd;
+                const before = node.value.substring(0, start);
+                const after = node.value.substring(end);
+                node.focus();
+                node.value = `${before}\t${after}`;
```
alexanderby, line 46: Let's replace `\t` with 4 spaces.

src/ui/devtools/style.less
```diff
@@ -65,6 +65,7 @@ header {
     outline: none;
     padding: @size-control-inner / 2 - @size-text-normal / 2;
     resize: none;
+    tab-size: 4;
```
alexanderby, line 68: We are not using Tab symbol in config, so there should be no need for this property.

src/ui/devtools/components/body.tsx
Gusted: For some reason it would just add 1 space... Render body is resetting it to 1 space.......

Gusted: image
Bad idea to use 4 spaces

src/ui/devtools/components/body.tsx
```diff
@@ -30,11 +30,23 @@ function Body({data, actions}: BodyProps) {
...
         }
+        node.addEventListener('keydown', (e) => {
+            if (e.key === 'Tab') {
+                e.preventDefault();
+                const start = node.selectionStart;
+                const end = node.selectionEnd;
+                const before = node.value.substring(0, start);
+                const after = node.value.substring(end);
+                node.focus();
+                node.value = `${before}  ${after}`;
```
alexanderby, line 46: Make it 4 spaces.

src/ui/devtools/components/body.tsx
```diff
@@ -30,11 +30,23 @@ function Body({data, actions}: BodyProps) {
...
+                node.setSelectionRange(start + 1, start + 1);
```
alexanderby, line 47: Fix the cursor position, play with the selection range to place it after the inserted 4 spaces.

PR #2494, opened 2020-05-11
Inverting Favicon

### LINKED ISSUE #2489: [Feature Request] Invert black favicons for dark browser themes
There are many websites that have a favicon that is just a black logo graphic, which is hard to see when using a dark browser theme and sometimes is a hindrance when trying to quickly find a tab. My idea is to have a list of those websites (like there already is for dark websites) and invert their favicon.

I already know this is possible to do, because I made a little test extension that does exactly what I described. You can find it [here on Github](https://github.com/Mertyn/favicon-invert), try it out for yourself and take a look at the code to see how it works internally
Here's a little before and after screenshot:
favicon-invert

The function can still be improved upon, like making it not just completely invert the whole image but only make black parts white for example.
I also tried to implement this functionality in darkreader myself, but I find the code pretty hard to understand since I am not an active developer here and have never really worked with typescript before.
I hope you think this could be a useful feature aswell and this helped make darkreader just that little bit better than it already is. If there is no interest at all in adding this to darkreader just let me know and I will break this off as my own extension project.

Thanks for reading and I am looking forward to hearing back :)

Gusted:
https://www.amd.com/en
https://unsplash.com/

Major credits to @Mertyn for the 95% of the code :D
Resolved #2489

### DIFF
--- a/src/inject/dynamic-theme/image.ts
+++ b/src/inject/dynamic-theme/image.ts
@@ -18,19 +18,50 @@ export interface ImageDetails {
 const FaviconSelector = "link[rel='icon'], link[rel='shortcut icon']";
 
 function invertImage(image: HTMLImageElement) {
+    const MAX_ANALIZE_PIXELS_COUNT = 32 * 32;
+
+    const naturalPixelsCount = image.naturalWidth * image.naturalHeight;
+    const k = Math.min(1, Math.sqrt(MAX_ANALIZE_PIXELS_COUNT / naturalPixelsCount));
+    const width = Math.max(1, Math.round(image.naturalWidth * k));
+    const height = Math.max(1, Math.round(image.naturalHeight * k));
+    
     const canvas = document.createElement('canvas');
-    canvas.width = image.width;
-    canvas.height = image.height;
+    canvas.width = width;
+    canvas.height = height;
     const ctx = canvas.getContext('2d');
+    ctx.imageSmoothingEnabled = false;
     ctx.drawImage(image, 0, 0);
     const imgData = ctx.getImageData(0, 0, image.width, image.height);
     const data = imgData.data;
 
-    for (let i = 0, len = data.length; i < len; i += 4) {
-        data[i] = 255 - data[i];
-        data[i+1] = 255 - data[i+1];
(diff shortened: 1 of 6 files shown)


Gusted:
TODO: Only actualy invert if the user have a black theme.

Mertyn:
I would maybe also make it only invert black (or almost black) pixels, so that if there are other colors and things in the graphic it doesn't invert them aswell.
Could be something like this in the inversion for-loop:
```typescript
for (let i = 0, len = data.length; i < len; i += 4) {
    if ( (data[i] == data[i+1] && data[i+1] == data[i+2]) && data[i] < blackThreshold) {
        data[i] = 255 - data[i];
        data[i+1] = 255 - data[i+1];
        data[i+2] = 255 - data[i+2];
    }
 }
```
So basically it should check if the rgb values are the same (or very similar maybe) and then if the pixel value is under some sort of threshold to be considered black. You can probably make the code a little cleaner and easier to understand.
Hope you understand what I have in mind :)

Gusted:
https://github.com/darkreader/darkreader/blob/master/src/inject/dynamic-theme/image.ts#L80 

This but than just to check if it's pass as an dark pixel :) Is possible @Mertyn

Mertyn:
> https://github.com/darkreader/darkreader/blob/master/src/inject/dynamic-theme/image.ts#L80
> 
> This but than just to check if it's pass as an dark pixel :) Is possible @Mertyn

Yeah right, that would basically do what I had in mind :)

Gusted:
> I would maybe also make it only invert black (or almost black) pixels, so that if there are other colors and things in the graphic it doesn't invert them aswell.
> Could be something like this in the inversion for-loop:

You actually brought me onto an idea to fix math formula's :)

AndrewDaGuy:
Is this a still work in progress? I would like this to get merged :]

Gusted:
> Is this a still work in progress? I would like this to be a feature :]

```
I see 2 high priorities: redesign and static CSS themes. Many features, like favicon inversion, seem like low priority, sorry.
```

I'd hope around this summer a lot of features can get an speed-up. But an strict timeline or even an lossy timeline isn't an guarantee.

PR #2508, opened 2020-05-13
Add Ipv6 support

### LINKED ISSUE #197: Toggle button doesn't work on IPv6 addresses
Pressing the "Toggle" button on [::1]:3000 (ipv6 version of localhost:3000) has no effect. The only way to disable the extension is to press the "off" button which globally disables Dark Reader :(

Could you please make the toggling work with IPv6 loopback addresses?

Gusted:
Resolved #197

### DIFF
--- a/src/utils/ipv6.ts
+++ b/src/utils/ipv6.ts
@@ -1,17 +1,17 @@
 /**
  * Compares if IPV6 addresses are the same.
- * @param firstIp First IP to compare with
- * @param secondIp Second IP to compare with
+ * @param firstIP First IP to compare with
+ * @param secondIP Second IP to compare with
  */
-export function CompareIPV6(firstIp: string, secondIp: string) {
-    firstIp = (firstIp
+export function compareIPV6(firstIP: string, secondIP: string) {
+    firstIP = (firstIP
         .replace(/^\^/, '') // Remove ^ at start
         .replace(/\$$/, '') // Remove $ at end
         .replace(/^.*?\/{2,3}/, '') // Remove scheme
         .replace(/\?.*$/, '') // Remove query
         .replace(/\/$/, '') // Remove last slash
     );
-    const ip = secondIp.toLowerCase();
-    const ip2 = firstIp.toLowerCase();
+    const ip = secondIP.toLowerCase();
+    const ip2 = firstIP.toLowerCase();
     return ip === ip2;
 }
--- a/src/utils/ipv6.ts
+++ b/src/utils/ipv6.ts
@@ -61,7 +61,7 @@ function validate(ip: string) {
     );
     let ip = decompress(ipv6.substr(0, ipv6.indexOf(']') + 1));
     let ip2 = decompress(orignalIp2.substr(0, orignalIp2.indexOf(']') + 1));
-    ip = '[' + ip + ipv6.substr(ipv6.indexOf(']'))
-    ip2 = '[' + ip2 + orignalIp2.substr(orignalIp2.indexOf(']'))
+    ip = '[' + ip + ipv6.substr(ipv6.indexOf(']')).toLowerCase()
+    ip2 = '[' + ip2 + orignalIp2.substr(orignalIp2.indexOf(']')).toLowerCase();
     return ip === ip2;
 }
--- a/src/utils/ipv6.ts
+++ b/src/utils/ipv6.ts
@@ -1,19 +1,17 @@
 /**
  * Compares if IPV6 addresses are the same.
- *
- * @param {string} orignalIp
- * @param {string} compareIp
+ * @param FirstIp First IP to compare with
+ * @param SecondIp Second IP to compare with
  */
-export function compareIPV6(orignalIp: string, compareIp: string) {
-    const orignalIp2 = (orignalIp
+export function compareIPV6(FirstIp: string, SecondIp: string) {
+    FirstIp = (FirstIp
         .replace(/^\^/, '') // Remove ^ at start
         .replace(/\$$/, '') // Remove $ at end
         .replace(/^.*?\/{2,3}/, '') // Remove scheme
         .replace(/\?.*$/, '') // Remove query
         .replace(/\/$/, '') // Remove last slash
     );
-    const ip = compareIp.toLowerCase();
-    const ip2 = orignalIp2.toLowerCase();
+    const ip = SecondIp.toLowerCase();
+    const ip2 = FirstIp.toLowerCase();
     return ip === ip2;
 }
-
--- a/src/utils/ipv6.ts
+++ b/src/utils/ipv6.ts
@@ -1,57 +1,10 @@
-function decompress(ipv6: string) {
-
-    ipv6 = ipv6.replace('[', '').replace(']','');
-
-    if (!validate(ipv6)) {
-       throw new Error('Invalid address: ' + ipv6);
-    }
-    ipv6 = ipv6.toLowerCase()
- 
-    const nh = ipv6.split(/\:\:/g);
-    if (nh.length > 2) {
-       throw new Error('Invalid address: ' + ipv6);
-    }
- 
-    let sections = [];
-    if (nh.length == 1) {
-       sections = ipv6.split(/\:/g);
-       if (sections.length !== 8) {
-          throw new Error('Invalid address: ' + ipv6);
-       }
-    } else if (nh.length == 2) {
-       const n = nh[0];
-       const h = nh[1];
-       const ns = n.split(/\:/g);
-       const hs = h.split(/\:/g);
-       for (let i in ns) {
-          sections[i] = ns[i];
-       }
-       for (let i = hs.length; i > 0; --i) {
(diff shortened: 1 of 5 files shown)


src/utils/url.ts
```diff
@@ -28,8 +29,12 @@ export function isURLInList(url: string, list: string[]) {
  * @param urlTemplate URL template ("google.*", "youtube.com" etc).
  */
 export function isURLMatched(url: string, urlTemplate: string): boolean {
-    const regex = createUrlRegex(urlTemplate);
-    return Boolean(url.match(regex));
+    if (url.indexOf('[') === -1 || urlTemplate.indexOf('[') === -1) {
```
alexanderby, line 32: Better use `String.includes`. I think that IP address is a rare case. IPv6 is even more rear case. Let's simplify it. Let's fast check if both URLs are IPv6 addresses. If some of them is, and other not, we can return `false`.

src/utils/url.ts
```diff
@@ -28,8 +29,12 @@ export function isURLInList(url: string, list: string[]) {
...
+        const regex = createUrlRegex(urlTemplate);
+        return Boolean(url.match(regex));
+    } else { //ipv6
```
alexanderby, line 35: This comment is pointless

src/utils/ipv6.ts
```diff
@@ -0,0 +1,67 @@
...
+    const orignalIp2 = (orignalIp
+        .replace(/^\^/, '') // Remove ^ at start
+        .replace(/\$$/, '') // Remove $ at end
+        .replace(/^.*?\/{2,3}/, '') // Remove scheme
+        .replace(/\?.*$/, '') // Remove query
+        .replace(/\/$/, '') // Remove last slash
+    );
+    let ip = decompress(ipv6.substr(0, ipv6.indexOf(']') + 1));
+    let ip2 = decompress(orignalIp2.substr(0, orignalIp2.indexOf(']') + 1));
+    ip = '[' + ip + ipv6.substr(ipv6.indexOf(']')).toLowerCase()
```
alexanderby, line 64: Is there a need in `[` symbol?

src/utils/ipv6.ts
```diff
@@ -0,0 +1,67 @@
...
+    ip2 = '[' + ip2 + orignalIp2.substr(orignalIp2.indexOf(']')).toLowerCase();
+    return ip === ip2;
+}
```
alexanderby, line 67: Missing new line

src/utils/ipv6.ts
```diff
@@ -0,0 +1,67 @@
...
+ 
+ function leftPad(section: string, padd: string, times: any) {
+    const padding = padd.repeat(times);
+    if (section.length < padding.length) {
+        section = padding.substring(0, padding.length - section.length) + section;
+    }
+    return section;
+ };
+
+ export function matchIPV6(orignalIp: string, ipv6: string) {
```
alexanderby, line 54: Let's compare IPv6 addresses in this function only. Rename arguments accordingly.

src/utils/ipv6.ts
```diff
@@ -0,0 +1,67 @@
...
+    const orignalIp2 = (orignalIp
```
alexanderby, line 55: Let's strip everything leaving IP and port only (the host part).

src/utils/ipv6.ts
```diff
@@ -0,0 +1,67 @@
...
+        .replace(/^\^/, '') // Remove ^ at start
+        .replace(/\$$/, '') // Remove $ at end
+        .replace(/^.*?\/{2,3}/, '') // Remove scheme
+        .replace(/\?.*$/, '') // Remove query
+        .replace(/\/$/, '') // Remove last slash
+    );
+    let ip = decompress(ipv6.substr(0, ipv6.indexOf(']') + 1));
```
alexanderby, line 62: If both are IPv6, then I think we don't need a decompression.

src/utils/ipv6.ts
```diff
@@ -0,0 +1,67 @@
...
+    let ip2 = decompress(orignalIp2.substr(0, orignalIp2.indexOf(']') + 1));
+    ip = '[' + ip + ipv6.substr(ipv6.indexOf(']')).toLowerCase()
+    ip2 = '[' + ip2 + orignalIp2.substr(orignalIp2.indexOf(']')).toLowerCase();
```
alexanderby, line 65: Is there a need in `[` symbol? Let's compare the IPv6 address and port.

src/utils/url.ts
```diff
@@ -1,4 +1,5 @@
 import {UserSettings} from '../definitions';
+import { matchIPV6 } from './ipv6';
```
alexanderby, line 2: Code style

tests/url.tests.ts
```diff
@@ -125,6 +125,23 @@ test('URL is enabled', () => {
...
+        {siteList: ['google.com'], siteListEnabled: [], applyToListedOnly: false} as UserSettings,
+        {isProtected: false, isInDarkList: false},
+    )).toBe(true);
+    expect(isURLEnabled(
+        '[::1]:8080',
+        {siteList: ['[::1]:8080'], siteListEnabled: [], applyToListedOnly: false} as UserSettings,
+        {isProtected: false, isInDarkList: false},
+    )).toEqual(false)
+    expect(isURLEnabled(
+        '[::1]:8080',
```
alexanderby, line 140: Let's add a test with save IPv6 and different ports.

src/utils/url.ts
```diff
@@ -28,8 +29,12 @@ export function isURLInList(url: string, list: string[]) {
  * @param urlTemplate URL template ("google.*", "youtube.com" etc).
  */
 export function isURLMatched(url: string, urlTemplate: string): boolean {
-    const regex = createUrlRegex(urlTemplate);
-    return Boolean(url.match(regex));
+    if (!(url.includes('[')) || !(urlTemplate.includes('['))) {
```
alexanderby, line 32: Let's save the result of `includes` into some variables. If one includes and other not, return false. If both include, check IPv6, otherwise check as usual.

Gusted: Isn't that what || is for

src/utils/url.ts
```diff
@@ -22,14 +23,36 @@ export function isURLInList(url: string, list: string[]) {
...
+
 /**
  * Determines whether URL matches the template.
  * @param url URL.
  * @param urlTemplate URL template ("google.*", "youtube.com" etc).
  */
 export function isURLMatched(url: string, urlTemplate: string): boolean {
-    const regex = createUrlRegex(urlTemplate);
-    return Boolean(url.match(regex));
+    if (isIPV6(url, urlTemplate)) {
```
alexanderby, line 50: I meant
```javascript
const isFirstIPV6 = ip1.includes('[');
const isSecondIPV6 = ip2.includes('[');
if (isFirstIPV6 && isSecondIPV6) {
    return isIPV6(url, urlTemplate);
} else if(!isFirstIPV6 && ! isSecondIPV6){
    const regex = createUrlRegex(urlTemplate);
    return Boolean(url.match(regex));
} else {
    return false;
}

File: src/utils/ipv6.ts [hunk 13]
```diff
@@ -0,0 +1,17 @@
+/**
+ * Compares if IPV6 addresses are the same.
+ * @param FirstIp First IP to compare with
+ * @param SecondIp Second IP to compare with
+ */
+export function compareIPV6(FirstIp: string, SecondIp: string) {
```
alexanderby, line 6: There is a convention, that only class names start from Upper Case. The exception is for JSX components.

src/utils/ipv6.ts
```diff
@@ -0,0 +1,17 @@
...
+    FirstIp = (FirstIp
+        .replace(/^\^/, '') // Remove ^ at start
+        .replace(/\$$/, '') // Remove $ at end
+        .replace(/^.*?\/{2,3}/, '') // Remove scheme
+        .replace(/\?.*$/, '') // Remove query
+        .replace(/\/$/, '') // Remove last slash
+    );
+    const ip = SecondIp.toLowerCase();
+    const ip2 = FirstIp.toLowerCase();
+    return ip === ip2;
```
alexanderby, line 16: The port (:8080 or whatever) should be compared too.

tests/url.tests.ts
```diff
@@ -125,6 +125,38 @@ test('URL is enabled', () => {
...
     )).toBe(false);
 
+    // IPV6 Testing
+    expect(isURLEnabled(
+        '[::1]:1337',
+        {siteList: ['google.com'], siteListEnabled: [], applyToListedOnly: false} as UserSettings,
+        {isProtected: false, isInDarkList: false},
+    )).toBe(true);
+    expect(isURLEnabled(
+        '[::1]:8080',
```
alexanderby, line 135: Add a test for `[::1]:8080` and `[::1]:8081` in site list, so that we can ensure that ports are compared.

src/utils/ipv6.ts
Gusted: Both are stripped with only host and port :)

Issue #1896, opened 2020-01-15
[Bug] Can't enable Dark Reader at all on some websites


KaKi87:
There are specific websites from which I can't enable Dark Reader at all : nothing happens when I click on the switch button and it stays on disabled state.

Here is the list :
- https://education.github.com/pack/offers
- https://github.community/

KaKi87:
up

alexanderby:
The first one requires login, the second works well with the latest build, wait for an update.
<img width="1199" alt="Screen Shot 2020-03-12 at 20 51 10" src="https://user-images.githubusercontent.com/9638588/76550565-62e76480-64a3-11ea-8b98-46c1179bcb8c.png">

Gusted:
@KaKi87 Can you comfirm that this works with the latest version of darkreader

KaKi87:
Yes, here is proof :

![2020-05-19-21-09-07](https://user-images.githubusercontent.com/21284089/82368234-60f6d000-9a15-11ea-902e-3779b99589eb.gif)

My version is 4.9.8, I don't know if it's the latest one, since there are no more git release tags since 2018...

Gusted:
What's your sitelist?
4.9.8 is the latest version of darkreader we don't publish anymore to release tab.

KaKi87:
> What's your sitelist?

I can only select it line-by-line, is there another way ?

> we don't publish anymore to release tab

Why so ? What's the purpose of versioning then ?

Gusted:
> I can only select it line-by-line, is there another way ?

Screenshot?

> Why so ? What's the purpose of versioning then ?

Wasn't my decision https://github.com/darkreader/darkreader/issues/2452#issuecomment-628196533

KaKi87:
> Screenshot?

Too much scroll height, here's another GIF ^^

![2020-05-19-22-54-30](https://user-images.githubusercontent.com/21284089/82377495-09ac2c00-9a24-11ea-9bd5-b366d45707e0.gif)

Gusted:
Their we got it an old but if you remove github.com temporarily from the list you can toggle subdomains

KaKi87:
Do you mean that disabling darkify on a domain prevents all subdomains to be darkified ? That's a hell of a bug !
What about `github.community` ? It's not a subdomain ! So what, it the site list works using `String.includes` ? 😅
[...]

PR #2517, opened 2020-05-14
Better URL Matching

### LINKED ISSUE #84: Sub domain are always matched by pattern
Hi,

I didn't find a way to exclude for instance [http://thetvdb.com/?tab=series&id=78804](http://thetvdb.com/?tab=series&id=78804) [must not be dark] without touching the sub-domain [https://forums.thetvdb.com/viewtopic.php?t=27159](https://forums.thetvdb.com/viewtopic.php?t=27159) [must be dark].

Is there a way or is it a bug?

### LINKED ISSUE #147: Sub domain control
If I use to invert the only mode, I add xxx.com to the site list, then I visit sub.xxx.com, I don't want to invert this domain. I can't exclude it. When I use the shortcut key to toggle the domain, extension just adds or remove the sub.xxx.com from site list, but the page has no effect.

I suggest a feature:
If current need add domain is a subdomain of a domain that already exists in the list. Then we do the opposite of the parent domain(add the subdomain to a exclude list?), or we remove the parent domain from the list(When user go back to the parent domain, he can re-add it to the list).

If current need add domain is a parent domain of a exists domain. Then we add it to the list, and don't change the exists domain. (When the parent remove from list, the old subdomain will be saved in the list, so the function of the subdomain has no effect)

In general, I want to add some features to avoid frequent visits to a website, toggle dark reader, but no effect problem.
(Sorry for my poor English)

### LINKED ISSUE #373: URL match patterns

### LINKED ISSUE #660: Subdomain of sites in global dark list may not be dark

### LINKED ISSUE #1614: Dynamic Dark Mode Theme reapplied on tab switch when turned off for a given site
Since Discord has its own dark mode theme which I'm used to (and this plugin gave its server icons a washed out makeover that made them barely visible), I toggled discord off on the extension. However, whenever I switch away to another tab and then switch back, the dark mode theme returns after a second or two. But it doesn't apply to the images for some reason. It's like half-applied.

The extension shouldn't even be touching the site at all. That's what that current site toggle is for, right? I have to toggle it enabled again for the current site (the images go weird again) and then disable it again for the site to return to normal. Switching between light and dark mode also re-disables the theme again.

Chrome 77, Windows 10.
I'm using the extension in the default blacklist mode, with the default dynamic theme.

### LINKED ISSUE #1689: remove support.discordapp.com from global dark list
image
it isnt dark.

### LINKED ISSUE #2055: SubDomains unable to work if main Domain is disabled
Example: If the plugin is disabled on https://yagpdb.xyz, it is unable to work on its subdomains like https://docs.yagpdb.xyz.

Video: 
https://streamable.com/nhjyq

### LINKED ISSUE #3359: [BUG] Domains and Subdomains Behave Differently on Site List
After further reading, this appears to be a bug, and not a Broken Website. I've updated the issue to reflect that.

**Dark Reader Version**: 4.9.16

### **Bug Description**

- In the case of toggling Dark Reader on a subdomain, the entry into the list is specific, i.e., only effecting `docs.gitlab.com`, and not `gitlab.com` or any other subdomains.
- In the case of toggling Dark Reader on a domain, the entry into the list is generic, i.e., it effects `gitlab.com`, **_as well as_** `docs.gitlab.com` and any other subdomain.

This behavior is inconsistent and seems more like a bug.

**Potential Bugfix**: Even adding a future that would allow you to manually input `//gitlab.com` to limit it to the domain (i.e., excluding sub-domains) would be a great way to compromise on this bug! :smile: 

A broader discussion on this started in [this comment below](#issuecomment-672573736).

------------------------------------------------------

### ORIGINAL POST

> **Website**
> https://docs.gitlab.com/ee/ci/quick_start/
> 
> **How to Reproduce**
> Steps to reproduce the behavior:
> An example of this is
> > 1. Go to Gitlab Docs website
> > 2. Attempt to activate Dark Reader
> 
> **Expected behavior**
> Dark Reader activates and Dark Reader generates a dark theme
> 
> **Actual behavior**
> docs.gitlab.com remains disabled
> 
> **Screenshots**
> image
> 
> 
> 
> **System info:**
> - OS: All OS's (screenshot is from Arch; tried on macOS and Windows as well)
> - Browser: Attempted with both Chrome and Firefox

### LINKED ISSUE #3729: [BUG] Similar domains being being recognized as the same one
### **Describe the bug**
A domain is being darkned because it is similar to another domain on the list.

### **Steps to reproduce**

- Go to www.wuxiaworld.co
- Add domain to the list
- Go to www.wuxiaworld.com

### **Expected behavior**
www.wuxiaworld.co should always get darkned by the extension but www.wuxiaworld.com should not.

### **Actual behavior**
Both domains are being darkned by the extension

### **System Information:**
 - OS: Windows 10
 - Browser: Chrome 85.0.4183.121 
 - Dark Reader version: 4.9.21

### LINKED ISSUE #3802: [BUG]
### **Describe the bug**

If Dark Reader is disabled for a URL such as 10.0.0.1, any other URL's that trail with the last digit (10.0.0.12, 10.0.0.135)
Also applies to any other combination of numbers at the end.

### **Steps to reproduce**

Disable Dark Reader for a site such as 10.0.0.1, then going to a site such as 10.0.0.123, will have the .123 site disabled, and no way to re-enable. Must enable for 10.0.0.1 first for Dark Reader to apply.

### **Expected behavior**

Each local IP should be a separate site as seen by Dark Reader

### **Actual behavior**

All local sites with a trailing number are disabled under the old site.

### **System Information:**
 - OS: Windows 10
 - Browser: Chrome 84
 - Dark Reader version: 4.9.21

### LINKED ISSUE #4263: [BUG] wildcards like foo*. do not work
### **Describe the bug**

Using a wildcard like `foo*.` in the config files does not work.

### **Expected behavior**

The config should apply to any url like `foobar.xyz` or `foo42.asd`.

### **Actual behavior**

Such websites do not get the custom fixes..

### **Screenshots**

Screenshot_20201124_104819

### **System Information:**

- OS: Manjaro
- Browser: Vivaldi-latest, Opera-latest
- Dark Reader version: 4.9.24

### LINKED ISSUE #4375: [BUG] Dark Reader handles IP addresses incorrectly
### **Describe the bug**

Dark Reader handles IP addresses incorrectly

### **Steps to reproduce**

- Add 192.168.1.1 to the exclusion list
- Make sure that 192.168.1.11 is not in the exclusion list
- Go to 192.168.1.11 

### **Expected behavior**

Dark Reader works on 192.168.1.11.

### **Actual behavior**

Dark Reader does not work on 192.168.1.11.

### **Screenshots**

image

### **System Information:**

- OS: Ubuntu 20.10
- Browser: Chrome 87.0.4280.66 (Official Build) (64-bit)
- Dark Reader version: 4.9.26

### **Additional context**

I tested this bug only on local network addresses.

### LINKED ISSUE #4675: [BUG] does not work if the query string ends with the site name from the exclude list
### **Describe the bug**

Dark Reader does not work if the query string ends with the site name from the exclude list.

### **Steps to reproduce**

- Add `example.com` to the exclude list
- Ensure that there is not `google.com` in the exclude list
- Go to 'https://www.google.com/search?q=site:https://www.example.com/'

### **Expected behavior**

Dark Reader works on `https://www.google.com/search?q=site:https://www.example.com/`

### **Actual behavior**

Dark Reader does not work on `https://www.google.com/search?q=site:https://www.example.com/`

### **Screenshots**

image

### **System Information:**

- OS: Ubuntu 20.10
- Browser: Chrome 87.0.4280.88 (Official Build) (64-bit)
- Dark Reader version: 4.9.26

### LINKED ISSUE #4731: [BUG] Problem with "[" in any URL
### **Describe the bug**

Any Site with "[" in the URL-Path will get the Darkmode applied (Maybe just uses the standard-mode).
In Addition you can't unset Darkmode (No Reaction when clicking).
Note: "Closing" with ']' doesnt have any diffrent effect.
']' alone is not affected by this Bug.

### **Steps to reproduce**

- Go to any site with "[" in the Path (Not in Query-String)
- example-URL: https://github.com/darkreader/darkreader/issues/[

### **System Information:**

- Browser: Chrome 87.0.4280.141
- Dark Reader version: 4.9.26

### LINKED ISSUE #6417: [Bug] Incorrectly parse URL
# Bug Report

## Bug Description
I have github.com in the "Not inverted list", if some other website's URL happens to include github.com then darkreader will think it's github.com and doesn't apply the dynamic theme.

## Steps To Reproduce

- Enable Dark mode
- Add github.com to "Not inverted list".
- Remove google.com from "Not inverted list" if it's in there.
- Search https://github.com in Google then click "Next" button

## Expected Behavior
google.com in dark theme.

## Actual Behavior
google.com is treated as github.com

## Screenshots
Screen Shot 2021-08-04 at 10 52 14

## System Information

- OS: macOS 11.5.1
- Browser: Firefox 90.0.2
- Dark Reader Version: 4.9.34

## Additional Context
Maybe the code in https://github.com/darkreader/darkreader/blob/master/src/utils/url.ts causes the bug?

### LINKED ISSUE #7206: [Bug] Broken when opening a local pdf file which name contains '[' or ']'
# Bug Report

## Bug Description

The extension doesn't work when opening a local pdf file which name contains '[' or ']'. 
Delete the  '[' or ']' form the filename then it works again.

## Steps To Reproduce

Open a local pdf file which name contains '[' or ']'

## Expected Behavior

The extension work

## Actual Behavior

The extension doesn't work

## System Information

- OS: Windows
- Browser: Edge 95
- Dark Reader Version: 4.9.39

Gusted:
Resolves #84 
Resolves #147 
Resolves #373
Resolves #660  
Resolves #1689
Resolves #1614
Resolves #2055 
Resolves #2103 
Resolves #3359
Resolves #3729
Resolves #3802
Resolves #4263
Resolves #4375
Resolves #4675
Resolves #4731
Resolves #6417
Resolves #7206


package.json
```diff
@@ -6,7 +6,7 @@
     "api": "node -e \"require('./tasks/bundle-api')()\"",
     "benchmark-server": "node tests/benchmark-server/index.js",
     "build": "node tasks/release.js",
-    "code-style": "eslint --fix \"src/**/*.ts\" \"src/**/*.tsx\" \"tasks/**/*.js\" \"tests/**/*.ts\" \"tests/**/*.js\" \"index.d.ts\"",
+    "code-style": "eslint --fix \"src/**/*.ts\" \"src/**/*.tsx\" \"tasks/**/*.js\" \"tests/**/*.ts\" \"tests/**/index.js\" \"index.d.ts\"",
```
Gusted, line 9: Benchmark

.eslintrc.js
```diff
@@ -39,7 +39,7 @@ module.exports = {
     },
     overrides: [
         {
-            files: ['tasks/**/*.js', 'tests/**/*.js'],
+            files: ['tasks/**/*.js', 'tests/**/index.js'],
```
Gusted, line 42: Benchmark

duredhel13:
Any news on getting this merged and released? I'm really looking forward to a fix to https://github.com/darkreader/darkreader/issues/660

Gusted:
This is a breaking change, changing a main component. A lot of system relies on this.
Also it's still unclear how this should work #373 give a good explanation. I've tried pushing this multiple times to get this merged. But their are more bug/features that is more wanted over this.

KaKi87:
Do you think it would be possible to publish on Chrome Web Store a beta version of Dark Reader that would include this PR and maybe others ?

Gusted:
> Do you think it would be possible to publish on Chrome Web Store a beta version of Dark Reader that would include this PR and maybe others ?

That will be a hell to manage.

Also seems like conflicts in this PR time to update.

KaKi87:
Okay, well, I see this PR merged in my dreams. xD

Gusted:
I hope that dream can be fillfulled this month. 
I've seen plans to get V5 by end of the month :eyes: released.

KaKi87:
Are you planning on writing proper commit messages ? :sweat_smile:

Gusted:
Uhm. If someone teaches it me 😜

Gusted:
And poof commit https://github.com/darkreader/darkreader/pull/2517/commits/8a0a646c51c8bab64e001bbe4d2f757c6e08e41b implements that the matching explicitly needs to be matched with it's domain so e.g. `stackexchange.com` won't match `codegold.stackexchange.com` and can independently toggled from each other.

Andre601:
What is the current state of this?
It would be nice to have this as I - like many others I assume - face the issue that putting a domain alone on the blacklist also includes any subpage without a way to exclude specific ones.

Gusted:
> What is the current state of this?

It's waiting to be reviewed, and expected to be merged before V5 and after simplifying the new UI.

> It would be nice to have this as I - like many others I assume - face the issue that putting a domain alone on the blacklist also includes any subpage without a way to exclude specific ones.

I also would like to have this merged, however it's probably the most breaking change that Dark Reader will occur due to the new matching system that also will differ from the current one e.g.  main domains won't match their subdomain and will be unexpected behaviour for users that don follow this issue.

So honestly? I really don't know, just @alexanderby can give more insight on this one.

Regards,
Gusted

Andre601:
> that also will differ from the current one e.g. main domains won't match their subdomain and will be unexpected behaviour for users that don follow this issue.

Isn't that the case right now?
If I set a main domain is the subdomain (not subpage) excluded and DR renders it dark.

Gusted:
Current implementation:
You have `google.com` in the sitelist it will also match mail.google.com

This implementation:

You have `google.com` in the sitelist it **WILL NOT** match mail.google.com

Andre601:
> Current implementation:
> You have `google.com` in the sitelist it will also match mail.google.com
> 
> This implementation:
> 
> You have `google.com` in the sitelist it **WILL NOT** match mail.google.com
> 

That's... not what happens for me.
DR always handles a subdomain like a completely different one.
For example, I have status.purrbot.site rendered with DR while purrbot.site is blacklisted.

Anothrr example I have is netlify.app
Any deploy preview (*.netlify.app) was having DR active while netlify wasn't....

Gusted:
> That's... not what happens for me.
> DR always handles a subdomain like a completely different one.
> For example, I have status.purrbot.site rendered with DR while purrbot.site is blacklisted.
> 
> Anothrr example I have is netlify.app
> Any deploy preview (*.netlify.app) was having DR active while netlify wasn't....

That is impossible with the current implementation other than having specific sites in your sitelist. I just tried on a fresh reinstall and had purrbot.site and saw that status.purrbot.site was disabled.

AllanChain:
Am I missing something? I expect github.community is rendered in dark mode even though github.com is not rendered. Should the following test pass?

```ts
    expect(isURLEnabled(
        'https://github.community/',
        {siteList: ['github.com'], siteListEnabled: [], applyToListedOnly: false} as UserSettings,
        {isProtected: false, isInDarkList: false},
    )).toBe(true);
```

However I added the test to `url.test.ts` but it failed.

Gusted:
Hi @AllanChain!

Thanks for testing this PR out and finding out about this bug!
I forgot that we also have domains that can start with the same letters .*com* .*com*munity and therefor was thinking it should match. 

I've fixed this edge case by https://github.com/darkreader/darkreader/pull/2517/commits/bdf0850629a7b203c7d19f5e201a742938dfed1e and added your test case to ensure it won't happen in the future.
Thank you!

Regards,
Gusted

mirsella:
does this PR is supposed to fix the problem that if for example 192.168.1.1 is in the blacklist site (because ya know router site are poorly made) and i got my server at 192.168.1.111 it will be blacklisted too even if not in the list. but when removing 192.168.1.1 from the blacklist i could finally reactivate dark reader for 192.168.1.111 or i open another issue ?

thanks

Yash-Singh1:
> does this PR is supposed to fix the problem that if for example 192.168.1.1 is in the blacklist site (because ya know router site are poorly made) and i got my server at 192.168.1.111 it will be blacklisted too even if not in the list. but when removing 192.168.1.1 from the blacklist i could finally reactivate dark reader for 192.168.1.111 or i open another issue ?
> 
> thanks

@Gusted I think this PR should attempt to resolve #4178 and #4375.

Gusted:
https://github.com/darkreader/darkreader/pull/2517/commits/5c3b7d8bd29f1e07b8d47abb51ec49752847864e

BlackDex:
First of all, thanks for this extension, and this PR.

Any idea when this will be merged?
It hurts my eye's on some sites :) (Like github.com vs docs.github.com)

mirsella:
@BlackDex you can probably merge it yourself and add it to your browser manually, but I don't know if this PR is working right now.
also for your particular example GitHub as a beta dark mode, don't know about docs.github

BlackDex:
> @BlackDex you can probably merge it yourself and add it to your browser manually, but I don't know if this PR is working right now.
> also for your particular example GitHub as a beta dark mode, don't know about docs.github

docs.github.com is white unfortunately.
Ill see if i'm able to test this.

mirsella:
> docs.github.com is white unfortunately.

also it should already work for both, just maybe it doesn't if you try to enable it on docs but disabled it on GitHub. for you you can't enable it ?

Gusted:
> It hurts my eye's on some sites :) (Like github.com vs docs.github.com)

Use `^github.com` in the sitelist.


> First of all, thanks for this extension, and this PR.
> 
> Any idea when this will be merged?

No idea :smile:



> @BlackDex you can probably merge it yourself and add it to your browser manually, but I don't know if this PR is working right now.
> also for your particular example GitHub as a beta dark mode, don't know about docs.github.com

It is working.

mirsella:
> Use ^github.com in the sitelist.

wait you can use regex for matching site ? so the example with 192.168.1.1 and  192.168.1.111 is no more valid ?

Gusted:
> wait you can use regex for matching site ?

In this PR, it will be supported to use custom regex. 

> so the example with 192.168.1.1 and 192.168.1.111 is no more valid ?

I've already fixed this within the PR to consider this as separate hosts.

BlackDex:
Great 

> > It hurts my eye's on some sites :) (Like github.com vs docs.github.com)
> 
> Use `^github.com` in the sitelist.
> 
> > First of all, thanks for this extension, and this PR.
> > Any idea when this will be merged?
> 
> No idea :)
> 
> > @BlackDex you can probably merge it yourself and add it to your browser manually, but I don't know if this PR is working right now.
> > also for your particular example GitHub as a beta dark mode, don't know about docs.github.com
> 
> It is working.

Great! That option i didn't know. Must have missed that somewhere in the docs :).
Thx.

aminomancer:
I have it set to enabled by default, and I'm using this ^github.com value to disable it on https://github.com, but enable it on docs.github.com. But then how do I add exceptions? I don't think you really need elaborate syntax in the blacklist/whitelist. You just need to have BOTH a blacklist and a whitelist. One that overrides the other. In other words, I could put ^github.com in my blacklist, but then put github.com/marketplace in my whitelist. Without that it seems like the only way you're ever going to achieve the same functionality is by implementing regular expression syntax.

Which would be fine, although this extension is consumed by a pretty broad audience, many of whom will probably not be familiar with regexp. Seems like a blacklist+whitelist would be easier to implement with the existing popup functionality too. Right now the toggle only pays attention to the domain. But you could add a second toggle underneath it that contains the full URL. The main toggle could blacklist/whitelist the domain, while the secondary toggle could make a narrow exception to that blacklist/whitelist rule.

This seems like the most performant way to go about it, because then instead of adding a bunch of checks or characters to your URL parser, you can just test the URL against the primary list, and then if it returns true, pass it to a test against the secondary list. Idk, haven't looked at your source code, but I can't really think of any reason that would ever be difficult to write or expensive to run. If you think that's too confusing for some users, then lock it behind a setting in the experimental layout like you did with the storage.sync toggle

Gusted:
> Which would be fine, although this extension is consumed by a pretty broad audience, many of whom will probably not be familiar with regexp.

Regexp is optionally the syntax will consistent of `^` `$` `!` which is indeed similar to regexp but used in a lossy way.

> Seems like a blacklist+whitelist would be easier to implement with the existing popup functionality too.

Agreed, #28

> This seems like the most performant way to go about it, because then instead of adding a bunch of checks or characters to your URL parser, you can just test the URL against the primary list, and then if it returns true, pass it to a test against the secondary list.

Uhm, this PR is primarily focused to a new matching system that is better than the current one, because the current problem is that the matching with all checks is very strict and not matching user's expected behaviour which will be solved by this PR. This problem will not remotely nor at all solve with having separate list.

> Idk, haven't looked at your source code, but I can't really think of any reason that would ever be difficult to write or expensive to run. If you think that's too confusing for some users, then lock it behind a setting in the experimental layout like you did with the storage.sync toggle

Oh it isn't expensive to write code, but it is to cover every edge case. I've spent a great time on this PR trying to cover everyone's expected behaviour and still 2 bugs where mentioned on this PR of edge cases that weren't right yet this kind of behaviour should be consistent with all users, power users can use the regexp syntax if they feel to. 

> edit: Oh also I really don't like the "google.com, !google.com/whatever" syntax. It seems fine in whitelist mode but in blacklist mode it would become confusing. Unless you changed the way "disabled on these websites" is described in the UI I guess. As it is, that would make it sound like google.com is disabled, and google.com/whatever is merely "not disabled," rather than explicitly "enabled." That's the reason I prefer the idea of a blacklist/whitelist.

It will be indeed confusing when we may or may not choose for a blacklist/whitelist setup but currently with 1 list it shouldn't be confusing at all an don't need to be currently a discussion point.

> Maybe you could then simplify it internally by maintaining a single list but giving members an extra boolean property to denote whether they're whitelist or blacklist rules.

That will performance wise a bad idea, if you e.g. want to just check if the site is in the whitelist you will need to loop over the list and then additionally check if the `property` is whitelist or blacklist and over a longer list e.g. the current dark-config site it could be expensive on performance and I honestly see no benefits for maintaining internally 2 list then 1. It would also avoid confusion when writing code.


Honestly the situations you describe fits #28 with exception of the syntax of the URL Matching. Your situation describes having a blacklist/whitelist. That situation does not fit this PR which will just improve how the matching of a certain URL with another URL should behave, please discuss this over at #28.

Regards,
Gusted

aminomancer:
Thanks, I hadn't seen that. While I'm here though, maybe I'm misunderstanding in the first place... is there any detailed documentation on the use of ^, $, ! in the site list for "enable by default" mode? ^ at the beginning of a string works in the sense of disabling subdomain URLs but the others seem to do nothing, at least in "enable by default" mode.

Gusted:
> is there any detailed documentation on the use of ^, $, ! in the site list for "enable by default" mode?

`^` and `$` will server no to little purpose in the new URL Matching. `!` simply indicates this 'entry' should be negated in the result.

Yash-Singh1:
Based on what @aminomancer said, there should be some detailed documentation, from which the site list is more understandable.

Gusted:
> Based on what @aminomancer said, there should be some detailed documentation, from which the site list is more understandable.

Originally we had #373 however it's not matching this PR due to treating subdomains differently off root domains. Their could be a wiki entry, but not for this moment. It's still unsure how it should behave, especially with subdomains/root domains.

Regards,
Gusted

aminomancer:
Okay, this is kind of hard to explain but it's pretty apparent that ! does not work for all use cases. Take the example I gave previously. I want to disable dark reader on github.com in general. But I want to enable it on github.com/marketplace. And I want to use "enable by default" mode. So you'd think one way to achieve that would be to set "enable by default" and then go to the site list and input github.com on one line, then input !github.com/marketplace on another line. But this doesn't work.

I read somewhere else that you could write "github.com, !github.com/marketplace" all on a single line, but that syntax doesn't seem to work at all. I think they might have misunderstood and were interpreting a list of comma-separated values for the site list overall as a single entry.

In any case, it seems like there's no way to make this work right now. Maybe if I used "disable by default" mode and specifically set github.com/marketplace in the sitelist that would be fine. But I don't want to have to manually enable dark reader on every new domain I visit. It seems like there should be a way to define a rule in the site list that supersedes a more general (lower-priority) rule in the list.

Gusted:
> I read somewhere else that you could write "github.com, !github.com/marketplace" all on a single line,

That's incorrect syntax, every entry has a new line.

>  But this doesn't work.

I will have a look why this doesn't work with this PR, but it should work.

Gusted:
image
```
github.com
!github.com/marketplace
```
It works, did you make sure you've side loaded this PR?

Regards,
Gusted

mirsella:
Sent from ProtonMail mobile

-------- Original Message --------

aminomancer:
Oh didn't think your PR had affected how rules supersede each other. I just built it now and it works correctly. With that I don't think I can think of any common scenarios where this syntax wouldn't be sufficient. Regexp or globs might be cool but all I really see myself using is this capability to set a general rule for a domain e.g. google.com, and negate it with a more specific rule e.g. google.com/search. Seems great to me

Gusted:
[binaries.zip](https://github.com/darkreader/darkreader/files/5843747/binaries.zip) https://github.com/darkreader/darkreader/pull/2517/commits/a8f0633cdbe6d159db818f77e5212c96ed900b7b

Yash-Singh1:
> > I read somewhere else that you could write "github.com, !github.com/marketplace" all on a single line,
> 
> That's incorrect syntax, every entry has a new line.
> 
> >

This is mentioned inside #373 as:

> ### URL lists
> URL lists are used in user's Site List config and fixes configurations (each record can have multiple URLs). `!` should reverse pattern result.
> 
> * `google.com, !*.google.com` should match google.com except it's subdomains.
> * `google.com, !mail.google.com` should match everything that matches `google.com` except everything that matches `mail.google.com`
> * `google.com, !mail.google.com, mail.google.com/compose` should not match everything that matches `mail.google.com` except everything that matches `mail.google.com/compose`

aminomancer:
@Yash-Singh1 yeah that's what i was thinking of

Gusted:
> This is mentioned inside #373 as:

That isn't at the moment a feature and probably won't be.
We have a nice SiteList UI in the V5 Preview, not sure why we should need/should have multiple records. The info at #373 is a bit outdated/not corresponding to this PR.

gianpaj:
What's the state of this PR? waiting for review? 
I see a merge conflict.

If this allows for #28 - it would be great!

Gusted:
> What's the state of this PR? waiting for review?

It's more or less on the backlog for now, as this will be a big and a breaking changes due to the way that the matching will be changing. Well this PR allows to have interesting things that will help and improve certain aspects of dark reader's performance, we're a waiting this one out until we really need this PR for our current/future improvements, mainly done by @bershanskiy.  

> If this allows for #28 - it would be great!

It will actually help to allow #28 as this PR will have logic and mechanism to have a deny and allow list(negated results) but doesn't implement the UI for it.

src/utils/matching.ts
```diff
@@ -0,0 +1,119 @@
...
+    // Make sure it's cached!
+    regexpCache.set(pattern, regObject);
+    return regObject;
+}
+
+/**
+ * Sanitazed the website so it ensures different input
+ * with the same meaning will have the same output here.
+ */
+function sanitazeInput(input: string) {
```
bershanskiy, line 48: I'm not sure we should perform this kind of magic transformation. It seems, every time someone does this kind of cleverness something else gets broken in mysterious ways. I remember Google got a lot of bad publicity from _cosmetically_ hiding what they called "trivial subdomains".

Gusted: I'm not sure how adding any of these characters could be "clever" I'm removing `$` and `^` as they don't have much effect anymore. `www.` is to account for old web pages which still uses `www.`, And it removes any query(We just don't want to mess around with it) and we remove the scheme.

src/utils/matching.ts
```diff
@@ -0,0 +1,119 @@
...
+    return (input.replace(/^\^/, '')
+        .replace(/\$$/, '')
+        .replace(/\?.*$/, '')
+        .replace(/^.*?\/{2,3}/, '')
```
bershanskiy, line 52: Are there any URLs with `///` (three slashes) in them?

src/utils/matching.ts
bershanskiy: I'm mostly worried about `www.` subdomain.

src/utils/matching.ts
Gusted: It's a copy paste of some regex somewhere which we've already used. Which most likely was then created by @alexanderby :) So probably no, but maybe yes.

src/utils/matching.ts
Gusted: I can see why someone would do it, but in such rare rare cases they can go for the regex solution.

Gusted: Let's remove the `www.` then.

bershanskiy:
@Gusted Would you mind if I took over this PR? Mostly to resolve conflicts and add a few tests.

Gusted:
@bershanskiy Sure thing, just be sure of modifying existing tests as most of them are hand-crafted and are edge-cases that we should cover.

src/utils/matching.ts
bershanskiy: I found out why it allows `///`: apparently, local files have URL like `file:///C:/stuff`. May be, we should add a comment or a unit test because I feel like this case could be lost during refactoring.

Gusted: Yeah, that's good to have a comment about that.

ackvf:
Will this support custom regex including look arounds?

Slowftw:
Will support lookahead?

neon-dev:
Is this PR currently incomplete or could it be merged?

joeldeteves:
It's probably too late for this but might I suggest taking a look at the way [Bitwarden](https://github.com/bitwarden) handles URL matching.



It seems like an elegant way to tackle this problem.

aranapatona:
> It's probably too late for this but might I suggest taking a look at the way [Bitwarden](https://github.com/bitwarden) handles URL matching.
> 
> 
> 
> It seems like an elegant way to tackle this problem.

lol I cant believe I came here to share exactly the same app with the same image with the same proposition.

Swivelgames:
@bershanskiy Any traction on this? This seems to be quite the undertaking, but certainly appears to be affecting a large number of people (while generating a sizeable number of issues).

Thanks for all of y'alls hard work! It's been awesome to see DarkReader explode in popularity and grow the way that it has!

mmseng:
> Maybe just a "whitelist" that over-rides the current blacklist? That way, we could whitelist specific subdomains and regexes.

That wouldn't address other issues like the need to blacklist subdomains, unless they also implemented the inverse solution. Even then, I don't think that would be flexible enough to allow for custom configurations per subdomain, only inclusion or exclusion from "higher" rules, although it would depend on the implementation I guess.

KaKi87:
In the meantime, I created a [Dark Reader dynamic blacklist](https://git.kaki87.net/KaKi87/userscripts/src/branch/master/darkReaderDynamicBlacklist/README.md) userscript that, among other things, manages this use case, using the [`darkreader-lock` meta tag](https://github.com/darkreader/darkreader/pull/9181).

For example, it will disable filtering on `github.com` & `gist.github.com` while keeping it enabled on `education.github.com`.

KaKi87:
> is it supposed to work automatically

Yes.

> I've been trying with alternativeto.net, and it doesn't detect the available-but-not-default dark theme

Are you using Violentmonkey, like recommended on my userscript homepage ?
By the way, I think we should discuss this at my repository, or on the Discord community linked on that repository, or at Violentmonkey's, but not here.

Jackenmen:
It seems universal enough for anyone who saw Kaki's comment that I figured I may as well mention it here. In order for it to work, you need to set "Theme generation mode" to Dynamic as it's the only one that supports `darkreader-lock` meta tag:
image
Additionally, you may want to make sure that your site list is configured such that it can actually make pages dark. One such configuration is to set it to "Not invert listed" with an empty list:
image

Francewhoa:
This is to confirm this bug with Dark Reader V4.9.62. Using [Firefox](http://mozilla.org/firefox) (FF) V102.7 or [Ungoogle Chromium](https://ungoogled-software.github.io/ungoogled-chromium-binaries/) (Chrome is powered by Chromium) V109.0.

For users facing this challenge, I published [this easy, quick, and temporary workaround](https://github.com/darkreader/darkreader/issues/1896#workaround).

For developers interested to contribute to resolving this bug, you might be interested in this:
* [Steps to reproduce](https://github.com/darkreader/darkreader/issues/1896#steps)
* Pull request at https://github.com/darkreader/darkreader/pull/2517

KaKi87:
Will the subdomains rule for the global dark list still be enforced once this PR is merged ?

sakgoyal:
is there an update on this?

iTrooz:
Is there anything we can do to help this get merged ?

If this PR is problematic to merge and is expected to stay open for a long time, I'm going to try to upload a fork of Dark Reader to https://addons.mozilla.org based on this branch (Don't worry, I'll change the name). That way, users can still get around the issue

alexanderby:
Hi @iTrooz, there are now 2 approaches: without regular expressions (like `www.google.*.*`) and with regular expressions (like `/www\.google\..*?\//`). Please leave a comment if anything is not covered here https://github.com/darkreader/darkreader/issues/373

KaKi87:
If `google.*` still matches `maps.google.com` then it's still a problem.

alexanderby:
Try `www.google.*` or `^google.*`

iTrooz:
> Hi @iTrooz, there are now 2 approaches: without regular expressions (like www.google.*.*) and with regular expressions (like /www\.google\..*?\//). Please leave a comment if anything is not covered here https://github.com/darkreader/darkreader/issues/373

Hey @alexanderby ! I'm sorry, but I don't think I understood your message. Could you reformulate ?

alexanderby:
This week a support for Regular Expressions was added. When a pattern starts with `/` and ends with `/` it means it is a RexExp. It would be nice to track feedback in a single place. This is a source issue https://github.com/darkreader/darkreader/issues/373

iTrooz:
I'm sorry, but I'm not interested in testing that feature if it doesn't help this PR get merged. I am interested in working on the code in this PR in order to it to get merged. My motivation for this is that I'd like to control subdomain dark themes independently from the main DarkReader UI (I'm affected by #1896)

iTrooz:
So, is there any way in which I can currently help, to get this PR merged ? (Or maybe you would accept a smaller PR to specifically handle subdomains independently ? But I'm not sure why this would be better, since nobody complained that this PR doesn't work ?)

alexanderby:
This PR can't be merged unfortunately, but some features can now be reimplemented in several smaller PRs. Particularly the root domain issue (e.g. GitHub.com and docs.github.com) can be fixed in 1 line and seems like a low risk change that can be done quickly.

The Site List separation into enabled/disabled has been done, and it is unclear if the negative patterns will be still necessary. Anyway it should be also in a separate PR.

I'm sorry that it takes so long, unfortunately my life situation didn't let me look into the problem 3 years ago.

alexanderby:
[This commit](https://github.com/darkreader/darkreader/commit/5546bd52e82f00dd1c436867f340170faaea868f) allows `www` as a subdomain, but not anything else. I know that domains with `www` and without are technically different, but this will allow us to easily handle site lists, fixes configs etc. @iTrooz Let me know if it solves your problem.

iTrooz:
This works perfectly, thank you !
As a site note, you may want to make the `www` subdomain and the main domain equivalent. If I go to https://example.com, disable the dark theme, and go to https://www.example.com, the "toggle theme" button doesn't do anything. You may want to make it toggle the dark theme for the main domain

> I'm sorry that it takes so long, unfortunately my life situation didn't let me look into the problem 3 years ago

Don't worry, it's perfectly okay ! You are doing this in your free time, and at no point did I want to imply that the work you are putting into this was bad. Thank you and the maintainers in general for making such a good extension, and devoting time for it. It's really useful for me on a daily basis, and I'm thankful for Dark Reader to exist

KaKi87:
In which version is this/will this be published to the stores ?
Thanks

alexanderby:
Thanks, I see this `www` toggling issue and will try to make a fix.

If no other issues are found, it can go live next week (hopefully Mozilla will review the previous release soon).

I'm attaching a build if you want to try:
[darkreader-firefox-4.9.70.1.zip](https://github.com/darkreader/darkreader/files/13464544/darkreader-firefox-4.9.70.1.zip)

PR #2518, opened 2020-05-15
Fix for VRT Nieuws (Belgian news site)

Tijs-B:
Add custom CSS for VRT nieuws (Belgian VRT news site). The header and some hover effects were bright white.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3599,7 +3599,7 @@ INVERT
 
 CSS
 article:hover::before {
-    background-color: #000000 !important;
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -3592,6 +3592,21 @@ body {
...
 ================================
 
+vrt.be/vrtnws
+
+INVERT
+.vrt-site-header__home>svg
+
+CSS
+article:hover::before {
+    background-color: #000000 !important;
```
Gusted, line 3602: Use `${Color to be darken}`

In this situation `${white}`

PR #2519, opened 2020-05-15
Cache AnalyzeImage

Gusted:
Well maybe you know AnalyzeImage is quite heavy program and caching the results would improve loading speeds and sites that request an image like 100x times *Looking at you google*
As it's 4 simple booleans it hopefully won't be heavy on memory usage.

We cache the .src as AnalyzeImage is being used at GetImageDetails that always pass an data: that reperesent the image data so if it's dynamic loaded it will pass another data:

*Linkedin improves speed on second reload and a little bit on first*

### DIFF
--- a/src/inject/dynamic-theme/image.ts
+++ b/src/inject/dynamic-theme/image.ts
@@ -4,6 +4,7 @@ import {getURLHost} from '../../utils/url';
 import {loadAsDataURL} from '../../utils/network';
 import {FilterConfig} from '../../definitions';
 
+
 export interface ImageDetails {
     src: string;
     dataURL: string;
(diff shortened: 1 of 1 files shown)


src/inject/dynamic-theme/image.ts
```diff
@@ -49,7 +50,12 @@ async function urlToImage(url: string) {
     });
 }
 
+const ImageCache = new Map();
```
alexanderby, line 53: These results are already cached, see [here](https://github.com/darkreader/darkreader/blob/master/src/inject/dynamic-theme/modify-css.ts#L227).

So, I think this PR has no effect.

Anyway some comments:
- Start variables from lowercase.
- Remember to clean the cache when necessary.

PR #2520, opened 2020-05-15
Use length Caching

Gusted:
https://jsben.ch/kidOp :)


alexanderby:
It turned out that `for` loop with length caching is 30% faster for `NodeList` (the result of `querySelectorAll`) and 3 times faster than `NodeList.forEach` in Chrome. So your changes make sense.

Gusted:
Big list = big memory;
Each time calculating and retrieving memory from big memory to get length = slow;
Optimization :)

PR #2522, opened 2020-05-15
fix for geogebra.org

xxyzz:
https://www.geogebra.org/graphing
Before:
Screen Shot 2020-05-16 at 07 42 49
Screen Shot 2020-05-16 at 07 46 57

After:
Screen Shot 2020-05-16 at 07 39 05
Screen Shot 2020-05-16 at 07 47 40

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1260,6 +1260,19 @@ img.listview__image, img.galleryview__image {
 
 ================================
 
+geogebra.org
+
+INVERT
+.icon-m
+.elemText canvas
+.gwt-Image
+.gwt-StackPanelItem img
+.GeoGebraMenuImage.menuImg
+.menuImg
+.buttonContent.stylebarButton
+
+================================
+
 get.todoist.help
 
 CSS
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -1260,6 +1260,19 @@ img.listview__image, img.galleryview__image {
...
+
+INVERT
+.icon-m
+.elemText canvas
+.gwt-Image
+.gwt-StackPanelItem img
+.GeoGebraMenuImage.menuImg
+.menuImg
+.buttonContent.stylebarButton
+
```
Gusted, line 1273: Let's add `canvas[class^="cursor"]` So the canvas will also be targetted

xxyzz: I tried to invert the main canvas but the lines were blinking when I moved around. The app supports change lines and background color in settings, that works well.

Gusted: I saw that when trying to just get the class `cursor_hit`
But when blinking it just changes the class after the _ so cursor_drag cursor_pen
So with CSS3 selectors we can target `startWith` with ^= so
canvas[class^="cursor"] means that it only target canvas with a class that start with cursor and then you don't have the blinking anymore :)

xxyzz: I used the selector `.EuclidianPanel canvas` earlier, I think the results are the same. After adding a few functions, you will notice the blinking at some part of the lines you added when you move the mouse or drag the canvas.

Gusted: .EuclidianPanel > canvas
Add that and then it will be merged

PR #2531, opened 2020-05-16
fix for facebook.com

Myshor:
Live streams with enabled native fb dark mode fix.

Before:
image

After:
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1084,7 +1084,7 @@ div[aria-label="Change Position"] .ha302278,
 ._8bb_ video, ._8bb_ ._w80, ._8bb_ ._7umt ._47yj {
     filter: none !important;
 }
-div[role="button"] > i.hu5pjgll:not(.sp_Jc8OKpJq5NW):not(.sp_T89CCTT7d9Z),
+div[role="button"] > i.hu5pjgll:not(.sp_Jc8OKpJq5NW, .sp_T89CCTT7d9Z),
 a[role="link"] > i.hu5pjgll, ._3w97, .q2y6ezfg, .mkbloq8g {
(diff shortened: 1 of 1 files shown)


Myshor:
Moved maps darkening into CSS - fixes white maps in native dark fb mode after last commit. 😄

Myshor:
Inline for svg icons fix:
image

Myshor:
Exclude some buttons for color fix to have it back white instead of gray.
Added force background image for maximize button instead of using blob.
image

mrturkmen06:
I was planning to open an issue regarding to facebook.com, however I saw these improvements and would like to add one more issue. (I could not see before after photo of this particular issue in your comments that's why I called it as issue)
Likes pop-up window seems does not work very well. The names are not visible in dark mode. @Myshor. I thought that you can take this issue since you already done with other issues of facebook.com site. 

Btw, thank you very much for your contribution ! 

Screenshot 2020-05-18 at 20 51 22

Myshor:
@mrturkmen06 You mean new facebook design? Cause I can not reproduce this on likes.
In new design I see:
image

On old design:
image

mrturkmen06:
No I do not want to mean new design, the name and surname was not properly displayed when I click likes section as you can see from the SS that I shared. 
 It could be depended on browser @Myshor, I am using 76.0.1 (64-bit) Firefox. I guess, you are using chrome.

Myshor:
If you can please create new issue with info that Firefox is used and with screen of your darkreader settings. I think I will merge this Pull Request today and I can forget to take a look in Firefox with all this fixes included. ;)

mrturkmen06:
> If you can please create new issue with info that Firefox is used and with screen of your darkreader settings. I think I will merge this Pull Request today and I can forget to take a look in Firefox with all this fixes included. ;)
Got it 👍🏻 
Thanks

PR #2533, opened 2020-05-16
Add MOBILE CSS 

Gusted:
Some sites are getting broken by the CSS Fixes on mobile view so fix them with an seperate section to not use complicated @media 

Example: https://tweakers.net (They are looking for a dark mode implementation so far I know ;) )

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3443,7 +3443,7 @@ CSS
 tweakers.net
 
 CSS
-.ankeiler>a.commentCount {
+.ankeiler > a.commentCount {
     background-image: url(../../../g/icons/commentcounter_small_bg.svg) !important;
     text-decoration: none !important;
 }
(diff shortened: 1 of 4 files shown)


alexanderby:
Are you sure there's a need in this change?

I understand that some websites can have different versions of their websites based on user agents. Some can have many stylesheets for different screen sizes. But we mostly deal with colors and images and I think these mobile and desktop fixes can be simply concatenated.

It seems like a rare use case and there's already a workaround to use media queries.

Gusted:
> It seems like a rare use case and there's already a workaround to use media queries.

Right, but existing CSS would interfere and rewriting copy paste and make it usable for mobile would take more space in the config

Gusted:
It's easier to hold it seperate and clean than have 1 big CSS section with complicated @media queries that can mess up other DPI displays etc.

Gusted:
> we mostly deal with colors and images and I think these mobile and desktop fixes can be simply concatenated.

Images, we set hardcoded image values and hardcoded offsets to make them work e.g. packed icons. If a site(all) use another path for mobile images with better scaling we will override and force the desktop variant as they are hardcoded in some CSS of site's

Yash-Singh1:
> It seems like a rare use case and there's already a workaround to use media queries.

There is a workaround but it is very complicated: https://css-tricks.com/snippets/css/media-queries-for-standard-devices/
I have seen some sites that have separate elements for mobile.

Gusted:
Yeah just closing this one as I don't consider it having it anymore.

PR #2534, opened 2020-05-17
Better Build script

Gusted:
Faster building. 
Production: false 
for caching 20 seconds building instead fo 2 minutes


tasks/build.js
```diff
@@ -0,0 +1,28 @@
+const bundleCSS = require('./bundle-css');
```
alexanderby, line 1: It looks very similar to `release.js` and `debug.js` files. So I suggest you reuse the code (maybe even merge everything into a single file, use `git mv` to keep the file history) and use command line arguments. Here's an [example of running a script with arguments](https://github.com/alexanderby/malevic/blob/master/package.json#L18) and [here how they are read](https://github.com/alexanderby/malevic/blob/master/tasks/build.js#L177) in Node.js script. So, later you will be able to create multiple tasks like `release`, `debug` and `debug-watch` that will run `node tasks/build.js --release` or `node tasks/build.js --debug --watch`.

Gusted: Will have a look and merge everything into a single file if I figure out to merge to single file with preserving history as git mv isn't the best option I thought.

alexanderby: You can move a single file and remove another one.

Gusted: You can't move into other files AFAIK

alexanderby: Yes, you have to remove the file you created, and then move the source file into new destination with `git mv`.

Gusted: Will try

Gusted:
Github having internal problems

package.json
```diff
@@ -3,13 +3,13 @@
   "version": "4.9.8",
   "description": "Dark theme for every website",
   "scripts": {
-    "api": "node -e \"require('./tasks/bundle-api')()\"",
+    "api": "node tasks/build.js --api",
     "benchmark-server": "node tests/benchmark-server/index.js",
-    "build": "node tasks/release.js",
+    "build": "node tasks/build.js --build",
```
alexanderby, line 8: The `build` script is confusing. The `api` script is also build, `release` is build, and `debug` is also build. Since `build` is used for debugging, I suggest you rename it into `debug`, and rename `debug` into `debug-watch`.

tasks/build.js
```diff
@@ -0,0 +1,89 @@
+//@ts-check
```
alexanderby, line 1: Put a ` ` between `//` and comment.

tasks/build.js
```diff
@@ -0,0 +1,89 @@
...
+const copy = require('./copy');
+const reload = require('./reload');
+const {runTasks} = require('./task');
+const {log} = require('./utils');
+const zip = require('./zip');
+
+/**
+ * @param {boolean} prod Pass option production true or false.
+ */
+async function release(prod) {
```
alexanderby, line 18: Release means building for production, there cannot be `prod` argument.

tasks/build.js
```diff
@@ -0,0 +1,89 @@
...
+            zip,
+        ], {production: prod});
+        log.ok('MISSION PASSED! RESPECT +');
+    } catch (err) {
+        log.error(`MISSION FAILED!`);
+        process.exit(13);
+    }
+}
+
+const debugTasks = [
```
alexanderby, line 38: Rename this array and reuse it in `release` and `debug` functions.

tasks/build.js
```diff
@@ -0,0 +1,89 @@
...
+    clean,
+    bundleJS,
+    bundleCSS,
+    bundleHTML,
+    bundleLocales,
+    copy,
+];
+
+async function debug() {
```
alexanderby, line 47: Add a `{watch}` flag.

tasks/build.js
```diff
@@ -0,0 +1,89 @@
...
+    log.ok('DEBUG');
+    try {
+        await runTasks(debugTasks, {production: false});
+        debugTasks.forEach((task) => task.watch());
```
alexanderby, line 51: Use `watch` flag to run watchers or not.

tasks/build.js
```diff
@@ -0,0 +1,89 @@
...
+    if (args.includes('--debug')) {
+        await debug();
+    }
+    if (args.includes('--api')) {
+        await api();
+    }
+
+}
+
+run();
```
alexanderby, line 89: No empty line at the end of file

tasks/bundle-api.js
```diff
@@ -21,11 +24,11 @@ async function bundleAPI() {
                 tsconfigOverride: {
                     compilerOptions: {
                         removeComments: true,
-                        target: 'es5',
+                        target: 'ES6',
```
alexanderby, line 27: This change should not be the part of this PR. We should analyze the real cases, if we should target old browsers and provide polyfills for them, or maybe we should target `esnext`, or maybe there should be multiple targets. Let's leave it for another issue.

tasks/bundle-api.js
```diff
@@ -42,4 +45,7 @@ async function bundleAPI() {
     });
 }
 
-module.exports = bundleAPI;
+module.exports = createTask(
+    'bundle-api',
+    bundleAPI,
+)
```
alexanderby, line 51: Missing ';'

tasks/bundle-api.js
Gusted: Okay, open vscode have master branch or something and use the API script :) Poof a error

alexanderby: I've fixed this error following the suggestion from the message https://github.com/darkreader/darkreader/pull/2571/files

tasks/build.js
```diff
@@ -0,0 +1,79 @@
...
+const clean = require('./clean');
+const copy = require('./copy');
+const reload = require('./reload');
+const {runTasks} = require('./task');
+const {log} = require('./utils');
+
+async function release() {
+    log.ok('RELEASE');
+    try {
+        await runTasks(standardTask, {production: true});
```
alexanderby, line 16: Some release tasks are missing, I suggest you to add them like `[...standardTasks, codeStyle, zip]`

package.json
```diff
@@ -3,13 +3,13 @@
   "version": "4.9.8",
   "description": "Dark theme for every website",
   "scripts": {
-    "api": "node -e \"require('./tasks/bundle-api')()\"",
+    "api": "node tasks/build.js --api",
     "benchmark-server": "node tests/benchmark-server/index.js",
-    "build": "node tasks/release.js",
+    "debug": "node tasks/build.js --build",
```
alexanderby, line 8: - Sort the tasks alphabetically.
- `build.js -- build` is strange. Use `build.js --debug`.

package.json
```diff
@@ -3,13 +3,13 @@
...
     "code-style": "eslint --fix \"src/**/*.ts\" \"src/**/*.tsx\" \"tasks/**/*.js\" \"tests/**/*.ts\" \"tests/**/*.js\" \"index.d.ts\"",
-    "debug": "node tasks/debug.js",
+    "debug-watch": "node tasks/build.js --build",
```
alexanderby, line 10: You can use 2 flags here, like `build.js --debug --watch`

tasks/build.js
```diff
@@ -0,0 +1,78 @@
...
+    }
+}
+
+async function run() {
+    const args = process.argv.slice(2);
+
+    if (args.includes('--release')) {
+        await release();
+    }
+    if (args.includes('--build')) {
```
alexanderby, line 69: Includes what?

Gusted: Debug, my comrade!

PR #2535, opened 2020-05-17
Fix restoring node position

### LINKED ISSUE #2065: Some sites don't work until darkreader is toggled off and back on
This issue is to replace [a closed issue](https://github.com/darkreader/darkreader/issues/1914) and provide more information.

These are sites that darkreader seems to work on, but if darkreader is already enabled when you navigate to the page, some or all elements are no longer in dark mode. These are sites where simply toggling the darkreader extension it suddenly works as expected.

Please add any website URLs along with screenshots, and any additional info you have that fits this description. I will be adding mine below as they arise.

Gusted:
- Fix's to move to good location

More Info & Resolves #2065

TODO: I need to know if this mess up removed styles?
Open up a benchmark server.
http://localhost:1357/?links=2&styles=10&rules=4
First replace the <link> to a new location.
Than delete <link>
Seems like it work's :)

### DIFF
--- a/src/inject/utils/dom.ts
+++ b/src/inject/utils/dom.ts
@@ -75,7 +75,7 @@ export function watchForNodePosition<T extends Node>(
     const MAX_ATTEMPTS_COUNT = 10;
     const ATTEMPTS_INTERVAL = getDuration({seconds: 10});
     const prevSibling = node.previousSibling;
-    const parent = node.parentNode;
+    let parent = node.parentNode;
     if (!parent) {
         // BUG: fails for shadow root.
         logWarn('Unable to watch for node position: parent element not found', node, prevSibling);
@@ -99,12 +99,11 @@ export function watchForNodePosition<T extends Node>(
         }
         if (prevSibling && prevSibling.parentNode === null) {
             stop();
+            return;
         }
         if (prevSibling.parentNode !== parent) {
-            logWarn(`Style was moved to another parent, changing node's position`, node, prevSibling, parent);
-            prevSibling.parentNode.insertBefore(node, prevSibling.nextSibling);
-            onRestore && onRestore();
-            return;
+            logWarn(`Style was moved to another parent`, node, prevSibling, parent);
+            parent = prevSibling.parentNode;
         }
         logWarn('Node was removed, restoring it\'s position', node, prevSibling, parent);
         parent.insertBefore(node, prevSibling ? prevSibling.nextSibling : parent.firstChild);
--- a/src/inject/utils/dom.ts
+++ b/src/inject/utils/dom.ts
@@ -97,10 +97,10 @@ export function watchForNodePosition<T extends Node>(
             start = now;
             attempts = 1;
         }
-        if (prevSibling && prevSibling.parentNode !== parent) {
+        if (prevSibling && (prevSibling.parentNode !== parent && prevSibling.parentNode !== null)) {
             // Moving to another location with not same parents anymore, but still exist.
             // https://github.com/darkreader/darkreader/issues/2065
-            logWarn('Style was moved to another location, restoring node\'s position', node, prevSibling, parent);
+            logWarn('Style was moved to another location with not the same parentNode, moving node\'s position', node, prevSibling, parent);
             prevSibling.parentNode.insertBefore(node, prevSibling.nextSibling);
             onRestore && onRestore();
             return;
--- a/src/inject/utils/dom.ts
+++ b/src/inject/utils/dom.ts
@@ -97,10 +97,10 @@ export function watchForNodePosition<T extends Node>(
             start = now;
             attempts = 1;
         }
-        if (prevSibling.parentNode === null) {
+        if (prevSibling && prevSibling.parentNode === null) {
             stop();
         }
-        if (prevSibling && prevSibling.parentNode !== parent) {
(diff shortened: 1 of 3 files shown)


src/inject/utils/dom.ts
```diff
@@ -97,9 +97,12 @@ export function watchForNodePosition<T extends Node>(
             start = now;
             attempts = 1;
         }
-        if (prevSibling && prevSibling.parentNode !== parent) {
-            logWarn('Unable to restore node position: sibling was removed', node, prevSibling, parent);
-            stop();
+        if (prevSibling && (prevSibling.parentNode !== parent && prevSibling.parentNode !== null)) {
+            // Moving to another location with not same parents anymore, but still exist.
```
alexanderby, line 101: These comments and reverence to issue can be removed.

src/inject/utils/dom.ts
```diff
@@ -97,9 +97,12 @@ export function watchForNodePosition<T extends Node>(
...
+            // https://github.com/darkreader/darkreader/issues/2065
+            logWarn('Style was moved to another location with not the same parentNode, moving node\'s position', node, prevSibling, parent);
```
alexanderby, line 103: Please, make this message shorter and grammatically correct.

alexanderby:
I'm not sure it resolves the whole issue, because it has no effect on some websites.

Gusted:
To be honest this just fixed some sites but not all can be related to the main issue that its change position to another parent all sites can have another problem were we need another fix for it. This is just a fix for the main site reported with that issue with it.

src/inject/utils/dom.ts
```diff
@@ -97,9 +97,10 @@ export function watchForNodePosition<T extends Node>(
             start = now;
             attempts = 1;
         }
-        if (prevSibling && prevSibling.parentNode !== parent) {
-            logWarn('Unable to restore node position: sibling was removed', node, prevSibling, parent);
-            stop();
+        if (prevSibling && (prevSibling.parentNode !== parent && prevSibling.parentNode !== null)) {
+            logWarn('Style was moved to another parent, moving node\'s position', node, prevSibling, parent);
+            prevSibling.parentNode.insertBefore(node, prevSibling.nextSibling);
```
alexanderby, line 102: `parent` variable should be updated with the new sibling's parent.

src/inject/utils/dom.ts
```diff
@@ -97,9 +97,10 @@ export function watchForNodePosition<T extends Node>(
             start = now;
             attempts = 1;
         }
-        if (prevSibling && prevSibling.parentNode !== parent) {
-            logWarn('Unable to restore node position: sibling was removed', node, prevSibling, parent);
-            stop();
+        if (prevSibling && (prevSibling.parentNode !== parent && prevSibling.parentNode !== null)) {
+            logWarn('Style was moved to another parent, moving node\'s position', node, prevSibling, parent);
```
alexanderby, line 101: I'm not sure, but I think you cannot *move the position*, you can *change the position* or *restore the position*. Also, since `\'` looks a little bit weird, you can use a template string quotes and `'` without the escaping.

src/inject/utils/dom.ts
```diff
@@ -97,9 +97,10 @@ export function watchForNodePosition<T extends Node>(
...
+            prevSibling.parentNode.insertBefore(node, prevSibling.nextSibling);
+            onRestore && onRestore();
```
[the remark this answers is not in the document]
Gusted: `return;`

src/inject/utils/dom.ts
Gusted: parent is a const outside the trothle and should be updated everytime a node is being changed

alexanderby: So, use `let`. The goal of this function is to stick a node to some previous element. So, the `parent` should be changed only when it is changed.

src/inject/utils/dom.ts
```diff
@@ -97,8 +97,11 @@ export function watchForNodePosition<T extends Node>(
             start = now;
             attempts = 1;
         }
-        if (prevSibling && (prevSibling.parentNode !== parent && prevSibling.parentNode !== null)) {
-            logWarn('Style was moved to another parent, moving node\'s position', node, prevSibling, parent);
+        if (prevSibling.parentNode === null) {
```
alexanderby: `prevSibling` can be `null` (when element is the first one), so add `prevSibling && ...`

src/inject/utils/dom.ts
```diff
@@ -97,8 +97,11 @@ export function watchForNodePosition<T extends Node>(
...
+            stop();
```
alexanderby, line 101: I'm not sure, but maybe it should `return`.

Gusted: 🤦

Gusted:
Works for reddit.com Seems like I see no issues with baidu.com however it's not updating? I have not a clue.

amanhigh:
I am now using latest Dark Reader 4.9.9 on firefox (77.0.1). However the url i use most is still having same issue. Dark Reader doesn't work until toggling.

While investing.com home page is fixed. Issue remains on many urls after login. One such example is login page itself.
https://www.investing.com/members-admin/login

Gusted:
You can open a new issue here https://github.com/darkreader/darkreader/issues/new/choose

PR #2539, opened 2020-05-18
Add matsci.org

wyphan:
Dynamic (`dynamic-theme-fixes.config`) and Filter (`inversion-fixes.config`) modes

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2012,6 +2012,18 @@ INVERT
 
 ================================
 
+matsci.org
+
+INVERT
+.d-header #site-logo
+
+CSS
+.category-logo.aspect-image img {
+  background-color: white !important;
+}
+
+================================
+
 medium.com
 
 INVERT
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -1191,6 +1191,18 @@ img
 
 ================================
 
+matsci.org
+
+NO INVERT
+.d-header #site-logo
+
+CSS
+.category-logo.aspect-image img {
+  background-color: white !important;
+}
+
+================================
+
 medium.com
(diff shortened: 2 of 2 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -2012,6 +2012,18 @@ INVERT
...
 ================================
 
+matsci.org
+
+INVERT
+.d-header #site-logo
+
+CSS
+.category-logo.aspect-image img {
+  background-color: white !important;
```
Gusted, line 2022: Use ${black} and check if that looks still good

src/config/inversion-fixes.config
```diff
@@ -1191,6 +1191,18 @@ img
...
 ================================
 
+matsci.org
+
+NO INVERT
+.d-header #site-logo
+
+CSS
+.category-logo.aspect-image img {
+  background-color: white !important;
```
Gusted, line 1201: Same as above

wyphan:
Unfortunately it doesn't work with `${black}`. The main page contains 3 transparent png and 4 white-background png logos. What's the best way to improve readability here? I already tried `CSS background-color: transparent`, but it didn't work for those 4 logos...

Here's the original page with darkreader turned off:
original

Using `${black}` messes up Dynamic mode on Light scheme:
dynamic-light

... as well as Filter mode on Dark theme:
filter-dark

src/config/dynamic-theme-fixes.config
wyphan: Unfortunately not

src/config/inversion-fixes.config
wyphan: Same as above

Gusted:
No suchs thing to get it transparent with CSS :(

PR #2544, opened 2020-05-18
Fix some typos

Kijek3:


src/_locales/pl.config
```diff
@@ -96,10 +96,10 @@ Zastosuj ustawienia tylko dla bieżącej strony
 Lista stron
 
 @invert_listed_only
-Odwórć tylko list.
+Odwróć tylko listę
 
 @not_invert_listed
-Nie odwracaj list.
+Nie odwracaj listy
```
Gusted, line 102: Missing a . after the sentence?

src/_locales/pl.config
```diff
@@ -96,10 +96,10 @@ Zastosuj ustawienia tylko dla bieżącej strony
 Lista stron
 
 @invert_listed_only
-Odwórć tylko list.
+Odwróć tylko listę
```
Gusted, line 99: Missing a . after the sentence?

src/_locales/pl.config
Kijek3: That dot was previously added only for shorthand purposes. "list" in polish is a plural form, so it doesn't "fit" good. "listy" is full single form and don't need that dot - it was made only for shorthand, not to finish a sentence.

src/_locales/pl.config
Kijek3: Same here. That dot was previously added only for shorthand purposes. "list" in polish is a plural form, so it doesn't "fit" good. "listę" is full single form and don't need that dot - it was made only for shorthand, not to finish a sentence.

PR #2548, opened 2020-05-18
Use sRGB to determine `pixel lightness`

### LINKED ISSUE #704: Dynamic filtering: pixel "lightness" measurement is wrong
Hello,

The way you [measure the pixel "lightness"](https://github.com/darkreader/darkreader/blob/master/src/inject/dynamic-theme/image.ts#L108) is wrong: you are taking only the two max components of RGB to compute a "lightness" value, entirely leaving out another component which may make the image brighter.

The proper way is to compute the pixel luminance, taking into accounts the human eyes color perception:
`L = (0.299*R + 0.587*G + 0.114*B)` (fast)
`L = sqrt( 0.299*R^2 + 0.587*G^2 + 0.114*B^2 )` (slower)
See this helpful StackOverflow thread: https://stackoverflow.com/a/596243/8861729

Formula is for RGB values in range [0-1]. You have to divide by 255 before applying the formula.

Gusted:
We also can use HSP instead of HSL in rest of our code :eyes: < If that is possible in every section of the code that use HSL at the moment.

Resolves #704

### DIFF
--- a/src/inject/dynamic-theme/image.ts
+++ b/src/inject/dynamic-theme/image.ts
@@ -88,9 +88,8 @@ function analyzeImage(image: HTMLImageElement) {
             if (a < TRANSPARENT_ALPHA_THRESHOLD) {
                 transparentPixelsCount++;
             } else {
-                // Use HSP to determine the `pixel Lightness`
-                // http://alienryderflex.com/hsp.html & https://stackoverflow.com/a/24213274/13569411
-                p = Math.sqrt(0.299 * r^2 + 0.587 * g^2 + 0.114 * b^2)
+                // Use sRGB to determine the `pixel Lightness`
+                p = (0.2126 * r + 0.7152 * g + 0.0722 * b) / 255
                 if (p < DARK_LIGHTNESS_THRESHOLD) {
                     darkPixelsCount++;
                 }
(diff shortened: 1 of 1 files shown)


src/inject/dynamic-theme/image.ts
```diff
@@ -76,7 +76,7 @@ function analyzeImage(image: HTMLImageElement) {
 
     let i: number, x: number, y: number;
     let r: number, g: number, b: number, a: number;
-    let l: number, min: number, max: number;
+    let p: number;
```
alexanderby, line 79: So, this variable is now `l` (for **L**uminance).

src/inject/dynamic-theme/image.ts
```diff
@@ -88,13 +88,13 @@ function analyzeImage(image: HTMLImageElement) {
             if (a < TRANSPARENT_ALPHA_THRESHOLD) {
                 transparentPixelsCount++;
             } else {
-                min = Math.min(r, g, b);
-                max = Math.max(r, g, b);
-                l = (max + min) / 2;
-                if (l < DARK_LIGHTNESS_THRESHOLD) {
+                // Use sRGB to determine the `pixel Lightness`
+                // https://en.wikipedia.org/wiki/Relative_luminance
+                p = (0.2126 * r + 0.7152 * g + 0.0722 * b) / 255
```
alexanderby, line 93: Missing semicolon `;`

Issue #535, opened 2018-05-27
Dark Reader making firefox slow when active


Jonatthu:
I am on Firefox version 61
Every time that I activate dark rider on firefox this one becomes slow
Test for example scrolling in this page. The scroll becomes not smoth.
https://news.google.com/?hl=en-US&gl=US&ceid=US%3Aen

And Firefox becomes slow as well.

alexanderby:
Most likely you are using Filter mode. It's performance is bad in Firefox, try switching to Dynamic mode. Some changes to CSS filters specification are prepared, maybe it's implementation will improve the scrolling performance in future.

DigitalPig:
I am using the Dynamic mode currently but still have the same slow down issue on twitter.com

rolagg:
I think it's not a bug, just low PC/mobile (especially) performance.

bloodgain:
Rolagg might be right, as I don't get any scrolling issues on these pages. However, in Firefox 60.1.0esr 32-bit, on my fairly powerful work laptop, I do get occasional messages from Firefox that "a script in Dark Reader is causing Firefox to run slow". I'm using Dynamic mode on all pages that aren't toggled off.

For what it's worth, I did manually install Dark Reader by dropping the XPI into my profile. That generally doesn't cause problems with extensions, but I didn't want to leave out that detail.

Orionsg:
It is definitely not just low PC/mobile performance!

I have noticed that web pages are extremely slow in rendering when Dark Reader is activated as compared to when it is turned off in the Firefox browser in the same device. For me, It is not so much the scrolling, but rather that the page simply does not show until after a long time. When I turn off Dark Reader, the page will load fast again on the same device. I have seen this behaviour on both Ubuntu Linux and on my Android phone.

It is rather a pity as it makes using Dark Reader unbearable when a web page take e.g. 20 seconds to load in extreme cases when Dark Reader is active. Also the error message "a script in Dark Reader is causing Firefox to run slow" is seen regularly. When opening the same page in Chrome, it loads fast with the same time for Dark Reader turned on and off -- thus it is a Dark Reader problem on Firefox.

I use Firefox Quantum 63.0 (32-bit) on Ubuntu.
On Android it is 62.0.3

As an extreme example I load this stackoverflow page:
https://stackoverflow.com/questions/7020136/free-c-sharp-qr-code-generator

[...]

PR #2550, opened 2020-05-18
Mutation Observer performance

alexanderby:
- Scan the whole DOM tree for changes when there's too large count of mutations.
- Don't handle unnecessary mutations.
- Another attempt for #535.

### DIFF
--- a/src/inject/utils/dom.ts
+++ b/src/inject/utils/dom.ts
@@ -145,3 +145,152 @@ export function iterateShadowNodes(root: Node, iterator: (node: Element) => void
         iterateShadowNodes(node.shadowRoot, iterator);
     }
 }
+
+export function isDOMReady() {
+    return document.readyState === 'complete' || document.readyState === 'interactive';
+}
+
+const readyStateListeners = new Set<() => void>();
+
+export function addDOMReadyListener(listener: () => void) {
+    readyStateListeners.add(listener);
+}
+
+export function removeDOMReadyListener(listener: () => void) {
+    readyStateListeners.delete(listener);
+}
+
+if (!isDOMReady()) {
+    const onReadyStateChange = () => {
+        if (isDOMReady()) {
+            document.removeEventListener('readystatechange', onReadyStateChange);
+            readyStateListeners.forEach((listener) => listener());
+            readyStateListeners.clear();
+        }
+    };
+    document.addEventListener('readystatechange', onReadyStateChange);
+}
+
+const HUGE_MUTATIONS_COUNT = 1000;
+
+function isHugeMutation(mutations: MutationRecord[]) {
+    if (mutations.length > HUGE_MUTATIONS_COUNT) {
+        return true;
+    }
+
+    let addedNodesCount = 0;
+    for (let i = 0; i < mutations.length; i++) {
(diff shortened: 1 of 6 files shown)


Gusted:
Poor CPU & Dark Reader(mutation-observer-performance branch) Vs big DOM website https://www.w3.org/TR/2018/WD-html53-20181018/

 Profiler link: https://perfht.ml/3dYYOat

Gusted:
Poor CPU & Dark Reader(mutation-observer-performance branch) Vs big DOM website https://www.w3.org/TR/2018/WD-html53-20181018/

https://perfht.ml/36c7urv

src/inject/utils/dom.ts
```diff
@@ -145,3 +145,153 @@ export function iterateShadowNodes(root: Node, iterator: (node: Element) => void
...
+
+const HUGE_MUTATIONS_COUNT = 1000;
+
+function isHugeMutation(mutations: MutationRecord[]) {
+    if (mutations.length > HUGE_MUTATIONS_COUNT) {
+        return true;
+    }
+
+    let addedNodesCount = 0;
+    for (let i = 0; i < mutations.length; i++) {
```
Gusted, line 182: Length caching

Gusted:
Poor CPU & Dark Reader(36a43b5) Vs big DOM website 
https://www.w3.org/TR/2018/WD-html53-20181018/

And this website is resolved 3 seconds until Interactive and some seconds later not more loading icon :D


https://perfht.ml/2WKnqy5

src/inject/utils/dom.ts
alexanderby: Let's trash what? What's the point of length caching?

Gusted: It was a joke :) 
But 
https://jsben.ch/kidOp 
Just that little performance boost

alexanderby: The performance boost can't be little :)

Here's what I see in Firefox
Screen Shot 2020-05-19 at 19 38 03
and after a 3rd relaunch in Chrome
Screen Shot 2020-05-19 at 19 37 11
So, it's obvious that `for..of` is 4 times slower in Firefox and 25% slower in Chrome. But all other functions are equal, and I choose what looks clearer (though some people prefer to use `Array.some` or `Array.every` to exit loop earlier, I think `for` loop is more clear).

Gusted: I don't get at all the same result........
Poor laptop

alexanderby: ... And maybe you've seen this saying of Donald Knuth
> Programmers waste enormous amounts of time thinking about, or worrying about, the speed of noncritical parts of their programs, and these attempts at efficiency actually have a strong negative impact when debugging and maintenance are considered. We should forget about small efficiencies, say about 97% of the time: premature optimization is the root of all evil. Yet we should not pass up our opportunities in that critical 3%

Gusted: If every time someone says `BUT THIS DROP PERFORMANCE` etc it start to get worried and I didn't though about performance at all if you look at my earliest PR's for darkreader with changes of codebase I used var =. I used a lot of functions what easy could be reduced to 1 etc.

PR #2557, opened 2020-05-19
Reddit.com fix --commentswrapper-gradient-color

tahlyn:
In a desktop browser, when visiting any subreddit post that has multiple comments, most of the comments are hidden and the blue "View Entire Discussion" button must be clicked to expand more comments.  Directly above this button is a white gradient.  You can see an example of this here:

https://www.reddit.com/r/ProtonVPN/comments/gh5fxd/ip_leak/

My fix is to set --commentswrapper-gradient-color to #ccc.


Gusted:
First I don't see anything like a view entire discussion button.
Second  < It's already being darken by Dark reader.

Issue #2560, opened 2020-05-19
[Feature Request] Stop inverting .leaflet-interactive in dynamic


austinhuang0131:
**Is your feature request related to a problem? Please describe.**
Sorta. Leaflet is the map library => https://leafletjs.com, which usually displays OpenStreetMap. However, since OSM itself is not dark-mode-friendly, it would make sense to not invert any lines drawn on it as well. Example: https://austinhuang.me/ogf-route-relations.html?map=C/15/37.33365/103.45317 (in this case I'm using OpenStreetMap standard layer for a different website; The thin bus lines were supposed to be #1F3D0C but was replaced with an invisible grey-ish color)

**Describe the solution you'd like**
Have an option to disable `--darkreader-inline-stroke`

**Additional context**

Gusted:
```
austinhuang.me

IGNORE INLINE STYLE
.leaflet-zoom-animated g path
```

Refresh the page after pressing apply

austinhuang0131:
Didn't realize `IGNORE INLINE STYLE` is a feature. I would suggest documenting it. But anyways, thanks.

Gusted:
Will make an PR with IGNORE INLINE STYLE but owner of Dark reader said it won't have a real world usage case but I will try to convience! Goodluck on fixing this site!

PR #2568, opened 2020-05-20
Add IGNORE INLINE STYLE documentation

Gusted:
Cough(Wearing face mask) #2560 It's usefull to let people know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,9 +53,6 @@ dynamic-theme-fixes.config
 
 example.com
 
-IGNORE INLINE STYLE
-.color-picker
-
 INVERT
 .icon
 
@@ -64,6 +61,9 @@ CSS
     background-color: ${white} !important;
     color: ${black} !important;
 }
+
+IGNORE INLINE STYLE
+.color-picker
 
 ```
 - `INVERT` rule inverts specified elements.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -53,6 +53,9 @@ dynamic-theme-fixes.config
 
 example.com
 
+IGNORE INLINE STYLE
```
alexanderby, line 56: Since it's rarely used, let's put it on the last place.

README.md
```diff
@@ -63,6 +66,7 @@ CSS
 }
 
 ```
+- `IGNORE INLINE STYLE` rule will not add any inline styles into matched selector's. (e.g. --darkreader-inline-bgcolor)
```
alexanderby, line 69: Imagine you don't know how Dark Reader works. Will you understand what `IGNORE INLINE STYLE` does after reading this? It should say, that it makes Dark Reader ignore inline styles for the element. Let's also move it after `INVERT` and `CSS` descriptions.

README.md
Gusted: That's now how our code work IGNORE INLINE STYLE I first one

alexanderby: Then I think the code should be fixed.

Gusted: Okay

PR #2575, opened 2020-05-21
Fix debug script

Gusted:


src/inject/utils/log.ts
```diff
@@ -1,10 +1,10 @@
-declare const __DEBUG__: boolean;
-const DEBUG = __DEBUG__;
+declare const __WATCH__: boolean;
```
alexanderby, line 1: DEBUG should be still there. DEBUG is used for logging info while you debug, WATCH is used for watching for code changes.

tasks/build.js
```diff
@@ -35,7 +35,7 @@ async function release() {
 async function debug({watch}) {
     log.ok('DEBUG');
     try {
-        await runTasks(standardTask, {production: false});
+        await runTasks(standardTask, {production: watch});
```
alexanderby, line 38: It should be `{debug: true, watch}`

tasks/bundle-api.js
```diff
@@ -31,7 +31,7 @@ async function bundleAPI({production}) {
                 cacheRoot: production ? null : `${fs.realpathSync(os.tmpdir())}/darkreader_api_typescript_cache`,
             }),
             rollupPluginReplace({
-                '__DEBUG__': 'false',
+                '__WATCH__': 'false',
```
alexanderby, line 34: `__DEBUG__` should be in this case

tasks/bundle-js.js
```diff
@@ -108,8 +108,8 @@ async function bundleJS(/** @type {JSEntry} */entry, {production}) {
                 cacheRoot: production ? null : `${fs.realpathSync(os.tmpdir())}/darkreader_typescript_cache`,
```
alexanderby, line 108: So, there should be no more `production` property, but `debug` with inverted logic (if debug, use cache etc).

tasks/bundle-js.js
```diff
@@ -108,8 +108,8 @@ async function bundleJS(/** @type {JSEntry} */entry, {production}) {
...
             }),
             rollupPluginReplace({
-                '__DEBUG__': production ? 'false' : 'true',
-                '__PORT__': production ? '-1' : String(PORT),
+                '__WATCH__': production ? 'true' : 'false',
```
alexanderby, line 111: Both `__DEBUG__` and `__WATCH__` should be there. Search for `production` all over the code and replace it with `debug` and add `watch` where necessary.

src/inject/utils/log.ts
Gusted: .... This getitng me headache

tasks/bundle-css.js
```diff
@@ -41,7 +41,7 @@ module.exports = createTask(
 ).addWatcher(
     ['src/**/*.less'],
     async () => {
-        await bundleCSS({production: false});
+        await bundleCSS({debug: false});
```
alexanderby, line 44: This should be `debug: true`. Later I will make it pass the options into this function.

tasks/bundle-html.js
```diff
@@ -98,7 +98,7 @@ async function rebuildHTML(changedFiles) {
     await Promise.all(
         pages
             .filter((page) => changedFiles.some((changed) => changed === getSrcPath(page.cwdPath)))
-            .map((page) => bundleHTMLPage(page, {production: false}))
+            .map((page) => bundleHTMLPage(page, {debug: false}))
```
alexanderby, line 101: This should be `debug: true`.

tasks/bundle-js.js
```diff
@@ -100,27 +100,27 @@ async function bundleJS(/** @type {JSEntry} */entry, {production}) {
...
                     compilerOptions: {
-                        removeComments: production ? true : false,
-                        sourceMap: production ? false : true,
+                        removeComments: debug ? false : true,
+                        sourceMap: debug ? true : false,
                     },
                 },
-                clean: production ? true : false,
-                cacheRoot: production ? null : `${fs.realpathSync(os.tmpdir())}/darkreader_typescript_cache`,
+                clean: debug ? false: true,
```
alexanderby, line 107: Run the `code-style`, it should fix it.

tasks/bundle-js.js
```diff
@@ -100,27 +100,27 @@ async function bundleJS(/** @type {JSEntry} */entry, {production}) {
...
+                cacheRoot: debug ? `${fs.realpathSync(os.tmpdir())}/darkreader_typescript_cache` : null,
             }),
             rollupPluginReplace({
-                '__DEBUG__': production ? 'false' : 'true',
-                '__PORT__': production ? '-1' : String(PORT),
+                '__DEBUG__': watch ? 'true' : 'false',
```
alexanderby, line 111: Use `debug` boolean value here

tasks/bundle-js.js
```diff
@@ -100,27 +100,27 @@ async function bundleJS(/** @type {JSEntry} */entry, {production}) {
...
+                '__PORT__': watch ? String(PORT) : '-1',
```
alexanderby, line 112: Add a 3rd `__WATCH__` parameter here and use if to enable or disable the WebSocket connection.

tasks/bundle-locales.js
```diff
@@ -51,7 +51,7 @@ module.exports = createTask(
     ['src/_locales/**/*.config'],
     async (changedFiles) => {
         for (const file of changedFiles) {
-            await bundleLocale(file, {production: false});
+            await bundleLocale(file, {debug: false});
```
alexanderby, line 54: This should be `debug: true`.

tasks/copy.js
```diff
@@ -55,8 +55,8 @@ module.exports = createTask(
     async (changedFiles) => {
         for (const file of changedFiles) {
             if (await fs.exists(file)) {
-                await copyFile(file, {production: false, firefox: false});
-                await copyFile(file, {production: false, firefox: true});
+                await copyFile(file, {debug: false, firefox: false});
```
alexanderby, line 58: This should be `debug: true`.

tasks/copy.js
```diff
@@ -55,8 +55,8 @@ module.exports = createTask(
...
+                await copyFile(file, {debug: false, firefox: true});
```
alexanderby, line 59: This should be `debug: true`.

PR #2578, opened 2020-05-22
Fixed link loading handling in Firefox

alexanderby:
- When `<link>` is loading in Firefox, the `sheet` property is not `null`, but an error is thrown when trying to access `sheet.cssRules`.
- Discovered this on links with `href="blob:"`, see issue #2065.


evilpie:
@alexalekseyenko please file a Firefox bug with a testcase.

PR #2582, opened 2020-05-22
add research.protocol.ai to dynamic-theme-fixes.config

hyperfekt:
FIxes the main text to be inverted instead of black.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2717,6 +2717,15 @@ CSS
 
 ================================
 
+research.protocol.ai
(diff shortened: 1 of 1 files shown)


Gusted:
Let's not do that as the main div is white and not visible

hyperfekt:
Good point. Given that the classes seem to be generated(?) it doesn't seem worth it to try and maintain the ones needed to select just the text on a black background.

Gusted:
Nope if they want to generate it each time Dark Reader also could fail at a point to darken it, I don't support non-normalized classes at all.

PR #2587, opened 2020-05-22

waterdrinkerAP:
There were big white boxes in the middle of the screen. Changed them to dark gray to better fit with the rest of the page. There was also some text in said boxes that was black and now is almost white.


Gusted:
Example url?

waterdrinkerAP:
The website is [https://www.wanikani.com/](url) but I believe you need to register in order to see what I am talking about.

PR #2588, opened 2020-05-23
Fix for p30download.com

ymg2006:
//fixed background color for p30download.com

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2542,6 +2542,15 @@ div.comment-op {
 
 ================================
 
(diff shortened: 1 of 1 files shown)


Gusted:
I see no changes in the site that are affected.

ymg2006:
> I see no changes in the site that are affected.

I forgot to put . (dot) before class tag, the edit would be :

p30download.com

CSS
.article-wrapper{
    color: ${#090702} !important;
}

attached images.
https://img.techpowerup.org/200523/disabled.png
https://img.techpowerup.org/200523/ednabled.png

PR #2589, opened 2020-05-23
Image processing performance

alexanderby:
- Reuse a single Canvas (5x times faster in Chrome and 20% in Firefox https://jsben.ch/ZPvNu).
- Offscreen Canvas is slow (available in Chrome only) https://jsben.ch/vc7Gm
- Clear Object URLs when necessary.


RanHum:
Clarification: in Firefox reusing canvas is 62% faster on software backend, at least on my i5-2520K

Issue #2101, opened 2020-03-10
Dynamic causing issues with Wikipedia


Thepenguin9:
Whilst using Vector skin on Wikipedia, the site almost universally functions as expected, though when notifications are received the container for the notification is white-on-white until it will fade to dark after a short time.
I will try and grab a screenshot/gif of this in action.

Thepenguin9:
![GIF of bug](https://light-theme-hurts.my-ey.es/5haZrc3.gif)

ElhemEnohpi:
It's related to an animation of the background colour between unseen and unread/read messages. Search the dev tools window for "wikipedia", and try adding this under CSS:

```
.mw-echo-ui-notificationItemWidget-initiallyUnseen.mw-echo-ui-notificationItemWidget-unread {
    background-color: rgb(24, 26, 27) !important;
}
```
There's maybe a more elegant way to do it, but this makes it readable at least.

Thepenguin9:
I can now make out the message without having to wait. Thanks for the ad-hoc fix, is it worth adding this to master? The behaviour is much better than without, and the messages still blink when they've been seen.

Gusted:
It's always nice to have this little things into the master. Only problem is I don't understand

> .mw-echo-ui-notificationItemWidget-initiallyUnseen.mw-echo-ui-notificationItemWidget-unread {
>     background-color: rgb(24, 26, 27) !important;
> }
If this fix the problem So yes? Could you give a gif so I can make a PR for this

Thepenguin9:
> If this fix the problem So yes? Could you give a gif so I can make a PR for this

A gif isn't necessary. It makes the unseen colour (the white) a shade of grey. Function is normal other than that. Matches everything else

ElhemEnohpi:
There's a CSS animation like this:
```
.mw-echo-ui-notificationItemWidget-initiallyUnseen {
    -webkit-animation-name: unseen-fadeout-to-read;
    animation-name: unseen-fadeout-to-read;
    -webkit-animation-duration: 0.5s;
    animation-duration: 0.5s;
    -webkit-animation-delay: 2s;
    animation-delay: 2s;
    -webkit-animation-timing-function: ease-out;
    animation-timing-function: ease-out;
    -webkit-animation-fill-mode: both;
    animation-fill-mode: both;
}

Keyframes unseen-fadeout-to-unread
0% {
    background-color: #dce8ff;
}

100% {
    background-color: #ffffff;
}
```

[...]

PR #2590, opened 2020-05-23
Add keyframe support

Gusted:
1. It's being more used in the wild.

Would solve #2101 But was manual with `@keyframes` they are now also dropped.
Resolve #2590

https://developer.mozilla.org/en-US/docs/Web/CSS/@keyframes < Documentation I followed to get everything worked of how the CSS should look like.

The example will show it will infinty fade in the `block`.
It works with Dakreader Off and with this Patch. Without it, it will just fade out as we don't target the colors.

Example I used
[n.zip](https://github.com/darkreader/darkreader/files/4671882/n.zip)

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3840,22 +3840,6 @@ a[href$=".svg"]:hover > img {
 .diff-deletedline .diffchange {
     background-color: ${#feeec8} !important;
 }
-@keyframes unseen-fadeout-to-unread {
-    from {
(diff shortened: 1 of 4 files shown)


Gusted:
Possible bug with ModifyColorWithCachr they the second color e.g. White or black so fat I know would return the same color but just in hex format...

ruffsl:
Could this be rebased with the merge conflicts resolved?

Gusted:
Rebasing is harder than just rewriting this PR from scratch with the new logic.

Gusted:
#3819 Done

PR #2613, opened 2020-05-25
fix for ipko.pl

Myshor:
ipko logo on login page invert
image


Myshor:
Logo after login invert add.
image

PR #2618, opened 2020-05-26
Update dark-sites.config

sprinkledcheese:
Added some sites.


Gusted:
That are quite full 😅 I will check sites carefully. Do I have permission to move some sites around so it will fit the alphabetic order? @sprinkledcheese

sprinkledcheese:
Yes @Gusted

PR #2621, opened 2020-05-26

neonr4in:
elearning.budiluhur.ac.id: Body text.


Gusted:
COLOR isn't a valid color to use.

PR #2644, opened 2020-05-28
www.realtek.com mobile site fix

sprinkledcheese:
Inverted header colour, also made "search", "globe" button transparent, and "MORE" button dark.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4025,6 +4025,28 @@ IGNORE INLINE STYLE
 
 ================================
 
+www.realtek.com
+
+CSS
+#t3-header
+{
+    background: none !important;
+}
+#t3-header > div > div.col-xs-3.visible-xs-block > button.search-toggle
+{
+    background-color: inherit !important;
+}
+#t3-header > div > div.col-xs-3.visible-xs-block > button.globe-toggle
+{
+    background-color: inherit !important;
+}
+body > div.t3-wrapper > div > div.section-wrap.bg-wrap1 > div > div > div.custom > div > a
+{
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -4025,6 +4025,28 @@ IGNORE INLINE STYLE
 
 ================================
 
+www.realtek.com
```
Gusted, line 4028: Remove www. to avoid any compatible issues.

src/config/dynamic-theme-fixes.config
```diff
@@ -4025,6 +4025,28 @@ IGNORE INLINE STYLE
...
+
+CSS
+#t3-header
+{
+    background: none !important;
+}
+#t3-header > div > div.col-xs-3.visible-xs-block > button.search-toggle
+{
```
Gusted, line 4036: `{` on the same line as the selector only different selectors on seperate lines.

src/config/dynamic-theme-fixes.config
sprinkledcheese: "realtek.com" doesn't exist.

Gusted: Yes but some files can be loaded from sub-domains etc. We only use www. for rare cases where it's needed.

src/config/dynamic-theme-fixes.config
sprinkledcheese: Done.

src/config/dynamic-theme-fixes.config
sprinkledcheese: ok.

PR #2651, opened 2020-05-29
Fixed 4pda.ru shadow

sprinkledcheese:


src/config/dynamic-theme-fixes.config
```diff
@@ -111,6 +111,9 @@ CSS
 .catend {
     background: ${#e8e8e8} !important;
 }
+div.description > span {
+    height: 0 !important;
```
Gusted, line 115: `display: none` Should be understandable why it isn't displaying.

sprinkledcheese: That works too.

PR #2653, opened 2020-05-29
Update dark-sites.config

sprinkledcheese:
Added humblebundle, humblegames. Removed bing since it's not dark and basho because it's inaccesible.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -23,7 +23,6 @@ asoftmurmur.com
 ast4u.me
 avengedsevenfold.com
 bash.im/comics
-basho.com/$
 batflix.org
 bbc.co.uk/iplayer
 beastskills.com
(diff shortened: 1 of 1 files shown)


src/config/dark-sites.config
```diff
@@ -36,7 +35,6 @@ beta.jdownloader.org
 betterttv.com
 bherila.net
 bin.disroot.org
-bing.com/$
```
Gusted, line 39: Home-page

Gusted: Is broken with darkreader on

src/config/dark-sites.config
```diff
@@ -169,6 +167,12 @@ heavybit.com
 hidive.com
 himovies.to
 hostedtalk.net
+humblebundle.com/$
```
Gusted, line 170: Why doing this well already refericing to other sub-url's?

src/config/dark-sites.config
sprinkledcheese: ok.

src/config/dark-sites.config
sprinkledcheese: Because those other sub-url's have dark mode.

PR #2669, opened 2020-05-30
Add tirto.id to dark-sites.config

neonr4in:
tirto.id has a toggle-able mode and with it a dark mode.

image


sprinkledcheese:
Must be dark by default.

PR #2679, opened 2020-05-31
fix for gazetawroclawska.pl

Myshor:
back with fix removed


Myshor:
bad page fixing... removing branch :)

PR #2682, opened 2020-06-01
fix for facebook

Myshor:
Removed one class to have icons in moves white in new native dark mode


Myshor:
Added INLINE IGNORE fix for dark stickers in messages.
image

PR #2687, opened 2020-06-01
Fixed instructure.com

BananaManCJ:
Removed .title_3bek::after from inverted list on instructure.com as the dynamic theme inverts it correctly now


Gusted:
Example link?(With sub-url as homepage doens't contain it)

Gusted:
@BananaManCJ  ^

PR #2696, opened 2020-06-02
📝 Add "discordtemplates.com", again!

eggsywashere:
Template defaults to dark now!


Gusted:
Just how it should be!
Dark by default.

PR #2715, opened 2020-06-03
fix for facebook.com

Myshor:
Dark icon on Use Application button. (forced to not use blob)
image


Myshor:
Add fix for emoiji in old pages when new native dark mode enabled.
They were inverted. :D
image

PR #2741, opened 2020-06-05
Add an INVERT for bestbuy.com Compare

erockce:
The default is a blinding bright white.  A simple INVERT makes it tolerable.  The default darker color choice was fine for me.

### DIFF
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -183,6 +183,13 @@ div.orb-nav-section.orb-nav-blocks > a > img
 
 ================================
 
+bestbuy.com
+
+INVERT
+#None > div > section > div.comparison-container.comparison-container-l.legacy-variant
+
+================================
+
 betterexplained.com
 
 NO INVERT


Gusted:
Example URL?

erockce:
There isn't a direct URL available.  The CSS is added when you add products to compare with one another.

Guess you could use the following steps for demonstration:
# [Roccat Vulcan 120 AIMO Wired Gaming Mechanical Keyboard with Back Lighting - Black](https://www.bestbuy.com/site/roccat-vulcan-120-aimo-wired-gaming-mechanical-keyboard-with-back-lighting-black/6237355.p?skuId=6237355)
# Click the *Compare* check-box under _Add to Cart_.  This will add the new Compare CSS layer on the page at the bottom
# Then scroll to the bottom to see the "People also viewed" section
# (Optional) Click *Compare* for other

After making the PR I realized that I did not verify any other aesthetics like preserving the existing images of the products in the Compare.  Having the change locally in Dev Tools, and changing the _Dynamic_ theme engine looks even better, but at the expense of more CPU and memory resources.

src/config/inversion-fixes.config
```diff
@@ -183,6 +183,13 @@ div.orb-nav-section.orb-nav-blocks > a > img
 
 ================================
 
+bestbuy.com
+
+INVERT
+#None > div > section > div.comparison-container.comparison-container-l.legacy-variant
```
Gusted, line 189: Let's add `.product-container` So the product itself also becomes inverted.

erockce: So in effect, inverting once for the `.comparison-container`, then inverting again on just the `.product-container`.

Thank you for the help.

I will update the branch and PR.

erockce:
I messed up by deleting the source fork/branch. See #2756.

PR #2747, opened 2020-06-06
Update dark-sites.config

FarewellAngelina:
tuganetgames.pt


Gusted:
Dark Reader prevent on this site to flash my eye's with an (I suggest) loading (White-ish background) image to be darken.

PR #2748, opened 2020-06-06
fix for web.whatsapp.com

Myshor:
Added alernative data-asset-intro-image-light.
Changed class for QR code.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4082,10 +4082,10 @@ CSS
 ._3Ye_R {
     color: gray !important
 }
-[data-asset-intro-image] {
+[data-asset-intro-image], [data-asset-intro-image-light] {
     background-image: url(/img/intro-connection_c98cc75f2aa905314d74375a975d2cf2.jpg) !important;
 }
-._11ozL {
+._1QMFu {
     border: 5px solid white !important;
 }
 span[data-icon="tail-in"] {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -4082,10 +4082,10 @@ CSS
 ._3Ye_R {
     color: gray !important
 }
-[data-asset-intro-image] {
+[data-asset-intro-image], [data-asset-intro-image-light] {
     background-image: url(/img/intro-connection_c98cc75f2aa905314d74375a975d2cf2.jpg) !important;
 }
-._11ozL {
+._1QMFu {
```
Gusted, line 4088: Changing every month? Let's see if we can use a `>` strict CSS3 to target this CSS to make it more reliabler in the next updates for whatsapp. Related what you did with google 9 dots in the right above corner?

Myshor: Saw inside class nice canvas with aria-label="Scan me!" but making border for it doesn't work and CSS still doesn't work for parents. 😉 but who knows... maybe I'll find some solution in the future.

Gusted: Let's not merge it yet I will have a look at it today to get a more global solution to this.

Gusted: `html[dir] .landing-main > :first-child > :nth-child(2) > :first-child`

Myshor:
We will see if it won't change. 😏

Gusted:
Than whatsapp will have a angry mail about normalized CSS 😠

PR #2755, opened 2020-06-06
fix calendar.google.com

starsep:
It inverts color of checkboxes in
My calendars/Other calendars panels

With Dark Reader disabled:
Dark Reader disabled

With Dark Reader currently:
Dark Reader currently

After invert:
After invert

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -515,6 +515,18 @@ INVERT
 
 ================================
 
+calendar.google.com
+
+CSS
+div[role="checkbox"] > div > div > div {
+    border: 1px solid ${black} !important;
+}
+
+IGNORE INLINE STYLE
+div[role="checkbox"]
+
+================================
+
 cheapshark.com
 
 CSS


src/config/dynamic-theme-fixes.config
```diff
@@ -515,6 +515,13 @@ INVERT
 
 ================================
 
+calendar.google.com
+
+INVERT
+.rq8Mwb
```
Gusted, line 521:
```
calendar.google.com

CSS
div[role="checkbox"] > div > div > div {
    border: 1px solid ${black} !important;
}

IGNORE INLINE STYLE
div[role="checkbox"]
```
Will be a better aproach to this.
1. Don't inline style the checkbox to not mess up the colors.
2. A more global solution to set a the border to white so other checkboxes in their will be white as well.
3. Not realing on a not normalized CSS class.

starsep: I used the snippet you posted, I confirm that it works.
Thanks for a help and clarification

Gusted: No problem Google have a kink on changing CSS selectors that are `jibberish`

PR #2763, opened 2020-06-07
Some tweaks to Jupyter notebook inverting

PyroGenesis:
These changes are helpful for me to work with Jupyter notebook with dark reader ON.

* The INVERT is for the Jupyter notebook logo
* CSS is for PNG images (matplotlib axes are black by default) and selected text highlighting
* Ignore output images

I'm not sure whether we are allowed to add localhost links, but the 88XX are the most common ports for local Jupyter notebooks.

PR #2767, opened 2020-06-07
Expose Customization to Dark Reader CSS

Gusted:
Follow-up for #2763


alexanderby:
Can you explain what does it make?

Gusted:
&{selectionBackground} in the devtools will be replaced by the current background selection that the user chooses. As some like jupyter playground use code mirror that use a class to simulate the selection

alexanderby:
An interesting idea. Maybe let's simplify it. This mechanism with `${}` is weird, people don't understand it. I'm not rushing you to implement it, just to think. Remember we were discussing CSS variables for static themes? What if we declare some variables, like `--darkreader-neutral-bg, --darkreader-neutral-text, --darkreader-selection-bg` etc? Then people can use it in their configs like `.selector { background: var(--darkreader-selection-bg) !important; }`.

Gusted:
Good idea actually with the variables

alexanderby:
Maybe let's start with 4 variables then.

You can declare
```html
<style class="darkreader darkreader--variables">
    :root {
      --darkreader-neutral-bg: ...;
      ...
    }
</style>

File: src/inject/dynamic-theme/index.ts [hunk 1]
```diff
@@ -91,6 +91,20 @@ function createStaticStyleOverrides() {
...
+    variableStyle.textContent = [
+        `:root {`,
+        `   --darkreader-neutral-bg: ${filter.mode === 0 ? lightSchemeBackgroundColor : darkSchemeBackgroundColor};`,
+        `   --darkreader-neutral-text: ${filter.mode === 0 ? lightSchemeTextColor : darkSchemeTextColor};`,
+        `   --darkreader-selection-bg: ${selectionColors.backgroundColorSelection};`,
+        `   --darkreader-selection-fg: ${selectionColors.foregroundColorSelection};`,
+        `}`
+    ].join('\n');
+    document.head.insertBefore(variableStyle, inlineStyle.nextSibling);
+    setupStylePositionWatcher(variableStyle, 'variable');
```
alexanderby, line 107: Let's use plural `variables`

src/inject/dynamic-theme/index.ts
```diff
@@ -91,6 +91,20 @@ function createStaticStyleOverrides() {
     overrideStyle.textContent = fixes && fixes.css ? replaceCSSTemplates(fixes.css) : '';
     document.head.appendChild(overrideStyle);
     setupStylePositionWatcher(overrideStyle, 'override');
+
+    const variableStyle = createOrUpdateStyle('darkreader--variable');
+    const selectionColors = getSelectionColor(filter);
+    const {darkSchemeBackgroundColor, darkSchemeTextColor, lightSchemeBackgroundColor, lightSchemeTextColor} = filter;
+    variableStyle.textContent = [
+        `:root {`,
+        `   --darkreader-neutral-bg: ${filter.mode === 0 ? lightSchemeBackgroundColor : darkSchemeBackgroundColor};`,
```
alexanderby, line 100: How do you think, if it is meant to be used by users, maybe we should use colors like `--darkreader-neutral-background` and `--darkreader-neutral-text`? Maybe `bg` tells something, but `fg` is less common.

src/inject/dynamic-theme/index.ts
```diff
@@ -91,6 +91,20 @@ function createStaticStyleOverrides() {
     overrideStyle.textContent = fixes && fixes.css ? replaceCSSTemplates(fixes.css) : '';
     document.head.appendChild(overrideStyle);
     setupStylePositionWatcher(overrideStyle, 'override');
+
+    const variableStyle = createOrUpdateStyle('darkreader--variable');
```
alexanderby, line 95: Let's use plural `darkreader--variables`

PR #2773, opened 2020-06-07
Fix flyzipline.com

starsep:
- Invert logo


Gusted:
This is already without any fixes.
Why do we need to invert it when it's already white.

starsep:
Logo looks like that for me without fixes:


With fix:

src/config/dynamic-theme-fixes.config
```diff
@@ -1347,6 +1347,13 @@ CSS
 
 ================================
 
+flyzipline.com
+
+INVERT
+#logo
+
```
Gusted, line 1354: Add
```
CSS
#logo {
    background-image: url(/blogo.svg) !important;
}
```

PR #2777, opened 2020-06-08
Fixed wikipedia.org

sprinkledcheese:
- Inverted loudspeaker thumbnail.
- Wikipedia logo in the footer of mobile version.


Myshor:
Would be nice to have some example URLs to be sure what to check. (Or screens at least.)

sprinkledcheese:
Example URL: https://en.m.wikipedia.org/wiki/Brussels

Myshor:
Thank You. 👍

PR #2790, opened 2020-06-08
Fixed enjen.net

sprinkledcheese:
- Removed background image.


sprinkledcheese:
Example URL: https://www.enjen.net/asn-blocklist/

PR #2791, opened 2020-06-09
Background for Kaggle notebook images

PyroGenesis:
I added a white background (in dark mode) so that I can see the graph axes.

I used `kaggle*` instead of `kaggle.com` because kaggle loads notebooks in an iframe with URL something like `kaggleusercontent.com`  

Before: 
image

After: 
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1957,6 +1957,18 @@ INVERT
 
 ================================
 
+kaggle*
+
+CSS
+div.output_png img {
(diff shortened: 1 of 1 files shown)


Myshor:
The code looks nice i think. Can you share some example URL to see the screen from example?

PyroGenesis:
Sure. I tested it on this: https://www.kaggle.com/kanncaa1/pytorch-tutorial-for-deep-learning-lovers

Myshor:
OK looks nice after fixes.

PR #2794, opened 2020-06-09
nitter instances are dark

B0pol:
I added the missing [nitter instances](https://github.com/zedeus/nitter/wiki/Instances).

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -325,9 +325,17 @@ newgrounds.com
 nexusmods.com
 nh-server.github.io/switch-guide/
 niebezpiecznik.pl
+nitter.13ad.de
+nitter.42l.fr
+nitter.cattube.org
+nitter.dark.fail
+nitter.fdn.fr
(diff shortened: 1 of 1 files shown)


src/config/dark-sites.config
```diff
@@ -455,6 +463,7 @@ supinic.com
 surviv.io
 svtplay.se
 symfony.com
+t.maisputain.ovh
```
Gusted, line 466: Down.

B0pol: I'm pretty sure not, just IPV6 only. Their invidious instance yt.maisputain.ovh is written IPV6 only here: t.maisputain.ovh

I'll try to ask the owner and confirm

B0pol: If you don't know what IPv6 is, please read the [wikipedia page](https://en.wikipedia.org/wiki/IPv6). If you don't have an IPv6 address, it will show the website as down, but I can promise you it's still up today (tell me if you want an evidence and which).

PR #2796, opened 2020-06-09
add some of the already dark peertube instances

B0pol:


src/config/dark-sites.config
```diff
@@ -351,6 +353,12 @@ parahumans.wordpress.com
 paranoid.email
 pathof.info
 pathofexile.com
+peertube.club
```
Gusted, line 356: Not dark

src/config/dark-sites.config
```diff
@@ -351,6 +353,12 @@ parahumans.wordpress.com
...
+peertube.co.uk
+peertube.nocturlab.fr
+peertube.pontostroy.gq
+peertube.servebeer.com
```
Gusted, line 360: 503?

B0pol: It prolly was a maintenance, it's up

PR #2804, opened 2020-06-10
Fix for notebook equations in GitHub preview

PyroGenesis:
LaTeX equations could not be seen in GitHub notebook previews before due to them being images. This inverts them to be visible again.
The domain filter is set to the default domain GitHub uses to serve their notebook preview from a linked iframe.

Test URL: https://github.com/bentrevett/pytorch-pos-tagging/blob/master/1%20-%20BiLSTM%20for%20PoS%20Tagging.ipynb

Before:
image

After:
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3086,6 +3086,14 @@ regex101.com
 INVERT
 canvas
 
+
+================================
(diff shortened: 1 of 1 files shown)


PyroGenesis:
I'm not sure why it's failing the tests

PR #2844, opened 2020-06-13
Fix databricks spark UI

chinwobble:
Dark mode off
image

Before CSS changes in this PR.
image

With new CSS
image


chinwobble:
> 
> 
> Use 4 spaces
> Use `${color to be darken}` so `${white}` Will become a dark color.

Thanks I changed to 4 spaces and used invert for the diagrams.
I don't think its worth the effort to change all the existing css.

chinwobble:
After changing all `color: white` to `color: ${black}` light theme looks like this:

image

Issue #2730, opened 2020-06-04
Conflict between the extension and the js script on Linux


n-peugnet:
I'm still having this problem with Firefox and Chrome on 4.9.9 so I made an issue for this specific problem as requested.
I am using the min version of the script from jsdeliver. You can check the problem with <a href="https://mers.club1.fr/film-copy" rel="nofollow">this page</a>.

With the extension disbled both firefox and chromium renders the page without any issue.
![Screenshot from 2020-04-06 16-19-13](https://user-images.githubusercontent.com/23519418/78569117-67ddd080-7823-11ea-9db6-3b326b9f1057.png)

With the extension enabled:
- On chromium the page loading never ends and there is only a dark background. After some time.
  ![Screenshot from 2020-04-06 16-20-14](https://user-images.githubusercontent.com/23519418/78568593-9c9d5800-7822-11ea-9479-f8136ef5d5ae.png)

- On firefox the page loading never ends and I get a strange render
  ![Screenshot from 2020-04-06 16-16-16](https://user-images.githubusercontent.com/23519418/78568621-a6bf5680-7822-11ea-82a5-f95fb3e3ec12.png)

_Originally posted by @n-peugnet in https://github.com/darkreader/darkreader/issues/2227#issuecomment-609825503_

Gusted:
This definitely a Linux issue when I profiled this on the original issue what you have linked. Both for Firefox and chromium didn't showed any slowdown on Windows 10 on a rather poor system specification. Guess I going to install Linux. Which Distribution do you currently use?

n-peugnet:
I am using Linux Mint 19.3 Cinnamon 64 bits with kernel version 5.3.0-51-generic

Mahanotrahul:
@n-peugnet Did you try this solution https://github.com/darkreader/darkreader/issues/2227#issuecomment-620218136
I had the similar problem. Detecting if the dark reader extension is on before enabling the darkreader via script on the website solves the problem it seems.

Gusted:
#2855

Issue #2227, opened 2020-04-03
Adding dark reader to website using npm package not works when the extension is on


fainir:
I'm building a website and tried to implement dark reader inside of it. it works fine except when the dark reader extension is on. Can you tell me how can I fix it? (I use react)
Thanks a lot

Gusted:
I assume you mean the npm package darkreader. So when you enable darkreader in the code and enable darkreader as extension. It doesn't display the site good.

fainir:
Yes, sorry if I was unclear. Do you know what the reason is or what could be done to fix it?

Gusted:
Well I cannot reproduce this easily but I think there's a bug in darkreader code. If you could provide your current site without any like critical information so I can reproduce this issue would help a lot.

fainir:
I can't share the project that the bug accrued in because it isn't my code and I can't share it. I tried to reproduce it on other projects and the bug didn't accrue. So unfortunately I can't help you reproduce it :(
If you suspect that certain information about the project could lead you to resolve it, please ask me.
Sorry for not sharing the code and thank you for your help.

Gusted:
No problem, will only a bit hard to understand why this problem could occur. Could you maybe try screenshot certain elements that aren't displayed correctly. And does the site use the normal darkreader.enabled() with the standard Filterconfig?

fainir:
Yes, I understand the difficulty. The page doesn't load at all. it keeps loading the page forever 
![image](https://user-images.githubusercontent.com/6260146/78452502-31258000-7694-11ea-8159-978f519db468.png)

fainir:
It doesn't show any errors in the log because it keep on loading the page.

Gusted:
Could you remove darkreader in the code and try to load the page so it's confirmed to be darkreader sides

Gusted:
@alexanderby bug label please

fainir:
It works fine when the enableDark in the code without the extension. Also, it works fine when  the extension is on and I comment out the enableDark in the code

Gusted:
It's like if the too will compete each other if it's enabled or not... Could you try to before making a darkreader call. To place a debugger; and than step for step see where in the code is the problem that wil help a lot

fainir:
I did it and it didn't stop at the debugger it freeze the tab. Actually I didn't understand what do you mean by finding the problem in the code because the problem occurs only when enableDark function is added.

fainir:
[...]

PR #2855, opened 2020-06-13
Extension Detection

Gusted:
Pretty sure before we `add` any elements to the DOM just like scripts, we already inject the firstElementChild 

Resolves: #2227
Resolves: #2730

Not Tested


src/inject/dynamic-theme/index.ts
```diff
@@ -329,12 +331,30 @@ function stopWatchingForUpdates() {
     stopWatchingForInlineStyles();
     removeDOMReadyListener(onDOMReady);
 }
+function checkMeta() {
```
alexanderby, line 334: Every function should have a single purpose. So I would split it into 2 functions: `isAnotherDarkReaderInstanceActive` and `createDarkReaderInstanceMarker` (maybe there's a better word than `Marker`). And also it should remove the marker, when `removeDynamicTheme` is called.

src/inject/dynamic-theme/index.ts
```diff
@@ -329,12 +331,30 @@ function stopWatchingForUpdates() {
...
+    const meta: HTMLMetaElement = document.querySelector('meta[name="darkreader"]');
+    if (meta) {
+        if (meta.content !== UUID) {
+            return true;
+        }
+        return false;
+    } else {
+        const metaElement: HTMLMetaElement = document.createElement('meta');
+        metaElement.name = "darkreader";
```
alexanderby, line 343: code-style

src/inject/dynamic-theme/uuid.ts
```diff
@@ -0,0 +1,15 @@
+// `hex` variable contain all available hex values into array.
+function hexify(number: Number) {
+    return number < 16 ? "0" + number.toString(16) : number.toString(16);
+}
+// Secure & Fast UUID generator implementation of https://tools.ietf.org/html/rfc4122 version 4.
+export function generateUUID() {
```
alexanderby, line 6: I've found out that this is invalid UUID. You can simply copy the last snippet from here https://stackoverflow.com/a/2117523/4137472

And another radical solution: we don't really need UUID, you can rename it to UID, simply generate 4 32bit numbers, convert them `toString(16)` and compose. That will have event less chance of collision than UUID and will not have weird dashes.

Also you can remove comments.

src/inject/dynamic-theme/index.ts
```diff
@@ -14,9 +14,11 @@ import {getCSSFilterValue} from '../../generators/css-filter';
 import {modifyColor} from '../../generators/modify-colors';
 import {createTextStyle} from '../../generators/text-style';
 import {FilterConfig, DynamicThemeFix} from '../../definitions';
+import {generateUID} from './uuid';
 import {createAdoptedStyleSheetOverride, AdoptedStyleSheetManager} from './adopted-style-manger';
 
 const variables = new Map<string, string>();
+const UUID = generateUID();
```
alexanderby, line 21: I think it's worth renaming it to `INSTANCE_ID`.

src/inject/dynamic-theme/uuid.ts
```diff
@@ -0,0 +1,9 @@
+
+export function generateUID() {
+    const s = window.crypto.getRandomValues(new Uint8Array(16));
+    return s[0].toString(16) + s[1].toString(16) + s[2].toString(16) + s[3].toString(16) +
```
alexanderby, line 4: Everything was OK with your `hexify` function, except that you could make it shorter like `(x < 16 ? '0' : '') + x.toString(16)`. Now you can have a situation when `f + ff == ff + f`.

If you decide to make it not compatible with the spec, you can simply use `Array.from(...).map(hexify).join('') to not repeat the same expression 16 times.

src/inject/dynamic-theme/uuid.ts
```diff
@@ -0,0 +1,9 @@
+
+export function generateUID() {
+    const s = window.crypto.getRandomValues(new Uint8Array(16));
```
alexanderby, line 3: `window.` is unnecessary, `crypto` is a global object.

Gusted:
BTW it's 80/20 for who is running first the check 80% it's the API so maybe their could be some improvements to let the Extension still handle as some have custom settings. Or we could have somehow getting the settings into the API from the extension so it will be actually efficienter.

swrobel:
I'd like to propose an alternative solution (a bit late, I know, as this is already implemented), but hear me out.

There are a few issues with this solution:
1. It uses a custom meta tag that is specific to Dark Reader, rather than a standards compatible existing tag.
1. It uses a custom meta tag that is technically invalid, and thus won't render through some frameworks (like Ruby on Rails, as I've discovered). The tag includes an invalid `name` value, `darkreader`, which is not on the [official list](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name). This causes Rails to silently drop the tag when rendering the page, presumably since it deems it invalid.

Instead, I propose using `<meta name="color-scheme">` which you can read about by scrolling a short ways down [this page](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name). This is easy to check with javascript, ie: `document.querySelector('meta[name="color-scheme"]').content.includes("dark")`.

You can test this js on my site, https://metasurfforecast.com/ which is where I've been trying to use the existing solution to disable Dark Reader.

Gusted:
> It uses a custom meta tag that is specific to Dark Reader, rather than a standards compatible existing tag.

This is, to check if any instance of Dark Reader is active on that web pages to not get conflicts with the Dark Reader API and let site disable Dark Reader on their site.

> It uses a custom meta tag that is technically invalid, and thus won't render through some frameworks (like Ruby on Rails, as I've discovered). The tag includes an invalid name value, darkreader, which is not on the official list. This causes Rails to silently drop the tag when rendering the page, presumably since it deems it invalid.

Oh, yeah that's quite unfortunate.

> Instead, I propose using <meta name="color-scheme"> which you can read about by scrolling a short ways down this page. This is easy to check with javascript, ie: document.querySelector('meta[name="color-scheme"]').content.includes("dark").

I think that could work. Just a question so I don't misinterpret the documentation. You can specify `dark light` inside the content of the meta. Whereby the documentation says about the following:
```
Indicating multiple color schemes indicates that the first scheme is preferred by the document, but that the second specified scheme is acceptable if the user prefers it.
```
What for me indicates the first value in content is being used as `default` value for a fresh new user.
As your site has `<meta content="light dark" name="color-scheme">` Whereby for me the default was actually a dark theme *not sure if it comes to the '(prefers-color-scheme: dark)' media Query*. 

I think a more globally solution which is reliable is actionable, thanks for the proposal!

swrobel:
@Gusted thanks for the thorough response! You are correct that my site does render by default in light theme for users that don't have `prefers-color-scheme` support, or when their system color scheme is light. The dark theme will render when `prefers-color-scheme` is `dark` as you noticed (and as specified by the spec).

Gusted:
@swrobel #3902

alexanderby:
@swrobel What are you trying to achieve? Are you trying to disable Dark Reader for your website?

swrobel:
> @swrobel What are you trying to achieve? Are you trying to disable Dark Reader for your website?

Yes, see https://github.com/darkreader/darkreader/pull/2855#issuecomment-713040345

alexanderby:
OK, I've left some comments there.

atwright147:
How about `<meta http-equiv="x-is-dark" content="true">`?

Prefixing with an `x-` is valid (see `<meta http-equiv="X-UA-Compatible" content="IE=9; IE=8; IE=7; IE=EDGE" />` as an example. This could also be set at a HTTP header level to work around the issue that Dark Reader might start before the `<head>` has started rendering.

hadialqattan:
A workaround to disable Darkreader extension: https://github.com/hadialqattan/no-darkreader

PeterHindes:
Hey btw found a bug. If you disable and then reenable dark reader, the meta tag is ignored.

PR #2857, opened 2020-06-13
Wikipedia IS wrapping PDF's Can I speak to the manager?

### LINKED ISSUE #2856: [Broken Website] Wikipedia PDFs
**Website**
https://en.wikipedia.org/wiki/File:ExtIPA_chart_(2015).pdf or any other PDF file on Wikipedia.

**How to Reproduce**
Open the *File:* page of any PDF on Wikipedia.

**Expected behavior**
Darkreader changes the page style the same way as on other *File:* pages, e.g. https://en.wikipedia.org/wiki/File:Molecular_Modeling.png.

**Actual behavior**
Darkreader does not change the page style, instead displaying *This is page is protected by browser* in its popup.

**Screenshots**
On PDFs:
Image 041
On other file pages:
Image 042

**System info:**
 - OS: Windows 10 x64 version 2004 build 19041.329
 - Browser: Firefox 77.0.1 x64
 - Darkreader Version: 4.9.9
 
**Additional context**
May be related to #373, but I'm not sure.
At first, I interpreted the message in Darkreader's popup as a problem with Firefox' extension permissions, so I wrote [this question](https://support.mozilla.org/en-US/questions/1291091) on the Mozilla support forums. However, I was told that the problem must be caused by Darkreader. One of the replies there mentions finding references to a setting regarding PDFs in Darkreader's code, but I could not find any setting of the sort in Darkreader's popup.

Gusted:
No the manager will not speak to you.

Well fine than I resolve #2856

### DIFF
--- a/src/utils/url.ts
+++ b/src/utils/url.ts
@@ -106,7 +106,6 @@ function createUrlRegex(urlTemplate: string): RegExp {
 }
 
 export function isPDF(url: string) {
-    url = url.replace(/^.*?\/{2,3}/, '') // Remove scheme
     if (url.includes('.pdf')) {
         if (url.includes('?')) {
             url = url.substring(0, url.lastIndexOf('?'));
@@ -114,9 +113,10 @@ export function isPDF(url: string) {
         if (url.includes('#')) {
             url = url.substring(0, url.lastIndexOf('#'));
         }
-        //Wikipedia's PDF'S
-        if (url.includes(':')) {
-            url = url.substring(0, url.lastIndexOf(':'));
+        if (url.match(/wikipedia.org/i)) {
+            if (url.match(/wikipedia.org\/.*\/[a-z]\:*.+\.pdf/i)) {
+                return false;
+            }
         }
         if (url.endsWith('.pdf')) {
             for (let i = url.length; 0 < i; i--) {
--- a/src/utils/url.ts
+++ b/src/utils/url.ts
@@ -106,6 +106,7 @@ function createUrlRegex(urlTemplate: string): RegExp {
 }
 
 export function isPDF(url: string) {
+    url = url.replace(/^.*?\/{2,3}/, '') // Remove scheme
     if (url.includes('.pdf')) {
         if (url.includes('?')) {
             url = url.substring(0, url.lastIndexOf('?'));
--- a/src/utils/url.ts
+++ b/src/utils/url.ts
@@ -113,7 +113,7 @@ export function isPDF(url: string) {
         if (url.includes('#')) {
             url = url.substring(0, url.lastIndexOf('#'));
         }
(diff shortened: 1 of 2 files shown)


mohe2015:
Isn't it possible to check the mime-type or does the check need to be done before page load?

Gusted:
We need to check beforehand.

supertassu:
This will not work on other language Wikipedias, for example the same file on the Finnish Wikipedia is located at https://fi.wikipedia.org/wiki/Tiedosto:ExtIPA_chart_(2015).pdf?uselang=en and has a translated namespace "Tiedosto" instead of "File"

Gusted:
Ugh Wikipedia and their weird names will have to make another fix.

supertassu:
Not just Wikipedia, it also includes other Wikimedia projects and every other website that uses MediaWiki :P

Gusted:
Pushed a commit that will resolve it globally

Gusted:
Only if it wouldn't fail -_-

supertassu:
> 
> 
> Seems like this solution is too broad. Maybe let's try using `/wikipedia.org\/.*\/[a-z]\:*.+\.pdf/i` RegExp for this specific case? If the URL matches it, it is **not** PDF.
> 
> Also add this URL to tests `https://fi.wikipedia.org/wiki/Tiedosto:ExtIPA_chart_(2015).pdf?uselang=en`

Now other [Wikimedia projects](https://meta.wikimedia.org/wiki/Complete_list_of_Wikimedia_projects) (for example https://commons.wikimedia.org/wiki/File:ExtIPA_chart_(2015).pdf) and every other site using [MediaWiki](https://www.mediawiki.org/wiki/MediaWiki) will be broken

supertassu:
Sorry for throwing more complications in, but `upload.wikimedia.org` should be excluded as it hosts the actual PDF files (such as https://upload.wikimedia.org/wikipedia/commons/5/56/ExtIPA_chart_%282015%29.pdf)

alexanderby:
But this regexp detects /[a-z]+: patterns (like `/File:doc.pdf`)

Gusted:
Uhm 
It shouldn't be detected as PDF, we have groups that detectes the first part.....

alexanderby:
Oh, wait, the test fails, the correct RegExp is in comment.

src/utils/url.ts
```diff
@@ -113,6 +113,9 @@ export function isPDF(url: string) {
         if (url.includes('#')) {
             url = url.substring(0, url.lastIndexOf('#'));
         }
+        if (url.match(/(wikipedia|wikimedia).org/i) && url.match(/(wikipedia|wikimedia)\.org\/.*\/[a-z]\:*.+\.pdf/i)) {
```
alexanderby, line 116: Here's a better RegExp `/(wikipedia|wikimedia)\.org\/.*\/[a-z]+\:[^\:\/]+\.pdf/i`

PR #2860, opened 2020-06-14
Automation patch

### LINKED ISSUE #1713: Automatic toggle Dark/Light, NOT full extension
Hello,

I LOVE this extension. I am a graduate student and am on my computer 10+ hours per day. This is so helpful.

I rarely turn *off* the extension, but I very frequently toggle it from "Dark" to "Light" modes. Can you add a checkbox/radio element to the "Automation" controls that controls whether "Automation turns on/off extension" or "Automation toggles Dark/Light Mode"?

Thank you!!!

### LINKED ISSUE #2102: Allow an option to switch dark/light themes based on time period.
The idea of having "active hours" in DarkReader is great, but having an option to switch themes for the rest of the time would be a greater idea. Like in the morning, you can have some sort of Sepia theme, and at the night, you can have a darker variant, or you can add some sort of "In this time period, I want a light theme, in this time period, I want a darker theme, in this period, I want completely black theme.", you get the idea.

### LINKED ISSUE #3243: [Feature Request] Active different global themes based on time (automation)
Allow the user to specify different dark themes, which can be automatically activated based on current time.

Gusted:
- Add option to switch On/Off or Scheme for when automation is effected.
- Resolve #1713
- Resolves #3243
- Resolves #2102


harshil15999:
Hey I would like to contribute to this feature @Gusted

Gusted:
Their isn't really to contribute to at this moment, as It could be merged right now(after patching up the conflicts) and be removed in a few weeks as I've mentioned a redesign which more or less will only use the current elements and move around some settings(so this will element will need a location as well etc).

MahdiNazemi:
@Gusted Are there any technical issues preventing this PR from being merged?

Gusted:
> 

Well, not really. In the spirit of Christmas let me merge it.

src/background/extension.ts
```diff
@@ -380,7 +380,23 @@ export class Extension {
...
             return;
         }
-        this.handleAutoCheck();
+        this.handleAutomationCheck();
+    };
+
+    private handleAutomationCheck = () => {
+        if (this.user.settings.automationBehaviour === 'Scheme') {
+            this.recalculateIsEnabled();
+            if (this.isEnabled) {
```
alexanderby, line 389: It doesn't seem to work, because when [the tab message is formed](https://github.com/darkreader/darkreader/blob/master/src/background/extension.ts#L567) and `this.isEnabled` is `false`, then the message to disable the app is formed. To my opinion `isEnabled` should show is the extension is on/off, and the theme.mode should be substituted during the forming of the tab message, instead of writing the color scheme into storage.

MahdiNazemi:
I have been testing this feature in the past few days, and it works great. The only issue is that PDFs are shown with a black background regardless of whether it is day or night. The expected behavior is to see the regular PDF file during the day and inverted colors at night.
P.S. I use the [Brave browser](https://brave.com/).

Gusted:
> The only issue is that PDFs are shown with a black background regardless of whether it is day or night. The expected behavior is to see the regular PDF file during the day and inverted colors at night.

PDF's are currently independent of the current theme.

Gusted:
> I think the previous behavior was to see PDFs with a white background in the morning and a black background at night. With this new version (and automation set to light/dark), the background is black all the time.
> If I revert automation back to on/off, the issue will be resolved.

Yeah, that seems like the correct behavior to me, ON/OFF is the global state of the extension and the scheme dark/light only impacts which color-scheme should be used.

PR #2869, opened 2020-06-15
invert bright home page image for wikidata.org

Discostu36:
dark mode for wikidata.org is quite good but the bright big header image of the homepage needs to be inverted

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4597,10 +4597,8 @@ INVERT
 
 wikidata.org
 
-CSS
-.wd-mp-headerimage {
-    filter: invert(100%);
-}
+INVERT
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -4595,6 +4595,15 @@ INVERT
 
 ================================
 
+wikidata.org
+
+CSS
+.wd-mp-headerimage {
```
Gusted, line 4601:
```
wikidata.org

INVERT
.wd-mp-headerimage

```
We prefer this so it will have a `hue-rotate(180deg)` So colors will be kind of 'originial'

Discostu36: Thanks, changed it

PR #2887, opened 2020-06-19
Remove discord sites from dark-sites

werymanen:
Because on some sites it is not actually dark
For example:
https://blog.discord.com
https://status.discord.com/
https://support.discord.com/hc/en-us

and more


Gusted:
#2517 already does this on the new correct way

werymanen:
Ok, sorry, didn't see that..

PR #2902, opened 2020-06-21
Update dark-sites.config

Nereuxofficial:
Added doc.rust-lang.org as it already supports multiple dark modes


Nereuxofficial:
It has to be noted that when you visit the site it will not be dark-themed but upon entering the books(where you will spend most of the time with this website) you will find pretty good different dark themes

Gusted:
Yeah not by default....

Nereuxofficial:
So 
https://doc.rust-lang.org/book/
and
https://doc.rust-lang.org/stable/rust-by-example/
?

PR #2903, opened 2020-06-21
Update dark-sites.config

Nereuxofficial:
Those two sites already have pretty good dark modes


Gusted:
The say is only for sites that have it **By Default**

PR #2934, opened 2020-06-24
Pre-Release and fixing bugs

### LINKED ISSUE #2660: [BUG] Toggle status is ignored if query string contains square braces
**Describe the bug**
If you navigate to a page you have disabled DarkReader for (via the toggle) and the URI for that page contains a `[` (in the query string) then the rule you have set will be ignored.  
When the braces are encoded this issue does not occur.

**To Reproduce**
1. Navigate to `google.co.uk`
1. Disable DarkReader via the quick toggle
1. Notice DarkReader isn't applied
1. Navigate to `google.co.uk?foo=[bar]`
1. Notice DarkReader is applied

**Expected behavior**
DarkReader should not be applied when it is toggled off

**Screenshots**
image
image

**System Information:**
 - MacOS 10.15.4
 - Chrome 83.0.4103.61

**Additional context**
Note: Between my first and second screenshot, DarkReader was **not** toggle manually

Gusted:
- Resolves #2660

### DIFF
--- a/src/utils/url.ts
+++ b/src/utils/url.ts
@@ -31,7 +31,7 @@ export function isURLInList(url: string, list: string[]) {
 export function isURLMatched(url: string, urlTemplate: string): boolean {
     const isFirstIPV6 = url.includes('[');
     const isSecondIPV6 = urlTemplate.includes('[');
-    if (isFirstIPV6 && isSecondIPV6) {
+    if (isFirstIPV6 && isSecondIPV6 && !(url.indexOf('?') < url.indexOf('['))) {
         return compareIPV6(url, urlTemplate);
     } else if (!isFirstIPV6 && ! isSecondIPV6){
         const regex = createUrlRegex(urlTemplate);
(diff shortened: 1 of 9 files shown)


src/utils/url.ts
```diff
@@ -31,9 +31,14 @@ export function isURLInList(url: string, list: string[]) {
 export function isURLMatched(url: string, urlTemplate: string): boolean {
     const isFirstIPV6 = url.includes('[');
```
alexanderby, line 32: `isFirstIPV6` and `isSecondIPV6` should be true, if they are IPv6. So, create a `function isIPv6(url: string): boolean`, other code should stay the same.

tests/url.tests.ts
```diff
@@ -167,6 +167,14 @@ test('URL is enabled', () => {
         {isProtected: false, isInDarkList: true},
     )).toEqual(false);
 
+    // Because some people who don't know about XSS still don't encode their URL's and have vulnerable [] in them sigh.
```
alexanderby, line 170: Shorten this comment just to note that unescaped `[]` can appear in URL query.

PR #2945, opened 2020-06-25
Fixed api.kde.org

Rolodophone:


src/config/dynamic-theme-fixes.config
```diff
@@ -272,6 +272,19 @@ CSS
 
 ================================
 
+api.kde.org
+
+INVERT
+.center
+.image
```
Gusted, line 279: Unable to find both selectors

src/config/dynamic-theme-fixes.config
```diff
@@ -272,6 +272,19 @@ CSS
...
+
+CSS
+.memdoc, .memproto {
```
Gusted, line 282: Unable to find as well...

Rolodophone:
Oh yeah sorry didn't realise this only applies to only some pages. As far as I've found it should apply to all pages under https://api.kde.org/appscomplete-api, https://api.kde.org/4.14-api, and https://api.kde.org/extragear-api. 

For example, https://api.kde.org/appscomplete-api/krita-apidocs/libs/libkis/html/classKrita.html, https://api.kde.org/4.14-api/kdelibs-apidocs/interfaces/kspeech/html/classKSpeech.html, and https://api.kde.org/extragear-api/graphics-apidocs/krita/libs/libkis/html/classKrita.html.

I'm not sure how to only make the rules apply to certain subpages of a website.

PR #2961, opened 2020-06-26
Update dark-sites.config

zoranzoki21:
Added my site kizule.xyz which supports dark mode.


sprinkledcheese:
Not dark by default.

PR #2964, opened 2020-06-27
Fix for Fontspring.com

LesterCovax:
Includes dynamic and filter/filter+ fixes
Properly renders the font preview images as white instead of black


src/config/dynamic-theme-fixes.config
```diff
@@ -1552,6 +1552,12 @@ INVERT
 .fiu-header__branding
 .fiu-sample-list__item img
 .fiu-gallery-head__text img
+================================
```
LesterCovax, line 1555: I accidentally an entire `/n` 🤦‍♂️ 

```suggestion

================================
```

PR #2967, opened 2020-06-28
Update dom.ts

Gusted:
Open github & Darkreader 4.9.11
Poof errors.


src/inject/utils/dom.ts
```diff
@@ -148,6 +148,9 @@ export function watchForNodePosition<T extends Node>(
 }
 
 export function iterateShadowNodes(root: Node, iterator: (node: Element) => void) {
+    if (root == null) {
```
alexanderby, line 151: This will fix in one place, but as I see it will break in the next place. Let's put this check here https://github.com/darkreader/darkreader/blob/master/src/inject/dynamic-theme/watch.ts#L208

PR #2988, opened 2020-07-01
Negative pattern sites and pages

sprinkledcheese:


sprinkledcheese:
OK, I'll remove it.

src/config/dark-sites.config
```diff
@@ -1,3 +1,14 @@
...
+!acs.msi.com
+!epicgames.com/store/*/about
+!epicgames.com/store/*/eula
+!eps.msi.com
+!ipc.msi.com
+!msi.com/case-studies
+!msi.com/language
+!msi.com/page/warranty
+!msi.com/support
+!rewards.msi.com
```
Yash-Singh1, line 11:
```suggestion
!rewards.msi.com
!status.jsfiddle.net
```

Yash-Singh1 (changes requested):
I think `status.jsfiddle.net` needs to be here

PR #3028, opened 2020-07-05
Grubhub

kimpurcell:
Popup text were not visible, made them white.


src/config/dynamic-theme-fixes.config
```diff
@@ -1911,6 +1911,10 @@ CSS
     color: rgb(24, 26, 27) !important;
 }
 
+label, h5, h6, header, .h5 {
+    color:white !important;
```
Gusted, line 1915: `color: ${black}`

PR #3063, opened 2020-07-10
Invert SVG figures at feynman lectures

LordMikey:
Invert SVG figures at feynmanlectures.caltech.edu

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1563,10 +1563,8 @@ CSS
 
 feynmanlectures.caltech.edu
 
-CSS
-div.figure > img {
-    background-color: white;
-}
+INVERT
+.figure > img
 
 ================================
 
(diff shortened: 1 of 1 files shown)


Gusted:
Example of a SVG? As your now also targetting normal images as well that don't need a invert at all.

LordMikey:
Example page:

https://www.feynmanlectures.caltech.edu/I_23.html

As far as I know all the images are SVG, but they appear in  tags

src/config/dynamic-theme-fixes.config
```diff
@@ -1561,6 +1561,13 @@ CSS
 
 ================================
 
+feynmanlectures.caltech.edu
+
+INVERT
+.figure > img
```
Gusted, line 1567: `.figure > img[src$=".svgz"]`
Have a look how that looks. This will hopefull only target the math images.

LordMikey: I checked about 25 chapters and it looks good.

PR #3084, opened 2020-07-11
Fix for google.*

Rakleed:
Examples: https://www.google.com, https://www.google.com/search?q=dark+reader

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5198,7 +5198,7 @@ INVERT
 .gb_0b
 #dictionary-modules img[src*="png"]
 a.gb_b > div
-.gb_zc
+a[href*="about/products"][title]
 .JOmIqc
 .hLcKi
 #EcMbV
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5198,7 +5198,7 @@ INVERT
 .gb_0b
 #dictionary-modules img[src*="png"]
 a.gb_b > div
-a[href*="about/products"]
+.gb_zc
 .JOmIqc
 .hLcKi
 #EcMbV


Rakleed:
And why in the address of the site at the beginning of the site is www, if without this everything works correctly?

Gusted:
And why changing it to a changing CSS selector? This will become a cat and mouse game with Google to stay up to date with this selector.

Gusted:
Poof it's already changed again. That's why we use a CSS3 selector.

Rakleed:
The rule `a[href*="about/products"]` inverts the icon on the site with the search result, although it does not need to be done there (for example, https://www.google.com/search?q=dark+reader).

Gusted:
I see... That's wow Didn't even noticed.

src/config/dynamic-theme-fixes.config
```diff
@@ -5198,7 +5198,7 @@ INVERT
 .gb_0b
 #dictionary-modules img[src*="png"]
 a.gb_b > div
-a[href*="about/products"]
```
Gusted, line 5201: `a[href*="about/products"][title]` The google main page has a title the search one has aria-label. We should preserve this selector as going down to the normal class selector will be litterly cat and mouse.

Rakleed:
> I see... That's wow Didn't even noticed.

It's sarcasm? After all, this is incorrect inversion.

> 
> Poof it's already changed again. That's why we use a CSS3 selector.

I still have the same name.
image

Gusted:
Google and dynamic classes :)

No it wasn't sarcasm I never knew that the 9 dots on search one was already bright enough.

src/config/dynamic-theme-fixes.config
Rakleed: Previously done by name (for example, `a[title="Google apps"]`, but in different language versions of the search this name will be in a different language.

Gusted: That's why we don't add the ="" But simple the `[title]` That way it just check if it exist.

Rakleed:
@Gusted And what's about?
> And why in the address of the site at the beginning of the site is www, if without this everything works correctly?

Gusted:
Sub Domains :)
docs.google.com
drive.google.com
etc etc etc

PR #3091, opened 2020-07-12
I88n patch

Gusted:
New UI :D


Gusted:
Failing as somehow we need the translations for the other languages as well.

PR #3093, opened 2020-07-12
Partial fix for Microsoft main logo

GalaxySH:
Microsoft Logo
[microsoft.com](https://microsoft.com)
Inverting the logo kind of works, but the background is black, not gray like the rest of the page (and it is a PNG, so I don't know how to fix background with CSS).

**Also,** when using light mode the image is now broken again, but the opposite of what it was before. I don't know how to fix this, it would be nice to just remove the background.

**Also also,** it seems like people have attempted to fix the Microsoft logo at least three times before, but they keep updating the image version. This is more of a temporary fix, I might see if I can do a more complex selector later.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2827,11 +2827,11 @@ img[src*="img-prod-cms-rt-microsoft-com.akamaized.net/cms/api/am/imageFileData/R
 img[src*="img-prod-cms-rt-microsoft-com.akamaized.net/cms/api/am/imageFileData/RE4pxBu"]
 img[src*="img-prod-cms-rt-microsoft-com.akamaized.net/cms/api/am/imageFileData/RE4pkvE"]
 img[src*="/static/images/inreplyto.svg"]
-img[src*="img-prod-cms-rt-microsoft-com.akamaized.net/cms/api/am/imageFileData/RE1Mu3b"]
+.clogo .cimage
 
 CSS
 .c-logo .c-image {
-    background: ${black} !important;
+    background: transparent !important;
 }
 
 ================================
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -665,8 +665,8 @@ cleantechnica.com
 
 CSS
 body {
-    background-image: none !important; 
-	background-color: #1d1e1f !important; 
+    background-image: none !important;
+	background-color: #1d1e1f !important;
 }
 
 ================================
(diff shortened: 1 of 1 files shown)


GalaxySH:
Funny how DR fixes the background on GitHub

src/config/dynamic-theme-fixes.config
```diff
@@ -2827,6 +2827,7 @@ img[src*="img-prod-cms-rt-microsoft-com.akamaized.net/cms/api/am/imageFileData/R
 img[src*="img-prod-cms-rt-microsoft-com.akamaized.net/cms/api/am/imageFileData/RE4pxBu"]
 img[src*="img-prod-cms-rt-microsoft-com.akamaized.net/cms/api/am/imageFileData/RE4pkvE"]
 img[src*="/static/images/inreplyto.svg"]
+img[src*="img-prod-cms-rt-microsoft-com.akamaized.net/cms/api/am/imageFileData/RE1Mu3b"]
```
Gusted, line 2830: Change it to `.c-logo .c-image`

and 
```
CSS
.c-logo .c-image {
    background: transparent !important;
}
```

sprinkledcheese: For fixing image on light mode use, **white** instead of **${black}**. 
Inverting the logo isn't a good idea since it makes **Microsoft** hard to read and colours are changed drastically.

sprinkledcheese:
Add this to INVERT: **img[src*="img-prod-cms-rt-microsoft-com.akamaized.net/cms/api/am/imageFileData/RE4pndL"]**.
It inverts **Shop for Surface Devices logo**

src/config/dynamic-theme-fixes.config
Gusted: Or we use the new Variable #2767 That I specially made to get the correct background/foreground as the ${} don't get parsed correctly due how it should work :/

Gusted:
I have a feeling that `RE4pndL` could be dynamicly generated and could be changed for a new website update.

sprinkledcheese:
@Gusted No it's not dynamically generated, the other icons still have the same thing.

Gusted:
Okay than we just have to update the ${} to var(--darkreader-neutral-background)

src/config/dynamic-theme-fixes.config
GalaxySH: But if it's a solid image how does changing the background fix it?

GalaxySH: I will add then @Gusted

GalaxySH:
> 
> 
> I have a feeling that `RE4pndL` could be dynamicly generated and could be changed for a new website update.

That is what I suggested

> 
> Also also, it seems like people have attempted to fix the Microsoft logo at least three times before, but they keep updating the image version. This is more of a temporary fix, I might see if I can do a more complex selector later.

PR #3106, opened 2020-07-13
Add remaing API Documentation

Gusted:


index.d.ts
```diff
@@ -73,6 +73,36 @@ declare namespace DarkReader {
          * Default 0.
          */
         textStroke: number;
+        /**
+         * Background color to use while dark mode.
```
alexanderby, line 77: Replace `while` with `for`.

PR #3115, opened 2020-07-14
Add documentation for the new introduced variables

Gusted:
- Added documentation to help understand how the variables should be used var(). 
- Added documention to explain what the variables does and mean.

https://github.com/darkreader/darkreader/blob/README-Patch/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,8 +87,9 @@ For **Dynamic mode** use `INVERT` only for dark images that are invisible on dar
 
 **Dynamic variables**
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -108,7 +144,6 @@ CSS
 }
 ```
 
-**Fixes for Filter and Filter+ modes**.
```
Gusted, line 111: Had been moved to line 122 so it will be understandable and more clearly sections.

PR #3119, opened 2020-07-15
blogs.windows.com

sprinkledcheese:
- Logo

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -571,7 +571,7 @@ blogs.windows.com
 
 CSS
 .c-logo .c-image {
-    background: var(--darkreader-neutral-background) !important;
+    background: white !important;
 }
 
 ================================
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -571,7 +571,7 @@ blogs.windows.com
 
 CSS
 .c-logo .c-image {
-    background: white !important;
+    background: ${white} !important;
 }
 
 ================================


sprinkledcheese:
--darkreader-neutral-background doesn't work since it inherits from the background. Only solution for ms logo is white background.

Gusted:
It the value that is used to set the background do the background color will automatically be updated when the user have a custom background color

Gusted:
Btw do you use stock darkreader if so the variables aren't implemented yet for it.

sprinkledcheese:
I'm on master build.

Gusted:
Than the variable should work and dynamically update to the correct background.

sprinkledcheese:
It just inherits the background color which the user has set, right? If I have dark background the logo will be just like before.

sprinkledcheese:
It should look like this: chrome_nEJslYoNXa

Gusted:
:/ it's tricky with Microsoft. Having background will improve visibility but makes it bit messy to have a random white background their.

sprinkledcheese:
Shouldn't the variable work only for background's that should be darker instead of light. I don't see any other use case.

sprinkledcheese:
If one uses ${} it messes in the other mode. If colour is used without ${}, the other mode doesn't have any issues.

Gusted:
My idea was generally for people with custom themes and certain fixes that the background color of those fixes are dark or white or whatever and don't fit into the custom background color. That was my thoughts with implementating it.

sprinkledcheese:
Ah nice. But for certain logos there should an exception.

alexanderby:
Try putting this selector into `INVERT` section

alexanderby:
...wait, it had to be `background: ${white} !important;`

sprinkledcheese:
transparent doesn't work. ${white} works in both modes, ${black} makes the logo black in light mode. In light mode the colour is inverted from the dark mode, is that expected behaviour. Shouldn't it just invert colour regardless of mode?

PR #3123, opened 2020-07-16
added bitwit.tech

richardfxr:
bitwit.tech now has a dark theme that utilizes the CSS prefers-color-scheme media query.


Gusted:
If it uses the prefers color scheme than it will be light for people who didn't set their OS to dark mode.

PR #3125, opened 2020-07-16
Fix Styling for virustotal

id0ntbyte:
There was an issue with dark mode of Virustotal where all the cards were shown as white. This change will take care of that issue.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4847,6 +4847,15 @@ body {
 
 ================================
 
+virustotal.com
+
+CSS
+vt-ui-file-card{
+    background-color: {black} !important;
+}
+
+================================
+
 vrt.be/vrtnws
 
 INVERT


Gusted:
Did you even tried this fix as I can see it will not even work.

sprinkledcheese:
I don't know why styling doesn't work properly on virustotal but if you apply any rule to the dev tools and apply 1-2 or maybe more times then style will be applied correctly. So, I think that 'retry PR' will fix virustotal @Gusted.

sprinkledcheese:
So I checked again and it's because of lazy-loading.
chrome_syAhXVXoju

Gusted:
Oh no it's a intresting bug with Variables and Shadow-roots.

id0ntbyte:
I've noticed if the styling is added after the website fully loads or with a few seconds delay, then it shows perfectly fine.

Gusted:
Yeah we got #4682 for tracking the main problem.

PR #3132, opened 2020-07-16
Workflow Patch

Gusted:
Rather effective to have this than going to branch and rebuilding it just to check/give to a person to check their issue/bug has been resolved.


Gusted:
it's running npm run test not sure why not npm run release :shrug:

PR #3133, opened 2020-07-16
added fix for bitwit.tech into dynamic-theme-fixes.config

richardfxr:
Added some CSS to fix the SVG illustrations on the site.  The site does have global CSS variables for colors but Dark Reader's default colors override them even when I use !important in the Dev tools. However, Dark Reader does a good enough job that this really isn't an issue.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -528,6 +528,54 @@ a[href="/product"]
 
 ================================
 
+bitwit.tech
+
+CSS
+.svg-primary {
+    fill: #375D69 !important;
+}
+.svg-primary-light {
+    fill: #4A7F8F !important;
+}
+.svg-primary-dark {
+    fill: #28444D !important;
+}
+.svg-secondary {
+    fill: #B88399 !important;
+}
+.svg-secondary-light {
+    fill: #E8A9C4 !important;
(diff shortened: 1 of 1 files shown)


Gusted:
I think it's just better to add it to the global dark sites. as it's already default by dark.

richardfxr:
Ah, I made the same mistake yesterday. The site actually used the prefers-color-scheme media query to activate it's dark mode. Sorry I didn't mention that here.

Gusted:
🤦 Forgot about it.

src/config/dynamic-theme-fixes.config
```diff
@@ -528,6 +528,54 @@ a[href="/product"]
...
+.svg-light {
+    fill: #ADCED9 !important;
+}
+.svg-outline {
+    fill: none !important;
+    stroke: #000000 !important;
+}
+.svg-primary-outline {
+    fill: none !important;
+    stroke: #ADCED9 !important;
```
Gusted, line 561: Have a look with `stroke: ${originial value} !important;`

richardfxr: The stroke color changes depending on prefers-color-scheme. I wanted the style to be consistent no matter what, so that's why I specified the values here.

PR #3131, opened 2020-07-16
Myshor wanted this

Gusted:
For any info contact @Myshor

PR #3139, opened 2020-07-18
fix for facebook.com

Myshor:
Big Blob breaking fixes.
Sorry no screens - it was too much.


Myshor:
With new functionality #3131 I have one class to exclude from blobs so I will refactor all this blobs fixes when it will be included in master and released. ;)

PR #3143, opened 2020-07-19
Fixed streamable.com

sprinkledcheese:
- Logos
- Svgs

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4150,6 +4150,9 @@ CSS
 streamable.com
 
 INVERT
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -4147,6 +4147,16 @@ CSS
 
 ================================
 
+streamable.com
+
+INVERT
+.landing .features-pane .feature span svg
```
Gusted, line 4153: Just to be sure with element you can use >
`.feature > span > svg`
You can do the same for img on line 4155.as elements are risky when using in this way.

PR #3161, opened 2020-07-19
fix for facebook.com

Myshor:
Next blobs in group.


Gusted:
Your busy with Blobs :P

Issue #1071, opened 2019-02-02
Bypassing cookie rules in Firefox


zng02:
DarkReader seems to be causing Firefox to save cookies when it shouldn't be able to. I had noticed that cookies for websites were staying on my system despite privacy settings, and disabling all add-ons fixed the problem. Re-enabling them one at a time, I found that DarkReader was the one allowing the cookies to be saved. I can verify this is happening on both Firefox 65.0 and Firefox Nightly, both on Ubuntu 64bit.

To replicate this bug, go into private browsing and use a cookie testing website like [html-kit ](http://www.html-kit.com/tools/cookietester/) or [whatarecookies](http://www.whatarecookies.com/cookietest.asp) . Private browsing shouldn't be able to save cookies normally, but will do so if DarkReader is enabled.

alexanderby:
Didn't notice anything like this and cannot imagine how this could be possible. Was not able to reproduce on specified websites.
<img width="540" alt="screen shot 2019-02-04 at 19 06 51" src="https://user-images.githubusercontent.com/9638588/52220596-55c77180-28b0-11e9-8742-23fafed85f64.png">

zng02:
Oddly enough, after re-enabling Dark Reader to test again, I can't replicate the bug either. For reference, the bug was encountered with Firefox Content Blocking set to Strict, not Custom with cookies fully disabled. However, it seems this was a weird problem isolated to my system. Sorry about the confusion, and thank you for the quick response.

zng02:
I apologize for the confusion, but I have done more testing with my extensions and tracked the cookie leaking issue back to Dark Reader. The bug seems to be inconsistent, which is why I couldn't replicate it before and closed the issue. It doesn't happen all the time when Dark Reader is enabled, but it never happens without it. This made identifying the culprit difficult, but I can verify now that it is Dark Reader.

To be clear, it does not seem to bypass custom Firefox cookie rules to disable all cookies, but it allows cookies from Private Browsing to persist if cookies are enabled in any way. This has been most thoroughly tested using custom content blocking rules set to block all third-party cookies. The "Do Not Track" signal is also enabled, and trackers are blocked in all windows using the level 1 block list.

This has been tested on Firefox 66.0.1, running on 64 bit Ubuntu 18.10. I tested this bug by disabling all extensions, then testing them one at a time over a long period of time.

Gusted:
I don't experience this bug. Also we don't have permissions to cookies so if this still persist it can't be dark reader.

PR #3162, opened 2020-07-20
Fix dynamic style bug for Github contribution display

E011011101001:
Colors of contribution icons do not seem good.

Using the latest version 4.9.15, it looks like this. Colors are hard to distinguish and are not corresponding with colors on the axis legend bar below.
Before Fix

And I find out it's because CSS selectors are trying to catch the wrong colors in the dynamic config for github.com. (Maybe those colors were right at some time, but Github changed them later on.)
Anyway, I replaced them with colors that I found is applying to the icons. After the fix, the contribution displayer looks like the snapshot below.
After Fix

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1903,21 +1903,21 @@ CSS
     fill: ${#ebedf0} !important;
     background-color: ${#ebedf0} !important;
 }
-.day[fill="#c6e48b"], .legend li[style*="background-color: rgb(198, 228, 139)"] {
-    fill: ${#c6d4cb} !important;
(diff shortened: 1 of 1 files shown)


Gusted:
Could have a look at IGNORE INLINE STYLE in README.md and make your fix according to it thst should prevent if the colors change again.

E011011101001:
I'm not sure how `IGNORE INLINE STYLE` could work on tags with no inline style. On the other hand, what the previous author did (#1071) is to avoid other bugs when Darkreader left the SVG rectangles out (#558).

I think it's hard without JavaScript to prepare for the possibility that Github changes its color again.

Gusted:
Agreedd....
 Pain in my eyes.

PR #3177, opened 2020-07-20
Fix for reddit.com

### LINKED ISSUE #669: Static, Dynamic: Reddit spoiler text is visible
[Example.](https://old.reddit.com/r/stevenuniverse/comments/948673/can_you_guess_which_su_episode_is_has_the_highest/)

Static:
default
Dynamic:
default

Gusted:
- Simulate spoiler behaviour
- Resolves #669


Gusted:
Not using important as that will overide the hover behaviour.

PR #3180, opened 2020-07-20
blob fixes for facebook.com

Myshor:
...


Gusted:
You know they will become useless ;)

Myshor:
No they won't... and they will be good start for refactor with new IGNORE FUNCTION 😝

PR #3182, opened 2020-07-20
Add IGNORE IMAGE ANALYSIS documentation

Gusted:
- Add explaination what IGNORE INMAGE ANALYSIS does

TODO: Fix grammer(I think so), Add more explaination.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ For **Dynamic mode** use `INVERT` only for dark images that are invisible on dar
 `!important` keyword should be specified for each CSS property to prevent overrides by other stylesheets.
 **Dynamic mode** supports `${COLOR}` template, where `COLOR` is a color value before the inversion (`white` will become `black` in dark mode).
 - `IGNORE INLINE STYLE` rule will not analyze the inline style of matched selectors. (e.g. `--darkreader-inline-bgcolor` or `<p style='color: red'>This will not be analyzed by Dark Reader if it has a matched selector</p>`)
-- `IGNORE IMAGE ANALYSIS` rule will not analyze background-image's of matched selectors.
+- `IGNORE IMAGE ANALYSIS` rule will not any images of matched selectors.
 
 **Dynamic variables**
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -77,13 +77,17 @@ CSS
...
+IGNORE IMAGE ANALYSIS
+.logo
 ```
 - `INVERT` rule inverts specified elements.
 For **Dynamic mode** use `INVERT` only for dark images that are invisible on dark backgrounds (icons, diagrams, charts, `<img>` and `<svg>` elements).
 - `CSS` rule adds custom CSS to a web page.
 `!important` keyword should be specified for each CSS property to prevent overrides by other stylesheets.
 **Dynamic mode** supports `${COLOR}` template, where `COLOR` is a color value before the inversion (`white` will become `black` in dark mode).
 - `IGNORE INLINE STYLE` rule will not analyze the inline style of matched selectors. (e.g. `--darkreader-inline-bgcolor` or `<p style='color: red'>This will not be analyzed by Dark Reader if it has a matched selector</p>`)
+- `IGNORE IMAGE ANALYSIS` rule will not analyze background-image's of matched selectors.
```
sprinkledcheese, line 90: just **images**

sprinkledcheese:
You should add some examples too, like fast.com's blob is fixed. Also nextdns.io logo is fixed.  This only works on the exact selector right? It doesn't work if I use the parent selector?

Gusted:
It should behave as a rule and not as a element so no just exact selectors.  The example is for how you could use it so should it behave as a class or as CSS or list of selectors etc. 0

README.md
```diff
@@ -87,7 +87,7 @@ For **Dynamic mode** use `INVERT` only for dark images that are invisible on dar
 `!important` keyword should be specified for each CSS property to prevent overrides by other stylesheets.
 **Dynamic mode** supports `${COLOR}` template, where `COLOR` is a color value before the inversion (`white` will become `black` in dark mode).
 - `IGNORE INLINE STYLE` rule will not analyze the inline style of matched selectors. (e.g. `--darkreader-inline-bgcolor` or `<p style='color: red'>This will not be analyzed by Dark Reader if it has a matched selector</p>`)
-- `IGNORE IMAGE ANALYSIS` rule will not analyze background-image's of matched selectors.
+- `IGNORE IMAGE ANALYSIS` rule will not any images of matched selectors.
```
sprinkledcheese, line 90: **analyze** 👍

PR #3187, opened 2020-07-21
Fix for blobs in facebook.com

Myshor:


Myshor:
I bet I'll find something more today so leaving on a branch to the end of the day. :)

PR #3198, opened 2020-07-22
Fixed color of recommended comments on theverge.com articles

Mertyn:
theverge


Gusted:
I think this is more of a personal preference as Dark Reader dynamic config should be considerd for fixes.

Mertyn:
Oh okay, I thought the goal was to kind of make websites more black and gray for the most part :D

Gusted:
We have grayscale filter for that ;) Dynamic should darken the background colors/Lighten text colors.

Issue #3215, opened 2020-07-23
Dynamic Theme css rule invert non-map related images on facebook


holysnipz:
*Not sure how to call this kind of issue properly in English, feel free to edit the title :)*

**Website**
facebook.com

**How to Reproduce**
> 1. Set Dark Reader's theme mode to Dynamic
> 2. Go view the problematic post [here](https://www.facebook.com/AComradeOfMine.II/posts/2681273225471113)
*(Not sure if related, but turned back to Old Design, or just open private tab)*
> 3. If facebook's system doesn't change the url of that image yet, you'll see the image got inverted

**Expected behavior**
Image shows normally without being inverted, since it's not a map (doesn't came from `static_map.php` or `map_tile.php`)

**Actual behavior**
It did inverted, because there the word **`map`** inside the url *(see image below)*

**Screenshots**
![image](https://user-images.githubusercontent.com/9103583/88318560-1a8b5d80-cd45-11ea-8b05-8832097911ad.png)

**System info:**
 - OS: Windows
 - Browser: Firefox 78.0.2 / Chrome 84.0.4147.89
 - Darkreader Version: 4.9.15

Myshor:
Easy fix to include only fb maps. Will check how it works with other maps in posts before merging. :)

holysnipz:
no problems with map in posts or full map view so far. but it remove the inverted filter from the small map under 'about' in pages. which i do not mind, but somebody else might does.


![image](https://user-images.githubusercontent.com/9103583/88331327-fdf82100-cd56-11ea-9319-e3e26848d7b6.png)

Gusted:
Should be fix-able if another selector for `map.php` will be added.

PR #3216, opened 2020-07-23
fix for facebook.com #3215

Myshor:
fix for images on facebook containing "map" in url
fix for #3215


Myshor:
Reverted fix in style as it fixes maps in events.

PR #3218, opened 2020-07-23
Toggle Icon based on site also.

### LINKED ISSUE #1671: Show status of current-site toggle in the icon
This request is an extension of https://github.com/darkreader/darkreader/issues/609. It is nice that the current-site toggle value is visible when I click on the icon, but I would like for it to be visible without having to click on the icon. 

Perhaps the goggles could change color or something.

### LINKED ISSUE #2620: Turn off the glow on glasses when dark reader in inactive.
[Feature Request] when on pages that are protected by the browser or are in the global dark list, can glasses on the dark reader icon not glow?

Gusted:
- Set the correct icon based off Site status.
- Resolves #2620
- Resolves #1671

### DIFF
--- a/src/background/extension.ts
+++ b/src/background/extension.ts
@@ -371,6 +371,7 @@ export class Extension {
     private getTabMessage = (url: string, frameURL: string) => {
         const urlInfo = this.getURLInfo(url);
         if (this.isEnabled() && isURLEnabled(url, this.user.settings, urlInfo)) {
+            this.icon.setActive();
             const custom = this.user.settings.customThemes.find(({url: urlList}) => isURLInList(url, urlList));
             const preset = custom ? null : this.user.settings.presets.find(({urls}) => isURLInList(url, urls));
             const theme = custom ? custom.theme : preset ? preset.theme : this.user.settings.theme;
@@ -422,6 +423,7 @@ export class Extension {
                 }
             }
         } else {
+            this.icon.setInactive();
             console.log(`Site is not inverted: ${url}`);
         }
         return {


jtrakk:
It would be nice if the current-site-toggled icon were different from the global-toggled icon so I know which button I need to press.

Gusted:
Not going to happend, we use the current icons to not get users confused.

jtrakk:
If there are three states (all enabled, all disabled, current disabled), but only two icon states, that is confusing and hard to understand which state I'm in. For example if the third icon "global enabled, local disabled" is half-light/half-dark, it would be clear.

AntonioRedondo:
Any idea when this is going to be merged? I'm really looking forward to this change to be merged because I find confusing the current icon approach, which will always be Dark Vader no matter Dark Reader is on or off for a specific site.

I'm happy with this 2-state approach with only active and inactive states.

Alternatively we could have a 3-state approach:

image

This is the same colour scheme used by uBlock Origin:

image

Gusted:
@alexanderby Master of the SVG and Inker. Mind have a look at the icons.

pdf:
You could also keep using just the current icons, but add an asterisk as a badge when the current site either has custom settings, or darkreader is disabled for the site, via [browserAction.setBadgeText()](https://developer.chrome.com/extensions/browserAction#method-setBadgeText) and [browserAction.setBadgeBackgroundColor()](https://developer.chrome.com/extensions/browserAction#method-setBadgeBackgroundColor).

So if the site is in dark mode with custom settings, add a badge and set the dark icon, and if the site has dark mode disabled, add a badge and set the light icon. Default is to set the badge to empty.

Gusted:
Not merging - Accurately detecting this requires a new "invasive" permission.

iamkroot:
> Not merging - Accurately detecting this requires a new "invasive" permission.

What permission is this? It seems that the extension already has the `tabs` permission.

bershanskiy:
> > Not merging - Accurately detecting this requires a new "invasive" permission.
> 
> What permission is this? It seems that the extension already has the `tabs` permission.

I'm trying to get rid of `tabs` permission, but there might be a way to do it without it. I'm not actively working on it, however.

Juraj-Masiar:
There may be some confusion here. 
You don't need `tabs` permission to use the `browser.browserAction.setIcon` API.
I use this API in some of my addons without having `tabs`, so I'm 100% sure it works. 
Also [docs](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/browserAction/setIcon) doesn't mention it.

PR #3222, opened 2020-07-24
fix for facebook.com

Myshor:
After yesterday fix for maps mapy.cz was not inverted.
Added fix for them. ;)

obraz


Myshor:
Added Endomondo training posts map darkening. ;)
obraz

PR #3238, opened 2020-07-26
Improve testing locales

wojtekmaj:
Improves the feedback from locales.test.js.

Before:

Something wrong... Somewhere 🤔 

Zrzut ekranu 2020-07-26 o 12 42 40

After:

Oh! These locales are wrong.

Zrzut ekranu 2020-07-26 o 12 57 29

But what precisely?

obraz

### DIFF
--- a/tests/config/locales.tests.ts
+++ b/tests/config/locales.tests.ts
@@ -1,65 +1,46 @@
-import {readFile as fsReadFile, readdir as fsReadDir} from 'fs';
+import {readFileSync as fsReadFile, readdirSync as fsReadDir} from 'fs';
 import {resolve as resolvePath} from 'path';
 
-function readDir(dir) {
-    return new Promise<string[]>((resolve, reject) => {
-        fsReadDir(resolvePath(__dirname, dir), (err, files) => {
-            if (err) {
-                reject(err);
-                return;
-            }
-            resolve(files);
-        });
-    });
+function readDir(dir: string) {
+    return fsReadDir(resolvePath(__dirname, dir));
 }
 
-function readLocale(name) {
-    return new Promise<string>((resolve, reject) => {
-        fsReadFile(resolvePath(__dirname, '../../src/_locales/', name), {encoding: 'utf-8'}, (err, data) => {
-            if (err) {
-                reject(err);
-                return;
-            }
-            resolve(data);
-        });
-    });
+function readLocale(name: string) {
+    return fsReadFile(resolvePath(__dirname, '../../src/_locales/', name), {encoding: 'utf-8'});
 }
 
-test('Locales', async () => {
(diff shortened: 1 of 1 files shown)


Gusted:
#3091 I already made a begin of this. and all the other languages need the translations as well in order to work...

Gusted:
Seems good that it should be organized into groups troug the [describe function](https://jestjs.io/docs/en/api#describename-fn) Potential we could change this for other unit tests also if it is possible.

tests/config/locales.tests.ts
```diff
@@ -1,65 +1,46 @@
-import {readFile as fsReadFile, readdir as fsReadDir} from 'fs';
+import {readFileSync as fsReadFile, readdirSync as fsReadDir} from 'fs';
```
Gusted, line 1: Not sure if we want using a Sync due to.
```
You should never call readFileSync in a node.js since it will tie up the single thread loop while I/O is performed. You want the node loop to process other requests until the I/O completes and your callback handling code can run.
```

tests/config/locales.tests.ts
```diff
@@ -1,65 +1,46 @@
...
-        return locales.every((loc, j) => {
-            const lines = loc.split('\n');
-            for (let i = 0; i < Math.min(lines.length, enLines.length); i++) {
-                if (!predicate(enLines[i], lines[i])) {
-                    console.error(`${files[j]}, line ${i + 1}`);
-                    return false;
-                }
+    function compareLinesToEnLocale(loc: string, predicate: (en: string, loc: string) => boolean) {
+        const lines = loc.split('\n');
+        for (let i = 0; i < Math.min(lines.length, enLines.length); i++) {
```
Gusted, line 19: Let's change this to a cache length loop.
`for (let i = 0, len =Math.min(lines.length, enLines.length); i < len; i++) {

tests/config/locales.tests.ts
```diff
@@ -1,65 +1,46 @@
...
             }
-            return true;
-        });
+        }
+        return true;
     }
 
-    // Line count is the same
-    expect(locales.every((loc) => loc.split('\n').length === enLines.length)).toBe(true);
+    it.each(files)('%s', (file) => {
```
Gusted, line 28: Let's add a comment to describe `it` as it's just [a alias](https://jestjs.io/docs/en/api#testeachtablename-fn-timeout)

tests/config/locales.tests.ts
wojtekmaj: Aw, that's a bummer. :/ describe() cannot be async, and it.each() can't have an array dynamically populated. So I can't use promises to define test structure at all.

Gusted: :/

PR #3259, opened 2020-07-27
Fix for getURLHost

### LINKED ISSUE #1119: Strip credentials from URLs
image

Gusted:
- Get better host.
- Strip credentials
- Resolve #1119


Gusted:
https://developer.mozilla.org/en-US/docs/Web/API/URL/host

PR #3269, opened 2020-07-28
fix for facebook.com

Myshor:
One more old blob fix remove.


Myshor:
Added traseo.pl map darkening when shared in post.

PR #3270, opened 2020-07-28
Fixed NASA JPL logo

TylerOlson:
Instead of inverting the black image to white and mess up the colors, force it to use the provided white image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2486,6 +2486,15 @@ INVERT
 
 ================================
 
+jpl.nasa.gov
+
+CSS
+.brand_area {
+   background-image: url("https://www.jpl.nasa.gov/assets/images/logo_nasa_trio_white@2x.png") !important;
+}
+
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -2486,6 +2486,15 @@ INVERT
 
 ================================
 
+jpl.nasa.gov
+
+CSS
+.brand_area {
```
Gusted, line 2492: Change it to

```
IGNORE IMAGE ANALYSIS
.brand_area
```
Should do the same affect but cleaner.

TylerOlson: I already tested it and `IGNORE IMAGE ANALYSIS` makes the text black on black. There are 2 provided images with the logo, one with black text and one with white text. When scrolling the logo moves from ontop of an image to a black background. Ignoring the analysis keeps black text on a dark background. Manually setting it to white retains the logo color and white text.

PR #3276, opened 2020-07-29

wanzengy:
The image in 'tianchi.aliyun.com' will be unclear under dark mode. So I set image 'no invert'.


Gusted:
`NO INVERT` is not a option in dynamic theme.
Have a look for the avaible commands in Dynamic theme
https://github.com/darkreader/darkreader/blob/master/README.md#use-dev-tools

src/config/dynamic-theme-fixes.config
```diff
@@ -6011,3 +6018,4 @@ zhihu.com
 
 INVERT
 img[eeimg="1"]
+
```
Gusted, line 6021: You can remove this new line.

src/config/dynamic-theme-fixes.config
```diff
@@ -4733,6 +4733,13 @@ INVERT
 
 ================================
 
+tianchi.aliyun.com
+
+NO INVERT
+img
```
Gusted, line 4739: `img` This is every image, try to just select certain images who need it.

wanzengy: Thank you for your reply. And actually all `image`s showed in `tianchi.aliyun.com` have some problems, for example the words in image can't see clearly under dark mode. So I set all images.

PR #3282, opened 2020-07-29
Fix for streamable.com

kyleseven:
Fixes play/pause button and track from being unreadable in the dark.


Gusted:
Any example link?

kyleseven:
Example: https://streamable.com/yows9s

PR #3310, opened 2020-08-02
Fix dndbeyond.com

atg22896:
When viewing a monster of item description page, the paper text background got completely garbled. Also fixed the website background image.

Before:
chrome_2020-08-02_04-38-50

After:
chrome_2020-08-02_05-16-54

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1186,21 +1186,13 @@ nav[aria-label="Servers sidebar"] foreignObject:hover {
 dndbeyond.com
 
 CSS
-.mon-stat-block {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -1183,6 +1183,30 @@ nav[aria-label="Servers sidebar"] foreignObject:hover {
 
 ================================
 
+dndbeyond.com
+
+CSS
+.mon-stat-block {
```
Gusted, line 1189: You can use multiple selectors like
```CSS
.mon-stat-block,
.mon-stat-block::before {
    background-image: none !important
}

PR #3324, opened 2020-08-03
Add docs.microsoft.com to dark-sites.config

jmkeol:


Gusted:
@sprinkledcheese Last time it was preffered-color-scheme, I just tested it with private window and TOR with light color. Seems like they have it now dark by default(You was the last time pointed it was still light).

sprinkledcheese:
Still uses prefers-color-scheme.

Gusted:
Ola.

jmkeol:
Could darkreader could detect the value of `prefers-color-scheme`?
Could it then be used somehow?

Gusted:
`prefers-color-scheme` is just a indication of the user have it's system to a dark or light color. So detecting wouldn't work much as I see that Microsoft likes to have it permanent somehow. They have a weird save system.

PR #3325, opened 2020-08-03
Another fix for mail.google.com

gabriele2000:
Upper-left corner, invert gmail logo (I think it's more consistent that way)

Before:
immagine

After:
immagine


Gusted:
Honestly, the Logo will looks worse than worse. Worser than [this](https://i.imgur.com/A9uCA12.png) But that was rather because it was really [invisble](https://i.imgur.com/HTy71a9.png). Let's not touch it unless majority of the users want this. As gmail is a popular website :)

PR #3332, opened 2020-08-03
fix for google maps timeline

Myshor:
fix for inverted satellite view in timeline
obraz

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2145,7 +2145,7 @@ button.section-directions-details-action-button
 div.section-loading-spinner
 a.gb_b > div
 a.gb_xc
-.gm-style img[role="presentation"]:not([src*="v=840"])
+.gm-style img[role="presentation"]:not([src*="v="])
 .i4ewOd-xl07Ob
 .i4ewOd-LQLjdd li::before
 .un1lmc-j4gsHd


src/config/dynamic-theme-fixes.config
```diff
@@ -2145,7 +2145,7 @@ button.section-directions-details-action-button
 div.section-loading-spinner
 a.gb_b > div
 a.gb_xc
-.gm-style img[role="presentation"]:not([src*="v=840"])
+.gm-style img[role="presentation"]:not([src*="v="])
```
Gusted, line 2148: Wha- How did I didn't notice this in my `git diff`.
Is it like preloaded already? But the `display` declaration just changes?

Myshor: Not sure how it works... I saw in my dev tools in firefox change flashes in source when I changed to satellite and compared what I see fixes. I had v=871 in my maps but I think it can change again to something else. Still as it is still used in satellite view (not sure what it is - version or sth) then it should work until google change satellite map links to totally different 😉)

Gusted: What- I'm switching to Firefox dev tools lmao. Nothing was flashing in Chrome's one.

Myshor: Maybe because of this:
obraz

Gusted: Developer edition, it should be somewhere among the other browsers in my list 😂. Just need to get opera installed.

Scutua (approved):

PR #3336, opened 2020-08-03
ProMods website

erkinalp:
The website is already light-on-dark, the only fix made is removing the main background image, which interferes with contrast settings.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3920,6 +3920,15 @@ body {
 
 ================================
 
+promods.net
+
+CSS
+html {
+    background-image: none !important;
+}
+
+================================
+
 prostovpn.org
 
 CSS
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -676,6 +676,15 @@ nav.pagination {
 
 ================================
 
+blog.promods.net
+
+CSS
+body.custom-background {
+    background: none !important;
+}
+
+================================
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -3920,6 +3929,15 @@ body {
 
 ================================
 
+promods.net
+
+CSS
+html {
+    background-image: none !important;
```
Gusted, line 3936: The background-image is already dark by default not sure why removing a non bright image.

src/config/dynamic-theme-fixes.config
```diff
@@ -676,6 +676,15 @@ nav.pagination {
 
 ================================
 
+blog.promods.net
+
+CSS
+body.custom-background {
+    background: none !important;
```
Gusted, line 683: The background-image is already dark by default, not sure why removing a non-bright image.

src/config/dynamic-theme-fixes.config
erkinalp: Because it messes with contrast settings. The image is originally calibrated to the website's grey background; however, it breaks with increased contrast.
Without dark reader
Dark reader, background image removal, max contrast
You see, this fix makes the website *darker*.

src/config/dynamic-theme-fixes.config
erkinalp: The same as for the main website, this fix makes it *darker* if you dial the contrast to max.

PR #3360, opened 2020-08-07
Improve 10fastfingers.com

fw623:
1. make text input field dark
1. make all backgrounds dark

I made the backgrounds darker than in #2189 because I find it uncomfortable when the important parts are darker than the background. If this is not something we want to do here, I'm open for discussion.\
These changes also fix the backgrounds of the `text-practice` and `top1000` modes.

### Top 1000 changes (Practice is similar)
top1000_old
old

top1000_new
new

### Typing and general background changes
The missing time is due to another personal customization which has nothing to do with these changes.

typing_old
old

typing_new
new

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -86,7 +86,12 @@ CSS
     background: var(--darkreader-neutral-background) !important;
     color: var(--darkreader-neutral-text) !important;
 }
-body, .container-modified > .row, #practice-main, #top1000-index-container, #text-practice, #content-bg {
+body, 
+.container-modified > .row, 
+#practice-main, 
+#top1000-index-container, 
+#text-practice, 
+#content-bg {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -79,14 +79,19 @@ html, body {
...
-}
 #speedtest-main .hide-time {
     color: transparent !important;
 }
 #inputfield {
-    color: #333 !important;
+    background: var(--darkreader-neutral-background) !important;
+    color: var(--darkreader-neutral-text) !important;
+}
+body, .container-modified > .row, #practice-main, #top1000-index-container, #text-practice, #content-bg {
```
Gusted, line 89:
```CSS
body, 
.container-modified > .row, 
#practice-main, 
#top1000-index-container, 
#text-practice, 
#content-bg {
```

PR #3363, opened 2020-08-07
Add feedly.com to dark-sites.config

ChristopherZhong:
Feedly.com natively supports a dark theme.


Gusted:
The list is only for site already by default dark.

PR #3373, opened 2020-08-10
Fixed heimu for moegirl.org.cn

leranjun:
Some users pointed out that the original CSS was not functional on the website.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3201,11 +3201,40 @@ IGNORE INLINE STYLE
 
 ================================
 
-moegirl.org
+moegirl.org.cn
 
 CSS
-.heimu, .heimu * {
-  color: #252525 !important;
+span.heimu a.external,
+span.heimu a.external:visited,
+span.heimu a.extiw,
+span.heimu a.extiw:visited {
+    color: #dadada !important;
+}
+.heimu,
+.heimu a,
+a .heimu,
+.heimu a.new {
+    background-color: #dadada !important;
+    color: #dadada !important;
+}
+body:not(.heimu_toggle_on) .heimu:hover,
+body:not(.heimu_toggle_on) .heimu:active,
+body:not(.heimu_toggle_on) .heimu.off {
+    color: black !important;
+}
+body:not(.heimu_toggle_on) .heimu:hover a,
+body:not(.heimu_toggle_on) a:hover .heimu,
+body:not(.heimu_toggle_on) .heimu.off a,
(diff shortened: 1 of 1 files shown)


Gusted:
Seems like both sites are offline.

leranjun:
Oh right, that's because they terminated the foundation site.

The current available sites are [zh.moegirl.org.cn](https://zh.moegirl.org.cn/), [en.moegirl.org.cn](https://en.moegirl.org.cn/), [ja.moegirl.org.cn](https://ja.moegirl.org.cn/), [library.moegirl.org.cn](https://library.moegirl.org.cn/) and [commons.moegirl.org.cn](https://commons.moegirl.org.cn/).

Should I change the main domain to *.moegirl.org.cn then?

Gusted:
> Should I change the main domain to *.moegirl.org.cn then?

Nah not needed.

Gusted:
Could you point out where is what changed. As the Page itself seems fine to me.

leranjun:
> Could you point out where is what changed. As the Page itself seems fine to me.

Right, so basically there is a template called "heimu" (kinda like the "spoiler" feature in Discord, if you know what I mean) that hides texts on a black background. However, this apparently is not easy to spot when Dark Reader is turned on. I inverted the colour of the background and the text so now it's black text on a grey background instead of white text on a black background.

The styles I added are based on [the default stylesheet](https://zh.moegirl.org.cn/MediaWiki:Common.css#L386). The original config file does not include the `:hover` pseudo-class, so the default styles (white on black) were applied, but then Dark Reader changed the text colour again, making it completely unreadable.

leranjun:
Oh, silly me. I'll provide a screenshot, wait a second…

leranjun:
### Dark Reader off
Screenshot 2020-08-10 at 9 50 14 PM

### Dark Reader on, before
Screenshot 2020-08-10 at 9 50 31 PM

### Dark Reader on, after
Screenshot 2020-08-10 at 9 51 30 PM

Hope that helps explaining things a bit?

Gusted:
Yep that make sense.
Could you have a look how it looks like with white scheme? If it's fine I will be ready to merge.

leranjun:
> Could you have a look how it looks like with white scheme? If it's fine I will be ready to merge.

Welp, considering the light version is black background with white font, I suppose the grey-black scheme should be acceptable ¯\\\_(ツ)\_/¯ (plus, I believe there is an option to turn off spoilers if they wish to)

Anyway, thanks a lot for helping!

Gusted:
>  I suppose the grey-black scheme should be acceptable

If the spoiler still have the same behavior it should be fine.

PR #3377, opened 2020-08-10
Fix background for dw.com

felixonmars:


src/config/dynamic-theme-fixes.config
```diff
@@ -1443,6 +1443,15 @@ CSS
 
 ================================
 
+dw.com
+
+CSS
+div#bodyMover {
```
Gusted, line 1449: `!important`

Gusted: Is `div` needed?

felixonmars: You are right. In my previous testing adding div avoids using `!important` but it's... unstable. Sorry!

PR #3382, opened 2020-08-11
fix for x-kom.pl

Myshor:
different selector for svg logo as they don't changed classes.
obraz


Myshor:
Added invert for hamburger menu:
obraz

PR #3394, opened 2020-08-12
Fix for alipay.com

felixonmars:
- Invert logo in various places across the site.
- Invert background of qrcode to make it easier to scan.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -265,6 +265,15 @@ CSS
 
 ================================
 
+alipay.com
+
+INVERT
+.alipay-logo
+#J_logoHomeUrl
+.global-logo
+
+================================
+
 aljazeera.com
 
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -265,6 +265,20 @@ CSS
...
+alipay.com
+
+INVERT
+.alipay-logo
+#J_logoHomeUrl
+.global-logo
+
+CSS
+.qrcode-detail-img {
+    background-color: white !important;
```
Gusted, line 277: Replace `white` with `${black}`

felixonmars: The QR code image itself is always black on white regardless of our theme, so I explicitly used `white` here instead of `${black}` and it works correctly. Using `${black}` here breaks it on Light theme.

PR #3410, opened 2020-08-14
Wikipedia formula fallback fix

erkinalp:
This one-line change fixes the math formula fallback images.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5799,6 +5799,7 @@ INVERT
 .mwe-math-element
 .mw-ext-score
 .mw-wiki-logo
+.mwe-math-fallback-image-inline
 .central-textlogo__image
 .svg-Wikimedia-logo_black
 .header .branding-box > a > span > img
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5800,6 +5800,7 @@ INVERT
 .mw-ext-score
 .mw-wiki-logo
 .mwe-math-fallback-image-inline
+.mwe-math-fallback-image-display
 .central-textlogo__image
 .svg-Wikimedia-logo_black
 .header .branding-box > a > span > img


Gusted:
Hi @erkinalp!

Could you provide a example URL?

Regards,
Gusted

erkinalp:
Any maths article, for example, https://en.wikipedia.org/wiki/Normal_distribution .

Gusted:
This made all math things black?

erkinalp:
Try again.

Gusted:
I'm not seeing any changes.

erkinalp:
Interestingly works for me.
resim

Gusted:
Are you on invert? Or have any custom settings?

erkinalp:
Yes, I have also changed the base colours of both the extension and the browser itself to light on dark (some sites are not darkened at all if I do not change the base colours using the browser preferences). Also changed the selection colour on the extension as well.

Gusted:
> I am on dynamic, I have also changed the base colours of both the extension and the browser itself to light on dark (some sites are not darkened at all if I do not change the base colours using the browser preferences). Also changed the selection colour on the extension as well.

Okay as majority of the people are on default settings all fixes should be according to the default settings.

erkinalp:
Do you plan adding a supplemental configuration file for keeping the local fixes separately? Because anyone on high contrast schemes will have similar problems with those fixes.
Can you also share a screenshot for the pre-PR situation?

Gusted:
`.mwe-math-element` is already inverting all math formula's.

> Do you plan adding a supplemental configuration file for keeping the local fixes separately? Because anyone on high contrast schemes will have similar problems with those fixes.

We are planning to add a `User CSS` to separate user fixes for situations like this, but for now I recommend a extension like Stylus. 
It won't be added soon(as V5 roadmap don't contain this feature), but there are plans about it.

erkinalp:
This is the fixup snippet for dark reader users with high contrast and dark browser base colour:
```CSS
@-moz-document domain("wikipedia.org") {
  .mw-body img.mwe-math-fallback-image-display,
  .mw-body img.mwe-math-fallback-image-inline {
    filter: invert(100%);
    border: none;
  }
}
```

PR #3425, opened 2020-08-15
Fix for syncsetting error

### LINKED ISSUE #3424: [BUG] Extension is stuck on "Loading Please Wait" and the dark mode is not working.
### **Describe the bug**

The extension does not seem to work anymore, after it working all of yesterday. Now, whenever I go to a site it wont load the dark mode, and in the extension itself it continuously says "Loading, please wait", it has been doing this for too long now for me to believe it will stop loading, since it usually loads so fast. This happens despite of restarting the browser and reinstalling Dark Reader.

### **Steps to reproduce**

Open the browser
Try to go to a site, dark mode will never load
Click on extension icon, and it is forever loading

### **Expected behavior**

Any given site to go into dark mode and for the extension to show the various options in its extension menu.

### **Actual behavior**

What happened above, dark mode is no longer working and the extension menu says its always loading.

### **Screenshots**

image

image

### **System Information:**
 - OS: Windows 8.1
 - Browser: Chrome 84.0.4147.125
 - Dark Reader version: 4.9.16

Gusted:
- Make sure to have a boolean in syncsettings. If local hasn't the setting, the default setting would be chosen.
- Resolves #3424

### DIFF
--- a/src/background/user-storage.ts
+++ b/src/background/user-storage.ts
@@ -23,6 +23,7 @@ export default class UserStorage {
     private loadSettingsFromStorage() {
         return new Promise<UserSettings>((resolve) => {
             chrome.storage.local.get(DEFAULT_SETTINGS, (local: UserSettings) => {
+                local.syncSettings = local.syncSettings || DEFAULT_SETTINGS.syncSettings;
                 if (!local.syncSettings) {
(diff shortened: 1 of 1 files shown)


Gusted:
@alexanderby I honestly don't know but it seems like when you load it, it don't load non-existing objects yet(So it returns null as value), but when resolve(local) it returns the settings good??

PR #3432, opened 2020-08-16
Added a fix for heise.de

PaulGoldschmidt:
I added a fix for the german technology-news-site "heise.de": It wasn't possible to see the results of the opinion-section of some news (a section where readers can vote for their opinion) because the frame was incorrectly inverted. I changed this behavior just by adding an "INVERT" rule, tested it in both modes and this small tweak works really good.


Gusted:
Hi @PaulGoldschmidt!

Could you provide a example URL where I can see the change?

Regards,
Gusted

PaulGoldschmidt:
Hey Gusted,

sure: https://www.heise.de/news/Trump-will-Begnadigung-von-Whistleblower-Snowden-pruefen-4871768.html

Best regards,
Paul

Gusted:
Hi @PaulGoldschmidt!

I'm not sure if this need a invert as it made it look even more invisible and the `advertisement` img to be inverted as well.

Regards,
Gusted

PaulGoldschmidt:
Hi @Gusted ,

agreed - the text shouldn't be inverted as well, but with the `INVERT` rule the percentages are readable at least..
I will check tomorrow if I can just invert the percentages (class `binPercentages`) so that they are readable and the rest stays the same.

Regards,
Paul

Gusted:
Hi @PaulGoldschmidt!

As it's actually in a Iframe the site has to be changed to Iframe source(That's how our code base work).

```CSS
compass.pressekompass.net

IGNORE INLINE STYLE
text

```
This should do it. It should looks like this(After a refresh) 

As for the current heide.de fix you can replace the class with `.logo-holder .brand-logo` To invert the right bottom logo.

PaulGoldschmidt:
Hi @Gusted !
Thanks for your suggestions regarding the Iframe.
The fix with the `compass.pressekompass.net` works perfectly, but I'm not sure where I should insert the `.logo-holder .brand-logo` part.
Could you tell me where you inserted the last `INVERT`-rule to fix the logo? It seems to me like the logo is also loaded from the IFrame.

Best regards and thank you very mutch for your help!
Paul

Gusted:
Hi @PaulGoldschmidt!

Indeed, the logo is loaded from the Iframe so you can insert the logo class also under INVERT of the `compass.pressekompass.net`

Gusted:
> Really appreciated your help

No problem 😃

PR #3449, opened 2020-08-17
easings.net has autodark

junpeikun:
easings.net has a theme chooser (at the bottom right of the page) with autodark working great.


Gusted:
That means people with light theme systems get a nice white flash in their eyes.

PR #3450, opened 2020-08-17
Fix for jobs.github.com

linsui:


src/config/dynamic-theme-fixes.config
```diff
@@ -2770,6 +2770,15 @@ h1.logo
 
 ================================
 
+jobs.github.com
+
+CSS
+#page {
+    background: none !important;
```
Gusted, line 2777: Please change `background` to `background-image`

PR #3453, opened 2020-08-17
Dynamic theme fixes for theinformation.com

michaelkotrous:
Invert the logo and tweak styles of the share button

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5097,6 +5097,20 @@ INVERT
 
 ================================
 
+theinformation.com
+
+INVERT
+.logo
+
+CSS
+article.single .sticky-header button.button-share,
+#bookmark-and-share button.button-share {
+    background-color: #21498c;
+    color: #e8e6e3;
+}
+
+================================
+
 thejakartapost.com
 
 INVERT
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5108,6 +5108,19 @@ article.single .sticky-header button.button-share,
     background-color: #21498c;
     color: #e8e6e3;
 }
+article.single .sticky-header button.button-share:hover,
+#bookmark-and-share button.button-share:hover {
+    background-color: #245098;
+}
+.locked section .insert-container {
+    filter: drop-shadow(20px 20px 20px #26292b);
+}
+.locked section .insert-container#app-screenshot-container {
+    filter: drop-shadow(20px -20px 20px #26292b);
+}
+
+IGNORE IMAGE ANALYSIS
+header.locked .logo
 
 ================================
 


src/config/dynamic-theme-fixes.config
```diff
@@ -5097,6 +5097,20 @@ INVERT
 
 ================================
 
+theinformation.com
+
+INVERT
+.logo
```
Gusted, line 5103: Seems to me it's already inverted without this fix

src/config/dynamic-theme-fixes.config
```diff
@@ -5097,6 +5097,20 @@ INVERT
...
+
+CSS
+article.single .sticky-header button.button-share,
+#bookmark-and-share button.button-share {
+    background-color: #21498c;
```
Gusted, line 5108: I assume this are the default values from the site itself?

The articles is after a paywall so I can't check for myself.

src/config/dynamic-theme-fixes.config
michaelkotrous: Good catch. I only looked at the article pages on the other side of the paywall. It looks like the gated homepage is the only page with a logo that inverts correctly. Adding some CSS will fix the box-shadows on the gated homepage too. Pasting my new devtools code below.

```suggestion
theinformation.com

INVERT
.logo

CSS
article.single .sticky-header button.button-share,
#bookmark-and-share button.button-share {
    background-color: #21498c;
    color: #e8e6e3;
}
article.single .sticky-header button.button-share:hover,
#bookmark-and-share button.button-share:hover {
    background-color: #245098;
}
.locked section .insert-container {
    -webkit-filter: drop-shadow(20px 20px 20px #26292b);
    filter: drop-shadow(20px 20px 20px #26292b);
}
.locked section .insert-container#app-screenshot-container {
    -webkit-filter: drop-shadow(20px -20px 20px #26292b);
    filter: drop-shadow(20px -20px 20px #26292b);
}

IGNORE IMAGE ANALYSIS
header.locked .logo

================================

```

Screenshot of article page with this code applied.

Screen Shot 2020-08-17 at 11 46 03 AM

src/config/dynamic-theme-fixes.config
michaelkotrous: These color values match other buttons on the site. You can confirm with the blue button on the [log-in page](https://www.theinformation.com/sessions/new).

src/config/dynamic-theme-fixes.config
Gusted: You can drop the `-webkit-filter` It's not needed for fixes for the rest it LGTM make sure check for `light` scheme in dark reader if it don't break the site.

michaelkotrous: Thanks. Double-checked the light theme and committed changes with `-webkit-filter` dropped.

PR #3454, opened 2020-08-17
Fixed compass.pressekompass.net

PaulGoldschmidt:
`compass.pressekompass.net` is a tool to let visitors of (mainly german) news websites vote for their opinion. The voting gets via an Iframe embedded into the website. This fix corrects the wrong invertation of the media outlet's logo and fixes the shown percentage after voting.
Example URL: [heise.de](https://www.heise.de/news/Trump-will-Begnadigung-von-Whistleblower-Snowden-pruefen-4871768.html)

Before:
grafik

After: 
grafik

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -964,6 +964,13 @@ INVERT
 
 ================================
 
+community.ntppool.org
+
+INVERT
+#site-logo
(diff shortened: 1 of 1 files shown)


PaulGoldschmidt:
Ah thanks, I got the position wrong in the config 🥇

PR #3465, opened 2020-08-18
Fix for bbs.chinauos.com

felixonmars:
- Invert topic tag icon (weird dark-white combination without the change)
- Fix main forum post text color (example page: https://bbs.chinauos.com/post/1930?id=1930&type_id=0)

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -575,6 +575,18 @@ INVERT
 
 ================================
 
+bbs.chinauos.com
+
+INVERT
+.post_tip
+
+CSS
+div[style="all: initial;"] {
+    all: unset !important;
+}
+
+================================
+
 bbs.thinkpad.com
 club.lenovo.com.cn
 


src/config/dynamic-theme-fixes.config
```diff
@@ -575,6 +575,18 @@ INVERT
 
 ================================
 
+bbs.chinauos.com
+
+INVERT
+.post_tip
+
+CSS
+div[style="all: initial;"] {
```
Gusted, line 584: Where can I find this element?

felixonmars: It's in the link I pasted above: https://bbs.chinauos.com/post/1930?id=1930&type_id=0

Gusted: Please use `^=` So it would match with elements that begin with that text. As dark reader is adding all their inline nonsense styles into it.

felixonmars: Updated :)

PR #3470, opened 2020-08-18
Fix for wikipedia.org

haruhi-s:
Deleting `.mwe-math-element` as inverting it would actually make math elements become completely black on a dark background.
[This page](https://en.wikipedia.org/wiki/Lattice_(group)) can be used as a reference to this issue, look for the color of _R^n_ in the first line.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -6004,7 +6004,6 @@ img[src="images/black.png"]
 wikipedia.org
 
 INVERT
-.mwe-math-element
 .mw-ext-score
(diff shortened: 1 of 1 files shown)


Gusted:
I don't see what you are saying. It's inverting correctly the R^n

haruhi-s:
> 
> 
> I don't see what you are saying. It's inverting correctly the R^n

Maybe it only happens on my end. It only displays correctly when I remove `.mwe-math-element` . I am using default dynamic.

Gusted:
Hi @haruhi-s!

Could you give it a try on default settings of Dark reader?

PR #3472, opened 2020-08-19
add jellyfin.org

muhlinux:
add jellyfin.org to dark-sites.config


Gusted:
my eyes https://jellyfin.org/docs/

PR #3475, opened 2020-08-19
missions.ingress.com is dark

felixonmars:


Gusted:
How in earth do you visit so many sites and make fixes on a day. At peak I get around 20 and that are from issues.

felixonmars:
> How in earth do you visit so many sites and make fixes on a day. At peak I get around 20 and that are from issues.

I just started using darkreader this month and am trying to fix every website I use :P

(Also learning CSS because of this, at the same time...)

Gusted:
> I just started using darkreader this month and am trying to fix every website I use :P

Well well 👏🏽 

> (Also learning CSS because of this, at the same time...)

Doing a great job so far. It's just getting along the syntax and than experience with various property's etc.

PR #3480, opened 2020-08-19
Fix Stack Exchange upvotes

### LINKED ISSUE #2602: [Broken Website] Upvote/Downvote/Accept button colours on StackOverflow
**Website**
StackOverflow, with site dark mode enabled.

**How to Reproduce**
An example for this is
> 1. Go to StackOverflow and enable both Darkreader and site dark mode.
> 2. Go to a question that you asked (so that you have permissions to accept/unaccept answers)
> 3. The upvote & downvote buttons do not change to orange on pressed.
> 4. The accept button does not change to green on accepted.

**Expected behavior**
Upvote and downvote buttons chnage colour to orange and back to gray when the answer is upvoted/downvoted or un-upvoted/un-downvoted, respectively.
Accept button changes colour to green and back to gray when the answer is accepted or unaccepted, respectively.

**Actual behavior**
All three buttons are gray, regardless of what you do.

**Screenshots**
| Expected | Actual |
| ---------- | ------- |
| Screencap-001-20200525-153750 | Screencap-001-20200525-153738 |

**System info:**
 - OS: Windows 10 1909
 - Browser: Chrome 83
 - Darkreader: 4.9.8

**Additional context**
The reason I'm using both StackOverflow's built in dark mode and Darkreader together, is as follows:

1. Using SO's dark mode alone causes text contrast to be too high. I need Darkreader to tune it down.
2. Using Darkreader alone causes some titles and some SVGs to be nearly unreadable against the dark background.

Therefore I enabled SO's built in dark mode to make all the logos and titles look nice, and then use Darkreader to reduce brightness and add sepia, so that the text don't hurt my eyes. Hopefully this use case makes sense.

### LINKED ISSUE #3478: [Broken Website]
**Website**
StackOverflow.com

**How to Reproduce**
This cannot be done via a new user.  You have to have rights to vote.
> 1. https://stackoverflow.com/
> 2. Login 
> 3. Find a good question and upvote it.
> 4. Turn on DarkReader

**Expected behavior**
The up arrow for voting would be filled in with a different color

**Actual behavior**
The up arrow looks the same as a non-voted up arrow

**System info:**
 - OS: Windows 10
 - Browser: chrome
 - Darkreader Version: version 4_9_16
 
**Additional context**
Sometimes (not every time) this also happens with comment up votes.

jacktose:
Remove !important that was overriding native color of up/down-voted arrows and selected checks.

- Resolves #2602 
- Resolves #3478

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -59,7 +59,7 @@ body {
     background-image: repeating-linear-gradient(0deg, transparent, transparent 13px, ${#e4e6e8} -13px, ${#e4e6e8} 21px) !important;
 }
 .c-pointer {
-    color: grey !important;
+    color: grey;
 }
 .js-accepted-answer-indicator.fc-green-500 {
     color: var(--green) !important;


Gusted:
Ah..... That makes a lot of sense now.

PR #3497, opened 2020-08-21
Better static mode

Gusted:
- Use pre-defined static themes.
- Use local files to store static themes.


src/ui/devtools/components/body.tsx
```diff
@@ -15,13 +15,8 @@ function Body({data, tab, actions}: BodyProps) {
     const previewButtonText = data.settings.previewNewDesign ? 'Switch to old design' : 'Preview new design';
     const {theme} = getCurrentThemePreset({data, tab, actions});
 
-    const wrapper = (theme.engine === ThemeEngines.staticTheme
+    const wrapper = (theme.engine === ThemeEngines.cssFilter || theme.engine === ThemeEngines.svgFilter 
         ? {
```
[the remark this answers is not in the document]
Gusted: Static is being removed for now by dev tools as it's going to get a whole rewrite and going to work on local files etc. So for now it's going to be removed by static and in a further PR being added.

src/static/themes/variables.css
```diff
@@ -0,0 +1,65 @@
+:root {
```
erkinalp, line 1: also export them in dynamic mode

erkinalp (changes requested):

src/static/themes/variables.css
Gusted: Already being taking care of by #2767 [see the code](https://github.com/darkreader/darkreader/blob/master/src/inject/dynamic-theme/index.ts#L101)

erkinalp: those six aren't going to be enough after we get more themes than just dark and light

Gusted: Uhm, Their will be just dark and light however, the new presets can have different colors and they are being dynamically computed by the current in use preset.

Gusted:
Doesn't seem must interest as well, so closing.

PR #3519, opened 2020-08-24
Add support to for external connections

### LINKED ISSUE #691: [enhancement]: Consider allowing external extensions to toggle theme.
I'd like to do some automation with dark reader, but I've found that external connections (using `chrome.runtime.sendMessage`) aren't handled.

Something like this in the constructor of Messenger:

```
chrome.runtime.onConnectExternal.addListener(function(port) {
  if (port.name === 'ui' && port.type === 'toggle-current-site') {
    port.onMessage.addListener(function() {
      this.adapter.toggleCurrentSite();
    });
  }
});
```

and this line to the `getMessengerAdapter` method in extension.ts:

``` 
toggleCurrentSite: () => this.toggleCurrentSite(),
```

Should be enough to allow this in other extensions:
```
chrome.runtime.sendMessage('eimadpbcbfnmbkopoojfekhnkhdbieeh', {name: 'ui', type:' toggle-current-site'});
```

Happy to add a pull request if this is considered worth adding.

### LINKED ISSUE #2168: [Idea] .Xresources/Pywal colors
Recently a plugin called [Pywalfox](https://github.com/Frewacom/Pywalfox) was released, which takes the colors generated by the program [pywal](https://github.com/dylanaraps/pywal) and uses them to theme various UI elements in Firefox.

The main problems this solves is using a custom script to make firefox aware of the system colors, and changing them without the need for a restart.

My idea is to take these colors and apply them to the whole web, like darkreader applies a universal dark mode. In theory it would look a bit like this:

This is a vim plugin I customized to preview markdown using the system colors. 

I have been looking at the codebase of both firefox plugins, and am unsure where to begin working on this. Would it be feasible to simply substitute darkreaders universal gray and white colors for the pywal colors? 

Pywalfox already provides an interface to select accent, highlights, etc. Would there be any "gotchas" that prevent us from simply hooking up the Pywalfox colors to darkreader?

Gusted:
- Let external extension make changes to dark reader for e.g. automation support.
- Add option(by default false) to Enable/Disable external connections.
- Resolves #691
- Resolves #2168

TODO: add some documentation in Reademe.MD and possible more functions to be added and exposed to external connections
[Example Extension for toggle extension](https://github.com/darkreader/darkreader/files/5120487/ahahaha.zip), It shows the basic Principe how it should work.


erkinalp (approved):

Yash-Singh1:
@Gusted Should there be an option to blacklist extensions?

erkinalp:
Indeed.

Gusted:
> @Gusted Should there be an option to blacklist extensions?

Little bit of a problem, we can only blacklist the ID's of the extension and for a `normal` user it won't make a sense at all, but I agree with this.

Yash-Singh1:
> > @Gusted Should there be an option to blacklist extensions?
> 
> Little bit of a problem, we can only blacklist the ID's of the extension and for a `normal` user it won't make a sense at all, but I agree with this.

@Gusted Take a look at https://developer.chrome.com/docs/extensions/reference/management
Seems like it `getAll` is supported in everything but safari: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/management

Gusted:
> @Gusted Take a look at https://developer.chrome.com/docs/extensions/reference/management
> Seems like it `getAll` is supported in everything but safari: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/management
> 
> The `management` API can make some sort of search interface in the popup, where the users can search through the installed extensions. Maybe there can be the extension icon and the extension name so the user can search through the installed extensions seeing the icon and the name. Later more properties can be taken advantage of: `version`, `homepageUrl`, etc.

If we can have the extension's name etc... We could possible build a new pop-up to mange this and might even restrict certain access to functions 🤷🏽‍♀️. It's possible it shouldn't be hard on the back-end, just the design.... I rather to not take care of.

Gusted:
[binaries.zip](https://github.com/darkreader/darkreader/files/5726687/binaries.zip) of https://github.com/darkreader/darkreader/pull/3519/commits/7d6c48ae3c9c92ad7ea6de549e89ed4fa0d4ed7e

erkinalp (approved):

bershanskiy:
> > @Gusted Should there be an option to blacklist extensions?
> 
> Little bit of a problem, we can only blacklist the ID's of the extension and for a `normal` user it won't make a sense at all, but I agree with this.

Perhaps, "access tokens" would be more elegant. Specifically, by default this feature is off, when user enables this feature extension generates a cryptographically strong random token and asks user to copy it. If external program does not have a valid token, then external request is ignored.

Gusted:
> Perhaps, "access tokens" would be more elegant. Specifically, by default this feature is off, when user enables this feature extension generates a cryptographically strong random token and asks user to copy it. If external program does not have a valid token, then external request is ignored.

That's possible, I will have to check that against some attack vector's.

golam71:
will they merge it?

Ranguna:
@Gusted  anything the community can help with to get this one merged ?

WhyNotHugo:
Was this closed signalling that this won't get merged? :c

alexanderby:
The idea of being able to control Dark Reader outside of it's UI is good. But this implementation may impose security risks. The app may also become disabled by default. I would think of moving the "headache" outside of the extension itself. Like you can try using Puppeteer:
```js
import {launch} from 'puppeteer-core';

async function run() {
    const executablePath = '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome';
    const userDataDir = '/Users/user/Library/Application Support/Google/Chrome';
    const darkReaderPath = `${userDataDir}/Default/Extensions/eimadpbcbfnmbkopoojfekhnkhdbieeh/4.9.81_0`;
    const darkReaderPopupURL = 'chrome-extension://eimadpbcbfnmbkopoojfekhnkhdbieeh/ui/popup/index.html';
    const toggleButtonSelector = '.header__app-toggle .toggle__btn:not(.toggle__btn--active)';
    const browser = await launch({
        executablePath,
        headless: true,
        userDataDir,
        args: [
            `--disable-extensions-except=${darkReaderPath}`,
            `--load-extension=${darkReaderPath}`,
        ],
    });
    const page = await browser.newPage();
    await page.goto(darkReaderPopupURL, {waitUntil: 'load'});
    await page.bringToFront();
    await page.waitForSelector(toggleButtonSelector);
    await page.click(toggleButtonSelector);
    await browser.close();
}

run();
```

PR #3520, opened 2020-08-25
Initial security policy draft

erkinalp:
This adds a formal security policy.
@alexanderby @Gusted feedback requested.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,15 @@
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 5.1.x   | :white_check_mark: |
+| 5.0.x   | :x:                |
+| 4.9.x   | :white_check_mark: |
+| < 4.9   | :x:                |
+
+## Reporting a Vulnerability
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -18,7 +18,7 @@ Latest version is supported.
 In case there is a vulnerability left behind, immediately notify
 to <darkreaderapp@gmail.com>; unless you can fix it yourself,
 in which case open a pull request on 
-<https://github.com/darkreader/darkrader/pulls>.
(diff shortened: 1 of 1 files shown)


erkinalp:
This adds a formal security policy.

SECURITY.md
```diff
@@ -0,0 +1,21 @@
...
+
+| Version   | Supported          |
+| --------- | ------------------ |
+| 4.9.16    | :white_check_mark: |
+| < 4.9.16  | :x:                |
+
+## Reporting a Vulnerability
+
+In case there is a vulnerability left behind, immediately notify
+to <darkreaderapp@gmail.com>; unless you can fix it yourself,
```
Gusted, line 19: > unless you can fix it yourself,

So that everyone can see the PR and abuse the bug right away?

SECURITY.md
```diff
@@ -0,0 +1,21 @@
+# Security Policy
+
+DarkReader is known to have no intentional security vulnerabilities.
```
Gusted, line 3: `Dark Reader` instead of `DarkReader`

SECURITY.md
```diff
@@ -0,0 +1,21 @@
...
+If reported, vulnerabilites will be fixed in the shortest time possible,
```
Gusted, line 4: `shortest time and highest priority as possible`

Gusted (changes requested):
Not sure how we even will be vulnerable, as it's a simple extension that only can be modified internal settings trough the UI that always have checks in place. However the API could be, but that's rather that it will load images that aren't needed in the first place. 🤷🏽‍♀️

SECURITY.md
erkinalp: I left that bit intentionally, to enable a copyleft relicense, which require the changes the published as soon as possible.

SECURITY.md
erkinalp: Fixing priority is deliberately left out as it would imply a warranty policy.

alexanderby (approved):

Gusted (approved):

Issue #2722, opened 2020-06-04
[BUG] auto not working on google chrome


phseiff:
**Describe the bug**
DarkReader.auto does not seem to work on google chrome, even though it behaves as expected on firefox (both desktop)

**To Reproduce**
I use Google Chrome, Version 83.0.4103.61 (Official Build) snap (64-bit), on Ubuntu 18.04, and [this](phseiff.github.io) website is shown in light mode even though my system, as well as my browser, use dark mode. On firefox, it works as expected.

**System Information:**
 - OS: Ubuntu 18.04 with Gnome
 - Browser: Google Chrome
 - Version: 83.0.4103.61 (Official Build) snap (64-bit)

Gusted:
Which automation mode do you use?

phseiff:
I took [darkreader.js](https://cdn.jsdelivr.net/npm/darkreader@4.9.10/darkreader.js) from jsDelivr (as suggested by README.md), embedded it into my page (not without including the license, of cause), and used

```DarkReader.auto({
    brightness: 100,
    contrast: 90,
    sepia: 10
});
```

as suggested by README.md.
And I used `DarkReader.setFetchMethod(window.fetch)` because it helped to fix some issues on firefox. The dark mode is applied whenever my os uses dark mode (on firefox), but not on chrome.

Is that what you meant when you asked which automation mode I use? If not, can you give me further instructions to find out which automation mode I am using?

Gusted:
Auto mode will bassicly watch for the user's `(prefers-color-scheme: dark)`. Dark -> Enabled | Light -> Disabled

phseiff:
Does this mean that the user's preferred color scheme can't be accessed on Chrome?

Gusted:
`matchMedia('(prefers-color-scheme: dark)').matches;` What does this return when you run it into your dev tools console?

phseiff:
`false`

Gusted:
Hmm than I'm sorry to say that this could be a browser bug as we just use this API as this is the only way to take a `dark mode` option from the system. I suggest to search with Ubuntu Gnome and dark mode Chrome etc. If you don't find any I suggest opening a chromium bug report about it.

phseiff:
That's a relief to hear since most people don't use this combination of os and browser, I guess. Thanks for your help in figuring it out!

Gusted:
If you like to you can post updates and share the chromium bug. If you need any more info for it I'm happy to hear and help with it.

[...]

PR #3533, opened 2020-08-27
Prevent whiteflash on inital document_start

### LINKED ISSUE #695: White background flashing when opening a web page
White background flashing when opening a web page
Although very short, it is too bright. Can it be turned into night mode?

### LINKED ISSUE #1219: White flash between page loads
When dark reader is turned on and installed into Firefox, there are white page load flashes between page loads where the all white default background is exposed before renders.
These flashes combined with dark reader have been giving me migraines as the page flip flops from over exposed bright white light to the mode dark reader creates when it does it's thing.

Is there any way dark reader could do something like change the default background in Firefox so that this flash does not occur?
It's really annoying.
Thanks. Great plugin BTW.

### LINKED ISSUE #1965: Version 4.8.8 white flash on page loads on Firefox
Dark Reader version: 4.8.8
Firefox version: 72.0.2 (64-bit)

**How to reproduce:**

1. Go to google.com
2. Refresh page

**What is expected to happen:**

Dark mode should load automatically.

**What actually happens:**

Non-dark mode blinds you for some milliseconds, before dark mode takes place.

### LINKED ISSUE #3307: [Broken Website]  white flash on git.sr.ht
**Website**
https://git.sr.ht and all hosted projects on it, e.g. https://git.sr.ht/~cadence/bibliogram
**How to Reproduce**
An example of this is
1. Go to the bibliogram project (2nd link) or any other project you know on git.sr.ht
2. Browse the website by clicking on links

**Expected behavior**
The page is dark instantly

**Actual behavior**
The page is dark after about one second

**System info:**
 - OS: Android
 - Browser: Firefox nightly
 - Darkreader Version: 4.9.16
 
**Additional context**
This is not related to mozilla-mobile/fenix#6313 which only happens when you load a website in a new tab. Except this first white flash which is due to a browser bug, there are no other white flashes on any website when you browse it **in the same tab** (not using a new tab). But on git.sr.ht it happens, therefore I think it's a darkreader bug.

### LINKED ISSUE #3523: [BUG] Lightmode for a split second while loading a page
### **Describe the bug**

While loading a new page, even on the same webite that has dark mode set, the page may go white mode for a split second, resulting in a very clear flash of glaring white.

### **Steps to reproduce**

- Browse around on, for example, Github.

### **Expected behavior**

Always stay darkmode if darkmode is enabled.

### **Actual behavior**

A flash of white, sometimes, when loading a page.

### **System Information:**
 - OS: Windows 10
 - Browser: Firefox 79
 - Dark Reader version: 4.9.16

Gusted:
- Adds a whiteflash killer style on document_start instead of checking all variables and do go trough functions.
- Adds fallback style a lot earlier to prevent white flash.
- Resolves #3523
- Resolves #1219 
- Resolves #3307 
- Resolves #1965
- Resolves #695

**DISCLAMER: HOWEVER THERE CAN BE STILL CASES WITH A WHITE FLASH CAN HAPPEN, THIS IS CAN BE HARD TO FIX AS DARK READER'S FIRST THING EXECUTING IS TO PREVENT THIS. IF THERE ARE 100% REPRODUCE ABLE CASES, OPEN UP A ISSUE WITH INFORMATION AND NO HATE AND DISRESPECT**

Even on a high-end system I noticed a nice white flash, however with this I didn't notice it. Also chrome has a nice feature to reduce memory usage and free memory of unused sites and give a even longer whiteflash, not sure how to fix that or if chrome need to do some magic to make sure to get it's dark-ish color.


Gusted:
For anyone to test this:

[build.zip for chromium based browsers](https://github.com/darkreader/darkreader/files/5136379/build.zip)
[build-firefox.zip for firefox](https://github.com/darkreader/darkreader/files/5136380/build-firefox.zip)


Instructions to side-load this:
- First unzip the `.zip` file and remember where the unpacked folder is located.

Chromium based browsers
- Open `chrome://extensions`
- Enable `Developer mode` in the right upper corner. 
- Click on `Load unpacked`. 
- Now go to the unpacked folder and select it. 
- Click on `Select folder`.
- Poof and side-loaded the extension congratulations.

Firefox
- Open the hamburger menu in the right upper corner. 
- Click on Addons. 
- Click on the setting icon. 
- Click on `Debug Add-ons`.  **WARNING SOMEHOW THIS IS A WHITE-ISH SITE**
- Click on `Load Temporary Add-On`. 
- Now go to the unpacked folder and select `manifest.json`. 
- Click on `Open`.
- Poof and side-loaded the extension congratulations.

Gusted:
Also @alexanderby should I add the script under [`updateContentScript`](https://github.com/darkreader/darkreader/blob/master/src/background/tab-manager.ts#L92)

src/inject/fallback.ts
```diff
@@ -0,0 +1,8 @@
+const fallbackStyle = document.createElement('style');
+fallbackStyle.textContent = 'html,body,body :not(iframe) { background-color:#181a1b!important;border-color:#776e62!important;color:#e8e6e3!important;';
```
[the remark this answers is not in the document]
Gusted: Could you explain?

This is the **fallback** style that means everything just has a base color so it won't be white in your eyess. It's reset after a the dynamic styles comes.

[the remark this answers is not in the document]
erkinalp: I mean, the base colours that are set in the browser preferences that are applied before any CSS overrides them.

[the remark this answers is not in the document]
Gusted: That's the problem. When the document is loaded the base colors are gone and it's switch to the website render. And the goal of darkreader is to make light pages dark so if it switches to the website it's a light flash. However with this being injected into the page it will not flash a white-ish color. Nothing is being changed the order of injecting this css is just now the first thing to do. Else it would be after some microcode that is causing the white flash.

src/inject/fallback.ts
```diff
@@ -0,0 +1,8 @@
...
+document.documentElement.appendChild(fallbackStyle);
```
alexanderby, line 3: Not sure if it works in FF.

src/inject/fallback.ts
```diff
@@ -0,0 +1,8 @@
...
+// After inserting into DOM no need for any performance reducing so all `meta/optional` things are being handled.
```
alexanderby, line 4: You can remove comments.

alexanderby (changes requested):

src/inject/dynamic-theme/index.ts
```diff
@@ -49,14 +49,9 @@ function stopStylePositionWatchers() {
 }
 
 function createStaticStyleOverrides() {
```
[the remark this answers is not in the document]
Gusted: Sure.

src/inject/fallback.ts
```diff
@@ -0,0 +1,8 @@
+const fallbackStyle = document.createElement('style');
```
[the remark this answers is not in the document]
Gusted: This is more micro-code which I exactly want to elimate. It should be as optimized/fast as it can be. Also Linux(part of it) is suffering a bug that this can't be dark #2722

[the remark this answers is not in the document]
alexanderby: I tried
```javascript
var start = performance.now();
var matches = matchMedia('(prefers-color-scheme: dark)').matches;
var linux = navigator.appVersion.includes('Linux');
var end = performance.now();
console.log(matches, linux, end - start);
```
and it gave `0.05ms` in Chrome and `0` in Firefox, so I think it's very fast. So use a check like (I have no Linux to check if it works)
```javascript
if (matchMedia('(prefers-color-scheme: dark)').matches || navigator.appVersion.includes('Linux')) {
    // ...
}
```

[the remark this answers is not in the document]
alexanderby: There's also `navigator.platform`, I don't know if it is more reliable.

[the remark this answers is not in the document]
Gusted: Well, whatever chromium has some native dark page loading color when it's detected for `(prefers-color-scheme: dark)`

Also a low system `0.15ms` in chrome `2ms` in firefox. So hold in mind a normal 60fps(Chrome/Firefox renders aims to) will need to get this code injected within `16.66...ms` Know as the first frame evaluted the CSS and than painted.

src/inject/fallback.ts
Gusted: Firefox is already know to prevent whiteflashes also have great customizations likes shadowFox what even prevent it more. It's rather chromium and it's flashes as

> Even on a high-end system I noticed a nice white flash, however with this I didn't notice it. 

This was on ungoogled chromium and stable google chrome. And the high end system is fast fast and renders things within seconds like gmail(with dark reader enabled on dynamic mode).

Gusted: Oh this for sure works to prevent #3523 On Firefox(Low-end system is very noticeable on firefox for lightmode, now It's not with this change for sure)

src/inject/fallback.ts
```diff
@@ -1,8 +1,8 @@
-const fallbackStyle = document.createElement('style');
-fallbackStyle.textContent = 'html,body,body :not(iframe) { background-color:#181a1b!important;border-color:#776e62!important;color:#e8e6e3!important;';
-document.documentElement.appendChild(fallbackStyle);
-// After inserting into DOM no need for any performance reducing so all `meta/optional` things are being handled.
-// Also using document.documentElement so it's no need to wait until the `head` element is loaded.
-fallbackStyle.media = 'screen';
-fallbackStyle.classList.add('darkreader');
-fallbackStyle.classList.add('darkreader--fallback');
+if (matchMedia('(prefers-color-scheme: dark)').matches) {
```
erkinalp, line 1: long time wanted feature again

erkinalp (approved):

src/inject/fallback.ts
erkinalp: but I think you have forgot a `!` before `matchMedia()`.

Gusted: This isn't a feature, but rather a check to see if it should block white flash or not.

erkinalp: `prefers-color-scheme: dark` check is incorrect, though. You should check the darkreader's theme preference to choose whether to apply whiteflash prevention instead.

erkinalp: prefers-color-scheme: dark check is incorrect, though. You should check the darkreader's theme preference to choose whether to apply whiteflash prevention instead.

erkinalp (changes requested):

src/inject/fallback.ts
Gusted: Okay see, that is the problem I have `16.66....ms` to get this code executed and evaluated etc. Now going to the chrome.storage API would first need to get local API(Fast enough), but than also to check if sync is enabled so than It will go to sync server and get the data. Well I can tell, that isn't fast enough the be executed within 1 Render frame. So going trough the initial start procedure Isn't going to work as with this PR I'm bypassing it for this exact reason.

Gusted: And the check is correct. It was suggested by the owner of Dark Reader @alexanderby https://github.com/darkreader/darkreader/pull/3533#discussion_r479426454

src/inject/dynamic-theme/index.ts
```diff
@@ -49,7 +49,8 @@ function stopStylePositionWatchers() {
...
-    const fallbackStyle = createOrUpdateStyle('darkreader--fallback');
+    const fallbackStyle = createOrUpdateStyle('darkreader--fallback', document);
+    createOrUpdateStyle('darkreader--fallback', document).remove();
```
alexanderby, line 53: Are you trying to remove it for placing into HEAD later? No need, `insertBefore` will do this.

src/inject/fallback.ts
```diff
@@ -0,0 +1,8 @@
...
+    const fallbackStyle = document.createElement('style');
+    fallbackStyle.textContent = 'html, body, body :not(iframe) { background-color: #181a1b !important; border-color: #776e62 !important; color: #e8e6e3 !important; }';
+    document.documentElement.appendChild(fallbackStyle);
+    fallbackStyle.classList.add('darkreader');
+    fallbackStyle.classList.add('darkreader--fallback');
+    fallbackStyle.media = 'screen';
+}
```
alexanderby, line 8: Missing new line at EOF.

alexanderby (changes requested):

src/inject/dynamic-theme/index.ts
Gusted: Got ya

Gusted:
Now this eliminate but adds black flash when it's in light scheme

erkinalp:
> it adds black flash when it's in light scheme

See my review. There should be a way to force preferences to be retrieved locally.

Gusted:
> See my review. There should be a way to force preferences to be retrieved locally.

See my comment, why I am not going to add that. Also it would be even harder now I know the fallback.js is not linked with the background-script(what only can retrieve this kind of data) so a intern connection should be made as well. And than we need async code or callbacks.

alexanderby (changes requested):
When the extension is disabled (turned Off through Dark Reader's menu), websites colors are broken.

alexanderby (changes requested):
This will not work if a website is in Site List. It creates a concurrency problem between messages from 2 scripts.

Maybe try removing fallback in `inject/index.ts` `onMessage` function where necessary.

Gusted:
@alexanderby.

Huh? It's already preventing white flash on sites in Site list. And the clean-up is always being called on sites that aren't being modified by dark reader e.g. Sites in Site list.

PR #3537, opened 2020-08-28
fix for matters.news

linsui:
Fix pop up sign in banner when scrolling to bottom in an article.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3508,8 +3508,8 @@ CSS
 matters.news
 
 CSS
-.jsx-178004874.container {
-  background-image: none !important;
+section[class$="container"][style^="background-image"] {
+    background-image: none !important;
 }
 
 ================================
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3505,6 +3505,15 @@ CSS
 
 ================================
 
+matters.news
+
+CSS
+.jsx-178004874.container {
+  background-image: none !important;
+}
+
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -3505,6 +3505,15 @@ CSS
 
 ================================
 
+matters.news
+
+CSS
+.jsx-178004874.container {
```
Gusted, line 3511: `.jsx-178004874.`
JSX seams to be [React JSX](https://reactjs.org/docs/introducing-jsx.html),
and `178004874` seems to be some randomized number.

So I guess a 
```CSS
section[class$="container"][style^="background-image"]
```
Also isn't a INVERT better?

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
linsui: Thanks! I'll test it later.

linsui: INVERT inverts its child elements.

```
INVERT
section[class$="container"][style^="background-image"]
section[class$="container"][style^="background-image"] > h2
section[class$="container"][style^="background-image"] > p
section[class$="container"][style^="background-image"] > div
```

sshot-009

or

```
CSS
section[class$="container"][style^="background-image"] {
}
```

sshot-008

which is better?

Gusted: Second one

linsui: Updated, thanks!

PR #3552, opened 2020-08-30
Wait for link element to be loaded.

### LINKED ISSUE #835: [Chrome] Dynamic only works after toggling twice on https://docs.python.org
I am using Dark Reader in Chrome 70.0.3538.77 with the "Dynamic" theme engine. 

When I open https://docs.python.org, the main frame is shown in white (there are some modifications in the background). This can be fixed by turning Dark Reader off and on again. 

Screencast: https://streamable.com/1br2s

Gusted:
- Wait until the link element has been loaded.
- Resolves #835


Gusted:
However, this should resolve the error and handling(what's being done). It doesn't show any progress on the styles itself.

erkinalp (approved):

Gusted:
@erkinalp Does it works for you? For me it's still persists as problem.

src/inject/dynamic-theme/css-rules.ts
```diff
@@ -17,13 +18,8 @@ export function iterateCSSRules(rules: CSSRuleList, iterate: (rule: CSSStyleRule
...
                     logInfo('An possible attempt on a unloaded CSSImportRule has been made.');
-                        (rule.parentStyleSheet.ownerNode as HTMLLinkElement).onload = () => {
-                        try {
-                            iterateCSSRules(rule.styleSheet.cssRules, iterate);
-                        } catch (err) {
-                            logWarn(err);
-                        }
-                    };
+                    await linkLoading(rule.parentStyleSheet.ownerNode as HTMLLinkElement);
+                    iterateCSSRules(rule.styleSheet.cssRules, iterate);
```
erkinalp, line 22: thanks

erkinalp (approved):

src/inject/dynamic-theme/css-rules.ts
Gusted: It's still a POC as it doesn't work for me so not sure where the fault relies.

11bit:
Hey guys, I have experimented a bit with this issue and found the change that seems to be the cause of the problem: https://github.com/darkreader/darkreader/commit/7185820f7d6865dbd908ee3196eaf080bf3c90b4#diff-3d04f9c5378d3d93c42df303885b05a5L144

Probably the change from `if (cssRules == null || accessError) {` to `if ((cssRules && !accessError) || isStillLoadingError(accessError)) {` is not correct because cssRules for links in load state in Chrom is null.

What do you think about changing the condition like this:
```diff
index 2b73dfc0..04d703d8 100644
--- a/src/inject/dynamic-theme/style-manager.ts
+++ b/src/inject/dynamic-theme/style-manager.ts
@@ -143,7 +143,7 @@ export function manageStyle(element: StyleElement, {update, loadingStart, loadin
                 logWarn(accessError);
             }

-            if ((cssRules && !accessError) || isStillLoadingError(accessError)) {
+            if ((cssRules === null && !accessError) || isStillLoadingError(accessError)) {
                 try {
                     await linkLoading(element);
                 } catch (err) {
```

Gusted:
`  if ((cssRules === null && !accessError)`

Hmm seems it's now checking again if it's null instead of checking if it exist. That's inverted logic. @alexanderby?

erkinalp (approved):

alexanderby:
Hmm, sounds reasonable, it used to check for `null` before, maybe I was not attentive. But then surprising how it works. You can simplify the condition with `if (!(cssRules || accessError) || isStillLoadingError(accessError))...`.

src/inject/dynamic-theme/css-rules.ts
```diff
@@ -3,7 +3,7 @@ import {parseURL, getAbsoluteURL} from './url';
 import {logWarn} from '../utils/log';
 
 export function iterateCSSRules(rules: CSSRuleList, iterate: (rule: CSSStyleRule) => void) {
-    forEach(rules, (rule) => {
+    forEach(rules, async (rule) => {
```
alexanderby, line 6: Is `async` needed there?

alexanderby (changes requested):

src/inject/dynamic-theme/css-rules.ts
Gusted: Not anymore

alexanderby:
Works unstable for me at https://docs.python.org/3/

Gusted:
Chromium is having a bug, which I partially understand but can't explain.  Tried to in https://github.com/darkreader/darkreader/pull/3552/commits/6d00650e25caaff8f93e9beb69d72689f3b9fe22

Gusted:
But their still seems to be some condition that the `sync` and `async` of iterateCSSType is conflicting and therefor still not loaded correctly.

erkinalp (approved):

Gusted:
Very unreliable - Overseeded by #6102

PR #3553, opened 2020-08-30
fix for flow.polar.com

Myshor:
Some of changeable highcharts had white background in after change to other - forced darkreader-inline-fill.

1st OK:
obraz

Change to next:
Before:
obraz

After:
obraz


Myshor:
Remove as it started to break other highcharts. :(

PR #3556, opened 2020-08-31
Fix for theguardian.com and added garamlee500.github.io to dark websites

garamlee500:
Fixed issue where logo was not inverted for some articles

PR #3557, opened 2020-08-31

garamlee500:
Fix for theguardian.com where certain news articles (mainly comment/opinion) didn't have logo inverted. Example:
https://www.theguardian.com/commentisfree/2020/aug/31/the-rights-culture-war-politics-rightwing-fantasy-elections

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5457,7 +5457,7 @@ theguardian.com
 
 INVERT
 .inline-the-guardian-logo__svg
-.css-t7dp8u
+a[data-link-name^="logo"] > svg
 
 ================================
 
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5457,6 +5457,7 @@ theguardian.com
 
 INVERT
 .inline-the-guardian-logo__svg
+.css-t7dp8u
 
 ================================
 
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5457,7 +5457,7 @@ theguardian.com
 
 INVERT
 .inline-the-guardian-logo__svg
-a[data-link-name$^="logo"] > svg
+a[data-link-name$="logo"] > svg
 
 ================================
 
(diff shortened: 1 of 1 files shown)


garamlee500:
Note: This is a Rule 4 compliant version of #3556

Gusted:
Hi @garamlee500!

I can't find the element that is supposed to be inverted.


Regards,
Gusted

garamlee500:
Hi @Gusted!
That's odd
image

image

For some reason it can't be searched from javascript

garamlee500:
Here's a demo of my fix:

Before:
image

After:
image

Gusted:
```javscript
document.getElementById
```
Only works for elements that have a `<p id="Hi_this_is_a_ID"></p>` id attribute.

Also it seems like all your classes are being normalized? Do you have any other extension enabled or something for the Guardian as the original class is suppose to be `inline-the-guardian-logo__svg` which is already being inverted.

garamlee500:
Good point about the id selector.

I disabled all extensions (exluding darkreader) and I still get the same black logo.
The homepage is ok without the fix though
image

For some reason, only articles within opinion seem to not be inverted

garamlee500:
Here's a better screenshot of the homepage without the fix (still works)
image

garamlee500:
For some reason, the problem doesn't seem to occur on private windows
(I'm logged in when i'm on my normal browser)

Not logged in:
image

Logged in:
image

Gusted:
Hmm weird, could you log into your account in a private window and check if the class is `css-t7dp8u`

garamlee500:
Seems so:
image

Gusted:
Okay, I think it's better to have the CSS.
```CSS
a[data-link-name^="logo"] > svg
```
Instead of some weird class that have random numbers.

garamlee500:
Do you think that CSS should also replace the original Inversion of
inline-the-guardian-logo__svg

Gusted:
No, just add it a non-logged in page have other elements that doesn't work with the logged-in CSS.

garamlee500:
I tested and fixed it!

garamlee500:
Hang on, it doesn't seem to work now suddenly

Gusted:
Oh change the `^` to `$`

garamlee500:
Thanks @Gusted it works now!

garamlee500:
Thanks @Gusted !

PR #3566, opened 2020-09-01

garamlee500:
Fix for chess.com where black pieces would be inverted by darkreader making them extremely white (only broken on daily matches for some reason)

View: https://www.chess.com/daily/game/274666002

Before fix:
image

After fix:
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -901,8 +901,8 @@ CSS
 
 chess.com
 
-INVERT
-div[class^='piece b']
+IGNORE IMAGE ANALYSIS
+.piece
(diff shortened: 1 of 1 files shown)


Gusted (changes requested):
Hi @garamlee500!

Good catch and the fix, however I think to even better improve to fix this bug. This CSS could be used:
```CSS
chess.com

IGNORE IMAGE ANALYSIS
.piece

```

Regards,
Gusted

garamlee500:
Hi @Gusted 

Good suggestion, I'll fix it now

Regards,
Garam Lee

garamlee500:
Thanks @Gusted !

Issue #779, opened 2018-10-17
Firefox not inverting Video color but Chrome does.


bmbeverst:
I have been using the Chrome version but now that you support Firefox I am moving back to it. I noticed that the video's on canvas.vt.edu are inverted in Chrome, which I like, but not in Firefox. It looks like the Chrome implementation is "broken" but I would like to "break" Firefox in the same way.

How can I configure DarkReader on Firefox to invert the colors of videos on canvas.vt.edu?

Thanks in advance.

P.S. This issue seems related: https://github.com/darkreader/darkreader/issues/313

alexanderby:
Hi, on what website this happens? Are you using Filter mode?

bmbeverst:
I am using the filter mode.

It is on the Virginia Tech college internal class site but the media is hosted on by Kaltura. I am not able to reproduce the inversion in with any of their other videos.

Here is a screen shot from Chrome. I would like Firefox to also invert the colors.
![screenshot_20181108_213335](https://user-images.githubusercontent.com/603632/48239719-fc79a000-e39d-11e8-920b-856f8233fc79.png)

bmbeverst:
Just noticed some thing, Firefox does correctly invert the colors when the video is full screen. This does not occur on Youtube, no inversion full screen or not.

Is there a way to control which sites have inverted video?

alexanderby:
Please reopen if there is still the same issue.

bmbeverst:
Thanks, I just graduated, so I can't retest. @snoop2head thanks for the PR!

PR #3573, opened 2020-09-02
Fix for yscec.yonsei.ac.kr

snoop2head:
### Dark Reader on Current version
Screen Shot 2020-09-02 at 12 51 50 PM

### Dark Reader on PR version
Screen Shot 2020-09-02 at 12 52 15 PM

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -7026,6 +7026,14 @@ paper-button.ytd-subscribe-button-renderer[subscribed] {
 
 ================================
 
+yscec.yonsei.ac.kr
+
+INVERT
+iframe
+#page-sidebar
+
+================================
+
 yuque.com
 
 INVERT
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -2518,3 +2518,11 @@ CSS
 #player-theater-container {
     background: none !important;
 }
+
+================================
+
+yscec.yonsei.ac.kr
+
+INVERT
+iframe
+#page-sidebar


Gusted:
Hi @snoop2head!

As this content is only available to students, I unfortunate can't see the fixes for myself.
So the side-bar is obvious, however the iframe what is inside the iframe, just some text for assignments. As Iframes can be used for a lot of things and could contain pictures/images(Not sure if this is the case) and they will be inverted as well.

Regards,
Gusted

snoop2head:
Thanks for the quick reply @Gusted!

The purpose was to make lecture videos on white background to become dark.
For me and other students like #779, inverting the colors of video is favorable.

Korean undergraduate students' consensus is that they are having sore eyes from watching white background videos for 20hrs/week. This is due to the social status to hold lectures online only, part of social distancing policy of the government.

As you have mentioned, iframe can be used for pop-up, images, video etc. 

However, yscec.yonsei.ac.kr uses iframe as providing video contents only. The website's iframe tag doesn't even have id or class name, since it is only used for video. Videos on website use iframe, whereas other medias are provided in form of downloadable file. There's another student portal that utilizes iframe for multiple purposes, but yscec isn't the case.

I hope the PR makes it to the update, so the other students can benefit from it.

Gusted:
> The purpose was to make lecture videos on white background to become dark.

I assume with lecture videos, just a power point presentation or something that way so that it's still visible to look at it.

snoop2head:
@Gusted Precisely!

Gusted:
Got ya, than I will merge it

Issue #3411, opened 2020-08-14
[Broken Website] onedrive


linsui:
**Website**
https://onedrive.live.com/?id=22FB491072D6A964!225&cid=22FB491072D6A964

**How to Reproduce**
An example of this is
> 1. Go to a onedrive file page
> 2. The menubar is blinking

**Screenshots**
![sshot-005](https://user-images.githubusercontent.com/36977733/90261790-fc44e900-de3c-11ea-8d33-3f4246b9e53a.png)
![sshot-006](https://user-images.githubusercontent.com/36977733/90261792-fcdd7f80-de3c-11ea-8bcd-302021ea7644.png)

**System info:**
 - OS: Windows 10
 - Browser: Firefox
 - Darkreader Version: 4.9.16
 
**Additional context**
Disable and re-enable dark reader, the problem will disappear.

Gusted:
Hi @linsui!

Thank you for reporting this issue,
it seems like that #3138 don't work for Microsoft apps yet(Just like online word is 1 big shining).
I will have a look soon to see what it's the cause of this issue!

Regards,
Gusted

linsui:
Thanks!

Gusted:
I determined kind of the source to the problem. As I'm a lazy person I will copy paste my message to the owner of dark reader(Alexander) and hopefully can help me out with this weird problem

```
Hey @CEO - Alexander Did you remember the flashing online word?
Well it seems like a trace to the problem has found.
As it was flashing in less than 2 seconds it seems to me that there were 
multiple observers on the node or something else that cause to rebuild the style sheet and shine word. 
https://github.com/darkreader/darkreader/blob/master/src/inject/dynamic-theme/style-manager.ts#L263 
Removing this line is making the flashy parts just leaving white https://i.imgur.com/mSdyd1j.png
I'm really unsure why this is causing the flashing the force part is just forcing to rebuild the stylesheet.
It shouldn't trigger the watcher for any reason and think it's has a new prev-sibling right?
Other than rebuilding the style so it gets it stylesheet back
```

Gusted:
```
alexanderby Today at 16:39
I've submitted an issue to merge-styles package https://github.com/microsoft/fluentui/issues/14735 that is the reason for removing Dark Reader's style.
```
Well it official became the waiting game for Microsoft and than for sites to update it's scripts.

Gusted:
https://codepen.io/Gusted/pen/gOrmXeL
[...]

PR #3574, opened 2020-09-02
Attempt to fix the restore logic

Gusted:
- Somehow we are detecting our own style being moved around by our own logic?
- Non resolved as this doesn't do the right fix.


Gusted:
@alexanderby the flashing has a repeated pattern in the mutationObject so detecting that and returning it seems to be fix in a lot of of places by having non styles at all, but for onedrive(#3411) it seems to be getting delayed?

PR #3614, opened 2020-09-08
Fix LinkedIn card prompt background

BurnhamG:
Before: 
firefox_RgTz3pfXtg

After:
firefox_oDl3bZXSdl

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3284,6 +3284,9 @@ li-icon[type="linkedin-logo"],
 .bug-text-color {
     fill: ${black} !important;
 }
+.pds-ge-entry-card__card {
+    background: transparent !important;
+}
 .vjs-load-progress {
     background-color: hsla(0,0%,100%,.3) !important;
 }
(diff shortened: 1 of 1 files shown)


Gusted:
Hi @BurnhamG!

Does the element have by default no background-color?

Regards,
Gusted

BurnhamG:
Hi @Gusted, the default was to have a linear gradient. I must apologize, I jumped the gun a bit with this PR. My latest commit preserves the linear gradient that was on the original card, using the correct CSS variables. Here's a new after image:

image

src/config/dynamic-theme-fixes.config
```diff
@@ -3284,6 +3284,15 @@ li-icon[type="linkedin-logo"],
 .bug-text-color {
     fill: ${black} !important;
 }
+.pds-ge-entry-card__card {
+  background: linear-gradient(
```
Gusted, line 3288: You can use the linear-gradient on 1 line. No need for spaces of this CSS property.

Gusted (changes requested):

PR #3634, opened 2020-09-13
More Firefox browser theme colors

### LINKED ISSUE #1940: New Tab Site White in Dark Mode after Restart
The Darkmode works fine for the about:newtab Site, after a restart all sites are as normal,
but the New Tab site is completely White, and I can only change it, when I reapply the Firefox Dark Theme. Without Dark Reader the Problem doesnt exists.

For now, you can change this Behavior, if you install the Addon "New Tab Override", there you can set a Dark Color for the New Tab.

Gusted:
- Add sidebar coloring
- Add about:home coloring
- Resolves #1940

Read more about the colors https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/theme#colors


Gusted:
Before:


After:

PR #3635, opened 2020-09-13
Fix memory leak for recursive imports

### LINKED ISSUE #3626: https://groupme.com/signin | Memory Leak
**Website**
https://web.groupme.com/signin
https://web.groupme.com/chats

**How to Reproduce**
Enabling DarkReader (in Dynamic Mode) on GroupMe causes an infinite increase in memory usage; this behavior is seen on both Firefox and Google Chrome. This _does not_ occur on the main landing page.

Google Chrome seemingly has protections halting tabs at ~4GB of RAM usage, while Firefox does not- causing the tab crash your machine from out of memory errors.

**Screenshots**
- [GroupMe tab using >24GB RAM.](https://i.gyazo.com/b4af4e47f1fc5693a40abda83de6c08d.gif)
- [Google Chrome's dev-console error.](https://i.gyazo.com/1745c752250c95f75958adc9be4dbd4c.png)

**System Info:**
 - OS: Windows 10 Pro : Version 2004, Build 19041.508
 - Browser: Firefox 80.0.1 (64-bit); Chrome 85.0.4183.102 (64-bit)
 - Dark Reader Version: 4.9.19
 - System: Ryzen 7 2700, 32GB RAM (3000Mhz), GTX 980

**Additional Context:**
I have tested disabling all other extensions, which validates it is Dark Reader causing the memory leak.

Gusted:
- Fix a memory leak whereby recursive the same file is being replaced by imports.
- Resolves #3626

Not sure whatever we should use the `absoluteURL` as key or `importURL` as key, but in both situation it seems to work for this issue. Also cleaning the alreadyReplaced map after using it so it will remove all the memory of the cached CSS.

### DIFF
--- a/src/inject/dynamic-theme/index.ts
+++ b/src/inject/dynamic-theme/index.ts
@@ -2,7 +2,7 @@ import {replaceCSSVariables, getElementCSSVariables} from './css-rules';
 import {overrideInlineStyle, getInlineOverrideStyle, watchForInlineStyles, stopWatchingForInlineStyles, INLINE_STYLE_SELECTOR} from './inline-style';
 import {changeMetaThemeColorWhenAvailable, restoreMetaThemeColor} from './meta-theme-color';
 import {getModifiedUserAgentStyle, getModifiedFallbackStyle, cleanModificationCache, parseColorWithCache, getSelectionColor} from './modify-css';
-import {manageStyle, getManageableStyles, StyleElement, StyleManager} from './style-manager';
+import {manageStyle, getManageableStyles, StyleElement, StyleManager, alreadyReplaced} from './style-manager';
 import {watchForStyleChanges, stopWatchingForStyleChanges} from './watch';
 import {forEach, push, toArray} from '../../utils/array';
 import {removeNode, watchForNodePosition, iterateShadowHosts, isDOMReady, addDOMReadyListener, removeDOMReadyListener} from '../utils/dom';
@@ -448,4 +448,5 @@ export function cleanDynamicThemeCache() {
     cancelRendering();
     stopWatchingForUpdates();
     cleanModificationCache();
+    alreadyReplaced.clear();
 }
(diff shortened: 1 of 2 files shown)


src/inject/dynamic-theme/index.ts
```diff
@@ -448,4 +448,5 @@ export function cleanDynamicThemeCache() {
     cancelRendering();
     stopWatchingForUpdates();
     cleanModificationCache();
+    alreadyReplaced.clear();
```
alexanderby, line 451: `alreadyReplaced` sounds strange out of context. Rename to `cssImportCache`.

src/inject/dynamic-theme/style-manager.ts
```diff
@@ -186,6 +186,7 @@ export function manageStyle(element: StyleElement, {update, loadingStart, loadin
             // so need to load CSS text and insert it into style element
             try {
                 const fullCSSText = await replaceCSSImports(cssText, cssBasePath);
+                alreadyReplaced.clear();
```
alexanderby, line 189: Is there a need to clear it here? Suppose 2 `<link>` elements are loading CSS files at the same time. You will clear all for them. Cleaning the cache sounds reasonable for freeing up the memory.

Maybe try restructuring the `replaceCSSImports` like:
```javascript
async function replaceCSSImports(cssText: string, basePath: string, cache: Map<string, string>) {
    // ...
    importedCSS = await replaceCSSImports(importedCSS, getCSSBaseBath(absoluteURL), cache);
    // ...
}
```
Then you will use this function like:
```javascript
const cssImportCache = new Map<string, string>();
const fullCSSText = await replaceCSSImports(cssText, cssBasePath, cssImportCache);
cssImportCache.clear(); // even if you leave it without clear(), Garbage Collector should clean up the memory, because `cssImportCache` variable looses all the references to it when the function exits.
```
Then there will be no need to export the function.

Or even better would be to clear it in `finally` block, that is called when there is an error or no error like:
```javascript
const cssImportCache = new Map<string, string>();
try {
    // ...
} catch(err) {
    // ...
} finally {
    cssImportCache.clear();
}
// or just do `cssImportCache.clear();` here, after `try...catch` without `finally`, should be no difference in this case.
```

alexanderby (changes requested):

src/inject/dynamic-theme/style-manager.ts
Gusted: I've just added a extra variable to pass along the cache. So the garbage collector will remove free the memory of it when the function is done 🤷🏽‍♀️.

alexanderby (approved):
Minor notice: `cache: Map<string, string> = new Map<string, string>()` can be shortened to `cache = new Map<string, string>()`

PR #3646, opened 2020-09-15
Update dynamic-theme-fixes.config - orf.at/corona/stories/daten - invisible tooltips

Markus87:
orf.at/corona/stories/daten - tooltips of diagram were not readable because the background was bright and the text color was white

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4431,6 +4431,13 @@ INVERT
 
 ================================
 
+orf.at/corona/stories/daten
+
+INVERT
+.bg
+
+================================
+
 overleaf.com
 
 INVERT


src/config/dynamic-theme-fixes.config
```diff
@@ -4431,6 +4431,13 @@ INVERT
 
 ================================
 
+orf.at/corona/stories/daten
```
Gusted, line 4434: Is their any specific reason the path is being used? I think `orf.at` is enough.

Gusted (changes requested):

Markus87:
Yeah, it is the only place where I noticed the problem. The rest of the page is news articles, only the corona page has complex graphs with popups that are very difficult to read.
Not sure if this could cause problems with the rest of the site.

Am 16. September 2020 19:07:59 MESZ schrieb Gusted <notifications@github.com>:

PR #3654, opened 2020-09-16
fix for leetcode-cn.com

linsui:
I didn't test on leetcode.com because of the redirect.

Before

sshot-015

After

sshot-014

I can see no difference with or without `.CodeMirror-cursor`. Is it still useful?

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3262,12 +3262,16 @@ leetcode.com
 leetcode-cn.com
 
 INVERT
+.cursor
 .CodeMirror-cursor
 .user-story-chapter-base .companies-showcase-base .logo
 
 CSS
 [class^=question-picker-detail] {
     background: none !important;
+}
+.monaco-editor, .monaco-editor-background, .monaco-editor .margin{
+    background-color: black;
 }
 
 ================================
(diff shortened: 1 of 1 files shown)


felixonmars:
The "Before" color looks dark to me already :(

linsui:
It's dark, but the dark yellow background is weird (and ugly IMO). It's white without dark theme and should be black in dark theme. I don't know why it's so yellow.

src/config/dynamic-theme-fixes.config
```diff
@@ -3262,13 +3262,17 @@ leetcode.com
...
+.cursor
 .CodeMirror-cursor
 .user-story-chapter-base .companies-showcase-base .logo
 
 CSS
 [class^=question-picker-detail] {
     background: none !important;
 }
+.monaco-editor, .monaco-editor-background, .monaco-editor .margin{
+    background-color: black;
```
Gusted, line 3274: `var(--darkreader-neutral-background) !important;`

src/config/dynamic-theme-fixes.config
```diff
@@ -3262,13 +3262,17 @@ leetcode.com
...
 INVERT
+.cursor
 .CodeMirror-cursor
 .user-story-chapter-base .companies-showcase-base .logo
 
 CSS
 [class^=question-picker-detail] {
     background: none !important;
 }
+.monaco-editor, .monaco-editor-background, .monaco-editor .margin{
```
Gusted, line 3273: `.margin{` -> `.margin {`

Gusted (changes requested):

PR #3661, opened 2020-09-17
Add support to CSSSupportsRule

### LINKED ISSUE #3648: [Broken Website]
**Website**
https://www.dagbladet.no/

**How to Reproduce**
View the website with Darkreader enabled.

**Expected behavior**
Should have dark background and light text.

**Actual behavior**
Shows the website with white background and black text, same as with Darkreader disabled.

**Screenshots**
image

**System info:**
 - OS: Windows 10
 - Browser: Microsoft Edge 85
 - Darkreader Version: 4.9.19

Gusted:
- Add support to iterate over CSSRules of CSSSupportsRule.
- Resolves #3648

### DIFF
--- a/src/inject/dynamic-theme/css-rules.ts
+++ b/src/inject/dynamic-theme/css-rules.ts
@@ -18,7 +18,9 @@ export function iterateCSSRules(rules: CSSRuleList, iterate: (rule: CSSStyleRule
                 logWarn(err);
             }
         } else if (rule instanceof CSSSupportsRule) {
-            iterateCSSRules(rule.cssRules, iterate);
+            if (CSS.supports(rule.conditionText)) {
+                iterateCSSRules(rule.cssRules, iterate);
+            }
         } else {
             logWarn(`CSSRule type not supported`, rule);
         }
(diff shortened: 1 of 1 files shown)


src/inject/dynamic-theme/css-rules.ts
```diff
@@ -17,6 +17,8 @@ export function iterateCSSRules(rules: CSSRuleList, iterate: (rule: CSSStyleRule
             } catch (err) {
                 logWarn(err);
             }
+        } else if (rule instanceof CSSSupportsRule) {
+            iterateCSSRules(rule.cssRules, iterate);
```
alexanderby, line 21: Great!

Now, there is `rule.conditionText`. Would be great to iterate style when the condition is supported (e.g. when there is a condition for IE, it will be not good to iterate styles).

I've checked, there is `[CSS.supports](https://developer.mozilla.org/en-US/docs/Web/API/CSS/supports)` function. So try something like:
```javascript
if (CSS.supports(rule.conditionText)) {
    iterateCSSRules(rule.cssRules, iterate);
}
```

alexanderby (changes requested):

alexanderby (approved):

PR #3668, opened 2020-09-18
Fix non-inverted text in confluence

ThinkChaos:


Gusted:
Hi @ThinkChaos!

Do you have a after/before screenshot?

Regards,
Gusted

ThinkChaos:
Sure!

Before:
before

After:
aftert

PR #3673, opened 2020-09-20
Update de.config

RedcodesDev:


src/_locales/de.config
```diff
@@ -17,8 +17,7 @@ Aus
 Aktuelle Seite umschalten
 
 @setup_hotkey_toggle_site
-Hotkey zum Umschalten der
-aktuellen Webseite einrichten
+Stelle einen Hotkey zum umstellen ein
```
Gusted, line 20: This actually has to be splits into 2 lines.
The first one should be to describe it's to setup the current site.
The second one should be a translation of `toggle hotkey`

Gusted (changes requested):

PR #3674, opened 2020-09-20
Improve shortcuts

### LINKED ISSUE #1957: Feature request: key shortcut to toggle between light/dark
Super useful extension, usually the first I download when I get a new computer. One thing that would make it even better is if there was a hotkey toggle between light and dark mode, instead of having to switch the extension off altogether.

Gusted:
- Added a new shortcut to switch between schemes.
- Changed `console.log` to `logInfo`.
- It will first check if the active Tab has a preset/custom theme and make changes according to it.
- Resolves #1957


Gusted:
Woops I've made a branch of lazyloading-Patch

PR #3691, opened 2020-09-23
Use a new mechanism to inject Fallback

### LINKED ISSUE #3593: [Broken Website] https://personal.vanguard.com
**Website**
https://personal.vanguard.com/us/faces/com/vanguard/logon/view/LogonLogoff.xhtml

**How to Reproduce**
 1. Go to https://personal.vanguard.com/us/faces/com/vanguard/logon/view/LogonLogoff.xhtml
2. Blank Page. 

**Expected behavior**
There should be content when I go to that page. Makes vanguard.com unusable. When you turn off Dark reader content renders fine.

**Actual behavior**
After login the website should render, but it does not.

**System info:**
 - OS:MacOS
 -  chrome
 - Darkreader Version: latest

### LINKED ISSUE #3653: [BUG] Roundcube Mail is broken if Dark Reader 4.9.17 or later is loaded (even if inactive)
### **Describe the bug**

Roundcube Mail web client cannot save or send email when Dark Reader 4.9.17 is loaded (even if inactive). First noticed with 4.9.19 as the updates from 4.9.17 onward happened while I wasn't using this system

### **Steps to reproduce**

This is tricky - you need an instance of Roundcube Mail to try this with, and it's not something you'd normally have public access to

### **Expected behavior**

Roundcube Mail should work normally

### **Actual behavior**

When sending an email or saving a draft, Roundcube just hangs until it finally times out.

### **System Information:**

Affects Roundcube 1.41 and 1.48 (I upgraded in the midst of debugging this) via Windows 10 Chrome 85.0.4183.83 (Official Build) (64-bit). Also fails via current 85.0.4183.102 (Official Build) (64-bit). This appears to have broken recently - within the last 10 days or less.

### **Additional context**

I've been using Dark Reader for years, and have never seen this happen. I first thought it was RC not working right with newer Chrome, but I narrowed it down to Dark Reader - even if I disable every other extension this one alone will break RC Mail in a way it didn't do 10 days or so ago when I last used it on this machine.

**I was able to find the original CRX fles for Dark Reader 4.9.16 through 4.9.19. After carefully editing the manifest to prevent upgrade, I was able to load each in turn unpacked, and confirm that 4.9.16 was the last version that worked properly in this regard.**

Gusted:
- Use the `createNodeAsap` logic to make sure the fallback will be injected correctly into the site.
- Use the logic to already create a `head` and append the fallback style into that.
- Resolves #3653
- Resolves #3593

### DIFF
--- a/src/inject/fallback.ts
+++ b/src/inject/fallback.ts
@@ -4,12 +4,13 @@ if (
     matchMedia('(prefers-color-scheme: dark)').matches &&
     !document.querySelector('.darkreader--fallback')
 ) {
-    createOrUpdateFallback('html, body, body :not(iframe) { background-color: #181a1b !important; border-color: #776e62 !important; color: #e8e6e3 !important; }');
+    const fallBackCSS = 'html, body, body :not(iframe) { background-color: #181a1b !important; border-color: #776e62 !important; color: #e8e6e3 !important; }';
+    createOrUpdateFallback(fallBackCSS);
 }
 
-export function createOrUpdateFallback(css: string) {
+function createOrUpdateFallback(css: string) {
     createNodeAsap({
-        selectNode: () => document.querySelector('darkreader--fallback'),
+        selectNode: () => document.querySelector('.darkreader--fallback'),
         createNode: (target) => {
             const fallbackStyle = document.createElement('style');
             fallbackStyle.textContent = css;
@@ -18,11 +19,7 @@ export function createOrUpdateFallback(css: string) {
             fallbackStyle.classList.add('darkreader--fallback');
             fallbackStyle.media = 'screen';
         },
-        updateNode: (existing) => {
-            if (css.replace(/^\s+/gm, '') !== existing.textContent.replace(/^\s+/gm, '')) {
-                existing.textContent = css;
-            }
-        },
+        updateNode: () => Function.prototype,
         selectTarget: () => document.head,
         createTarget: () => {
             const head = document.createElement('head');
--- a/src/inject/fallback.ts
+++ b/src/inject/fallback.ts
@@ -1,19 +1,21 @@
 import {createNodeAsap} from './utils/dom';
 
+const fallBackCSS = 'html, body, body :not(iframe) { background-color: #181a1b !important; border-color: #776e62 !important; color: #e8e6e3 !important; }';
+
 if (
     matchMedia('(prefers-color-scheme: dark)').matches &&
-    !document.querySelector('.darkreader--fallback')
+    !document.querySelector('.darkreader--fallback') &&
+    document.documentElement instanceof HTMLHtmlElement
 ) {
-    const fallBackCSS = 'html, body, body :not(iframe) { background-color: #181a1b !important; border-color: #776e62 !important; color: #e8e6e3 !important; }';
-    createOrUpdateFallback(fallBackCSS);
+    createOrUpdateFallback();
 }
 
-function createOrUpdateFallback(css: string) {
+function createOrUpdateFallback() {
     createNodeAsap({
         selectNode: () => document.querySelector('.darkreader--fallback'),
         createNode: (target) => {
             const fallbackStyle = document.createElement('style');
-            fallbackStyle.textContent = css;
+            fallbackStyle.textContent = fallBackCSS;
             target.appendChild(fallbackStyle);
             fallbackStyle.classList.add('darkreader');
             fallbackStyle.classList.add('darkreader--fallback');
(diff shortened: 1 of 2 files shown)


src/inject/fallback.ts
```diff
@@ -1,11 +1,34 @@
...
     matchMedia('(prefers-color-scheme: dark)').matches &&
     !document.querySelector('.darkreader--fallback')
 ) {
-    const fallbackStyle = document.createElement('style');
-    fallbackStyle.textContent = 'html, body, body :not(iframe) { background-color: #181a1b !important; border-color: #776e62 !important; color: #e8e6e3 !important; }';
-    document.documentElement.appendChild(fallbackStyle);
-    fallbackStyle.classList.add('darkreader');
-    fallbackStyle.classList.add('darkreader--fallback');
-    fallbackStyle.media = 'screen';
+    createOrUpdateFallback('html, body, body :not(iframe) { background-color: #181a1b !important; border-color: #776e62 !important; color: #e8e6e3 !important; }');
```
alexanderby, line 7: Put the CSS string into some constant.

src/inject/fallback.ts
```diff
@@ -1,11 +1,34 @@
...
+        selectNode: () => document.querySelector('darkreader--fallback'),
+        createNode: (target) => {
+            const fallbackStyle = document.createElement('style');
+            fallbackStyle.textContent = css;
+            target.appendChild(fallbackStyle);
+            fallbackStyle.classList.add('darkreader');
+            fallbackStyle.classList.add('darkreader--fallback');
+            fallbackStyle.media = 'screen';
+        },
+        updateNode: (existing) => {
```
alexanderby, line 21: I think `updateNode` can be an empty NOOP function in this case, like `Function.prototype`. If there will be a fallback style, it will not do anything.

src/inject/fallback.ts
```diff
@@ -1,11 +1,34 @@
...
-    document.documentElement.appendChild(fallbackStyle);
-    fallbackStyle.classList.add('darkreader');
-    fallbackStyle.classList.add('darkreader--fallback');
-    fallbackStyle.media = 'screen';
+    createOrUpdateFallback('html, body, body :not(iframe) { background-color: #181a1b !important; border-color: #776e62 !important; color: #e8e6e3 !important; }');
+}
+
+export function createOrUpdateFallback(css: string) {
+    createNodeAsap({
+        selectNode: () => document.querySelector('darkreader--fallback'),
```
alexanderby, line 12: Missing dot `.` in the selector

src/inject/fallback.ts
```diff
@@ -1,11 +1,34 @@
...
-    const fallbackStyle = document.createElement('style');
-    fallbackStyle.textContent = 'html, body, body :not(iframe) { background-color: #181a1b !important; border-color: #776e62 !important; color: #e8e6e3 !important; }';
-    document.documentElement.appendChild(fallbackStyle);
-    fallbackStyle.classList.add('darkreader');
-    fallbackStyle.classList.add('darkreader--fallback');
-    fallbackStyle.media = 'screen';
+    createOrUpdateFallback('html, body, body :not(iframe) { background-color: #181a1b !important; border-color: #776e62 !important; color: #e8e6e3 !important; }');
+}
+
+export function createOrUpdateFallback(css: string) {
```
alexanderby, line 10: Is there a need to export the function?

alexanderby (changes requested):

src/inject/fallback.ts
```diff
@@ -1,11 +1,31 @@
...
-    fallbackStyle.textContent = 'html, body, body :not(iframe) { background-color: #181a1b !important; border-color: #776e62 !important; color: #e8e6e3 !important; }';
-    document.documentElement.appendChild(fallbackStyle);
-    fallbackStyle.classList.add('darkreader');
-    fallbackStyle.classList.add('darkreader--fallback');
-    fallbackStyle.media = 'screen';
+    const fallBackCSS = 'html, body, body :not(iframe) { background-color: #181a1b !important; border-color: #776e62 !important; color: #e8e6e3 !important; }';
+    createOrUpdateFallback(fallBackCSS);
+}
+
+function createOrUpdateFallback(css: string) {
```
alexanderby, line 11: I don't see a purpose of passing `css` argument, it is always the same.

src/inject/fallback.ts
```diff
@@ -1,11 +1,31 @@
+import {createNodeAsap} from './utils/dom';
+
 if (
```
alexanderby, line 3: Let's add `document.documentElement instanceof HTMLHtmlElement` condition, so that it will not inject anything into SVGs and other strange files.

alexanderby (changes requested):

alexanderby (approved):

PR #3695, opened 2020-09-24
Added thingiverse.com

MikeRatcliffe:


Gusted:
Hi @MikeRatcliffe!

What's the purpose of this fix? As I see every `p, div, span` element is being targeted. 
As I the default color of thingiverse seems to be gray(which is a variant of the original color)

Regards,
Gusted

MikeRatcliffe:
I can hardly read the text if it is grey... the contrast just isn't high enough.

If you don't agree then of course you are free to close this pull request.

Gusted:
> If you don't agree then of course you are free to close this pull request.

Well I agree with you on your statement, but rather how the fix is made.

I think it's better to just select those elements that are really gray(Unless really everything is gray)>
And instead of `#fff` use `var(--darkreader-neutral-text) !important;`

Regards,
Gusted

MikeRatcliffe:
> 
> 
> > If you don't agree then of course you are free to close this pull request.
> 
> Well I agree with you on your statement, but rather how the fix is made.
> 
> I think it's better to just select those elements that are really gray(Unless really everything is gray)>
> And instead of `#fff` use `var(--darkreader-neutral-text) !important;`
> 
> Regards,
> Gusted

If only it was that easy!

Thingiverse is a React app so it's elements look like this:

```html
<div class="ThingPage__blockTitle--3ZdLu">Summary</div>
```

Targetting `.ThingPage__blockTitle--3ZdLu` is pointless because part of the class name is generated by some React plugin, probably Modernizr and the suffix will likely change in the future.

I will switch to the CSS variable though.

Gusted:
> If only it was that easy!

Well read about some [examples](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors#Examples) of current CSS selectors for attribute.

> Targetting .ThingPage__blockTitle--3ZdLu is pointless because part of the class name is generated by some React plugin, probably Modernizr and the suffix will likely change in the future.

Yep. I've seen it too many by now, but knowing only the last part changes.
```CSS
div[class^="ThingPage__blockTitle"]
```
Should do the job and doesn't break when the last suffix changes as `^=` means for `starts with`.

Regards,
Gusted

Issue #1933, opened 2020-01-27
extension creates 404 requests for images in css files with relative path


Bhoft:
This extension creates a lot of 404 requests on web servers which uses background-images in CSS files with an relative path.

e.g. i have following files:
`/test.html`
`/sub/test.html`
`/css/style.css`
`/css/images/background.png`

Content of /test.html  (for the test.html in sub directory href="../css/style.css"

    <!DOCTYPE html>
    <html>
    <head>
        <link rel="stylesheet" type="text/css" href="css/style.css">
        <title>test in root</title>
    </head>
    <body>
    <div class="my-class">
        this should have a background
    </div>
    </body>
    </html>

content of /css/style.css

    .my-class {
        height:500px;
        width: 500px;
        display: block;
        background: url("images/background.png") repeat scroll 0 0 green;
    }

e.g. if i open`/test.html`

**without dark reader enabled:**
![image](https://user-images.githubusercontent.com/1777147/73173893-b9246200-4106-11ea-9190-a8841e454d01.png)

**with dark reader enabled:**
![image](https://user-images.githubusercontent.com/1777147/73173865-a6aa2880-4106-11ea-91b5-9db7169cfae5.png)

the background image is requested 3 times!
http://localhost/test/image-test/css/images/background.png (correct path from the html code)
http://localhost/test/image-test/css/images/background.png (by the index.js code from the extension)
http://localhost/test/image-test/images/background.png  => 404 html error

> Relative URLs are allowed, and are relative to the URL of the stylesheet (not to the URL of the web page).

[CSS Url Syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/url#Syntax)

Same issue of course also for html requests which are located in sub directories e.g. `/sub/test.html`.

matthijskooijman:
I'm seeing exactly the same issue when using the Django default admin interface. I can't share a URL, since it's password-protected, but it looks like exactly the problem @Bhoft has reported: The relative URL in a CSS file is resolved relative to the main HTML file instead of to the CSS file.

Gusted:
#3055

mppfiles:
I'm still experiencing this bug in version 4.9.16.
I'm using FF 79, but confirmed in latest Chrome and Opera.

Here I have setup a [simple site](https://mppfiles.github.io/test-static-page/) to test the issue.
If you change the theme mode from dynamic to something else, the issue dissapears.

mppfiles:
[...]

PR #3696, opened 2020-09-24
fix wrong requests #1933

3axapp:


Gusted:
@alexanderby Was their a specific reason for `value == null ? sourceValue`?

alexanderby:
As far as I remember it is used for asynchronous styles: images that have to be loaded and analyzed. Later this source value is replaced by modified image (SVG with a filtered image, converted into a Blob URL) or the source image if inversion was not necessary.

alexanderby:
To fix the problem try finding the place where the URL becomes wrong.

PR #3698, opened 2020-09-24
Migrate automation settings

### LINKED ISSUE #1958: Toggling the extension should not reset my settings.
Actual behavior:
My settings is to use the system's color scheme to decide whether pages are light or dark. Sometimes I need to override this, so I naturally turn the extension off with the most prominent toggle switch in the popover. This resets the mentioned system color setting, which is not expected given that I have not requested any settings to change.

Expected behavior:
My settings are preserved and the extension continues to react to external color change events as soon as it's switched back on again. My preferences do not change regardless of how many times I use the toggle switch to force pages to render normally.

Forgot to mention: big fan of this project. Can't live without it now.

### LINKED ISSUE #3268: [Feature Request] The automation settings are lost if I toggle on/off manually. Can they be saved?
**Is your feature request related to a problem? Yes**

The automation settings are lost if I toggle on/off manually.

I have set Dark Reader to turn on automatically based on sunset at a specified location and it works fine. But when I manually turn it on during daytime briefly and switch it off, the location based on/off stops working. 

Turning off reverts to auto on/off mode based on settings saved.

**Additional context**

Gusted:
- Migrate the automation settings to a object.
- Don't lose the automation mode on switching `on`/`off`.
- Resolves #1958
- Resolves #3268


anohren:
I don't think this resolves #1958, unless you're back to a two-state switch.

Gusted:
> I don't think this resolves #1958, unless you're back to a two-state switch.

I don't think so, the problem states switching the mode(in the newer UI) to `on/off` would reset your saved mode(location, system, time) this Patch fix that problem by saving it into a object which has the `mode` and `enabled state` seperated.

I'm not sure which switching back to a two-state switch is fixing anything it's actually going to create more problems.

anohren:
You may be right, it does touch upon that problem. I might have mixed two issues into one there. 

The issue is not about the new UI though, since I wrote it back in February. I couldn't find an option to revert to the old UI to refresh my memory so I guess we'll consider it fixed.

johnyoonh:
Any plan to merge this soon?

Gusted:
> 
> 
> Any plan to merge this soon?

Once it get reviewed by @alexanderby

Gusted:
> Just built and installed this and it works fine until I close my browser and open it later, it switches to according to system default theme again. (Firefox 90, W10)

Have you correctly install it? By default it will remove side-loaded extension after the browser closes.

Gusted:
Superseded by #9210

PR #3718, opened 2020-09-26
Fix Amazon Package Tracker

MikeRatcliffe:
amazon-before-after

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -350,6 +350,22 @@ CSS
     background-color: ${white} !important;
 }
 
+div.milestone.notReached .milestone-marker::before {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -350,6 +350,22 @@ CSS
     background-color: ${white} !important;
 }
 
+div.milestone.notReached .milestone-marker::before {
```
Gusted, line 353:
```CSS
div.milestone.notReached .milestone-marker::before {
    border-color: var(--darkreader-neutral-text) !important;
}
```
Use 4 spaces and `!important` for every CSS declaration. Use this as well for the other ones.

Gusted (changes requested):

PR #3726, opened 2020-09-26
Added in configuration for calibre-ebook.com

### LINKED ISSUE #3723: [Broken Website] calibre-ebook.com
**Website**
https://calibre-ebook.com/

**How to Reproduce**

> 1. Go to the  site
> 2. Hover mouse over top menu

**Expected behavior**
Dark background

**Actual behavior**
Light background

**System info:**
 - OS: Windows,
 - Browser: Chrome
 - Darkreader Version: 4.9.21

Yash-Singh1:
This resolves #3723

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -918,6 +918,18 @@ div[role="checkbox"]
 
 ================================
 
+calibre-ebook.com
+
+INVERT
+.tooltip
+
+CSS
+.tooltip {
+   color: rgb(1, 1, 1) !important;
+}
+
+================================
+
 carmax.com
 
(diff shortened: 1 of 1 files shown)


Gusted (changes requested):
Hi @Yash-Singh1!

I think a better fix for this site should be

```CSS
calibre-ebook.com

IGNORE IMAGE ANALYSIS
.tooltip
#content-wrapper

```

As both problems mentioned are just images that get `inverted` by dark reader what actually shouldn't be happening.

Yash-Singh1:
Ok, looks like the tooltip was black in the first place. While testing, it looked like it was white, probably cause I had darkreader on.

PR #3728, opened 2020-09-27
Added PDF Viewer Invert on PDF Viewer Only

Yash-Singh1:
All this pull request does is the following:
Say I want to configure inside my Dynamic Configuration for only PDFs. Then inside the following:
```
mydomain.com
thisotherstuff.com

CSS
...

IGNORE IMAGE ANALYSIS
.ignore-this
#ignoreme

INVERT
.wrong-logo

```
I could add in `isPDF` in the URL Patterns section:
```
mydomain.com
thisotherstuff.com
isPDF

CSS
...

IGNORE IMAGE ANALYSIS
.ignore-this
#ignoreme

INVERT
.wrong-logo

```
And then whenever you open up a PDF, the configuration will be applied.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,13 @@ or **[inversion-fixes.config](https://github.com/alexanderby/darkreader/blob/mas
 
 Automatically syncing the above files to every Dark Reader user was disabled because the GitHub team doesn't allow using GitHub as a CDN. Storing these files and making requests to other resources would be expensive and look suspicious. As such, changes are included with each new Dark Reader release.
 
+The URL patterns work like Regular Expressions with a few improvements specifically for URLs. Here are some examples:
+* `*.google.com`
+* `amazing.com`
+* `new.example.com/*`
+
+In Dynamic Mode, you can match PDF files with the `isPDF` pattern.
+
 ### Use Dev Tools!
 
 Dev Tools is designed to **fix minor issues** on a web page
@@ -184,8 +191,8 @@ Submit a **pull request**, and wait for **review**.
 You can install the extension from a file.  
 Install [Node.js LTS](https://nodejs.org/en/). Download the source code (or check out from git).  
 Open terminal in root folder and run:  
-- `npm install`  
-- `npm run release`  
+- `npm install`
+- `npm run release`
 
 This will generate `build.zip` for use in Chromium browsers and `build-firefox.xpi` for use in Firefox.
 
--- a/README.md
+++ b/README.md
@@ -31,13 +31,6 @@ or **[inversion-fixes.config](https://github.com/alexanderby/darkreader/blob/mas
 
 Automatically syncing the above files to every Dark Reader user was disabled because the GitHub team doesn't allow using GitHub as a CDN. Storing these files and making requests to other resources would be expensive and look suspicious. As such, changes are included with each new Dark Reader release.
 
-The URL patterns work like Regular Expressions with a few improvements specifically for URLs. Here are some examples:
-* `*.google.com`
-* `amazing.com`
-* `new.example.com/*`
-
-In Dynamic Mode, you can match PDF files with the `isPDF` pattern.
-
 ### Use Dev Tools!
 
 Dev Tools is designed to **fix minor issues** on a web page
--- a/package-lock.json
+++ b/package-lock.json
@@ -1193,7 +1193,7 @@
     },
     "@types/filewriter": {
       "version": "0.0.28",
-      "resolved": "https://registry.npmjs.org/@types/filewriter/-/filewriter-0.0.28.tgz",
+      "resolved": "http://registry.npmjs.org/@types/filewriter/-/filewriter-0.0.28.tgz",
       "integrity": "sha1-wFTor02d11205jq8dviFFocU1LM=",
       "dev": true
     },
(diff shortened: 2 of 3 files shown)


Gusted:
Hi @Yash-Singh1!

Could you tell me what the purpose of this exactly is?

Regards,
Gusted

Yash-Singh1:
> Could you tell me what the purpose of this exactly is?

Sure. This can have two purposes:
1. As you can see at [Line 60](https://github.com/darkreader/darkreader/blob/7b94f09e401ce5fa012c2bfd971a06506a43fc75/src/generators/dynamic-theme.ts#L60), we have an if statement that checks whether PDFs are enabled and concatenates to the array, some configuration for PDFs. If someone has `HTML` like:
```html
<!DOCTYPE html>
<html>
   <head>
      <script src="myscript"></script>
      ...
   </head>
   <body>
      <embed type="application/pdf">
      ...
   </body>
</html>
```
And they want to configure it so that Dark Reader does not invert the embed and want to keep PDF mode on, that is impossible because that configuration is hardcoded inside the TypeScript code. This could be used for such an edge case and we could remove that if statement and add the configuration for PDF files inside the DarkReader DevTools.

2. Say someone simply wants to change around their PDF Viewer configuration, they can simply use the `isPDF` inside the DevTools and apply that configuration to all PDFs.

Gusted:
>     1. Say someone simply wants to change around their PDF Viewer configuration, they can simply use the `isPDF` inside the DevTools and apply that configuration to all PDFs.

Yep, and that isn't for DevTools but for a extension like Stylus. However the chromium PDF implemented is 1 big object and not well structruted like pdf.js from Firefox, so much PDF configuration is not even possible. What's your perspective on this?



> And they want to configure it so that Dark Reader does not invert the embed and want to keep PDF mode on.

Wait, the enabled for PDF option is EXACTLY this to disable/enable. Nvm after a quick search it seems @alexanderby had this feature removed in https://github.com/darkreader/darkreader/pull/2603/commits/ae859d3175d86ca22f6b95ad10803950537b6614. So the site-toggle will now show just the site even if it contains a PDF okay got it.

> his could be used for such an edge case and we could remove that if statement and add the configuration for PDF files inside the DarkReader DevTools.

And why do we want to configure a PDF as like I said, that's impossible.

I'm still unsure why we having this, as any changes to PDF configuration is the best way trough 
`*` as Chromium has a forced layout for PDF documents and not individual sites.

Yash-Singh1:
@Gusted Agreed, looking at a sample PDF, I realized that you can't configure anything inside a PDF:
Screenshot from 2020-09-27 12-32-49
Then I think that it would just be used for reason 1, to invert PDFs inside PDF Viewer only. For reason 2, I don't think any configuration will be done by us inside the configuration file, but it will give users an option to change, for example, the scrollbar color, maybe. The main reason I think that there should be a configuration for PDFs is for Firefox users:
Screenshot from 2020-09-27 12-42-01
As you can see, there is a lot that can be done for Firefox users. I have not checked Safari, Edge, and others, but I think Firefox will be the main use for this.

Gusted:
>As you can see, there is a lot that can be done for Firefox users. I have not checked Safari, Edge, and others, but I think Firefox will be the main use for this.

As I know, Firefox still restrict web Extension to modify PDF's as it's considered a `restricted object`.

Yash-Singh1:
Ok, then I guess that Chrome will be the only usage. I don't think you can anyways do too much with this other than reason 1.

Gusted:
I think their are a lot better solutions to solve 1.

Yash-Singh1:
Then maybe we can change the condition at [Line 60](https://github.com/darkreader/darkreader/blob/7b94f09e401ce5fa012c2bfd971a06506a43fc75/src/generators/dynamic-theme.ts#L60) to something like `(enabledForPDF && isPDF(frameURL || url))`, should I add this in this PR, or should I close this one and create a new one for that?

Yash-Singh1:
Inside PDF Viewer, the CSS is applied with PDF Viewer enabled:
Screenshot from 2020-09-30 15-16-48
When not in PDF Viewer, the CSS is gone:
Screenshot from 2020-09-30 15-17-20

Yash-Singh1:
Though the `isPDF` pattern for URLs isn't too important, I do think that `!isPDF` will be important. We can add the following into the Dynamic Configuration:
```
!isPDF

INVERT
embed[type="application/pdf"]
```
This will allow a user to switch between inverting embedded PDF Viewers when they are not looking at another PDF file.

Gusted:
> Any thoughts on adding that in?

Wait, so why are you using the dev tools as `configuration for something` it's just for Sites fixes and I think it should only exist for that sole purpose.

Yash-Singh1:
I use the dev tools `10%` of the time for configuration and little things that I like on my own browsing experience. Then I guess we should leave out this on the configuration side since not many others do so. Anyways, I fixed the problem where it doesn't work on PDF Viewer only.

Gusted:
> I fixed the problem where it doesn't work on PDF Viewer only.

Wait, I'm missing something here?

Yash-Singh1:
The problem is that not only PDF Viewer gets inverted, but other embed tags with pdfs get inverted. The UI specifically says:
Screenshot from 2020-10-03 17-34-02
That does not mean that we invert PDFs embedded on other sites. It says that we will embed specifically on PDF files.

Gusted:
>That does not mean that we invert PDFs embedded on other sites. It says that we will embed specifically on PDF files.

It is currently just inverting embed PDF's. As It's after all a PDF document. It doesn't exclude embedded ones with any meaning.

Yash-Singh1:
Ok, agreed.

PR #3740, opened 2020-09-28
Ignore non-existing images.

Gusted:
- Do not analyze images that are 0x0.

[Example page](https://privacytools.io)

### DIFF
--- a/src/inject/dynamic-theme/image.ts
+++ b/src/inject/dynamic-theme/image.ts
@@ -73,6 +73,9 @@ function analyzeImage(image: HTMLImageElement) {
         createCanvas();
     }
     const {naturalWidth, naturalHeight} = image;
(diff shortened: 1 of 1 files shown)


src/inject/dynamic-theme/image.ts
```diff
@@ -73,6 +73,9 @@ function analyzeImage(image: HTMLImageElement) {
         createCanvas();
     }
     const {naturalWidth, naturalHeight} = image;
+    if (naturalHeight > 0 && naturalWidth > 0) {
```
alexanderby, line 76: Did you mean `naturalHeight === 0 || naturalWidth === 0`?

Also worth adding a warning `logWarn(`Image is empty ${url}`)` and checking if it is really empty.

alexanderby (changes requested):

alexanderby (approved):

PR #3751, opened 2020-09-30
added one useful tip.

mayuresh4700:


README.md
```diff
@@ -20,6 +20,7 @@ See the list of [language codes](https://developer.chrome.com/webstore/i18n#loca
 
 If a website is **already dark** (has all pages dark by default), you can **add it to the [dark-sites.config](https://github.com/alexanderby/darkreader/blob/master/src/config/dark-sites.config) file**
 *(please, preserve alphabetical order)*.
+If sometimes a website is not properly visible in dark reader then you can turn off dark reader when you are using that website.
```
Gusted, line 23: I don't think the solution should be turning dark reader off, but instead reporting it as a issue.

I think editing [this part](https://github.com/darkreader/darkreader#fix-a-wrong-inversion) a bit to include. 

```
If some parts of a web page are wrongly inverted, you can fix this by specifying appropriate CSS selectors in dynamic-theme-fixes.config (for Dynamic Theme mode) or inversion-fixes.config (for Filter and Filter+ modes) (please, preserve alphabetical order by URL, use short selectors, and preserve code style).
```
Whereby as alternative option to submit a issue is given.

Gusted (changes requested):

PR #3754, opened 2020-10-01
Improve wuffs.org dark mode

LRFLEW:
This website has a tool called MeteoNook. One of the pages of it shows a monthly summary with highlights on certain days (namely weekends and days with meteor showers). I found the text on the bright-yellow highlighted rows to be very unreadable for me. This may be due in part to my colorblindness, which is why I labeled this an "improvement" instead of a "fix" (please feel free to give feedback on the changes I'm making).

I lowered the brightness of both yellow highlightcolors so that the light text is more visible. I also increased the brightness of both the background and day label for the weekends to make them a bit more visible.

**Before:**
Before

**After:**
After

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -7103,6 +7103,9 @@ CSS
 .wday-6 .mWeekday {
     color: ${#000059} !important;
 }
+select:focus {
+    color: var(--darkreader-selection-text) !important;
+}
 
 ================================
 
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -7090,10 +7090,19 @@ span.name a
 wuffs.org
 
 CSS
-.heavyShower { background-color: #660; }
-.lightShower { background-color: #990; }
-.wday-0, .wday-6 { background: #333; }
-.wday-0 .mWeekday, .wday-6 .mWeekday { color: #CCF; }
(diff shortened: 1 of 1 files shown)


LRFLEW:
I reread the README, and realized the quick fix I wrote for my use case wasn't really sufficient for a full fix. I updated it to use the `${COLOR}` syntax so that the fix is affected by the user settings. I tested the new settings in the full range of brightness/contrast to make sure it all worked. I also inverted the relative brightnesses of the highlights, as disabling the plugin for the site to compare showed that the plugin was not correctly inverting it. Here's a comparison of what the site looks like with these fixes with the plugin disabled and enabled:

**Disabled:**
Disabled

**Enabled:**
Enabled

Looking at it, I feel like I might have made the yellow highlights *too* dark. Removing the `.heavyShower` rule and changing the `.lightShower` rule to use `${ff8}` (or similar) might be better, but it made some text harder to read at higher contrast settings. Thoughts?

LRFLEW:
Added another quick fix. Selects in the tool (primarily on the Seed page) had the text become darker and hard to read after making a selection and giving the select element focus. I simply set the color attribute to `var(--darkreader-selection-text)`, which fixed the issue.

Gusted:
Hi @LRFLEW!

Fix seems good to me, do you have anything else to add?

Regards,
Gusted

Gusted (approved):

PR #3760, opened 2020-10-01
Fix for pcpartpicker.com

VFDan:
+ Fixed legend color for graph

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4636,6 +4636,13 @@ body {
 
 ================================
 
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -4636,6 +4636,13 @@ body {
 
 ================================
 
+pcpartpicker.com
+
+INVERT
```
Gusted, line 4641: Replace `INVERT` to `IGNORE INLINE STYLE`

Gusted (changes requested):

PR #3771, opened 2020-10-03
Added 5 websites which have a dark mode.

max-programming:
I added 5 websites which have dark mode. But one of them, [Hacktoberfest](hacktoberfest.digitalocean.com), determines dark mode by checking the system theme. If the system is dark, the site is dark, else light. But with dark reader, it looks like this
With the default light mode
image

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -108,7 +108,6 @@ coolmathgames.com
 corepacks.com
 coriolis.io
 countermail.com
-courses.codestackr.com
 cracked.to
 crackingking.com
 crackwatch.com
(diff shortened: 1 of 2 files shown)


src/config/dark-sites.config
```diff
@@ -107,6 +108,7 @@ coolmathgames.com
 corepacks.com
 coriolis.io
 countermail.com
+courses.codestackr.com
```
Gusted, line 111: Not dark

src/config/dark-sites.config
```diff
@@ -267,6 +271,7 @@ hackersforcharity.org
 hackertyper.com
 hackforums.net
 hackthebox.eu
+hacktoberfest.digitalocean.com
```
Gusted, line 274: Not dark by default due to system color checking.

Gusted (changes requested):

src/config/dark-sites.config
max-programming: Done

PR #3777, opened 2020-10-04
Google Translate Verified Translation button fix

kaixokkiten:
If the page is zoomed in, the verified translation button would be zoomed in but the right color. This fixes it.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -6254,6 +6254,11 @@ INVERT
 .ita-hwt-grip
 .gt-ex-quote
 
+CSS
+.trans-verified-button {
+    background-size: cover;
+}
+
 IGNORE IMAGE ANALYSIS
 .copybutton .jfk-button-img
 .speech-button .jfk-button-img


Gusted:
Hi @kaixokkiten!

I don't see a difference of the color when zoomed in.

Regards,
Gusted

kaixokkiten:
No, the color is right! It is just that the verified checkmark is zoomed in because it is a span with a background that changes size. 
image
with my edit: 
image
it appears to have something to do with the default zoom in chrome being more than 100% or zooming in, or reducing the width of the screen to less than 720 causing some weirdness that resets the size of the checkmark.

src/config/dynamic-theme-fixes.config
```diff
@@ -6254,6 +6254,11 @@ INVERT
 .ita-hwt-grip
 .gt-ex-quote
 
+CSS
+.trans-verified-button {
+    background-size: cover;
```
Gusted, line 6259: Add `!important;`

Gusted (changes requested):

likuilin:
Hi, I'm running into this issue too, and I can reproduce it on Chrome 85.0.4183.121: 

* Create a new profile
* Install Dark Reader
* Go to https://translate.google.com/
* Put in a common translation, like English to Spanish word "yes"
* Zoom in slightly using Ctrl-+

The above patch does fix it. Since `.trans-verified-button` is the most specific CSS selector, adding `!important` wouldn't change anything, but it does help with futureproofing the fix.

PR #3781, opened 2020-10-04
Fix locale time

Gusted:
- Don't use ISO method to get current time as it doesn't get custom time-zones in accounting.


alexanderby (approved):

alexanderby (changes requested):
Use padStart to add zeros

Gusted:
If you mean formatting the numbers so it has a `0` before a single digit number, that's already being handled by node.

PR #3784, opened 2020-10-05
Add the official github dependabot implementation

alazymeme:
This will add dependabot - that updates dependencies. A pr label will need to be made called `dependencies` in order for it to function correctly.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    labels:
+      - "dependencies"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,6 +3,9 @@ updates:
   - package-ecosystem: "npm"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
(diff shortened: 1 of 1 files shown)


Gusted:
Hi @alazymeme!

This will probably cause more spam than good. As I know some dependency's we use update every 2 days.
The manual NPM updates are good for now so we can test if all functionally still works and watches major updates of certain dependecy's.

Regards,
Gusted

alazymeme:
> This will probably cause more spam than good. As I know some dependency's we use update every 2 days.

There are multiple configuration options that can be set within the dependabot yml file.

> The manual NPM updates are good for now so we can test if all functionally still works and watches major updates of certain dependecy's.

The bot makes a branch that should allow you to test these changes prior to merging.

You should seek more information about a topic rather than jumping to conclusions: https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates perhaps configuring it more was the solution, but instead you have chosen to close the pr and not explore this option.

alazymeme:
Specifically you should look at the versioning strategy option: https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates#versioning-strategy

Gusted:
I will think about it.

PR #3791, opened 2020-10-05
Improve Code Quality

Gusted:
- Improve certain code.


src/generators/modify-colors.ts
```diff
@@ -194,7 +194,7 @@ function modifyFgHSL({h, s, l, a}: HSLA, pole: HSLA) {
     }
 
     let hx = h;
-    let lx = l;
+    let lx: number;
```
Gusted, line 197: `lx` will always be override so their is no need to initialize it already.

alexanderby (approved):

PR #3817, opened 2020-10-09
fix for miktex.org

linsui:


src/config/dynamic-theme-fixes.config
```diff
@@ -4000,6 +4000,23 @@ body {
...
+.icon-bar
+
+CSS
+body,
+#site-preheader,
+#site-header,
+#site-postheader,
+#site-content,
+.site-footer {
+    background: none !important;
```
Gusted, line 4015:
```CSSS
```

Let's not override the original colors as well and just remove the background-image.

Gusted (changes requested):

linsui:
The original color is ok but the inverted color looks a little wierd... But it's up to you.

Gusted (approved):

PR #3836, opened 2020-10-12
fix for habr.com

lazycrazy07c2:
it invert some svg icons

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2872,6 +2872,9 @@ CSS
 html {
     text-shadow: none !important;
 }
+[data-darkreader-inline-fill] {
+    fill: ${black} !important;
+}
 
 ================================
 


src/config/dynamic-theme-fixes.config
```diff
@@ -2872,6 +2872,9 @@ CSS
 html {
     text-shadow: none !important;
 }
+[data-darkreader-inline-fill] {
```
Gusted, line 2875: This isn't how a fix is sussposed to be working.

lazycrazy07c2: may i ask, what is wrong with it. works by me. and is shorter than make it for each element.
else it has to be like this suggestion

```suggestion
#dropdown-control svg {
    fill: ${black} !important;
}
.icon-svg {
```

Gusted: `may i ask, what is wrong with it`
It's overriding a dark reader style CSS(the `--inline` style).

`and is shorter than make it for each element.`
Than let it be for each element as that style shouldn't be override as it will mess up other innocent elements.

lazycrazy07c2: thanks for response. the suggestion above is edited.

PR #3838, opened 2020-10-13
discover.manjaro.org bottom-left tux fix

jrd:
The bottom-left little tux is invisible without the fix

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1638,6 +1638,23 @@ path[class^="wavePath"] {
 
 ================================
 
+discover.manjaro.org
+
+CSS
+html, body {
+    background-image: url("/static/images/tux.svg") !important;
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -1638,6 +1638,23 @@ path[class^="wavePath"] {
 
 ================================
 
+discover.manjaro.org
+
+CSS
```
Gusted, line 1643: This fix can be removed the `IGNORE IMAGE ANALYSIS` will do it's work and make it visible.

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
jrd: Does not work [here](https://discover.manjaro.org/applications/abiword) for instance if I remove the CSS fix

PR #3854, opened 2020-10-14
Remove fallback after another CSS

### LINKED ISSUE #3811: [Feature Request] Remove the White Flash Between Web Pages while Navigating in Dark Mode
Hello,

I notice there is often a white flash that occurs when I am in dark mode and navigate between web pages. This is quite noticeable in certain settings like dark rooms. I expect this has something to do with some kind of latency between chrome loading the new page and then dark reader being applied. 

I really enjoy using this extension, but having a feature to keep the screen dark while navigating between web pages would make it that much better.   

Thanks

Gusted:
- Fix a bug where the white flash will still present by other modes than Dynamic.
- Resolves #3811


src/inject/style.ts
```diff
@@ -9,6 +10,7 @@ export function createOrUpdateStyle(css: string) {
             style.type = 'text/css';
             style.textContent = css;
             target.appendChild(style);
+            removeDynamicTheme({removeFallback: true});
```
alexanderby, line 13: It should not call anything from Dynamic mode here.

alexanderby (changes requested):
I think that fallback should not be considered as a part of Dynamic theme. Maybe all other modes should know how to deal with the fallback (remove it when main style node is created).

Gusted:
Filter uses in newer Chromium rgb(255,255,255) on html node, which probably cause the exisiting white-flash.

@alexanderby ?

src/inject/fallback.ts
```diff
@@ -1,3 +1,5 @@
+import {removeNode} from './utils/dom';
```
alexanderby, line 1: Not sure, but I think this import will make the fallback bundle grow. Please check if so.

src/inject/style.ts
```diff
@@ -9,6 +10,15 @@ export function createOrUpdateStyle(css: string) {
             style.type = 'text/css';
             style.textContent = css;
             target.appendChild(style);
+            const wait = (animationFrame: number) => requestAnimationFrame(() => {
```
alexanderby, line 13: What's the purpose of this? Wait for 60 frames (about 1 second)? I think just removing the fallback should work.

alexanderby (changes requested):

src/inject/style.ts
Gusted: AFAIK laptops with integrated GPU onto the CPU takes more frames to get the filter applied over it and rendered. Also not sure how much it, it improves it due to the html still having a white background.

PR #3859, opened 2020-10-15
Fix for reddit.com

kyleseven:
- Make reddit logo text visible.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5142,7 +5142,7 @@ IGNORE IMAGE ANALYSIS
 reddit.com
 
 INVERT
-svg._1bWuGs_1sq4Pqy099x_yy-
+a[aria-label="Home"] > svg:nth-child(2)
 
 CSS
 [style^="--pseudo-before-background"] {
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5141,6 +5141,9 @@ IGNORE IMAGE ANALYSIS
 
 reddit.com
 
+INVERT
+svg._1bWuGs_1sq4Pqy099x_yy-
+
 CSS
 [style^="--pseudo-before-background"] {
     --pseudo-before-background: ${#DAE0E6} !important;


src/config/dynamic-theme-fixes.config
```diff
@@ -5141,6 +5141,9 @@ IGNORE IMAGE ANALYSIS
 
 reddit.com
 
+INVERT
+svg._1bWuGs_1sq4Pqy099x_yy-
```
Gusted, line 5145: The class is actuall dynamiclly generated and therefor not a reliable fix.
I think something like
```CSS
IGNORE INLINE STYLE
header a svg > g
header a svg > g > path
```
Should fix the issue.

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
```diff
@@ -5141,6 +5141,9 @@ IGNORE IMAGE ANALYSIS
 
 reddit.com
 
+INVERT
```
Gusted, line 5144: This will invert the logo, however this will interfere when you don't have the native dark mode enabled along side dark reader

kyleseven: Interesting, I don't seem to have this issue on my end.

Dark Reader + Reddit Dark Mode Disabled
dark_disabled

Dark Reader + Reddit Dark Mode Enabled
dark_enabled

Gusted: Hmm, weird .
Now I got it inverted back to almost non-visible in light mode.
 The result you got shouldn't be possible. As the dark/light mode of reddit changes the colors of the logo.

kyleseven: Yeah, this is really strange. I can't seem to replicate what you're seeing, even in Firefox.
firefox_2020-10-15_16 56 26

Gusted: I can't even put my finger on why this is happening. Maybe going to this fix helps?  https://github.com/darkreader/darkreader/pull/3859#discussion_r505350067

kyleseven: I tried that first, but it only worked when Reddit dark mode was enabled.

Gusted: Maybe add another CSS to correct the color so dark reader will first not interfere(ignore inline style) and CSS to force it it's original color in dark mode.

PR #3860, opened 2020-10-15
Update dark-sites.config

PyroGenesis:
https://create-react-app.dev/ has a global dark mode built into their website


Gusted:
It's toggled based on system preferred color scheme.

PR #3869, opened 2020-10-17
9gag.com

akhyarrh:
There's a toggle in top navigation bar, right side


Gusted:
The site is meant for sites that have a dark mode by default and not part of it.

PR #3873, opened 2020-10-17
Optimize platform/browser checking

### LINKED ISSUE #3870: Firefox 81.0.2 Says that Dark Reader is slowing the browser down
As the title says, firefox shows the little yellow drop down with the message that Dark reader is causing the browser to slow down with the webpage being very slow to load for a moment before popping in.  The message is only there for a moment so if it has some other means of telling me what is wrong from the dialogue box I wouldn't know. It comes up the most on youtube and gmail, but the pop up alert has shown up on other sites as well, just not as consistently.

Gusted:
- Nuke the checking so it will only be checked once and cache results.
- Attempt to fix #3870


src/utils/platform.ts
```diff
@@ -1,56 +1,38 @@
...
-}
-
-export function isMacOS() {
-    if (typeof navigator === 'undefined') {
-        return null;
+import {PlatformData} from '../definitions';
+
+export const platformData: PlatformData = {} as any;
+
+export function runPlatformTest() {
```
alexanderby, line 5: Is there a need in this function? You could just assign everything in the top level scope of this file.

```typescript
export const isFirefox = userAgent.includes('firefox');
```

If the tests will fail (they should not in JSDom environment), substitute `navigator` with some object.

src/utils/platform.ts
```diff
@@ -1,56 +1,38 @@
...
+    } else {
+        platformData.chromiumVersion = '';
     }
-    return navigator.platform.toLowerCase().startsWith('mac');
-}
-
-export function isMobile() {
-    if (typeof navigator === 'undefined') {
-        return null;
+    try {
```
alexanderby, line 23: Try self-executing function to hide logic out of root scope
```typescript
export const isDefinedSelectorSupported = (() => {
    try {	
        document.querySelector(':defined');	
        return true;	
    } catch (err) {	
        return false;	
    }
})();
```
Note () at the end.
Same for similar cases.

alexanderby (changes requested):

src/utils/platform.ts
```diff
@@ -1,57 +1,47 @@
...
+export const isShadowDomSupported = typeof ShadowRoot === 'function';
+export let chromiumVersion: string = null;
+let m = userAgent.match(/chrom[e|ium]\/([^ ]+)/);
+if (m && m[1]) {
+    chromiumVersion = m[1];
+} else {
+    chromiumVersion = '';
+}
+
+// Clean up memory
```
alexanderby, line 23: This is a micro optimization :) They are rather small strings and Garbage Collector will remove it from memory after some time, because no other code references them.

src/utils/platform.ts
```diff
@@ -1,57 +1,47 @@
...
+export const isFirefox = userAgent.includes('firefox');
+export const isVivaldi = userAgent.includes('vivaldi');
+export const isYaBrowser = userAgent.includes('yabrowser');
+export const isOpera = userAgent.includes('opr') || userAgent.includes('opera');
+export const isEdge = userAgent.includes('edg');
+export const isWindows = platform.startsWith('win');
+export const isMacOS = platform.startsWith('mac');
+export const isMobile = userAgent.includes('mobile');
+export const isShadowDomSupported = typeof ShadowRoot === 'function';
+export let chromiumVersion: string = null;
```
alexanderby, line 15: I would wrap it into self-executing function too to hide this `m` variable from the scope. I think it will make the code a little bit more clean.

alexanderby (changes requested):

alexanderby (approved):
One merge conflict left

PR #3874, opened 2020-10-17
Optimize Temp Style usage

### LINKED ISSUE #3870: Firefox 81.0.2 Says that Dark Reader is slowing the browser down

Gusted:
- Re-use so it won't need to rely on slow DOM 'document.append'.
- Attempt to fix #3870


src/inject/dynamic-theme/stylesheet-modifier.ts
```diff
@@ -20,17 +20,22 @@ function getThemeKey(theme: Theme) {
     return themeCacheKeys.map((p) => `${p}:${theme[p]}`).join(';');
 }
 
-function getTempCSSStyleSheet(): {sheet: CSSStyleSheet; remove: () => void} {
+let tempStyle: HTMLStyleElement = null;
+
+function getTempCSSStyleSheet(): CSSStyleSheet {
+    if (tempStyle) {
+        return (tempStyle as HTMLStyleElement).sheet;
```
alexanderby, line 27: Check if `tempStyle.parent === document.head`, if not, then it needs to be appended.

alexanderby (changes requested):
Please check if it works in

src/inject/dynamic-theme/stylesheet-modifier.ts
Gusted: Why is their a check needed? I've modified so it will only be appended to `document.head` As if you need a temp StyleSheet the DOM(head) is already loaded

src/inject/dynamic-theme/stylesheet-modifier.ts
```diff
@@ -73,16 +81,18 @@ export function createStyleSheetModifier() {
...
             if (variables.size > 0 || cssText.includes('var(')) {
                 const cssTextWithVariables = replaceCSSVariables(cssText, variables);
                 if (rulesTextCache.get(cssText) !== cssTextWithVariables) {
                     rulesTextCache.set(cssText, cssTextWithVariables);
                     textDiffersFromPrev = true;
                     vars = getTempCSSStyleSheet();
-                    vars.sheet.insertRule(cssTextWithVariables);
-                    varsRule = vars.sheet.cssRules[0] as CSSStyleRule;
+                    vars.insertRule(cssTextWithVariables);
+                    varIndex = vars.cssRules.length - 1;
```
Gusted, line 94: Not sure if we need dynamically `varIndex`. As this code should runs synchronous and can't be overridden mid-way by some MutationObserver, right?

alexanderby: I think yes, there can be only a single rule.

Gusted: Great, removed it alongside re-using the CSSStyleSheet as well.

Gusted:
Somehow making a copy of the `.sheet` property can act as a valid CSSStyleSheet even if the original Node has been removed and garbage from memory. I've implemented this in https://github.com/darkreader/darkreader/pull/3874/commits/1bf33eb42c35812dab68b5ad7ac0d62d518fccf8

alexanderby (approved):
This one is great and minimalistic. Let's check if it works in Firefox.

Gusted:
Seems to be working.

PR #3878, opened 2020-10-20
Update dynamic-theme-fixes.config for coursera.org fix

jacobhutchinson:
CSS element was blocking video, different element was inverting colors on it. Both of those bugs have been eliminated and I added .rc-VideoMiniPlayer CSS Selector to the inversion list for posterity. Works in light and dark on each page I have tested it on.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1355,6 +1355,24 @@ a {
 
 ================================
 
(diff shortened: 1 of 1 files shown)


jacobhutchinson:
'Run npm test' is reporting it cannot download the log, and to please try again later. This sounds like it wasn't to do with the code I added, so I will close for now.

PR #3886, opened 2020-10-20
Fix Deezer seek bar color

mrn:
The seek bar had a dark gray color and was invisible.
This fix uses Deezer's default theme colors to make the seek bar have a gray-to-red gradient.
The colors are referenced using variables, so in case Deezer changes its theme Dark Reader will use the new colors.

Current:
image

Fixed:
image

### DIFF
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -315,6 +315,11 @@ deezer.com
 INVERT
 .page-sidebar
 
+CSS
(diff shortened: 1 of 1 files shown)


mrn:
I am not sure why the test says formatting is wrong:

image

mrn:
I made the correction to the wrong file (for the inverted theme instead of dynamic). Correct version here: https://github.com/darkreader/darkreader/pull/3887

PR #3887, opened 2020-10-20
Fix Deezer seek bar color

mrn:
The seek bar in Deezer sometimes takes the average color of the playlist. This was designed with the assumption that the background is white, so in the cases when the album cover is in grayscale, the seek bar can be dark gray and appear invisible when using Dark Reader.

This fix uses Deezer's default theme colors to make the seek bar have a gray-to-red gradient all the time.
The colors are referenced using variables, so in case Deezer changes its theme Dark Reader will use the new colors.

Current:
image

Fixed:
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1520,6 +1520,15 @@ body {
 
 ================================
 
+deezer.com
+
+CSS
+.slider-track .gradient-default {
+    background-image: linear-gradient(1deg, var(--color-dark-grey-800) 13%, var(--color-coral-500));
+}
+
+================================
+
 dennisbareis.com
 
 CSS
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1524,7 +1524,7 @@ deezer.com
 
 CSS
 .slider-track .gradient-default {
-    background-image: linear-gradient(1deg, var(--color-dark-grey-800) 13%, var(--color-coral-500));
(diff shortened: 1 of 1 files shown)


mrn:
Interestingly, non-black seek bar colors are preserved:
image
The fix only affects the seek bar when the album cover is grayscale.

src/config/dynamic-theme-fixes.config
```diff
@@ -1520,6 +1520,15 @@ body {
 
 ================================
 
+deezer.com
+
+CSS
+.slider-track .gradient-default {
+    background-image: linear-gradient(1deg, var(--color-dark-grey-800) 13%, var(--color-coral-500));
```
Gusted, line 1527: Hmm, I just checked with gradient-default, but it doesn't seem like it was a colored gradient, but rather the Black/White one.

mrn: Ah, so this only targets the dark gradient and makes it red

Gusted: Yes, their seems to be a lot of gradients e.g. `.gradient-yellow`

mrn: This is good then, since only the dark gradient was the problematic one. Other gradients are left untouched

Gusted: I think it's better to just have it darkened and use `var(--color-white...`

mrn: How is this?

image

Gusted: That seems better as it's darkening the orginial color(dark)

Issue #2552, opened 2020-05-19
API to disable dark reader on your site


Jonatino:
It would be nice to be able to let darkreader know your site already handles dark theme. That way darkreader can whitelist your site. Saves a lot of cpu time on large sites.

Gusted:
Add them to https://github.com/darkreader/darkreader/blob/master/src/config/dark-sites.config

Gusted:
@Jonatino ^

Jonatino:
I feel like that list is unpractical and will quickly get out of hand with the growing amount of sites supporting dark theme out of the box.

I think adding some sort of known html meta tag that darkreader looks for or even a javascript api to disable it manually is a better solution than hardcoding a list of urls.

Gusted:
Well that would be also unpractical as not all dark sites want to add some meta into their html.

Jonatino:
Less impractical than keeping a list though.

Gusted:
I'm sure if we replace the list and ask all sites to add meta only 5% will do it

Jonatino:
I never suggested replacing the list, I simply want another way to do it rather than rely on a third party project to merge a pull request, wait for them to push their next update, and wait for that update to propagate to all installs of dark reader. Vs just adding a known tag that I know will work right away across any darkreader version that supports it.

Gusted:
Adding this feature to the code base also will be pushed to next update and I'm certain that owner of darkreader will simply close the pull request. We are the first party we determine if the site is actually dark theme to be sure and than just merge it its not an external party does it.

Gusted:
Hi @Jonatino 

I've been rethinking about this, it should be a pretty good idea.
How do you suggest the `tag` to be look be?

Regards,
Gusted

Jonatino:
Eh I was just thinking using a metatag.
```
<meta name="darktheme_supported" content="true">
```

or having some sort of javascript alternative.

Gusted:
I will think about the name.

Gusted:
`<meta name="darkreader" instance="someUIDhere">` 
#2855

swrobel:
@Gusted what you've presented is an invalid meta tag. There is no `instance` [attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#Attributes) ... did you mean `content="UID"`?

Gusted:
> 
> 
[...]

PR #3902, opened 2020-10-21
Add global detection patch for color-scheme

Gusted:
- User color-scheme meta instead of the unofficial `darkreader` one.
- Resolves https://github.com/darkreader/darkreader/pull/2855#issuecomment-713040345


Gusted:
Maybe TODO: a Observer for `meta` as it's possible dark reader will start before the `meta` has loaded?

alexanderby:
So, the user actually wants an ability to disable Dark Reader. Here are some of my thoughts.

1. Having the author's dark theme detection would be nice.
2. `matchMedia(... dark)` checks for system dark mode, not for a web-page theme.
3. The actual dark theme detection could be reading the `meta[color-scheme]` along with detecting HTML or BODY computed background color.
4. We can experiment with this detection, if it will fail quite often, it should be configurable by user.
5. Still, to support standards, we can use some element instead of custom meta tag, e.g. `<style id="darkreader-instance" data-id="xyz"></style>`.

Gusted:
> Having the author's dark theme detection would be nice.

This will require some logic and probably the most reliable way is to detect changes into the `body` or `html` element for background-color. Probably hooking into this code [here](https://github.com/darkreader/darkreader/blob/master/src/inject/dynamic-theme/stylesheet-modifier.ts#L98) but not sure how much of a performance impact can have and as well, how to stop dark reader from continuing analyzing the page?
2. Oh it's only the case if the `dark` and `light` is inside the content and therefor if the user has a system-dark-mode the site should by documentation render a dark color-scheme.
3. Read 1.
4. Meta observer was suggested ^
5. Sure things will add this into the PR! I think to detect the author's dark theme should be a separate PR.

swrobel:
@alexanderby why not just disable when `meta[color-scheme]` includes `dark`?

chase-moskal:
@alexanderby @Gusted 

hey folks, i just made a nice dark theme website, but darkreader unfortunately trashes it by default for me and all my cool friends, who of course, all have the good taste to be using darkreader.

i've read through these related threads, and i'm left wondering the same as @swrobel  -- shouldn't this be really simple? a quick and easy check should do it:

```js
const websiteAlreadyDark = document.querySelector(`meta[name="color-scheme"]`)
  ?.content
  ?.includes("dark")

if (websiteAlreadyDark)
  disableDarkReader()
```

if dark reader would respect the [color-scheme](https://web.dev/color-scheme/) meta tag, then this would solve all my problems and make me very happy!

Gusted:
> if dark reader would respect the [color-scheme](https://web.dev/color-scheme/) meta tag, then this would solve all my problems and make me very happy!

The problem with this is that I've seen sites using this `meta` tag and not having specified behavior(e.g. the order matters). Even when `dark` is specified, it doesn't mean the site is dark, it should be dark when the user's system's color is also preferred dark. And certain sites don't have any dark mode at all, and still add the `dark` keyword(which is mainly the reason why I haven't merged this yet).

chase-moskal:
@Gusted understood. and thanks for the swift response.

so, then, what can i do as a website author, to fix my website for darkreader users?

1. should i install "no-darkreader" onto my website? https://github.com/hadialqattan/no-darkreader
2. can i use `<meta name="darkreader" content="someUIDhere">` as [others have asked about?](https://github.com/darkreader/darkreader/issues/2552#issuecomment-930768618)
    - if so, what should i put in place of `someUIDhere`?

Gusted:
> 1. should i install "no-darkreader" onto my website? https://github.com/hadialqattan/no-darkreader

You could do that, but it's better to do step 2(as it's easier and smaller to use)



> * if so, what should i put in place of `someUIDhere`?

Anything 😄 Their is no restrictions on that.

TruncatedDinosour:
> > 1. should i install "no-darkreader" onto my website? https://github.com/hadialqattan/no-darkreader
> 
> You could do that, but it's better to do step 2(as it's easier and smaller to use)
> 
> > * if so, what should i put in place of `someUIDhere`?
> 
> Anything smile Their is no restrictions on that.

Hey, sorry for replying but

I added `<meta name="darkreader" content="false">` to my site (as you said you can put `someUIDhere` as anything) and I created a completely fresh profile in my browser (firefox 97.0.2), installed dark reader and dark reader still is on by default, why is that?

Gusted:
> I added `<meta name="darkreader" content="false">` to my site (as you said you can put `someUIDhere` as anything) and I created a completely fresh profile in my browser (firefox 97.0.2), installed dark reader and dark reader still is on by default, why is that?

Enabled and actually applying is something different. Is dark reader actually doing anything?

TruncatedDinosour:
> > I added `<meta name="darkreader" content="false">` to my site (as you said you can put `someUIDhere` as anything) and I created a completely fresh profile in my browser (firefox 97.0.2), installed dark reader and dark reader still is on by default, why is that?
> 
> Enabled and actually applying is something different. Is dark reader actually doing anything?

yes, it's completely transforming the site's colourscheme, it's by default brown-ish (like https://files.ari-web.xyz/) and it turns almost black with the extension

ModProg:
> > > I added `<meta name="darkreader" content="false">` to my site (as you said you can put `someUIDhere` as anything) and I created a completely fresh profile in my browser (firefox 97.0.2), installed dark reader and dark reader still is on by default, why is that?
> > 
> > 
> > Enabled and actually applying is something different. Is dark reader actually doing anything?
> 
> yes, it's completely transforming the site's colourscheme, it's by default brown-ish (like https://files.ari-web.xyz/) and it turns almost black with the extension

It does seam to work with chrome... but in FF even the the newest version here from GitHub it has a kind of opposite effect of switching to a mode where I only have white on Gray...


With the tag:
image

without the tag:
image


For some reason, chrome doesn't correctly apply the prefere-dark media tag, therefor it is white in the right window.

chase-moskal:
it appears that the meta tag is not working in chrome 99?

despite having `<meta name="darkreader" content="dark"/>`, users are finding that darkreader is messing up our web game ui at https://benevolent.games/humanoid -- and the worst part is that users are often confused and think the game is broken, not realizing that it's dark reader causing the problems

does _any_ user interaction with dark reader (enabling/disabling the website in darkreader) override the `<meta>` technique? perhaps users (myself included) had toggled the website on/off before on a version of the website that didn't have the `<meta>`, thus now overriding the meta?

not sure what's going on 🤷‍♂️

Gusted:
You do realize this is an **open** pull request? This means the current behavior isn't like this.

chase-moskal:
@Gusted oh no, i hadn't realized this — does this mean there's currently _no way whatsoever_ for a website to signal to disable darkreader?

is there something we can do to get this PR merged and resolve this issue?

Gusted:
> does this mean there's currently _no way whatsoever_ for a website to signal to disable darkreader?

https://github.com/darkreader/darkreader/blob/master/src/config/dark-sites.config

Gusted:
> > 
> 
> open source is funny sometimes (note: i am more than a fan and an advocate for open source)

Not sure what you're implying. @q1800 Is simply notifying me of outdated pull requests (which is fine, as I can then check them and likely close them).

PR #3906, opened 2020-10-22
Force fallback to be opacity: 1

### LINKED ISSUE #3884: [BUG] Fallback is not reacting on 0 opacity
[Site](https://searchengineland.com/google-no-longer-supports-relnext-prev-314319)

Bypasses the fallback by adding `opacity: 0 ` to the `html` node.

And removing after the DOMContent is loaded.

Because, the browser switches to the HTML background-color as soon it's get loaded and `opacity: 0` it will return a white background-color as the HTML doesn't exist to be rendered.

Reported: https://github.com/darkreader/darkreader/issues/3811#issuecomment-712895413

Gusted:
- This bypasses the bypass of #3884.
- Force `opacity: 1` and therefore the html/body will be visible for rendering and return a dark color in order to still prevent the white-flash.
- Resolves #3884

### DIFF
--- a/src/inject/fallback.ts
+++ b/src/inject/fallback.ts
@@ -3,7 +3,7 @@ if (
     matchMedia('(prefers-color-scheme: dark)').matches &&
     !document.querySelector('.darkreader--fallback')
 ) {
-    const css = 'html, body, body :not(iframe) { background-color: #181a1b !important; border-color: #776e62 !important; color: #e8e6e3 !important; opacity: 1 !important; }';
+    const css = 'html, body, body :not(iframe) { background-color: #181a1b !important; border-color: #776e62 !important; color: #e8e6e3 !important; } html, body { opacity: 1 !important; }';
     const fallback = document.createElement('style');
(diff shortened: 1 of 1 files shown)


src/inject/fallback.ts
```diff
@@ -3,7 +3,7 @@ if (
     matchMedia('(prefers-color-scheme: dark)').matches &&
     !document.querySelector('.darkreader--fallback')
 ) {
-    const css = 'html, body, body :not(iframe) { background-color: #181a1b !important; border-color: #776e62 !important; color: #e8e6e3 !important; }';
+    const css = 'html, body, body :not(iframe) { background-color: #181a1b !important; border-color: #776e62 !important; color: #e8e6e3 !important; opacity: 1 !important; }';
```
alexanderby, line 6: It can reveal some unwanted content, maybe try apply it to html and body only instead?

alexanderby (changes requested):

alexanderby (approved):
Great, check if it fixes the problem.

PR #3919, opened 2020-10-24
Add bongo.cat

KTibow:
This adds https://bongo.cat which is already inverted.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -75,6 +75,7 @@ blog.ja-ke.tech
 blokada.org
 blox.link
 bogleech.com
+bongo.cat
 bonsaihd.live
 botboy.snaz.in
 breaksforeyes.app


Gusted:
Uses `(prefers-color-scheme: dark)` Which doesn't meet the requirements of dark page by default.

KTibow:
Oh. Sorry, didn't notice. What does that mean?

KTibow:
It does make a difference, though:
Original:
image
Inverted:
image

Gusted:
If you use a color-scheme that preffers a light theme e.g. on Windows 10  The site will fail on 
```javascript
matchMedia('(prefers-color-scheme: dark)');
```
And therefor load a light colored page.
It's easy to detect it as you can see it only works on a `(prefers-color-scheme: dark)`

KTibow:
Got it. I suppose it's a bad practice to apply that (even if the page is designed to use it when dark mode is detected) in dynamic mode.

KTibow:
Any way for something like this to be implemented?
https://stackoverflow.com/questions/55838230/can-browser-extensions-overwrite-css-media-features-query

Gusted:
> https://stackoverflow.com/questions/55838230/can-browser-extensions-overwrite-css-media-features-query 

Bad practice. Follow a existing #1285

PR #3925, opened 2020-10-25
Improve Image analyzing. 

Gusted:
- It's improved? It still works and I noticed using the options for specific coordinates is a lot slower than using the basic coordinates.
- This PR is mostly targetted for website e.g. Gmail which has a great impact because getImageData and drawImage is slow.

### DIFF
--- a/src/generators/modify-colors.ts
+++ b/src/generators/modify-colors.ts
@@ -20,25 +20,35 @@ function getFgPole(theme: Theme) {
 }
 
 const colorModificationCache = new Map<ColorFunction, Map<string, string>>();
-const colorParseCache = new Map<string, HSLA>();
+const HSLAParseCache = new Map<string, HSLA>();
 
 function parseToHSLWithCache(color: string) {
-    if (colorParseCache.has(color)) {
-        return colorParseCache.get(color);
+    if (HSLAParseCache.has(color)) {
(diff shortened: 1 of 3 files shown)


alexanderby:
Strange PR. So, my solution was to set the canvas at maximum of 32x32 pixels for performance reasons. If image is smaller, it will be drawn without losses and native size pixels will be analyzed. If it is larger than 32x32, it will be downscaled to fit there. Natural size can be huge, so analyzing every pixel should be very slow.

Gusted:
> Natural size can be huge, so analyzing every pixel should be very slow.

Downside of that somehow it's decreasing the performance of canvas functions.

alexanderby:
Do you mean that downscaling an image is slow? If so, then the image could be painted in its real size, but some rows and columns can be skipped, e.g. `const xStep = Math.max(1, Math.floor(width / 32)); ... x += xStep`.

But a proper benchmark should be done to ensure this is a problem.

Gusted:
> Do you mean that downscaling an image is slow? 

It seems to be. Or just the rapid repainting of different images. 

> But a proper benchmark should be done to ensure this is a problem.

Agreed.

Gusted:
Will be succeeded by an better and more improved PR.

PR #3915, opened 2020-10-23
Replace Inline CSS Variables

Gusted:
- Replace the inline CSS variable with it's corresponding value
- Make sure a conversion from background -> background-color is happening in order for `iterateCSSDeclarations` to understand it.
- Resolves #3914

PR #3928, opened 2020-10-26
fix for github.com

linsui:
sshot-010
sshot-011


Gusted:
#3915 Will have a better fix for this issue.

PR #3944, opened 2020-10-27
Add more browser test

Gusted:
- Ability to restore a style when it was removed.
- Ability to watch for a source style and link positions.
- Ability to remove override when style was removed.
- Ability to react on dynamic style creation.
- Ability to react on inserting something into a style dynamically (sheet.insertRule(...)).

### DIFF
--- a/tests/browser/dynamic/style-override.tests.ts
+++ b/tests/browser/dynamic/style-override.tests.ts
@@ -48,4 +48,160 @@ describe('Style override', () => {
         await expect(page.evaluate(() => getComputedStyle(document.querySelector('h1')).color)).resolves.toBe('rgb(232, 230, 227)');
         await expect(page.evaluate(() => getComputedStyle(document.querySelector('h1 strong')).color)).resolves.toBe('rgb(255, 26, 26)');
     });
+
+    it('should restore override', async () => {
+        await loadTestPage({
+            '/': multiline(
+                '<!DOCTYPE html>',
+                '<html>',
+                '<head>',
+                '   <script>',
+                '       const styleElement = document.createElement("style");',
+                '       styleElement.classList.add("testcase-style")',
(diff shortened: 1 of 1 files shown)


tests/browser/dynamic/style-override.tests.ts
```diff
@@ -48,4 +48,160 @@ describe('Style override', () => {
...
         await expect(page.evaluate(() => getComputedStyle(document.querySelector('h1 strong')).color)).resolves.toBe('rgb(255, 26, 26)');
     });
+
+    it('should restore override', async () => {
+        await loadTestPage({
+            '/': multiline(
+                '<!DOCTYPE html>',
+                '<html>',
+                '<head>',
+                '   <script>',
```
alexanderby, line 58: I would suggest to use `page.evaluate()` where possible, e.g.
```javascript
await page.evaluate(() => {
    const styleElement = document.createElement("style");
    styleElement.classList.add("testcase-style");
    document.head.append(styleElement);
    styleElement.sheet.insertRule("h1 { color: gray } ")
    styleElement.sheet.insertRule("strong { color: red } ")
})`
```

Please note that the function passed as an argument is converted to a string, and is executed in the page using `eval()`, follow the docs if you need to pass something into this function.

Don't forget about `code-style`

alexanderby (changes requested):

tests/browser/dynamic/style-override.tests.ts
Gusted: Oh woops. That was something for testing purposes...

Gusted:
Also should `npm run test` includes browser tests?

Gusted:
Oh yeah, we have Github Actions...

alexanderby (approved):
I think it's okay.

PR #3961, opened 2020-10-28
fix white gradient in BigBlueButton

serkonda7:
**Before**
image

**After**
image

The url cannot be specified further because `bbb` has different numbers (the room number) as suffix.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -620,7 +620,7 @@ INVERT
 
 ================================
 
-bbb*
+bbb*.
 
 CSS
 div[class*="scrollableList"] {
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -621,6 +621,7 @@ INVERT
 ================================
 
 bbb*.
+demo*.bigbluebutton.org
 
 CSS
 div[class*="scrollableList"] {
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -620,6 +620,15 @@ INVERT
 
 ================================
 
+bbb*
+
+CSS
+.scrollableList--Z2s6Her {
+    background: none;
+}
+
+================================
+
 bbc.co.uk/weather
 bbc.com/weather
 
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -623,7 +623,7 @@ INVERT
 bbb*
 
 CSS
-.scrollableList--Z2s6Her {
+div[class*="scrollableList"] {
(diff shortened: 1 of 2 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -620,6 +620,15 @@ INVERT
 
 ================================
 
+bbb*
```
Gusted, line 623: What's an example URL?

src/config/dynamic-theme-fixes.config
```diff
@@ -620,6 +620,15 @@ INVERT
...
+
+CSS
+.scrollableList--Z2s6Her {
```
Gusted, line 626: Hmm change it into something like 
```CSS
div[class*="scrolableList"]
```
As after the -- are mostly automatic generated and change over time.

src/config/dynamic-theme-fixes.config
Gusted, line 623: What's an example URL?

Gusted (changes requested):

erkinalp:
Also confirmed by me.

src/config/dynamic-theme-fixes.config
erkinalp: Log in to a demo at https://demo.bigbluebutton.org.

src/config/dynamic-theme-fixes.config
serkonda7: Thanks for the tip, I'm just a beginner with CSS.

src/config/dynamic-theme-fixes.config
serkonda7: I can't give an example URL. I discovered the issue in the privately hosted BigBlueButton instance of my university. All the rooms have a URL like this: `https://bbb42.fbi.h-da.de` but the number can be different.

Gusted: Ah I see, I think the URL can be changed to bbb*.*.* in order to just specify subdomains that start with `bbb` as something like bbba1234.com could be targetted as well.

erkinalp: You can also try a demo at https://demo.bigbluebutton.org, using the `demo[nn].bigbluebutton.org` pattern.

serkonda7:
@erkinalp how does the URL of the website look like, that you used?

erkinalp:
@serkonda7 https://demo2.bigbluebutton.org/html5client/join

serkonda7:
@Gusted  ready to merge from my side

erkinalp:
I still have gradient issue even with this applied.

serkonda7:
@erkinalp the fix isn't released yet.
If you copy my updated config into your DarkReader DevTools it should work.

PR #3984, opened 2020-10-30
Run tests in Karma

alexanderby:
- Use `npm run test:inject` for a single run in all browsers.
- Use `npm run test:inject:debug` to run tests in Chrome with live reloading.
- Coverage reports can be found in `tests/inject/coverage`.


Gusted:
Fixed conflicting file due to #4007

PR #3988, opened 2020-10-30
Fix for genius.com

Grelo4ka:


src/config/dynamic-theme-fixes.config
```diff
@@ -2599,6 +2599,13 @@ img.listview__image, img.galleryview__image, .variant__header__dropdown__item__i
 
 ================================
 
+genius.com
+
+INVERT
+.texmath
```
Gusted, line 2605: Example URL couldn't find this.

Grelo4ka: Any song, really. Just scroll a bit down and see how the Apple Music bottom bar looks like.
https://genius.com/Luis-fonsi-despacito-lyrics

PR #3989, opened 2020-10-30
Fix for brainly.pl

Grelo4ka:


src/config/dynamic-theme-fixes.config
```diff
@@ -949,6 +949,14 @@ INVERT
 
 ================================
 
+brainly.pl
+
+INVERT
+[id^="TexFormula"]
```
Gusted, line 955: Where can I find this?

Grelo4ka: https://brainly.pl/zadanie/18770149
https://brainly.pl/zadanie/18770085
https://brainly.pl/zadanie/18770321
https://brainly.pl/zadanie/18770210 (that's why "not" is there)
https://brainly.pl/zadanie/18770196 (again, that's why "not" is there)

PR #3992, opened 2020-10-31
Fix for bab.la

Grelo4ka:


src/config/dynamic-theme-fixes.config
```diff
@@ -613,6 +613,13 @@ CSS
 
 ================================
 
+bab.la
+
+INVERT
+.logo-flash
```
Gusted, line 619: This element seems to be invisible for me. Any specific scenario where this is visible?

Grelo4ka: Just scroll down and look at the top left corner

PR #3993, opened 2020-10-31
Fix for computerhope.com

Grelo4ka:


src/config/dynamic-theme-fixes.config
```diff
@@ -1232,6 +1232,13 @@ text
 
 ================================
 
+computerhope.com
+
+INVERT
+#text-tool > p > img
```
Gusted, line 1238: Couldn't find this element.

Grelo4ka: https://www.computerhope.com/cgi-bin/text-tool.pl

PR #4003, opened 2020-10-31
Fix for myaccount.google.com

Rakleed:
Fix icons in the left menu.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4394,7 +4394,7 @@ CSS
 myaccount.google.com
 
 INVERT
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -4391,6 +4391,13 @@ CSS
 
 ================================
 
+myaccount.google.com
+
+INVERT
+.b3qkse .YPzqGd
```
Gusted, line 4397: As the class names seems to be generated please change it to
```CSS
c-wiz figure img
```

Gusted (changes requested):

PR #4005, opened 2020-11-01
Better Idle performance

### LINKED ISSUE #1125: checkForUpdate polling causes significantly increased CPU usage on some sites
#### Overview
I recently noticed a significant increase in Chromium's idle CPU usage and identified darkreader as the culprit. My methodology involves observing the relevan't Tab's CPU usage in the Chromium task manager before and after switching from the Dynamic theme to Static/Filter but also verified it by disabling the offending code, see below.

#### Impact
The CPU usage on Google Search/Youtube homepages goes up from about 2-4% in idle to 15-20% in idle (as seen in Chromium's task manager).

Due to the use of `requestAnimationFrame` to schedule polling this only seems to happen when the page is currently visible in a browser window limiting the impact somewhat.

#### Reproducing
I belive the polling of stylesheet changes in `checkForUpdate` is causing this high load. If I comment out the `requestAnimationFrame(checkForUpdate)` call in `src/inject/dynamic/style-manager.ts` the CPU usage reliably goes back to baseline levels.

#### Fix
I'm not really sure what kind of changes are being watched for here, is this only for first time load stuff or is this also intended to catch changes after the site finished loading? I would send a patch but my javascript foo is very rusty, but given some direction I would be happy to help implementing a fix.

Would it be possible to get notified of these stylesheet changes instead of polling? If not throttling the polling as the site reaches a steady state might be a decent fix.

Personally I might be happy just being able to set the theme engine per-site but that seems crude as a fix :)

### LINKED ISSUE #2070: [Bug] DR causes massive CPU issues when active on inspected pages (Chromium devtools)
Google Chrome Version 80.0.3987.132 (Official Build) (64-bit)
DR 4.8.9
Win 8.1

This has been a problem I've been running into for a long time, and I finally got around to troubleshooting. To my surprise, seems like DR was the problem.

Basically, when you inspect most pages DR is active on, once you hover the DOM tree, CPU spikes through the roof, at least an additional 50-70%, often more, and never releases. I made a [screen record](https://streamable.com/vjd48) to demo, first without DR enabled, then with it enabled. CoreTemp is active in the system tray, and the green indicator is CPU. Bear in mind, OBS is running which averages about 5-10% CPU.

As you'll see, without DR active, hovering the DOM in devtools will very briefly spike ~5-10% when the mouse is active, and almost immediately normalize once the mouse stops. When DR is active, hovering the DOM tree causes CPU to skyrocket and never release. OBS actually throws an encoding warning, the CPU spikes so dramatically.

This is not true for every page DR is active on. In my testing, a few seemed unaffected (Github for example), but the vast majority are affected.

Steps to repro:

* New Google Chrome profile

* New DR install

* Visit:  https://example.com/

* Open devtools and hover DOM tree nodes while observing CPU

### LINKED ISSUE #3905: [BUG] checkIfDefined is doing nothing but wasting CPU
### **Describe the bug**

The current implementation can and is endlessly looping so long the elements are not defined which in some cases won't be and due to the logic used in this function it will somehow use noticeable (on low-medium spec CPU) CPU Utilization.

e.g. https://mail.google.com

### **Steps to reproduce**

- Go to mail.google.com
- Open performance tab in Dev tools
- Get a fresh performance `ctrl +e` after the page has been fully loaded.
- See that even that checkIfDefined is still being called in a animationFrame alongside checkForUpdate(expected)

### **Expected behavior**

Not using CPU or some other implementation to prevent this, not even sure what `:defined` has to do with Shadow DOM's.

### **Actual behavior**

It's using CPU utilization even in `idle` when not needed

### **System Information:**
 - OS: Windows 10
 - Browser: Vivaldi(3.4.2066.76)
 - Dark Reader version: 4.9.21

### **Additional context**

More info about this https://github.com/darkreader/darkreader/issues/3811#issuecomment-712695274

Gusted:
- Uses a prototype override to detect changes in the style.
- This will improve idle CPU consumption of the browser as it won't have to run `requestAnimationFrame` to detect changes.
- Most of the understanding of prototype overriding came from @tophf's [XRay-Styler](https://github.com/tophf/XRayStyler/). Credits to him!
- NO WINDOW POLLUTING!!
- Resolves #2070
- Resolves #1125
- Resolves #3905

### DIFF
--- a/src/inject/dynamic-theme/index.ts
+++ b/src/inject/dynamic-theme/index.ts
@@ -210,7 +210,7 @@ function createManager(element: StyleElement) {
         }
     }
 
-    function update() {
+    const update = () => {
         const details = manager.details();
         if (!details) {
             return;
@@ -222,6 +222,8 @@ function createManager(element: StyleElement) {
             throttledRenderAllStyles();
         }
     }
+
+    (window as any).update = update;
 
     const manager = manageStyle(element, {update, loadingStart, loadingEnd});
     styleManagers.set(element, manager);
--- a/src/inject/dynamic-theme/index.ts
+++ b/src/inject/dynamic-theme/index.ts
@@ -210,7 +210,7 @@ function createManager(element: StyleElement) {
         }
     }
 
-    const update = () => {
+    function update() {
         const details = manager.details();
         if (!details) {
             return;
@@ -222,8 +222,6 @@ function createManager(element: StyleElement) {
             throttledRenderAllStyles();
         }
     }
-
-    (window as any).update = update;
 
     const manager = manageStyle(element, {update, loadingStart, loadingEnd});
     styleManagers.set(element, manager);
--- a/src/inject/dynamic-theme/index.ts
+++ b/src/inject/dynamic-theme/index.ts
@@ -16,7 +16,7 @@ import {FilterConfig, DynamicThemeFix} from '../../definitions';
 import {generateUID} from '../../utils/uid';
 import {createAdoptedStyleSheetOverride, AdoptedStyleSheetManager} from './adopted-style-manger';
 import {isFirefox} from '../../utils/platform';
-import {injectProxy, removeProxyOnPrototype} from './stylesheet-proxy';
+import {injectProxy} from './stylesheet-proxy';
 
 const variables = new Map<string, string>();
 const INSTANCE_ID = generateUID();
@@ -26,18 +26,27 @@ let filter: FilterConfig = null;
 let fixes: DynamicThemeFix = null;
 let isIFrame: boolean = null;
 
-function createOrUpdateElement(className: string, root: ParentNode = document.head || document, elementType = 'style') {
-    let element = root.querySelector(`.${className}`);
+function createOrUpdateStyle(className: string, root: ParentNode = document.head || document) {
+    let element: HTMLStyleElement = root.querySelector(`.${className}`);
     if (!element) {
-        element = document.createElement(elementType);
+        element = document.createElement('style');
         element.classList.add('darkreader');
         element.classList.add(className);
-        if (element instanceof HTMLStyleElement) {
-            element.media = 'screen';
-        }
+        element.media = 'screen';
     }
     return element;
 }
+
+function createOrUpdateScript(className: string, root: ParentNode = document.head || document) {
+    let element: HTMLScriptElement = root.querySelector(`.${className}`);
+    if (!element) {
+        element = document.createElement('script');
+        element.classList.add('darkreader');
+        element.classList.add(className);
+    }
+    return element;
+}
+
 
 const nodePositionWatchers = new Map<string, ReturnType<typeof watchForNodePosition>>();
 
@@ -52,17 +61,17 @@ function stopStylePositionWatchers() {
 }
 
 function createStaticStyleOverrides() {
-    const fallbackStyle = createOrUpdateElement('darkreader--fallback', document);
+    const fallbackStyle = createOrUpdateStyle('darkreader--fallback', document);
     fallbackStyle.textContent = getModifiedFallbackStyle(filter, {strict: true});
     document.head.insertBefore(fallbackStyle, document.head.firstChild);
     setupNodePositionWatcher(fallbackStyle, 'fallback');
 
-    const userAgentStyle = createOrUpdateElement('darkreader--user-agent',);
+    const userAgentStyle = createOrUpdateStyle('darkreader--user-agent');
     userAgentStyle.textContent = getModifiedUserAgentStyle(filter, isIFrame, filter.styleSystemControls);
     document.head.insertBefore(userAgentStyle, fallbackStyle.nextSibling);
     setupNodePositionWatcher(userAgentStyle, 'user-agent');
 
-    const textStyle = createOrUpdateElement('darkreader--text');
+    const textStyle = createOrUpdateStyle('darkreader--text');
     if (filter.useFont || filter.textStroke > 0) {
         textStyle.textContent = createTextStyle(filter);
     } else {
@@ -71,7 +80,7 @@ function createStaticStyleOverrides() {
     document.head.insertBefore(textStyle, fallbackStyle.nextSibling);
     setupNodePositionWatcher(textStyle, 'text');
 
-    const invertStyle = createOrUpdateElement('darkreader--invert');
+    const invertStyle = createOrUpdateStyle('darkreader--invert');
     if (fixes && Array.isArray(fixes.invert) && fixes.invert.length > 0) {
         invertStyle.textContent = [
             `${fixes.invert.join(', ')} {`,
@@ -87,17 +96,17 @@ function createStaticStyleOverrides() {
     document.head.insertBefore(invertStyle, textStyle.nextSibling);
     setupNodePositionWatcher(invertStyle, 'invert');
 
-    const inlineStyle = createOrUpdateElement('darkreader--inline');
+    const inlineStyle = createOrUpdateStyle('darkreader--inline');
     inlineStyle.textContent = getInlineOverrideStyle();
     document.head.insertBefore(inlineStyle, invertStyle.nextSibling);
     setupNodePositionWatcher(inlineStyle, 'inline');
 
-    const overrideStyle = createOrUpdateElement('darkreader--override');
+    const overrideStyle = createOrUpdateStyle('darkreader--override');
     overrideStyle.textContent = fixes && fixes.css ? replaceCSSTemplates(fixes.css) : '';
     document.head.appendChild(overrideStyle);
     setupNodePositionWatcher(overrideStyle, 'override');
 
-    const variableStyle = createOrUpdateElement('darkreader--variables');
+    const variableStyle = createOrUpdateStyle('darkreader--variables');
     const selectionColors = getSelectionColor(filter);
     const {darkSchemeBackgroundColor, darkSchemeTextColor, lightSchemeBackgroundColor, lightSchemeTextColor} = filter;
     variableStyle.textContent = [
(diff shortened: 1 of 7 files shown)


src/inject/dynamic-theme/style-manager.ts
```diff
@@ -318,44 +318,67 @@ export function manageStyle(element: StyleElement, {update, loadingStart, loadin
...
+
+                // Create a new Event so it can be dispatched.
+                const event = new Event('updateSheet');
+
+                // Create the proxy function that should be identicial to the native one's.
+
+                function proxyAddRule(selector, style, index) {
+                    addRuleFunction.value.call(this, selector, style, index);
+                    this.ownerNode.dispatchEvent(event);
+                    // Check why it returns -1 https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/addRule#Return_value.
```
Gusted, line 347: https://drafts.csswg.org/cssom/#dom-cssstylesheet-deleterule check definition 8.
https://searchfox.org/mozilla-central/source/layout/style/StyleSheet.cpp#640 Firefox return 1;
https://github.com/chromium/chromium/blob/master/third_party/blink/renderer/core/css/css_style_sheet.cc#L442 Chromium return 1;

src/inject/dynamic-theme/style-manager.ts
```diff
@@ -318,44 +318,67 @@ export function manageStyle(element: StyleElement, {update, loadingStart, loadin
...
+                    return -1;
+                }
+                function proxyInsertRule(rule, index) {
+                    // Cache the return value of insertRule as insertRule has the chance to not return the originial index.
```
Gusted, line 351: https://github.com/chromium/chromium/blob/master/third_party/blink/renderer/core/css/css_style_sheet.cc#L322-L367 can return `0` due to errors.
https://searchfox.org/mozilla-central/source/layout/style/StyleSheet.cpp#588-597 can return `0` due to errors.

Gusted:
[darkreader.zip](https://github.com/darkreader/darkreader/files/5470038/darkreader.zip)
Debug copies of this Branch.

And a test page it should make the background-color `rgb(89, 0, 16)`
[aaa.zip](https://github.com/darkreader/darkreader/files/5470039/aaa.zip)

ddnexus:
Just tried the debug copy on the same system that was consuming quite a lot on idle, and it works as expected now. 
Great job! :1st_place_medal:

Gusted:
> 
> 
> Just tried the debug copy on the same system that was consuming quite a lot on idle, and it works as expected now.
> Great job! 🥇

Great to hear that!

Gusted:
Cleaned up a lot of code and made it less messy and more understandable also fixed some bugs. https://github.com/darkreader/darkreader/pull/4005/commits/9e7b381b5efef2f9e85ccde6745731b9bd8e5450

Improved the clean up of the prototype override. https://github.com/darkreader/darkreader/pull/4005/commits/1044344498d9d836d0aa73339634386f9b4b6754

here the build binaries for Firefox & Chromium.
[darkreader.zip](https://github.com/darkreader/darkreader/files/5471852/darkreader.zip)

It should be stable enough now.

ddnexus:
I retried both builds, the difference between dynamic and non-dynamic mode on idle is almost 0 almost on all pages that I tried.

However that doesn't seem to apply to the gmail inbox page. In that case the improvement is just marginal (i.e. CPU difference on idle is still about 10% more for dynamic).

Maybe that's worth some more analysis, because it could be a more general case not limited to gmail.

Here is the profile:
[Profile-20201102T105255.zip](https://github.com/darkreader/darkreader/files/5472731/Profile-20201102T105255.zip)

Gusted:
> However that doesn't seem to apply to the gmail inbox page. In that case the improvement is just marginal (i.e. CPU difference on idle is still about 10% more for dynamic).
> 
> Maybe that's worth some more analysis, because it could be a more general case not limited to gmail.
> 
> Here is the profile:
> [Profile-20201102T105255.zip](https://github.com/darkreader/darkreader/files/5472731/Profile-20201102T105255.zip)

The faulty code is https://github.com/darkreader/darkreader/blob/master/src/inject/dynamic-theme/watch.ts#L47

ddnexus:
Good that you already know the cause, but is there also a solution?

Gusted:
> Good that you already know the cause, but is there also a solution?

I have a idea for a solution that requires a rewrite of this `defined` logic so expect a new commit to fix this within a few days.

src/inject/dynamic-theme/index.ts
```diff
@@ -106,16 +109,28 @@ function createStaticStyleOverrides() {
         `}`
     ].join('\n');
     document.head.insertBefore(variableStyle, inlineStyle.nextSibling);
-    setupStylePositionWatcher(variableStyle, 'variables');
+    setupNodePositionWatcher(variableStyle, 'variables');
+
+    const proxyScript = createOrUpdateElement('darkreader--proxy', undefined, 'script');
+    proxyScript.textContent = `// First get the originial 'native code' of the function.
```
alexanderby, line 115: I suggest you to move the whole function into a separate file. Avoid polluting window scope.
E.g. use something like
```javascript
import {injectProxy} from './stylesheet-proxy';
proxyScript.textContent = `(${goodFunctionName})()`;
```
It will convert the function to a string and execute.

src/inject/dynamic-theme/index.ts
```diff
@@ -422,6 +437,18 @@ export function createOrUpdateDynamicTheme(filterConfig: FilterConfig, dynamicTh
     }
 }
 
+function removeProxy() {
+    const proxy = document.head.querySelector('.darkreader--proxy');
+    if (proxy) {
+        const Dproxy = document.createElement('script');
```
alexanderby, line 443: `Dproxy`? :) Try some different meaningful lowercase name.

src/inject/dynamic-theme/index.ts
```diff
@@ -106,16 +109,28 @@ function createStaticStyleOverrides() {
...
+    proxyScript.textContent = `// First get the originial 'native code' of the function.
+        window['addRuleFunction'] = Object.getOwnPropertyDescriptor(CSSStyleSheet.prototype, 'addRule');
+        window['insertRuleFunction'] = Object.getOwnPropertyDescriptor(CSSStyleSheet.prototype, 'insertRule');
+        window['deleteRuleFunction'] = Object.getOwnPropertyDescriptor(CSSStyleSheet.prototype, 'deleteRule');
+        window['removeRuleFunction'] = Object.getOwnPropertyDescriptor(CSSStyleSheet.prototype, 'removeRule');
+        ${injectProxy.toString()}
+        ${removeProxyOnPrototype.toString()}
+        injectProxy();`;
+    document.head.insertBefore(proxyScript, variableStyle.nextSibling);
+    setupNodePositionWatcher(proxyScript, 'proxy');
```
alexanderby, line 124: I don't think there's a need to watch for its position.

src/inject/dynamic-theme/index.ts
```diff
@@ -25,50 +26,52 @@ let filter: FilterConfig = null;
...
 let isIFrame: boolean = null;
 
-function createOrUpdateStyle(className: string, root: ParentNode = document.head || document) {
-    let style = root.querySelector(`.${className}`) as HTMLStyleElement;
-    if (!style) {
-        style = document.createElement('style');
-        style.classList.add('darkreader');
-        style.classList.add(className);
-        style.media = 'screen';
+function createOrUpdateElement(className: string, root: ParentNode = document.head || document, elementType = 'style') {
```
alexanderby, line 29: Let's create 2 functions: one for style and one for script, because it is not obvious that calling `createOrUpdateElement(className)` returns a style element.

src/inject/dynamic-theme/style-manager.ts
```diff
@@ -318,50 +318,22 @@ export function manageStyle(element: StyleElement, {update, loadingStart, loadin
...
-                update();
-            }
-            rulesCheckFrameId = requestAnimationFrame(checkForUpdate);
-        };
-        checkForUpdate();
-    }
-
-    function unsubscribeFromSheetChanges() {
-        cancelAnimationFrame(rulesCheckFrameId);
+        element.addEventListener('updateSheet', update);
```
alexanderby, line 322: There's a small chance of events name collision. Add some `darkreader` prefix to the event name.

src/inject/dynamic-theme/stylesheet-proxy.ts
```diff
@@ -0,0 +1,43 @@
+export function injectProxy() {
+    // Create a new Event Object that can be dipatched on elements.
+    const updateSheetEvent = new Event('updateSheet');
+
+    // Create the proxy function that should be identicial to the native one's.
+    function proxyAddRule(selector?: string, style?: string, index?: number): number {
+        window['addRuleFunction'].value.call(this, selector, style, index);
+        this.ownerNode.dispatchEvent(updateSheetEvent);
+        // Should always returns -1 https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/addRule#Return_value.
```
alexanderby, line 9: You can remove all comments except this one.

src/inject/dynamic-theme/stylesheet-proxy.ts
```diff
@@ -0,0 +1,43 @@
...
+        this.ownerNode.dispatchEvent(updateSheetEvent);
+    }
+
+    function proxyRemoveRule(index?: number): void {
+        window['removeRuleFunction'].value.call(this, index);
+        this.ownerNode.dispatchEvent(updateSheetEvent);
+    }
+
+    // Define the prototype of the CSSStyleSheet to use the proxy function.
+    Object.defineProperty(CSSStyleSheet.prototype, 'addRule', {...window['addRuleFunction'], value: proxyAddRule});
```
alexanderby, line 31: So, try doing the same without polluting the global scope.

narcolepticinsomniac:
Needless to say, this PR is awesome as well. Previously, Chromium devtools were practically unusable when the dynamic theme was active on a page. This resolves whatever weird conflict element highlighting has with `requestAnimationFrame`, and the non-stop, ridiculous CPU consumption that comes along with it. Also resolves some other weird bugs I've encountered on a somewhat regular basis on certain sites.

src/inject/dynamic-theme/stylesheet-proxy.ts
```diff
@@ -0,0 +1,43 @@
...
+    Object.defineProperty(CSSStyleSheet.prototype, 'insertRule', {...window['insertRuleFunction'], value: proxyInsertRule});
+    Object.defineProperty(CSSStyleSheet.prototype, 'deleteRule', {...window['deleteRuleFunction'], value: proxyDeleteRule});
+    Object.defineProperty(CSSStyleSheet.prototype, 'removeRule', {...window['removeRuleFunction'], value: proxyRemoveRule});
+}
+
+export function removeProxyOnPrototype() {
+    // Define the prototype of the CSSStyleSheet to the orginial function.
+    Object.defineProperty(CSSStyleSheet.prototype, 'addRule', {...window['addRuleFunction']});
```
alexanderby, line 39: Instead try launching the cleanup by dispatching a custom event on document, or saving a removal function into window scope like `__darkreader__removeCSSStyleSheetProxy__()` (more ugly).

alexanderby (changes requested):

alexanderby:
One ugly thing about adding style is that some sites can have CSP no inline style etc.

src/inject/dynamic-theme/stylesheet-proxy.ts
```diff
@@ -0,0 +1,46 @@
+export function injectProxy() {
+    if (!window['__darkreader__addRuleFunction']) {
+        window['__darkreader__addRuleFunction'] = Object.getOwnPropertyDescriptor(CSSStyleSheet.prototype, 'addRule');
```
alexanderby, line 3: You can just store the descriptors in the function scope like `const addRuleDescriptor = Object.getOwnPropertyDescriptor(CSSStyleSheet.prototype, 'addRule')` :)

alexanderby (changes requested):

alexanderby (approved):
Great!

Now the new problem arises: if some style sheet is changed synchronously many times, the update will be called too often. The `throttle` function can come into rescue. Or maybe `queueMicrotask` can help there. Let it be a problem of another PR.

Gusted:
> Let it be a problem of another PR.

It will be a issue. I have spend too many hours on this PR as well.

Gusted:
> Good that you already know the cause, but is there also a solution?

#4055 It's possible however I still know half of what the defined Elements, so it's up to @alexanderby for this one.

PR #4009, opened 2020-11-01
Fix for kinoart.ru

Rakleed:
Fix logos.

Examples: https://kinoart.ru/, https://kinoart.ru/reviews/ot-sokrata-k-boratu-mozhet-li-borat-2-chto-to-izmenit.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3516,6 +3516,16 @@ CSS
 
 ================================
 
+kinoart.ru
+
+INVERT
+._2nHMA
+._1Hgym
+.sd8Hf
+._1YCgN
+
+================================
+
 komputerswiat.pl
 
 INVERT
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3519,10 +3519,10 @@ CSS
 kinoart.ru
 
 INVERT
-._2nHMA
-._1Hgym
-.sd8Hf
-._1YCgN
+main > div > svg
+header > a > svg
+footer > div > svg
+header > div > a > svg
 
 ================================
 


src/config/dynamic-theme-fixes.config
```diff
@@ -3516,6 +3516,16 @@ CSS
 
 ================================
 
+kinoart.ru
+
+INVERT
```
Gusted, line 3521: All are generated and likely to change in the future on a predictable basis.
The best is to use DOM based selector e.g. for `._2nHMA`
```CSS
main > div > svg`
```

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
Rakleed: As I understand it, there shouldn't be a "\`" at the end. But even without it, this rule only works for the logo next to the header. In this case, the logo next to the footer is not fixed, and on the second page the logos next to the header are not fixed either. I tried `svg`, but then other elements change too, which shouldn't change.

Gusted: It was rather a example to use the dom based selectors. As like I said using the selector can break in a near future.

Rakleed: I tried to pick a DOM based selector myself that would only invert the logos, but I couldn't. If you can find one, it will be great.

Gusted:
```CSS

kinoart.ru

INVERT
main > div > svg
footer > div > svg
header > div > a > svg
```

?

Rakleed: Thanks. And what is the `main` tag, because it is not in the page code? What is it used for?

Rakleed: Didn't notice. Thanks again!

PR #4017, opened 2020-11-02
Add arewewaylandyet.com to dark-sites

mpsq:
[arewewaylandyet.com](https://arewewaylandyet.com/) has a dark mode enabled by default when `prefers-color-scheme` is set to dark.


Gusted:
> has a dark mode enabled by default when `prefers-color-scheme` is set to dark.

It has to be dark regardless of the prefers-color-scheme.

PR #4019, opened 2020-11-02
Update Gradient regex

Gusted:
- Faster and cleaner Regexp and has protection against backtracking regexp.

### DIFF
--- a/src/inject/dynamic-theme/modify-css.ts
+++ b/src/inject/dynamic-theme/modify-css.ts
@@ -230,7 +230,7 @@ function getColorModifier(prop: string, value: string): string | CSSValueModifie
     }
 }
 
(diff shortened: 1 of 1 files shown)


alexanderby:
Not sure if it will work in Firefox, could you please add some tests for changing gradient colors? (Karma tests)

Gusted:
Hmm seems like I can't execute browsers test from either puppeteer or Karma.

alexanderby:
Did you try Linux? Strange that Karma doesn't work. By the way, the windows should be focused, or Dark Reader will just wait for focus in Dynamic mode. For Puppeteer I used random paths from Stack overflow, didn't check.

Gusted:
https://pastebin.com/raw/KSUkkjtg is the error log. And puppeteer seems not to connect to Firefox even with the workaround that was posted in the puppeteer issue where I saw you had a comment =).

PR #4055, opened 2020-11-06
Move definedElements to darkreader's proxy

Gusted:
- Instead of using rAF use the page's context to check when the custom element is defined and let the content script know it can be resolved.

Is their a simple test case where this is actually helpful? as I tried reading the code and understand it and wasn't able to have a simple test case to test this against.

### DIFF
--- a/src/inject/dynamic-theme/stylesheet-proxy.ts
+++ b/src/inject/dynamic-theme/stylesheet-proxy.ts
@@ -10,17 +10,17 @@ export function injectProxy() {
         Object.defineProperty(CSSStyleSheet.prototype, 'deleteRule', deleteRuleDescriptor);
         Object.defineProperty(CSSStyleSheet.prototype, 'removeRule', removeRuleDescriptor);
         document.removeEventListener('__darkreader__cleanUp', cleanUp);
-        document.removeEventListener('__darkreader__addUndefinedResolver', (e: any) => addUndefinedResolver(e.details.tag));
+        document.removeEventListener('__darkreader__addUndefinedResolver', (e: CustomEvent<{tag: string}>) => addUndefinedResolver(e));
     };
 
-    const addUndefinedResolver = ($tag: string) => {
-        customElements.whenDefined($tag).then(() => {
-            document.dispatchEvent(new CustomEvent('__darkreader__isDefined', {detail: {tag: $tag}}));
+    const addUndefinedResolver = (e: CustomEvent<{tag: string}>) => {
+        customElements.whenDefined(e.detail.tag).then(() => {
+            document.dispatchEvent(new CustomEvent('__darkreader__isDefined', {detail: {tag: e.detail.tag}}));
         });
     };
 
     document.addEventListener('__darkreader__cleanUp', cleanUp);
-    document.addEventListener('__darkreader__addUndefinedResolver', (e: any) => addUndefinedResolver(e.details.tag));
+    document.addEventListener('__darkreader__addUndefinedResolver', (e: CustomEvent<{tag: string}>) => addUndefinedResolver(e));
 
     const updateSheetEvent = new Event('__darkreader__updateSheet');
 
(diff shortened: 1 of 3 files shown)


src/inject/dynamic-theme/watch.ts
```diff
@@ -37,24 +37,24 @@ function collectUndefinedElements(root: ParentNode) {
...
+const resolvers = new Map<string, () => void>();
+
+const handleIsDefined = (e: any) => {
+    if (resolvers.has(e.details.tag)) {
+        const resolve = resolvers.get(e.details.tag);
+        resolve && resolve();
+    }
+}
+
+function customElementsWhenDefined($tag: string) {
```
alexanderby, line 49: Why did it switch from `tag` to `$tag`? Usually `$` is a variable prefix in PHP. Also `$` was used as a global jQuery variable, and there was a convention for a jQuery selector object like `var $elements = $('.some-class').` I sometimes prepend `$` for some encoded strings, or when a variable is later processed into another one without `$`, maybe it is a bad practice.

src/inject/dynamic-theme/watch.ts
```diff
@@ -37,24 +37,24 @@ function collectUndefinedElements(root: ParentNode) {
         });
 }
 
-function customElementsWhenDefined(tag: string) {
+const resolvers = new Map<string, () => void>();
+
+const handleIsDefined = (e: any) => {
```
alexanderby, line 42: Worth using type `CustomEvent<{tag: string}>`.

I also suggest you to think about replacing `const` with `function`. Some would say that `const` is not overridable, uses a proper `this` scope when defined in classes, but I find it more readable and shorter, especially in root scope. `() => {}` are good when they are one-liners or passed as an argument into another function.

Funny article about this topic https://www.freecodecamp.org/news/constant-confusion-why-i-still-use-javascript-function-statements-984ece0b72fd/

src/inject/dynamic-theme/stylesheet-proxy.ts
```diff
@@ -10,8 +10,17 @@ export function injectProxy() {
         Object.defineProperty(CSSStyleSheet.prototype, 'deleteRule', deleteRuleDescriptor);
         Object.defineProperty(CSSStyleSheet.prototype, 'removeRule', removeRuleDescriptor);
         document.removeEventListener('__darkreader__cleanUp', cleanUp);
+        document.removeEventListener('__darkreader__addUndefinedResolver', (e: any) => addUndefinedResolver(e.details.tag));
     };
+
+    const addUndefinedResolver = ($tag: string) => {
```
alexanderby, line 16: `tag` argument seems better, you can put it into `detail: {tag}`

src/inject/dynamic-theme/stylesheet-proxy.ts
```diff
@@ -10,8 +10,17 @@ export function injectProxy() {
...
+        customElements.whenDefined($tag).then(() => {
+            document.dispatchEvent(new CustomEvent('__darkreader__isDefined', {detail: {tag: $tag}}));
+        });
+    };
+
     document.addEventListener('__darkreader__cleanUp', cleanUp);
+    document.addEventListener('__darkreader__addUndefinedResolver', (e: any) => addUndefinedResolver(e.details.tag));
```
alexanderby, line 23: Worth using type `CustomEvent<{tag: string}>`

alexanderby (changes requested):
Hmmm, the content script talks with the page script using the event? I'm afraid the potential problem could be if element will be defined before the message will be delivered and the subscription will happen. But let's see it in real world. Other messaging type could be `window.postMessage`, but it can have the same problem.

**This messaging can introduce vulnerabilities, so always think about not sharing any user's data or giving ability to hackers to interact with the background page!**

src/inject/dynamic-theme/watch.ts
Gusted: Probably to see the difference in

```typescript
 {detail: {tag: tag}}));
````

src/inject/dynamic-theme/watch.ts
Gusted: >  but I find it more readable and shorter, especially in root scope. `() => {}` are good when they are one-liners or passed as an argument into another function.

I agree with you on this and that's why I used a `const` for this one as it's going to be passed.

Gusted:
image
Seems to be working when I import #4109 and test it with Chrome(Which doesn't have the Custom Elements available)

alexanderby (approved):
Speaking about `customElements.define` availability, in our Karma tests Dynamic mode runs in page context, so the method is available. It is possible to test it out in Puppeteer, so you can add such a test if you wish.

PR #4077, opened 2020-11-09
Added startpage.com to ignore list

True-cc:


Gusted:
Hi @True-cc!

Startpage detects if the system prefers a `dark` color-scheme. if this is the case startpage is indeed dark, however if your system prefers a `light` color-scheme it will show a light theme 
image Therefore it isn't dark by default.

Regards,
Gusted

PR #4083, opened 2020-11-10
Fixed menu item icons in Google Drive

bendavis78:
Menu item icons were not inverting in google drive. This fixes it.


Gusted:
Good catch!

PR #4096, opened 2020-11-11
Add isEnabled function to API

ZeProf2Code:


Gusted:
Hi @ZeProf2Code!

I'm not quite sure if Dark Reader should provide an `toggle` option. 
As website who like to use this, already have their own toggle way with e.g. localStorage and based on that they enable/disable Dark Reader.

Regards,
Gusted

alexanderby:
Could you please tell more about why do you need this function?

alexanderby:
@ZeProf2Code?

Yash-Singh1:
I don't think that the `toggle` option is really needed but the `enabled` boolean could be useful.

src/api/index.ts
```diff
@@ -22,10 +23,20 @@ export function enable(themeOptions: Partial<Theme> = {}, fixes: DynamicThemeFix
         throw new Error('Theme engine is not supported.');
     }
     createOrUpdateDynamicTheme(theme, fixes, isIFrame);
+    enabled = true;
+}
+
+export function toggle(themeOptions: Partial<Theme> = {}, fixes: DynamicThemeFix = null) {
```
alexanderby, line 29: Let's expose `isEnabled()` function instead`

alexanderby (changes requested):

src/api/index.ts
```diff
@@ -6,6 +6,7 @@ import ThemeEngines from '../generators/theme-engines';
 import {createOrUpdateDynamicTheme, removeDynamicTheme} from '../inject/dynamic-theme';
 import {collectCSS} from '../inject/dynamic-theme/css-collection';
 
+let isEnabled = false;
```
Gusted, line 9: Variable name is inconsistent.
also this will now collude with the `isEnabled` function.
I think `isDarkReaderActive` is a good variable or `isDarkReaderEnabled`.

Yash-Singh1:
I tried this locally and it worked:
Screenshot from 2020-11-19 09-31-20

Gusted:
I think only some changes should be made to README.md and index.d.ts file to add the new function and then should be ready to go.

PR #4104, opened 2020-11-12
Add more tests

Gusted:
- Add Media-Query karma tests.
- Add Shadow-Dom karma tests.


alexanderby (approved):

alexanderby:
Some ideas for future tests:
- Style elements with `@media ... { ... }` CSS.
- Custom Elements. If you need an example of how to define a custom element, [here's one](https://stackoverflow.com/a/62932376/4137472). So you can test when element is defined, and when it is not yet defined (call define() later) to test if it is properly searching for styles inside the custom element.

alexanderby:
I will later add some tests for Link elements and image analysis.

Gusted:
Sure things!

PR #4109, opened 2020-11-13
Add more Karma test

Gusted:
- Add defined elements test.
- Add `@media {}` test.


tests/inject/dynamic/shadow-dom.tests.ts
```diff
@@ -88,4 +88,53 @@ describe('Handle Shadow-DOM', () => {
...
+        }
+        customElements.define('svg-container', SVGContainer);
+
+        createOrUpdateDynamicTheme(theme, null, false);
+
+        await timeout(100);
+        expect(getComputedStyle(document.querySelector('svg-container').shadowRoot.querySelector('text')).color).toBe('rgb(255, 26, 26)');
+    });
+
+    it('should react to defined elements', async () => {
```
Gusted, line 115: This is currently failing. `--sync` contains 0 rules no error has been throwned.

tests/inject/dynamic/shadow-dom.tests.ts
```diff
@@ -88,4 +88,53 @@ describe('Handle Shadow-DOM', () => {
...
+            '       <style class="testcase-style">#svg-text { color: red; }</style>',
+            '       <text id="svg-text" y="20">I am SVG</text>',
+            '   </svg>',
+            '</svg-container>',
+        );
+        class SVGContainer extends HTMLElement {
+            constructor() {
+                super();
+                const shadowRoot = this.attachShadow({mode: 'open'});
+                shadowRoot.append(...this.childNodes as any);
```
alexanderby, line 104: Well, that was a bad example :) No need to copy anything, it's a logic for another purpose highlighted in that stack overflow answer. So, you can append some style to the shadow root inside the constructor.

tests/inject/dynamic/shadow-dom.tests.ts
```diff
@@ -88,4 +88,53 @@ describe('Handle Shadow-DOM', () => {
...
+            }
+        }
+
+        createOrUpdateDynamicTheme(theme, null, false);
+
+        await timeout(100);
+        customElements.define('svg-container2', SVGContainer);
+        await timeout(100);
+        expect(getComputedStyle(document.querySelector('svg-container2').shadowRoot.querySelector('text')).color).toBe('rgb(255, 26, 26)');
```
alexanderby, line 137: If the test doesn't work, you first need to determine why, and if nothing helps remove it.

By the way, each test is independent and you can remove `-2` postfixes from class names etc.

tests/inject/dynamic/shadow-dom.tests.ts
```diff
@@ -88,4 +88,53 @@ describe('Handle Shadow-DOM', () => {
...
+            '       <style class="testcase-style">#svg-text { color: red; }</style>',
+            '       <text id="svg-text" y="20">I am SVG</text>',
+            '   </svg>',
+            '</svg-container2>',
+        );
+        class SVGContainer extends HTMLElement {
+            constructor() {
+                super();
+                const shadowRoot = this.attachShadow({mode: 'open'});
+                shadowRoot.append(...this.childNodes as any);
```
alexanderby, line 128: Same as above, It is common to define styles inside a custom element's constructor.

alexanderby (changes requested):

tests/inject/dynamic/shadow-dom.tests.ts
Gusted: > By the way, each test is independent

Wrong, it is on the same page and the CustomElementRegistry isn't cleared therefore obviously so removing `-2`.
```
NotSupportedError: Failed to execute 'define' on 'CustomElementRegistry': the name "svg-container" has already been used with this registry
```

tests/inject/dynamic/media-query.tests.ts
```diff
@@ -77,10 +77,30 @@ describe('Handle Media Queries', () => {
...
         await timeout(100);
-        
+
         expect(getComputedStyle(document.querySelector('h1')).backgroundColor).toBe('rgb(0, 102, 0)');
         expect(getComputedStyle(document.querySelector('h1 strong')).color).toBe('rgb(255, 174, 26)');
         expect(getComputedStyle(document.body).backgroundColor).toBe('rgb(0, 0, 0)');
     });
-    
+
+    it('should check for CSS @media', async () => {
```
alexanderby, line 86: I remember you had a test with `<style media="...">`. I think it's worth adding it too along with testing that `<style media="print">` is ignored.

tests/inject/dynamic/shadow-dom.tests.ts
```diff
@@ -88,4 +88,67 @@ describe('Handle Shadow-DOM', () => {
         expect(getComputedStyle(shadowRoot.querySelector('p')).color).toBe('rgb(255, 26, 26)');
     });
 
+    it('should handle defined elements', async () => {
+        container.innerHTML = multiline(
+            '<svg-container>',
```
alexanderby, line 93: You can remove the content inside the custom element, it is irrelevant to the test. Also rename `svg-container` to something neutral.

tests/inject/dynamic/shadow-dom.tests.ts
```diff
@@ -88,4 +88,67 @@ describe('Handle Shadow-DOM', () => {
...
+            '</svg-container>',
+        );
+        class SVGContainer extends HTMLElement {
+            constructor() {
+                super();
+                const shadowRoot = this.attachShadow({mode: 'open'});
+                const style = document.createElement('style');
+                style.textContent = 'p { color: pink }';
+                const text = document.createElement('p');
+                text.textContent = 'Some text content that should be pink.';
```
alexanderby, line 107: text.textContent? Rename `text` to something else, like `p` or `element` etc.

tests/inject/dynamic/shadow-dom.tests.ts
```diff
@@ -88,4 +88,67 @@ describe('Handle Shadow-DOM', () => {
...
+    it('should handle defined elements', async () => {
+        container.innerHTML = multiline(
+            '<svg-container>',
+            '   <svg viewBox="0 0 100 30">',
+            '       <style class="testcase-style">#svg-text { color: red; }</style>',
+            '       <text id="svg-text" y="20">I am SVG</text>',
+            '   </svg>',
+            '</svg-container>',
+        );
+        class SVGContainer extends HTMLElement {
```
alexanderby, line 100: Worth using some neutral name instead of `SVGContainer`, because SVG is not relevant to this test.

tests/inject/dynamic/shadow-dom.tests.ts
```diff
@@ -88,4 +88,67 @@ describe('Handle Shadow-DOM', () => {
...
+    });
+
+    it('should react to defined elements', async () => {
+        container.innerHTML = multiline(
+            '<svg-container-2>',
+            '   <svg viewBox="0 0 100 30">',
+            '       <style class="testcase-style">#svg-text { color: red; }</style>',
+            '       <text id="svg-text" y="20">I am SVG</text>',
+            '   </svg>',
+            '</svg-container2>',
```
alexanderby, line 129: Pay attention that tags don't match. Use the same suggestions as for a test above.

alexanderby (changes requested):

PR #4112, opened 2020-11-13
Update font files

Gusted:
- Update fonts to better compressed ones.
- Reduce size of fonts.

Fetched from https://fonts.google.com/download?family=Open%20Sans


alexanderby:
Hmmm, strange, the version number looks the same. Maybe they've thrown away something, like Hebrew, or Chinese characters, who knows. Worth comparing if possible.

Gusted:
> Hmmm, strange, the version number looks the same. Maybe they've thrown away something, like Hebrew, or Chinese characters, who knows. Worth comparing if possible.

They are rendered correctly however don't seems to be affected by the font itself.

However checking the info of the TFF it says 938 glyphs whereby the ones with `covering the Latin, Greek and Cyrillic alphabets with a wide range of diacritics. ` are 897 so it's safe to assume it covers a lot of alphabets.
image

alexanderby (approved):

PR #4117, opened 2020-11-14
Handle multiple running instances of Dark Reader

### LINKED ISSUE #4073: [BUG] Inconsistent settings sync in Vivaldi
### **Describe the bug**

The site list (and possibly other settings, but I don't use them) syncs inconsistently across my multiple Vivaldi installations on different windows machines. Sometimes my changes propagate, sometimes my whitelists just keep diverging over time (measured over minutes, hours, and weeks). Sync otherwise works fine for everything else in my browser, even disabling/enabling dark reader as an addon, sometimes instantaneously.

Manually running syncs after whitelist changes shows changes being detected and shared by the browser but not picked up/applied by the addon on the receiving end. Sometimes restarting both vivaldis at the same time gets the whitelists straightened out, though which version of the list wins is uncertain.

Sorry this isn't a very precise description but it's pretty random/unpredictable to me. Since vivaldi is a chrome clone, I wouldn't expect the issue to be their sync implementation.

### **Steps to reproduce**

- add/remove sites to/from the whitelist on separate instances of vivaldi in sequence/parallel and with browser restarts here and there (sync manually to save your time), i.e. just fast-forward emulate normal multi-device browsing behavior, you'll see the lists diverge after a while

### **Expected behavior**

The whitelist converges to the same list across multiple devices over time, especially when syncs occur

### **Actual behavior**

Lists diverge and stay diverged.

### **System Information:**
 - OS: windows 10
 - vivaldi 3.4.2066.94 (Stable channel) (64-bit)
 - Dark Reader version: 4.9.22

Gusted:
- Make sure all settings are consistent across running instances of Dark Reader.
- Resolves #4073


src/background/user-storage.ts
```diff
@@ -2,13 +2,18 @@ import {DEFAULT_SETTINGS, DEFAULT_THEME} from '../defaults';
 import {debounce} from '../utils/debounce';
 import {isURLMatched} from '../utils/url';
 import {UserSettings} from '../definitions';
-import {readSyncStorage, readLocalStorage, writeSyncStorage, writeLocalStorage} from './utils/extension-api';
+import {readSyncStorage, readLocalStorage, writeSyncStorage, writeLocalStorage, isWriting} from './utils/extension-api';
```
alexanderby, line 5: Maybe I miss something, maybe it works the other way in transpiled code, but I'm sure that `isWriting` will always be a constant value. Use `someObject.isWriting` instead. Anyway no need in it, see below...

src/background/user-storage.ts
```diff
@@ -2,13 +2,18 @@ import {DEFAULT_SETTINGS, DEFAULT_THEME} from '../defaults';
...
 
 const SAVE_TIMEOUT = 1000;
 
+interface UserStorageOptions {
+    onSettingsChange: () => any;
+}
+
 export default class UserStorage {
-    constructor() {
+    constructor({onSettingsChange}: UserStorageOptions) {
```
alexanderby, line 14: Maybe let's rename to `onRemoteSettingsChange` or something like this? So it will be clear that the callback is triggered when settings are changed from the outside.

src/background/user-storage.ts
```diff
@@ -2,13 +2,18 @@ import {DEFAULT_SETTINGS, DEFAULT_THEME} from '../defaults';
...
         this.settings = null;
+        chrome.storage.onChanged.addListener(() => !isWriting && this.loadSettings() && onSettingsChange());
```
alexanderby, line 16: Put this into `extension-api.ts` file and expose in some function like `subscribeToOuterSettingsChange` or similar.

alexanderby (changes requested):

src/background/user-storage.ts
```diff
@@ -2,13 +2,18 @@ import {DEFAULT_SETTINGS, DEFAULT_THEME} from '../defaults';
...
 
+interface UserStorageOptions {
+    onRemoteSettingsChange: () => any;
+}
+
 export default class UserStorage {
-    constructor() {
+    constructor({onRemoteSettingsChange}: UserStorageOptions) {
         this.settings = null;
+        subscribeToOuterSettingsChange(this.loadSettings && onRemoteSettingsChange);
```
alexanderby, line 16: Not sure if it will work now. `this.loadSettings &&` acts as boolean and will be always true. Did you mean something else? Don't forget that `this.loadSettings` is `async`.

src/background/utils/extension-api.ts
```diff
@@ -64,15 +60,17 @@ export function writeSyncStorage<T extends {[key: string]: any}>(values: T): Pro
...
     return new Promise<void>((resolve) => {
         chrome.storage.local.set(values, () => {
-            isWriting = false;
             resolve();
         });
     });
 }
 
+export const subscribeToOuterSettingsChange = (callback: () => void) => {
+    chrome.storage.onChanged.addListener(callback);
```
alexanderby, line 71: Previous use of `isWriting` seemed to be valid. So you could use `!isWriting && ...` to ensure the callback is called for remote changes only. By the way, I don't know if `chrome.storage.onChanged` reacts on both local and outer changes.

alexanderby (changes requested):

src/background/utils/extension-api.ts
Gusted: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/onChanged Says it could be from "managed" "remote" "local"

Gusted:
I've used `setTimeout` to reset the value as after the `resolve()` the callback will be actaully triggered and then the isWriting value should be still `true`, not sure if this work.

alexanderby (approved):

PR #4127, opened 2020-11-15
Added keywords

Yash-Singh1:


package.json
```diff
@@ -32,6 +32,17 @@
   "bugs": {
     "url": "https://github.com/darkreader/darkreader/issues"
   },
+  "keywords": [
+    "dark-theme",
+    "chrome-extension",
```
Gusted, line 37: `chrome-extension` and `firefox-addon` can be removed as this is a NPM package and not the full extension.

Gusted (changes requested):

package.json
Yash-Singh1: Done in e71245a

PR #4138, opened 2020-11-15
 Remove manger when Style becomes disabled

### LINKED ISSUE #3841: [Broken Website] infinitysearch.co
**Website**
infinitysearch.co

**How to Reproduce**
An example of this is
> 1. Go to infinitysearch.co

**Expected behavior**
sshot-005

**Actual behavior**
sshot-004

After reenable dark reader it will look like the above one.

**System info:**
 - OS: Windows 10
 - Browser: Firefox
 - Darkreader Version: 4.9.17

Gusted:
- Remove the manager of an StyleElement when the disabled attribute is being set.
- Add test cases to not handle disabled StyleElements.
- Fixes bug where the previous link wasn't removed.
- `document.head` -> `container` in `should move override` test.
- Resolves #3841

### DIFF
--- a/src/inject/dynamic-theme/watch.ts
+++ b/src/inject/dynamic-theme/watch.ts
@@ -166,7 +166,7 @@ export function watchForStyleChanges(currentStyles: StyleElement[], update: (sty
             if (m.target.isConnected) {
                 if (shouldManageStyle(m.target)) {
                     updatedStyles.add(m.target as StyleElement);
-                } else if (shouldManageStyle(m.target, {disableDisabled: true})) {
+                } else if ((m.target as HTMLLinkElement).disabled && shouldManageStyle(m.target, {disableDisabled: true})) {
                     removedStyles.add(m.target as StyleElement);
                 }
             }
--- a/src/inject/dynamic-theme/watch.ts
+++ b/src/inject/dynamic-theme/watch.ts
@@ -163,11 +163,12 @@ export function watchForStyleChanges(currentStyles: StyleElement[], update: (sty
         const updatedStyles = new Set<StyleElement>();
         const removedStyles = new Set<StyleElement>();
         mutations.forEach((m) => {
-            if (m.target.isConnected) {
-                if (shouldManageStyle(m.target)) {
-                    updatedStyles.add(m.target as StyleElement);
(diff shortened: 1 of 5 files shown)


src/inject/dynamic-theme/watch.ts
```diff
@@ -161,16 +161,21 @@ export function watchForStyleChanges(currentStyles: StyleElement[], update: (sty
...
     function handleAttributeMutations(mutations: MutationRecord[]) {
         const updatedStyles = new Set<StyleElement>();
+        const removedStyles = new Set<StyleElement>();
         mutations.forEach((m) => {
-            if (shouldManageStyle(m.target) && m.target.isConnected) {
-                updatedStyles.add(m.target as StyleElement);
+            if (m.target.isConnected) {
+                if (shouldManageStyle(m.target)) {
+                    updatedStyles.add(m.target as StyleElement);
+                } else if ((m.target as HTMLLinkElement).disabled && shouldManageStyle(m.target, {disableDisabled: true})) {
```
alexanderby, line 169: Very nice, but ... is there a need in options.disableDisabled? Funny name BTW, should we disable disabled or enable disabled? 🤣 

Maybe replace with `} else if (m.target instanceof HTMLLinkElement && m.target.disabled) {`, it is not an expensive operation and code looks simpler.

src/inject/dynamic-theme/watch.ts
```diff
@@ -161,16 +161,21 @@ export function watchForStyleChanges(currentStyles: StyleElement[], update: (sty
 
     function handleAttributeMutations(mutations: MutationRecord[]) {
         const updatedStyles = new Set<StyleElement>();
+        const removedStyles = new Set<StyleElement>();
         mutations.forEach((m) => {
```
alexanderby, line 165: Worth using destructuring `{target}` by the way.

alexanderby (changes requested):

src/inject/dynamic-theme/watch.ts
Gusted: > Very nice, but ... is there a need in options.disableDisabled? Funny name BTW, should we disable disabled or enable disabled? 

`anPossibleCheckToExcludeDisableFromBeingTestedInTheCondition`

alexanderby (approved):

PR #4141, opened 2020-11-15
Remove unused store

Gusted:
- The `store` variable isn't being utilized and therefore it can be removed.


alexanderby (approved):
I can't get how it was supposed to work :)

PR #4142, opened 2020-11-15
Enforce a stricter Code-Style

Gusted:
- Enforce a stricter Code-style to get a cleaner and consistent code-style across the code base.
- Disable various ESLint plugins and replace them with Typescript-ESLint.
- Get some compiler optimizations due to stricter code-style/Type Checking. *Notice-able on special machines*.


.eslintrc.js
```diff
@@ -30,11 +24,11 @@ module.exports = {
         'space-in-parens': ['error', 'never'],
         'space-infix-ops': 'error',
         'spaced-comment': ['error', 'always', {exceptions: ['-']}],
-        'quotes': ['error', 'single', {
-            allowTemplateLiterals: true,
-            avoidEscape: true,
-        }],
-        'eol-last': ["error", "always"],
+        'quotes': 'off',
+        'eol-last': ['error', 'always'],
```
alexanderby, line 28: Sort alphabetically

.eslintrc.js
```diff
@@ -43,13 +37,36 @@ module.exports = {
         '@typescript-eslint/no-object-literal-type-assertion': 'off',
         '@typescript-eslint/no-use-before-define': 'off',
         '@typescript-eslint/explicit-module-boundary-types': 'off',
-        '@typescript-eslint/no-duplicate-imports': 'error',
+        '@typescript-eslint/no-duplicate-imports': 'warn',
```
alexanderby, line 40: Let it be an error. Warnings are usually used when ESLint is attached to old code base, and it is painful to change everything.

.eslintrc.js
```diff
@@ -43,13 +37,36 @@ module.exports = {
...
+        '@typescript-eslint/array-type': ['warn', {
```
alexanderby, line 41: I find it more suitable to be `string[]`, or `MyType[]` or `Promise<MyType[]>` than `Array<string>`, `Array<MyType>` and `Promise<Array<MyType>>`. But something like `(() => void)[]` can look strange. Let's try a magic [`array-simple` option](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/array-type.md#array-simple).

.eslintrc.js
```diff
@@ -43,13 +37,36 @@ module.exports = {
...
+            default: 'generic',
+        }],
+        '@typescript-eslint/consistent-type-assertions': 'error',
+        '@typescript-eslint/consistent-type-imports': 'error',
+        '@typescript-eslint/member-delimiter-style': 'error',
+        '@typescript-eslint/no-implicit-any-catch': 'warn',
```
alexanderby, line 47: I don't think there's a point in `catch (err: unknown)`. It will always be unknown. JS doesn't support multiple catches based on error type. One may throw anything, like `throw 'message'`, but it will not generate stack trace, so people normally do `throw new Error('message')`

alexanderby (changes requested):

alexanderby (changes requested):
I suggest you to merge the latest master, resolve conflicts and run the Code Style again.

PR #4145, opened 2020-11-16
fix for linkedin.com

Myshor:
One more fight with black text.
Messing with linkedin variables.
Fixing artdeco-pill in jobn search tab

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3820,13 +3820,10 @@ CSS
 
 linkedin.com
 
-INVERT
-.artdeco-button
-.jobs-search-box__input-icon
-.artdeco-pill
-.artdeco-tabs .artdeco-tab
-
 CSS
+:root {
+    --color-text-low-emphasis: $(gray);
+}
 .js-job-card-company-logo {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -3820,13 +3820,10 @@ CSS
...
 
-INVERT
-.artdeco-button
-.jobs-search-box__input-icon
-.artdeco-pill
-.artdeco-tabs .artdeco-tab
-
 CSS
+:root {
+    --color-text-low-emphasis: $(gray);
```
alexanderby, line 3825: Did you mean `--color-text-low-emphasis: ${gray} !important;`?

Myshor: yup... still checking how it will work with all places around linkedin

Myshor: normaly there was some dark color set as variable. If it will make some text too dark somewhere then I will try to make the original one included here.

alexanderby: It will not work with round brackets.

PR #4177, opened 2020-11-18
Create .browserslistrc

Yash-Singh1:
This adds in browser support data for the DarkReader API. Data brought from `caniuse`: https://caniuse.com/css-filters

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,4 +1,0 @@
-# Browser support for the DarkReader API
-last 2 major versions
-not ie > 0
-not ie_mob > 0
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,0 +1,4 @@
+# Browser support for the DarkReader API
+last 2 major versions
+not ie > 0
+not ie_mob > 0
--- a/package.json
+++ b/package.json
@@ -41,6 +41,20 @@
     "accessibility",
     "eye-care"
   ],
+  "browserslist": [
+    "supports es5",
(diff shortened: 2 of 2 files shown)


Gusted:
Hi @Yash-Singh1!

What's the purpose of having an `.browserslistrc` file?

Regards,
Gusted

Yash-Singh1:
@Gusted `Browserslist` is a method of sharing target browsers of a front-end tool. If you go to the root of the repository of `darkreader` now, and run `npx browserslist`, it will show the supported browsers of the `darkreader` API. Take a look at their organization: https://github.com/browserslist/browserslist. This tool is very popular and I have seen a `.browserslistrc` in about 60% of front-end tools. The `.browserslistrc` is one method of storing the `browserslist` queries. If you want, we can store the `browserslist` queries in the `package.json`.

.browserslistrc
```diff
@@ -0,0 +1,4 @@
+# Browser support for the DarkReader API
+last 2 major versions
```
Gusted, line 2: Hmm `last 2 major versions`, our DarkReader API should support ES5 as it's gets transpiled to ES5
https://github.com/darkreader/darkreader/blob/master/tasks/bundle-api.js#L27 not sure how this can be interpenetrated in the browserlist syntax.

Yash-Singh1: @Gusted I made sure that the browser supports `es5` and `css-filters` in 5d8da73.

PR #4189, opened 2020-11-18
Added in issue configuration

Yash-Singh1:


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1 @@
+blank_issues_enabled: true
```
Gusted, line 1: Is this an problem?
You can open up Blank_Issues
image

I actually think disabling blank issues is an better idea, because their is no real use for it. The issues template provide most information for issues.

Yash-Singh1: Ok

PR #4190, opened 2020-11-18
Added several restricted schemes in Chrome and Edge

Yash-Singh1:
**Chrome**
- `about:blank` is restricted and maybe others starting with `about`
- `data:*` can't be edited, but can be edited into dark mode in `firefox`
- `view-source:*` is restricted in chrome and edge

**Edge**
- `about:blank` is restricted and maybe others starting with `about`
- `data:*` can't be edited, but can be edited into dark mode in `firefox`
- `view-source:*` is restricted in chrome and edge
- `devtools:*` is restricted in chrome and edge

### DIFF
--- a/src/background/utils/extension-api.ts
+++ b/src/background/utils/extension-api.ts
@@ -19,10 +19,14 @@ export function canInjectScript(url: string) {
     }
     if (isEdge) {
         return (url
+            && !url.startsWith('about')
             && !url.startsWith('chrome')
+            && !url.startsWith('data')
+            && !url.startsWith('devtools')
             && !url.startsWith('edge')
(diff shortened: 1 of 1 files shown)


Gusted:
Hey @Yash-Singh1!

Everything looks good to me, however the `data` will cause issues with blank iframes.
It's an certain feature of browsers that blank iframes will have `about:blank` url/src. However with this patch that won't be styled while it actually should. [More info about this](https://stackoverflow.com/questions/41408936/can-anyone-explain-that-what-is-the-use-of-match-about-blank-in-chrome-extensi#41410165)

Regards,
Gusted

PR #4194, opened 2020-11-19
Temporary fix for Github.com

alexanderby:
- Variables' values are taken from a selector not used in page.


Andre601:
I want to point out that this did fix the light-grey for comment boxes and alike.
However, I also want to point out, that the [Notification-tab](https://github.com/notifications) seems to be unaffected by this and still displays the light-grey color.

image

PR #4206, opened 2020-11-20
Fix for psprices.com

Rakleed:
- Invert PS Store logo.
- Fix background.

Example: https://psprices.com/region-us/game/2947059/cyberpunk-2077

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5625,6 +5625,18 @@ CSS
 
 ================================
 
+psprices.com
+
+INVERT
+.d-inline-block
+
+CSS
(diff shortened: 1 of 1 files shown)


Rakleed:
I also thought to invert the Canvas chart with prices. But if I add the `game-price-dynamics` identifier to INVERT, then everything is inverted well, except for one. When I hover over the chart, a pop-up window appears, in which the text color turns black. Which makes it less readable compared to when it was white before inversion.

Rakleed:
@Gusted the Canvas chart with prices cannot be correctly inverted?

Gusted:
image
 Seems read-able to me?

PR #4213, opened 2020-11-20
add fix for google play audiobook player

PhilippIRL:
This just fixes some font color issues

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5468,6 +5468,18 @@ img[src^="data:image/png;"]
 
 ================================
 
+play.google.com/books/listen
+
+INVERT
+.gb_ua
+
+CSS
+.chapter-item:not(.iron-selected) .chapter-title {
+  color: var(--darkreader-neutral-text) !important;
+}
+
+================================
+
 play.google.com/music
 
 INVERT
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -5468,6 +5468,18 @@ img[src^="data:image/png;"]
 
 ================================
 
+play.google.com/books/listen
+
+INVERT
+.gb_ua
```
Gusted, line 5474: Change it to `a[href*="about/products"]` this selector is known to change every x weeks.

src/config/dynamic-theme-fixes.config
```diff
@@ -5468,6 +5468,18 @@ img[src^="data:image/png;"]
...
+
+CSS
+.chapter-item:not(.iron-selected) .chapter-title {
+  color: var(--darkreader-neutral-text) !important;
```
Gusted, line 5478: Use 4 spaces as ident.

src/config/dynamic-theme-fixes.config
PhilippIRL: `a[href*="about/products"]` doesn't seem to work in this case. I used `a[href*="books/audiobooks"]` instead.

src/config/dynamic-theme-fixes.config
PhilippIRL: okay, done

PR #4220, opened 2020-11-21
Modify variables to match theme settings

### LINKED ISSUE #4182: [Broken Website] Github has grey colors
**Website**
https://github.com

**How to Reproduce**

Visit any repository.

**Expected behavior**

Before, GitHub was dark as it should, but I guess they changed some styles, so .. now things don't work as expected anymore.

**Actual behavior**
Everything is grey instead of black.

**Screenshots**

image

image

**System info:**

- OS: MacOS
- Browser: Firefox 82.0.3 (64-bit)
- Darkreader Version: 4.9.23

**Additional context**
N/A

Gusted:
- If the theme settings are different from the default one, modify the variables to match the settings.
- Resolves #4182

### DIFF
--- a/src/inject/dynamic-theme/index.ts
+++ b/src/inject/dynamic-theme/index.ts
@@ -10,13 +10,14 @@ import {logWarn} from '../utils/log';
 import {throttle} from '../utils/throttle';
 import {clamp} from '../../utils/math';
 import {getCSSFilterValue} from '../../generators/css-filter';
-import {modifyColor} from '../../generators/modify-colors';
+import {modifyBackgroundColor, modifyColor, modifyForegroundColor} from '../../generators/modify-colors';
 import {createTextStyle} from '../../generators/text-style';
 import {FilterConfig, DynamicThemeFix} from '../../definitions';
 import {generateUID} from '../../utils/uid';
 import {createAdoptedStyleSheetOverride, AdoptedStyleSheetManager} from './adopted-style-manger';
 import {isFirefox} from '../../utils/platform';
 import {injectProxy} from './stylesheet-proxy';
+import { parse } from 'utils/color';
 
 const variables = new Map<string, string>();
 const INSTANCE_ID = generateUID();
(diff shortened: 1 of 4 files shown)


Gusted:
Compiled code:
[builds.zip](https://github.com/darkreader/darkreader/files/5577176/builds.zip)

src/inject/dynamic-theme/index.ts
```diff
@@ -108,11 +109,22 @@ function createStaticStyleOverrides() {
 
     const variableStyle = createOrUpdateStyle('darkreader--variables');
     const selectionColors = getSelectionColor(filter);
-    const {darkSchemeBackgroundColor, darkSchemeTextColor, lightSchemeBackgroundColor, lightSchemeTextColor} = filter;
+    const {darkSchemeBackgroundColor, darkSchemeTextColor, lightSchemeBackgroundColor, lightSchemeTextColor, mode, brightness, contrast, sepia, grayscale} = filter;
+    let schemeBackgroundColor = mode === 0 ? lightSchemeBackgroundColor : darkSchemeBackgroundColor;
+    let schemeTextColor = mode === 0 ? lightSchemeTextColor : darkSchemeTextColor;
+    if (brightness !== 100 || contrast !== 100 || sepia !== 0 || grayscale !== 0) {
```
alexanderby, line 115: Worth replacing this block with
```javascript
schemeBackgroundColor = modifyBackgroundColor(parse(schemeBackgroundColor), filter);
schemeTextColor = modifyForegroundColor(parse(schemeTextColor), filter);
```

It looks like shorter and equivalent. You can also remove `...brightness !== 100 || ...` because it looks like micro optimization here (called just 1 time).

alexanderby (changes requested):

nulh (commented):
[darkreader/darkreader] End-to-end tests (#4267)

src/inject/dynamic-theme/index.ts
Gusted: Forgot that we don't need to check because modifying the pole color with default settings will return the pole color.

alexanderby (approved):

PR #4228, opened 2020-11-22
Ignore image analysis of stars on book.douban.com

liuhao326:


liuhao326:
Hi👋!

There is an error here, but I don't know how to fix it.

I would be very grateful if you could tell me what should I do.

Thanks!

PR #4237, opened 2020-11-22
Improved build method

Gusted:
- Add an extra `build` script for just building dark reader.
- Make sure the release doesn't contain any Linting errors.
- A unified function inside build.js


package.json
```diff
@@ -5,11 +5,12 @@
   "scripts": {
     "api": "node tasks/build.js --api",
     "benchmark-server": "node tests/benchmark-server/index.js",
+    "build": "node tasks/build.js --build",
```
alexanderby, line 8: `build.js --build`? It's quite common to have Release and Debug building schemes, so let's replace with `--release`.

alexanderby (changes requested):

tasks/build.js
```diff
@@ -21,21 +21,16 @@ const standardTask = [
...
-}
+const buildTask = [
+    ...standardTask,
+    codeStyle,
+    zip
+];
 
-async function debug({watch}) {
-    log.ok('DEBUG');
+async function release({debug, watch}) {
```
alexanderby, line 30: Let's rename this function into `build()`, because `release(debug)` is strange

tasks/build.js
```diff
@@ -64,10 +59,7 @@ async function run() {
     const args = process.argv.slice(2);
 
     if (args.includes('--release')) {
-        await release();
-    }
-    if (args.includes('--debug')) {
-        await debug({watch: args.includes('--watch')});
+        await release({watch: args.includes('--watch'), debug: args.includes('--debug')});
```
alexanderby, line 62: Now there's a dilemma. Let's use `--release` (`build({debug: false, watch: false})`) and `--debug` (`build({debug: true, watch: args.includes('--watch')})`) separately.

alexanderby (changes requested):

alexanderby (approved):

PR #4238, opened 2020-11-22
Fix a few icons not displaying correctly

kikiwako:
Inverts a couple of icons in the character sheets that were black-on-black.

For some reason, when activating dark mode, a specific icon's source URL gets scrambled. (.i-type-bludgeoning)
So I forced the correct URL to be loaded.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1871,6 +1871,18 @@ INVERT
 
 dndbeyond.com
 
+INVERT
+.i-ritual
+.i-concentration
+.ddbc-damage-type-icon
+.ddbc-combat-attack__icon-img--spell-school-conjuration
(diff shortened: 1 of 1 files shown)


Gusted:
Hi @kikiwako!

Do you have an example URL?

Regards,
Gusted

kikiwako:
Yep.

The icon's source gets changed like this.
```
.i-type-bludgeoning {
    background-image: url(blob:https://www.dndbeyond.com/299eb5dc-2098-4a73-8147-2c8a4749fc8a);
}
```

If you go to the url (including the "blob:") it displays the icon correctly. But in dark mode it is display as a white or black square.

kikiwako:
It looks like this by default with the dark mode
image

And like this after setting the url in the syles
image

Gusted:
I rather mean the URL where I can observe the changes.

nulh (approved):

PR #4247, opened 2020-11-23
CodeMirror in DevTools and stylesheet-editor

### LINKED ISSUE #4108: [Feature Request] Consider switching to CodeMirror for DevTools

Currently the stylesheet editor and the DarkReader DevTools use `textarea` to render themselves. It does not have any syntax highlighting, current line highlighting, line numbers, etc. I think that CodeMirror could provide more features for the editors.

Yash-Singh1:
- Resolves #4108
- Implemented CodeMirror into `DevTools` and the `stylesheet-editor` giving a richer user interface including: `active line`, `syntax highlighting`, `line numbers`, `line wrapping`, and more


Gusted:
Hey @Yash-Singh1!

Good PR, however a few things.

- 'Hard coded' codemirror inside the source, this means whenever a update of codemirror is we won't get it as it's `hardcoded`, that's a problem as we like to have the latest updates.
- I'm seeing some 'insecure' `setTimeout` and `setInterval` functions, where I don't see the purpose for it.

For the rest it's good, however the codemirror should possible get somehow as `devDependency`(as the devtools is not inside the Dark Reader API) and therefore get the latest updates. Also it's increases the git repo with 10mB, luckily compressed build files aren't affected by this.

Regards,
Gusted

Yash-Singh1:
> 'Hardcoded' `codemirror` inside the source, this means whenever an update of `codemirror` is we won't get it as it's hardcoded, that's a problem as we like to have the latest updates.

Yes, I added in the `dependencies/` folder temporarily, I will create a commit that will use `npm`.

> I'm seeing some 'insecure' `setTimeout` and `setInterval` functions, where I don't see the purpose for it.

The `setTimeout` function with the second argument of `0` is required for some reason when loading `codemirror`. I tried removing it, but it just creates an empty `div` inside the `DOM`. Also, I was getting some errors for the `syntax highlighting` so I did the following:
```
 
*

INVERT
.jfk-bubble.gtx-bubble
.captcheck_answer_label > input + img

CSS
.vimvixen-hint {
    background-color: ${#ffd76e} !important;
    border-color: ${#c59d00} !important;
    color: ${#302505} !important;
}
::placeholder {
    opacity: 0.5 !important;
}

```
First, it will consider URL patterns as a string. Then, the CSS highlighting will start at the first keyword: `INVERT` in this case and end at the `===..` seperator. In the middle, all of the keywords will be consistently changed into their own token: `atom` in a `setInterval`. So, the `setInterval`'s and the `setTimeout`'s are required to make everything run.

Gusted:
> Yes, I added in the `dependencies/` folder temporarily, I will create a commit that will use `npm`.

Aha, okay!



> The `setTimeout` function with the second argument of `0` is required for some reason when loading `codemirror`. I tried removing it, but it just creates an empty `div` inside the `DOM`. Also, I was getting some errors for the `syntax highlighting` so I did the following:
> 
> ```
>  
> *
> 
> INVERT
> .jfk-bubble.gtx-bubble
> .captcheck_answer_label > input + img
> 
> CSS
> .vimvixen-hint {
>     background-color: ${#ffd76e} !important;
>     border-color: ${#c59d00} !important;
>     color: ${#302505} !important;
> }
> ::placeholder {
>     opacity: 0.5 !important;
> }
> 
> ================================
> ```
> 
> First, it will consider URL patterns as a string. Then, the CSS highlighting will start at the first keyword: `INVERT` in this case and end at the `===..` seperator. In the middle, all of the keywords will be consistently changed into their own token: `atom` in a `setInterval`. So, the `setInterval`'s and the `setTimeout`'s are required to make everything run.

Ah, I see... I will tinker with this a bit around, as I try to not have those functions in code as it can be unreliable as well e.g. the current `setTimeout` will run when the callstack is empty and that can be even after codemirror already has been loaded or before it was loaded due to some `performance rules` made by Javascript engine what's noticeable on low-end machines.
As for the `setInterval` every 100th of an second it will collect all codemirror lines and wrap the theme around it. Which will cause CPU usage to go brrrrrrrr. 

Regards,
Gusted

Yash-Singh1:
> As for the setInterval every 100th of an second it will collect all codemirror lines and wrap the theme around it. Which will cause CPU usage to go brrrrrrrr.

Since this just updates the keywords, I think 0.5 seconds could work. What do you suggest?

Yash-Singh1:
Currently, the configuration uses the 4 spaces rule. Now that `codemirror` has been implemented, we can switch to tabs. Would you like to disable tabs and use 4 spaces instead or go through the config (maybe with a script) and replace all instances of 4 spaces with a tab?

Yash-Singh1:
@Gusted 6dee80d will remove the `setInterval` and add in the `update` event instead.

Gusted:
> @Gusted [6dee80d](https://github.com/darkreader/darkreader/commit/6dee80d31910c4a6532b357a2d0c0f0e555fe5e7) will remove the `setInterval` and add in the `update` event instead.

Seems good.

Please merge the latest `master` branch and run `npm run code-style`.

Yash-Singh1:
@Gusted Done: 5d779bc

Gusted:
Had a quick view over the weird.
LGTM, not sure if we should ignore an ESLint warning tho, but that isn't a big issue.
Guess it up to @alexanderby to review this monster 😛

PR #4255, opened 2020-11-23
Fetch error handling

### LINKED ISSUE #4246: [Broken Website] Google Search
**Website**
https://www.google.com/

**How to Reproduce**
The whole page isn't changed, nothing to reproduce.

**Expected behavior**
I had Dark Reader like a year ago where Google worked just fine.
Now I am using it again and it works everywhere except for this site.

**Actual behavior**
It's unchanged.

**Screenshots**
https://i.imgur.com/ZLZdtDw.png

**System info:**
- OS: Windows 10 20H2
- Browser: Opera GX (Chromium) LVL2 (core: 71.0.3770.323)
- Darkreader: 4.9.24

**Additional context**
I do not have any custom settings, I reinstalled the extension, I don't have any websites blacklisted, I activated the new design of the extension and reset everything in the dev tools.
It literally doesn't work.

Gusted:
- Handle errors when fetch fails.
- Resolves #4246

### DIFF
--- a/src/utils/network.ts
+++ b/src/utils/network.ts
@@ -1,13 +1,20 @@
+import {logWarn} from '../inject/utils/log';
 import {isFirefox} from './platform';
 
 async function getOKResponse(url: string, mimeType?: string) {
-    const response = await fetch(
-        url,
-        {
-            cache: 'force-cache',
-            credentials: 'omit',
-        },
-    );
+    let response: Response;
+    try {
+        response = await fetch(
+            url,
+            {
+                cache: 'force-cache',
+                credentials: 'omit',
+            },
+        );
+    } catch (err) {
+        logWarn(err);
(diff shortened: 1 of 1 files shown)


alexanderby:
I don't see how catching an error and throwing it can fix anything. Background fetch is located in [background/tab-manager.ts](https://github.com/darkreader/darkreader/blob/master/src/background/tab-manager.ts#L88) and is wrapped in try-catch. The error is sent to content script and processed there.

The described bug could be fixed by CSP fallback (using `requestAnimationFrame` when inline script is blocked).

Gusted:
> The described bug could be fixed by CSP fallback (using `requestAnimationFrame` when inline script is blocked).

https://github.com/darkreader/darkreader/issues/4246#issuecomment-733753348
It didn't, special mention he use Opera GX some special brand consuming less memory and made for alongside gaming. That seems to be based on Chromium v71

PR #4262, opened 2020-11-24
Fix for systemuicons.com

mikerouxfr:
Make icons visible


Gusted:
Hi @mikerouxfr!

The fix should be made to https://github.com/darkreader/darkreader/blob/master/src/config/inversion-fixes.config

Regards,
Gusted

mikerouxfr:
done on #4326 😄

Issue #4236, opened 2020-11-22
[Broken Website] Twitter.com


privacyguy123:
**Website**
https://www.twitter.com

**How to Reproduce**
1. Go to Twitter.
2. Everything is fine
3. Refresh page/click through a link 
4. Bugged colors

**Screenshots**
![image](https://user-images.githubusercontent.com/50337995/99906038-86b6a780-2ccc-11eb-8a88-0d276b63d186.png)

Chat bubbles in PMs also have a bright white background for the person I'm talking to like the search bar in the screenshot.

**System info:**

- OS: Windows 10
- Browser: FireFox 83
- Darkreader Version: 4.9.24

Gusted:
Hi @privacyguy123!

Does this bug occur when you try on a fresh profile with Dark Reader, it might be some `about:config` or strict tracking protection of Firefox triggering this.

Regards,
Gusted

privacyguy123:
Hi @Gusted!

Yes, DarkReader reset and none of the FireFox tracking protection on - only AFTER a page refresh on Twitter however.

Gusted:
Hmm weird behaviour.

I only get bugged colors for the `main thread` as it's not getting styled right.
![image](https://user-images.githubusercontent.com/25481501/99906706-d8fdc580-2cd8-11eb-8002-2e413f842108.png)

privacyguy123:
Yes that's what I see - also in chat bubble DMs too.

Gusted:
Might found the error, in order to reduce idle CPU we made a script to be injected inside the site however it seems like twitter is throwing errors for it #4005 
![image](https://user-images.githubusercontent.com/25481501/99906887-e36c8f00-2cd9-11eb-9f70-ca1b01b1fbee.png) and therefore might not detect changes into the site's style.

privacyguy123:
Is there a quick fix for this?

Gusted:
> Is there a quick fix for this?

Revert to 4.9.23 🤷🏽‍♀️ I will take a look tomorrow to see if we can `bypass` the error.

privacyguy123:
Oh right yeah, reverting back all these changes would fix everything - derp!

privacyguy123:
Fixed!

PR #4270, opened 2020-11-24
Use RAF when proxy script blocked by CSP

alexanderby:


Gusted:
This Resolves #4236

alexanderby:
Oh, yes, I don't know why there's no CSP error in Chrome.

Gusted:
I have a such feeling it might be disabled by default to show. Also uBlock origin has an option to hide CSP messages, you might have this option enabled?

alexanderby:
I don't know how it works.

PR #4273, opened 2020-11-24
Don't invert company logos on LinkedIn

sd2048:


src/config/dynamic-theme-fixes.config
```diff
@@ -4123,7 +4123,8 @@ CSS
 linkedin.com
 
 INVERT
-img[src*="company-logo"]
+img[src*="company-logo"],
```
Yash-Singh1, line 4126: Don't add a comma over here, this will fail the following test: https://github.com/darkreader/darkreader/blob/master/tests/config/config.tests.ts#L64
It expects that all selectors don't have a comma

sd2048:
Looks like the official file doesn't have the problem anymore.

PR #4290, opened 2020-11-25
Correct dynamic inversion for roblox.com

adwalters:
The checked and unchecked inbox checkboxes on the Messages page were both dark (couldn't tell if it was checked or not).

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -6058,6 +6058,15 @@ CSS
 
 ================================
 
+roblox.com
+
+CSS
+.checkbox input[type=checkbox]:checked + label::before {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -6058,6 +6058,15 @@ CSS
 
 ================================
 
+roblox.com
+
+CSS
+.checkbox input[type=checkbox]:checked + label::before {
+    background-color: #aaa !important;
```
Gusted, line 6065: Please use `var(--darkreader-neutral-text)` instead of `#aaa`

Gusted (changes requested):

Gusted (approved):

PR #4291, opened 2020-11-25
Improve variable support

### LINKED ISSUE #3241: Collect variables
Hmm a intresting variable bug with the collection of it.

[Test-Case.zip](https://github.com/darkreader/darkreader/files/4977888/Test-Case.zip)

It replace the variable with the wrong variable as it's into a css selector it's never match against the element due to the codebase as it's expects it's just somewhere in :root However this variable should change on dark/light mode but dark reader just grabs the variable without matching the selector and putting the selector into it, it never changes.

### LINKED ISSUE #3267: [Broken Website] Quip
**Website**
https://www.quip.com

**How to Reproduce**

1. Go to the site and open / add a doc with any default text
2. Text is purple instead of white

Text is inverted as purple instead of white (old behavior). This seems to have just started very recently for me (within the last day).

**Expected behavior**
Text is white

**Actual behavior**
Text is purple

**Screenshots**
image

image

**System info:**
 - MacOS
 - Chrome
 - Darkreader Version: 4.9.16
 
**Additional context**

### LINKED ISSUE #3618: [Broken Website] docs.microsoft.com switches to "High Contrast" theme even if excluded
**Website**
https://docs.microsoft.com/en-us/

**How to Reproduce**
1. Go to the site above
2. Refresh once or twice
3. See that it switches to the "High contrast" scheme (which is also available at the bottom "Theme" menu) instead of keeping the "Dark" or "Light" theme.
4. Even if this domain is excluded, the High Contrast scheme is always picked after a refresh or immediately.

**Expected behavior**
The website should stick to the actually selected theme and not magically switch to "High contrast"

**Actual behavior**
The website always switches to "High contrast".

**Screenshots**
Note how the
- website is excluded
- website thinks "Dark" is selected
- obvious "High contrast" theme is active, however
image

**System info:**
 - OS: Windows 10 2004
 - Browser: edge-chromium 85.0.564.51
 - Darkreader Version: 4.9.16 - this started happening not more than a month ago.
 
**Additional context**
Of course, high contrast mode is _not_ enabled in my operating system, or this issue would be redundant.
I reset and reinstalled my browser, and tried in Incognito: But as soon as I install / enable Darkreader, docs.microsoft.com forces the High contrast scheme on me.

### LINKED ISSUE #3972: [BUG] DarkReader sometimes ignores Stylus CSS variable overrides
### **Describe the bug**

DarkReader seems to ignore CSS variable overrides set with Stylus. My overrides work when DarkReader is off, but not when it is on. For example, if the site has the following CSS:

```css
:root {
    --code-color: red;
}
code {
    color: var(--code-color);
}
```
and I use this CSS in Stylus:
```css
:root {
    --code-color: green !important;
}
```
The `code` text will appear green when DarkReader is off, but when DarkReader is on, it will appear as the inversion of the red color. 

### **Steps to reproduce**

I tried to reproduce this on GitHub by overriding the `--color-border-secondary` color, but it worked. I don't know what's happening with this other site. It is our local-intranet Confluence page.

### **System Information:**
 - OS: Windows 10 1809
 - Browser: Chrome 85
 - Dark Reader version: 4.9.22

### **Additional context**

Even if I "uncheck" all of the relevant CSS variables in the developer tools, the "constructed stylesheet" from DarkReader does not change. You would expect it to change when the underlying variables change, no?

### LINKED ISSUE #4182: [Broken Website] Github has grey colors

### LINKED ISSUE #4225: [Broken Website] GitHub has black colors
**Website**
Issue Link just like: <https://github.com/darkreader/darkreader/issues/4182>
And repository Link just like: [darkreader](https://github.com/darkreader/darkreader)

**How to Reproduce**
An example of this is
>
> 1. Go to <https://github.com/darkreader/darkreader/issues/4182> or [darkreader](https://github.com/darkreader/darkreader)
> 2. Choose the Light theme

**Actual behavior**
It works well before, but there are some black colors on the website now.

<details>
			<summary><strong>Screenshots</strong></summary>

image

image

</details>

**System info:**

- OS: Windows
- Browser: chrome
- Darkreader Version: 4.9.24

**Additional information:**

I reinstalled Darkreader and closed other extensions, but it can't solve the problem.

### LINKED ISSUE #4363: [Broken Website] github.com
**Website**
The link to the broken website
https://github.com

**How to Reproduce**
An example of this is
>
> 1. Go to x site
> 2. Hover over x button
> 3. See that when hovering it isn't changing colors

1. Go  to issue with thumbs up
2. Put the thumb up

**Expected behavior**
Background should differ significantly, to show that you have reacted.

**Actual behavior**
A clear and concise description of what happened.
Background differs so slightly, so it's impossible to say if you have reacted.

**Screenshots**
image

**System info:**

- OS: [e.g. Windows, MacOS, Linux] linux
- Browser: [e.g. chrome, safari] firefox
- Darkreader Version: [e.g. 4.9.9] 4.9.26

**Additional context**

### LINKED ISSUE #4534: github and stackoverflow light mode problem
**Website**
https://github.com
https://stackoverflow.com

**How to Reproduce**
1. Go to site
2. Set Light mode
3. the site background color is black, not light color.

图片

**Expected behavior**
I want using light color for the site background, like this:
图片

**Actual behavior**
the site background color is black in light mode:
图片
图片

**Screenshots**

- OS: Windows 10
- Browser: Firefox 84
- Darkreader Version: 4.9.26

### LINKED ISSUE #4798: [Broken Website] chromestatus.com
**Website**
https://www.chromestatus.com/features/5678152091172864

**Expected behavior**
Text inverted to be readable

**Actual behavior**
Text not inverted

**Screenshots**
image

**System info:**

- OS: Windows
- Browser: Firefox 84.0.2
- Darkreader Version: 4.9.26

### LINKED ISSUE #4800: [Broken Website] wikiteq.com
The site of the [WikiTeq](https://www.wikiteq.com) got redesign too, and this extension makes the site look very bad.

In dark mode by browser (with disabled this extension), the page looks much better. See screenshots below:

| Look of the site when this extension is enabled | Look of the site when this extension is disabled |
| ------------- | ------------- |
| scrnli_18-1-2021- 19-31-44 | scrnli_18-1-2021- 19-30-26 |

**System info:**

- OS: Windows 10
- Browser: Opera GX - LVL2 (core: 72.0.3815.465)
- Darkreader Version: 4.9.26

### LINKED ISSUE #4801: [Broken Website] redbubble.com
**Website**
https://www.redbubble.com

**How to Reproduce**

1. Go to site
2. Search for anything
3. See that the link bar at the very top and the one below the search box still have a light background

**Expected behavior**
All parts of the page have a dark background

**Actual behavior**
The link bar at the very top and the one below the search box still have a light background

**Screenshots**
image

**System info:**

- OS: Windows 10 Pro
- Browser: Firefox version 84.0.2 (64-bit)
- Darkreader Version: 4.9.26

Gusted:
- Change fundamentally how Dark Reader handles variables.
- Create for background-color/color/border an individual variable and modify variables to Dark Reader's variables.
- Makes sure that the variable is only active when the originals variable selectors are met.
- Resolves #3241
- Resolves #3267 
- Resolves #3618
- Resolves #3972
- Resolves #4182 
- Resolves #4225
- Resolves #4363
- Resolves #4534
- Resolves #4798
- Resolves #4800
- Resolves #4801


src/inject/dynamic-theme/index.ts
```diff
@@ -252,16 +258,54 @@ function createManager(element: StyleElement) {
...
+                `   --darkreader-bg${value.property}: ${modifiedBackground};`,
+                `   --darkreader-text${value.property}: ${modifiedText};`,
+                `}`
+            ].join('\n'));
+        } else {
+            let parsedValue: RGBA;
+            try {
+                parsedValue = parse(value.value);
+            } catch (err) {
+                logWarn(err);
```
Gusted, line 290: Only fails for nested variables.
e.g.

```CSS
    --color-red: #ccc;
    --proxy-color-red: var(--color-red);
```
Will throw an error as `var(--color-red)` should be replaced with the value of Dark Reader's modified one however. It doesn't have to be in order so.

```CSS
    --proxy-color-red: var(--color-red);
    --color-red: #ccc;
```
And Dark Reader won't have a single knowledge about --color-red *yet* and should be placed on some list.

alexanderby:
Whatever you are doing start with tests for the old implementation in a separate PR. This way we will ensure that the new implementation can do everything the old one could.

Then use the best practices of Test-Driven Development. Add some test cases for the new implementation. Use `npm run test:inject:debug` to speed up the process. You can debug the code in a browser. You can filter out particular test cases in `karma.conf.js` to speed up even more.

Gusted:
@alexanderby Thanks for the suggestions I've fixed some bugs and ported all old test cases to dynamic one and will shortly have the old one's back to ensure that their is an expected output.

Gusted:
[binaries.zip](https://github.com/darkreader/darkreader/files/5619047/binaries.zip)
Working variable support.

alexanderby:
@Gusted Let's start with putting tests for old variables implementation into another PR. This way we will ensure that tests for the old implementation are valid.

Gusted:
> @Gusted Let's start with putting tests for old variables implementation into another PR. This way we will ensure that tests for the old implementation are valid.

What you mean by putting the test into another PR? The current implementation has already every edge case covered and this PR has new tests that works with the new implementation. Do you mean to have the tests for the new implementation having a separate PR?

Gusted:
So I've found the bug that causes to remove the variables on settings changes I've fixed this in https://github.com/darkreader/darkreader/pull/4291/commits/2561640aee192063cdb7374da69ba1110e6e44c0
As this was the only bug reported so far It should now not contain any know bugs.

Binaries to be side-loaded: [Binaries.zip](https://github.com/darkreader/darkreader/files/5625688/Binaries.zip)

CC @liuhao326 @kkoomen Could you both confirm this has indeed fixed the issue with changing settings.

Regards,
Gusted

liuhao326:
Hi @Gusted!

The bug mentioned in [#4225（comment）](https://github.com/darkreader/darkreader/issues/4225#issuecomment-736639919) has been fixed successfully. And there are no other issues found.

kkoomen:
@Gusted I also confirm that the problem has been fixed.

alexanderby:
So, here's how the process normally looks like:
1. The feature is added (the old variables implementation).
2. Tests for the feature are added (it was untested in our case, only narrowly scoped unit tests).
3. Another feature is added or improved (what you've done in this PR).
4. Old tests are run to ensure nothing was broken.
5. Tests for new capabilities are added.

So in our case paragraph 2 was skipped and we don't know if the new tests are valid for the old implementation. Please put tests for the old implementation into a separate PR, we will ensure they run, then we will merge master into this PR an I will add some more tests to check how the new implementation works.

Gusted:
Done #4369

Yash-Singh1:
@Gusted I tested this and it seems to be working in `microsoft-edge-dev`.

src/inject/dynamic-theme/index.ts
```diff
@@ -182,7 +193,7 @@ function createDynamicStyleOverrides() {
         .filter((details) => details && details.variables.size > 0)
         .map(({variables}) => variables);
     if (newVariables.length === 0) {
-        styleManagers.forEach((manager) => manager.render(filter, variables, getIgnoreImageAnalysisSelectors()));
+        styleManagers.forEach((manager) => manager.render(filter, legacyVaribales, getIgnoreImageAnalysisSelectors()));
```
kkoomen, line 196: `legacyVaribales` must be `legacyVariables`

kkoomen (commented):
Just found a typo.

src/inject/dynamic-theme/index.ts
Gusted: 👀 Does vscode have some spellchecking?

alexanderby:
I've merged tests for old implementation, 1 test case and 2 tests are failing:
1. Tests for `replaceCSSVariables` function, did the format change? Is this function still used?
2. Reacting on variables change.
3. `<html>` element variables. I understand that handling variables in inline style attribute can be tricky, we can skip it for this PR. But I remember many inline variables in root element were used in Outlook mail, maybe somewhere else. Would be good to make it supported if it is easy, but if not, let's not make this PR too difficult.

Gusted:
>     1. Tests for `replaceCSSVariables` function, did the format change? Is this function still used?

Yes, it's will now return an array.
First item is the replacedCSSText and the second item is a object with all `variables values` to `orginial variable`.

> Reacting on variables change.

This should work right?



>     3\. `<html>` element variables. I understand that handling variables in inline style attribute can be tricky, we can skip it for this PR. But I remember many inline variables in root element were used in Outlook mail, maybe somewhere else. Would be good to make it supported if it is easy, but if not, let's not make this PR too difficult.

I've added support this one as well, not tested.

Gusted:
@alexanderby I've found the bug that causes to be failing not reacting to changes, I made a little mistake in my caching logic in particular deleting a cached variable...
I've also updated the CSSRules test cases to pass.

I just want to mention that https://github.com/darkreader/darkreader/pull/4291/commits/755ea517410a6da44d4223ab1c94d68656e5fef3 will break test cases as the `body` in the test cases will expect the default-colour however now it will actually have that color and will break the test cases.

alexanderby:
@Gusted Glad to see it works. Could you please explain how it works?

Don't worry about that body color. Karma runs tests in iframe with the same origin, so it has access to parent Window and set's the background there without touching the iframe itself.

Gusted:
> @Gusted Glad to see it works. Could you please explain how it works?

Bare with me, I'm not the best explainer. I will try step for step explain the thinking process to have a good understanding what's going on.

### Why was this PR originally created?

Because, the current implementation of the variable is lazy.

### And why is it lazy?

It doesn't take selectors in account and that is causing problems with sites that uses variables to switch between light theme <-> dark theme. It will always use the value of the last `collected` variable this can be either the light theme or the dark theme one. A good and simple example is:
```html
<style>
.red {
    --text: red;
}
.green {
    --text: green;
}
h1 {
    color: var(--text); 
}
</style>
<h1 class="red">Red CSS variable</h1>
<h1 class="green">Green CSS variable</h1>
```

So, dark reader(`updateVariables` function) will see the variables
```json
{
"--text": "red",
"--text": "green"
}
```
You see the problem? The selector's is not being taken into account.
And because `--text: green;` overrides the `--text: red`;
Both h1.red and h1.green will end up being styled with a modified green.

### So how does this PR fix this?

Well, first we make sure we collect the according selector-text of the rule whereby the variable is created.
And when we have the variable + selectorText we simply modify the possible bg/text/border of the variable's value.
Then having the modifiedColors we place it into this new stylesheet `darkreader--dynamicVariables`. Now it actually takes into account the selector.

### How does the PR work on the example?

Assume the last example scenario.

```html
<style>
.red {
    --text: red;
}
.green {
    --text: green;
}
h1 {
    color: var(--text); 
}
</style>
<h1 class="red">Red CSS variable</h1>
<h1 class="green">Green CSS variable</h1>
```

Dark Reader(`updateVariables` function) will now see the following:
```json
{
".red;--text": "red",
".green;--text": "green"
}
```
And having some logic in place it will correct remove the selector and make sure this is saved for later.
And now the inserted logic will seems as follows into the `darkreader--dynamicVariables` element.
```CSS
.red {
    --darkreader-bg--text: #cc0000;
    --darkreader-text--text: #ff1a1a;
    --darkreader-border--text: #b30000;
}
.green {
    --darkreader-bg--text: #006600;
    --darkreader-text--text: #72ff72;
    --darkreader-border--text: #00d900;
}
```
And the modified CSS for:
```CSS
h1 {
    color: var(--text); 
}
```
Looks like
```CSS
h1 {
    color: var(--darkreader-text--text);
}
```
And poof that's expected behaviour.

### Side-notes
- It doesn't take in parent rules with their selector. e.g. media rules. *It's fix-able*
- Possible performance problem to create for every variable an modified bg/text/border, however can be optmized at a later moment and I don't know a performance problem with the modify function and should be considered efficient enough.
- The only caveat of when this break, is when an variable name is not being used in the same context.
```CSS
.red {
    --text: red; */ As <color> value */
}
.green {
    --text: 0.1; */ As <opacity> value */
}
```

Because behind the scenes we still use the old method to make sure the variable is replaced with the actual value so we can recognize if it's an color or not. And in the scenario here above it will be replaced to 0.1 instead of the color `red` and therefore won't be detected right. 
I might want to notice that such case is very edge and not logical at all and not be considered to be worth fixing.

@alexanderby I hope you understand it else please ask.

Regards,
Gusted

Gusted:
> Don't worry about that body color. Karma runs tests in iframe with the same origin, so it has access to parent Window and set's the background there without touching the iframe itself.

Debug version of it doesn't run tests in the iframe 
image

Gusted:
> - It doesn't take in parent rules with their selector. e.g. media rules. It's *fix-able*

Added support https://github.com/darkreader/darkreader/pull/4291/commits/f67bd0c1a82c214049fdbc9bdb2202cd9af58f43 and test-cases https://github.com/darkreader/darkreader/pull/4369/commits/75278cd827ae68ce948f6383c4e4ab5e74a9cd8f

src/inject/dynamic-theme/css-rules.ts
```diff
@@ -41,23 +42,33 @@ function isCSSVariable(property: string) {
...
+    }
+    stack.push(parentRule.conditionText);
+    return getParentGroups(parentRule, stack);
+}
+
 export function getCSSVariables(rules: CSSRuleList) {
-    const variables = new Map<string, string>();
+    const variables = new Map<string, Variable>();
     rules && iterateCSSRules(rules, (rule) => {
+        const parentGroups = getParentGroups(rule);
```
alexanderby, line 57: The result should be the same for each sibling. Try optimizing it. E.g. use WeakMap to cache results for each `parentRule`, but only in `getCSSVariables` function scope to prevent unpredictable bugs.

Maybe `iterateCSSRules` is not perfect for iterating trees and should be refactored in future.

src/inject/dynamic-theme/index.ts
```diff
@@ -130,6 +130,14 @@ function createStaticStyleOverrides() {
     const proxyScript = createOrUpdateScript('darkreader--proxy');
     proxyScript.textContent = `(${injectProxy})()`;
     document.head.insertBefore(proxyScript, variableStyle.nextSibling);
+
+    const dynamicVariableStyle = createOrUpdateStyle('darkreader--dynamicVariable');
```
alexanderby, line 134: Let's follow our CSS selectors convention and use spinal case `darkreader--dynamic-variables`

src/inject/dynamic-theme/css-rules.ts
```diff
@@ -41,23 +42,33 @@ function isCSSVariable(property: string) {
...
+    if (!parentRule || !(parentRule instanceof CSSMediaRule)) {
+        return stack;
+    }
+    stack.push(parentRule.conditionText);
+    return getParentGroups(parentRule, stack);
+}
+
 export function getCSSVariables(rules: CSSRuleList) {
-    const variables = new Map<string, string>();
+    const variables = new Map<string, Variable>();
```
alexanderby, line 55: I suggest you to use `Map<string, Map<string, Variable>>`. The first string is a selector, and the nested key string is a custom property (CSS variable name). This way you will not need to split by ";" that can be unpredictable (e.g. some attribute value selector), and will not produce a single rule for each variable, but instead many variables will be packed inside a single selector.

src/inject/dynamic-theme/css-rules.ts
```diff
@@ -41,23 +42,33 @@ function isCSSVariable(property: string) {
...
     rules && iterateCSSRules(rules, (rule) => {
+        const parentGroups = getParentGroups(rule);
         rule.style && iterateCSSDeclarations(rule.style, (property, value) => {
             if (isCSSVariable(property)) {
-                variables.set(property, value);
+                variables.set(`${rule.selectorText};${property}`, {property, value, parentGroups});
```
alexanderby, line 60: So, as described above, this should be replaced with
```javascript
if (!variables.has(rule.selectorText)) {
    variables.set(rule.selectorText, new Map());
}
variables.get(rule.selectorText).set(property, {value, parentGroups: []});
```

I'm doubtful about `parentGroups`, but let it be with the first iteration.

src/inject/dynamic-theme/css-rules.ts
```diff
@@ -41,23 +42,33 @@ function isCSSVariable(property: string) {
...
     return variables;
 }
 
 export function getElementCSSVariables(element: HTMLElement) {
-    const variables = new Map<string, string>();
+    const variables = new Map<string, Variable>();
     iterateCSSDeclarations(element.style, (property, value) => {
         if (isCSSVariable(property)) {
-            variables.set(property, value);
+            variables.set(`:root;${property}`, {property, value, parentGroups: []});
```
alexanderby, line 71: Hmmm, since it will be used for `<html>` element only at the moment, let's use `:root`, but later something other should be invented (e.g. assigning some `data-` attribute to make it a selector, like it is currently made for overriding inline colors).

src/inject/dynamic-theme/index.ts
```diff
@@ -460,6 +457,7 @@ export function removeDynamicTheme() {
         removeNode(document.head.querySelector('.darkreader--invert'));
         removeNode(document.head.querySelector('.darkreader--inline'));
         removeNode(document.head.querySelector('.darkreader--override'));
+        removeNode(document.head.querySelector('.darkreader--dynamicVariable'));
```
alexanderby, line 460: Don't forget about the selector name convention here too.

src/inject/dynamic-theme/modify-css.ts
```diff
@@ -36,6 +36,17 @@ export function getModifiableCSSDeclaration(property: string, value: string, rul
         property === 'stroke' ||
         property === 'stop-color'
     ) {
+        if (value.startsWith('var(')) {
```
alexanderby, line 39: Please explain what happens here.

tests/inject/dynamic/variables.tests.ts
```diff
@@ -207,4 +208,256 @@ describe('CSS Variables Override', () => {
...
+        );
+        createOrUpdateDynamicTheme(theme, null, false);
+        expect(getComputedStyle(container).backgroundColor).toBe('rgb(0, 0, 0)');
+        expect(getComputedStyle(container.querySelector('.red')).color).toBe('rgb(255, 26, 26)');
+        expect(getComputedStyle(container.querySelector('.green')).color).toBe('rgb(255, 255, 255)');
+        expect(getComputedStyle(container.querySelector('.orange')).color).toBe('rgb(255, 255, 255)');
+    });
+});
+
+describe('Managing CSS rules', () => {
```
alexanderby, line 320: These tests are already in `tests/inject/dynamic/css-rules.tests.ts`, so you can remove them here.

src/inject/dynamic-theme/variables.ts
```diff
@@ -0,0 +1,107 @@
...
+    value: string;
+    parentGroups: string[];
+}
+
+export function updateVariables(newVars: Map<string, Variable>, theme: Theme) {
+    if (newVars.size === 0) {
+        return;
+    }
+    newVars.forEach((value, key) => {
+        legacyVariables.set(value.property, value.value);
```
alexanderby, line 36: Replace `value.value` with `variable.value`

src/inject/dynamic-theme/variables.ts
```diff
@@ -0,0 +1,107 @@
...
+        cachedVariables.delete(key);
+        variables.set(key, value);
+    });
+
+    legacyVariables.forEach((value, key) => {
```
alexanderby, line 41: Why there are these `legacyVariables` and why they are passed into Style Managers?

src/inject/dynamic-theme/variables.ts
```diff
@@ -0,0 +1,107 @@
...
+    const declarations: VariableDeclaration[] = [];
+    variables.forEach((variable, key) => {
+        const selector = key.split(';')[0];
+        const {property, value, parentGroups} = variable;
+        const standardDeclaration = {selector, key, property, parentGroups};
+        if (cachedVariables.has(key)) {
+            const {modifiedBackground, modifiedText, modifiedBorder} = cachedVariables.get(key);
+            declarations.push({...standardDeclaration, modifiedBackground, modifiedText, modifiedBorder});
+        } else if (value.includes('var(')) {
+            const [modifiedBackground, modifiedText, modifiedBorder] = ['bg', 'text', 'border'].map((value) => `--darkreader-${value}${property}`);
```
alexanderby, line 74: Isn't there a mistake? I don't fully understand yet, did you suppose to get `--darkreader-bg--some-variable` value? Maybe it should be wrapped into `var(--darkreader-bg...)`? Also, the `variable` value has different meaning in the same scope, replace it with `v` or `prefix` in this loop to make it easier to understand.

alexanderby (changes requested):
Nice approach, it will cover 90% cases I think. I will add some more comments after you clarify some moments.

src/inject/dynamic-theme/variables.ts
Gusted: So, because in plain CSS any variable e.g. `var(--text)` is unknown what the value would be, that's why I preserve the old implementation of replacing the variable which is value. Because we switches it later back to the original value it's pure for dark reader to know what the `value type` is e.g. an <color> or an <opacity>. An based on that won't style an opacity, but only an color variable. It still has the flaw of overwritten by a new selector, but the value type won't change most likely and shouldn't be a problem.

src/inject/dynamic-theme/modify-css.ts
Gusted: So if a property is detected that uses an variable as value. The appropriate variable name is being assigned based on it's property. e.g background-color will become darkreader-bg--some-variable. 

Because we don't know which color it will be *due to selectors* we simply parse it to the correct variable and let the browser assign the correct value to the variable. Dark Reader now plays on this to assign the correct values for different variables based on it's value and their desired selectors.

Gusted:
The new data structure will now use O(n2) to interact with all property's and their value/parentGroups.

Gusted:
[binaries.zip](https://github.com/darkreader/darkreader/files/5662494/binaries.zip)
Latest binaries for anyone who like to test out the latest commits. It passes the test and github.com didn't break for me so no bugs known so far.

Gusted:
[binaries.zip](https://github.com/darkreader/darkreader/files/5662519/binaries.zip)

Latest binaries. Passes test and doesn't break for me on github.com(using light theme preference on github.com) no known bugs.

Yash-Singh1:
@Gusted `github.com` seems to be working on `firefox` and `microsoft-edge-dev`

liuhao326:
Hi @Gusted!

`github.com` works well on `Chrome`. But there is an issue on `Stack Overflow` when I choose the light theme:

<details>
<summary>Details</summary>

image

</details>

As we can see, there are some black colors.

Besides, when trying to click an input area on `github.com` after switching the tab, it needs to cost a short period of time to focus on it. This is tolerable in general. But it would be better if it can be solved, actually, the old version doesn't have this problem.

<details>
<summary>Details</summary>

input area

</details>

Gusted:
> But there is an issue on Stack Overflow when I choose the light theme:

#4452

> Besides, when trying to click an input area on `github.com` after switching the tab, it needs to cost a short period of time to focus on it. This is tolerable in general. But it would be better if it can be solved, actually, the old version doesn't have this problem.

It seems like chrome is doing something CPU-Intensive in the background that causes, I know dark reader doesn't have special code after it's darken. Does this problem also occur with dark reader disabled?

liuhao326:
I found that this problem has disappeared now. @Gusted

alexanderby:
I've added a test that fails, because different media queries can have the same selectors. Maybe that's why the light mode doesn't work well in StackOverflow.

So, a quick temporary solution could be using a full path like `@supports ... -> @media screen -> @media ... -> .selector` as a key for the Map.

A good solution could be to not create another `.darkreader--dynamic-variables` style for variables, but put them into `.dynamic--sync` where necessary, but it is more complex and brain-cracking.

Gusted:
> So, a quick temporary solution could be using a full path like `@supports ... -> @media screen -> @media ... -> .selector` as a key for the Map.

Could be possible.
PS: I think I forgot to add support for `@supports`

> A good solution could be to not create another `.darkreader--dynamic-variables` style for variables, but put them into `.darkreader--sync` where necessary, but it is more complex and brain-cracking.

Hmm, that will require some argument pass-trough to make sure we have the right `darkreader--sync`, but it's possible I will have a look in a later PR.

alexanderby:
`@supports` is a rare case, it will be fixed by my second proposal.

The second proposal means that variables should get into this `--sync` just as usual color overrides do.
```typescript
export function getModifiableCSSDeclaration(property: string, value: string, rule: CSSStyleRule, ignoreImageSelectors: string[], isCancelled: () => boolean): ModifiableCSSDeclaration {
    const important = Boolean(rule && rule.style && rule.style.getPropertyPriority(property));
    const sourceValue = value;
    if (property.startsWith('--')) {
        // Instead of `return null` do the work here, e.g. like this:
        // return null;
        try {
            parse(value);
        } catch (err) {
            return;
        }
        const modifier = getVariableModifier(property, value, isCancelled);
        if (modifier) {
            return {property, value: modifier, important, sourceValue};
        }
    }
    // ...
}
```

Gusted:
> ```ts
>  const modifier = getVariableModifier(property, value, isCancelled);
> ```

What do you propose that getVariableModifier does? We don't know the exact case yet e.g. bg or color or shadow or border.

alexanderby:
If the variable will be used just for one type of color (bg, border or text), there could be variableTypes map. When the property starts with --, the modifier will check if type is known and will return a modified color according to it's type, otherwise it will return a Promise. The resolve of the promise will be stored somewhere. When the variable will be met in var(), the Promise will be fulfilled, and the correct modified variable value could be substituted. var() will not have to be replaced with anything. But the variable could be used for different type of colors in this unideal world.

So, currently you could try invent some kind of modifier or something, that will insert 3 values, as you proposed for each color type. My assumptions could be wrong too. And it looks rather complex to achieve.

Gusted:
> If the variable will be used just for one type of color (bg, border or text), there could be variableTypes map. When the property starts with --, the modifier will check if type is known and will return a modified color according to it's type, otherwise it will return a Promise. The resolve of the promise will be stored somewhere. When the variable will be met in var(), the Promise will be fulfilled, and the correct modified variable value could be substituted. var() will not have to be replaced with anything. But the variable could be used for different type of colors in this unideal world.
> 
> So, currently you could try invent some kind of modifier or something, that will insert 3 values, as you proposed for each color type. My assumptions could be wrong too. And it looks rather complex to achieve.

I'd like this idea as it will make it more optimized to achieve the same, however like you said, A variable can be re-used for a different type. So I actually want to have this in a next PR so we can merge this monster after 2 weeks and ship a new version as `github.com` is now burning my eyes with this color scheme of theirs. I will commit the new data structure to resolve the duplicate selector fix.

alexanderby:
The test is still failing. Please note, that you are now using an Array as a key, not a string, but some reference in memory on a particular Array instance, that will be different each time you create an Array.

alexanderby:
Regarding the blinking colors in Karma, use `window === window.top` to check if the page is an iframe. Personally I just open Chrome dev tools on the index.html page. To make it not blink in `debug.html` I have no ideas.

Gusted:
> The test is still failing.

image


>  Please note, that you are now using an Array as a key, not a string, but some reference in memory on a particular Array instance, that will be different each time you create an Array.

I'm not sure how to do this differently. As we can't use a string for a path other than using some kind of `splitter` like `;`

alexanderby:
Use something like `>>>` that has less chance to be met, or `/darkreader/` :)

Gusted:
Updated the data structure once again, I realized we could use the key to just identify THAT specific rule and store all the data into the Variable structure to actual use that data.

alexanderby:
Ai-yai-yai, you shifted the groups away. StackOverflow is about to work in light mode now, I guess there's an issue with nesting variables, `--theme-content-background-color: var(--white)`, I guess this `--white` is dark somewhere in `prefers-color-scheme:dark` media.
Screen Shot 2020-12-10 at 00 45 58

Gusted:
> Ai-yai-yai, you shifted the groups away.

I'd like to declare that I treated every group equally 😄.



>  StackOverflow is about to work in light mode now, I guess there's an issue with nesting variables, `--theme-content-background-color: var(--white)`, I guess this `--white` is dark somewhere in `prefers-color-scheme:dark` media.

image
image
image

Correcto

Gusted:
@alexanderby When you requested the prefix I forgot to update the other references as well also, for some stupid reason myself replaced any variable with an value from legacyVariables, this causes to have the old bug and don't use references to the correct variable.

https://github.com/darkreader/darkreader/pull/4291/commits/2bee7dda0a59a72f9ff0827469ecaf4f989f5b18 fixes this and https://github.com/darkreader/darkreader/pull/4291/commits/0c43a20f15c9b982b1488d867e9ed1ac4ab39217 removes the stackoverflow.com fix

Gusted:
> Regarding the blinking colors in Karma, use `window === window.top` to check if the page is an iframe. Personally I just open Chrome dev tools on the index.html page. To make it not blink in `debug.html` I have no ideas.

It's not by default and probably only used by me for now to debug why something doesn't work as I prefer without the iframe to test certain conditions and code. So it should be fine also did some small optimizations to run the test cases what faster(under 3s on both browsers for my machine)

alexanderby:
Unfortunately there are still 2 bugs:
1. Background images are ignored, seems like can be fixed, if value starts with `url(`. Seen on YouTube.
2. Variables can be not only at the start of the value, but also inside, like this one from Linkedin `0px -1px 1px var(--color-border-faint),1px 0px 1px var(--color-border-faint),-1px 0px 1px var(--color-border-faint),0px 1px 1px var(--color-border-faint)`. That causes to picking an invalid variable name.

src/inject/dynamic-theme/variables.ts
```diff
@@ -0,0 +1,134 @@
...
+    variables.forEach((properties, key) => {
+        properties.forEach((variable, property) => {
+            const {value, selectorText, parentGroups} = variable;
+            const standardDeclaration = {selectorText, property, parentGroups, key};
+            if (cachedVariables.has(key) && cachedVariables.get(key).has(property)) {
+                const {modifiedBackground, modifiedText, modifiedBorder} = cachedVariables.get(key).get(property);
+                declarations.push({...standardDeclaration, modifiedBackground, modifiedText, modifiedBorder});
+            } else if (value.includes('var(')) {
+                // Removes var() from the value and thereby preseve any fallback specified.
+                const variableProperty = value.substring(4, value.length - 1);
```
alexanderby, line 71: So, I think the problem lays out here. E.g. when value is something like `1px solid var(--x), 1px solid var(--y, var(--z, black))`. Maybe it can be fixed by something like `replaceCSSVariables`.

alexanderby (changes requested):

Gusted:
@alexanderby Do you have a simple test case for the `url(`? So I can make sure it will work my fix. For now commenting a fix to make sure https://github.com/darkreader/darkreader/pull/4291/commits/6a8d62027b7bfc7829271a0bce9498f52a7dc514 are working properly.

Gusted:
@alexanderby I can find a few background-images, however they aren't getting styled anyway. It seems rather complex to support it now unless I rewrite the variable.ts file I prefer to add support in this in a follow-up PR and for now disable URL convert back to a darkreader-v variable.

Gusted:
I added some support for background-images anyway https://github.com/darkreader/darkreader/pull/4291/commits/58409c76beb2ba4fc6077b4a42a7a915d4191a3a I will see if I can add some test cases. 
[xd.zip](https://github.com/darkreader/darkreader/files/5681995/xd.zip) Used simple test case to make sure it's being parsed.

Gusted:
@alexanderby darkreader/darkreader/pull/4291/commits/cb79a7fef0495a1492cc424e72e526e1e5d614c9 Add support for the background-images you over had and I'm willing to optimize it to not have it to a new `CSSRule` and add it back into the desired selectorGroup. However not a big deal.
https://github.com/darkreader/darkreader/pull/4291/commits/7434688e6f5bf68ff163bf0541c84c0575359b00 add supports for gradients and fixes some edge cases with Firefox's `regex.exec` and also finally got around to support make sure everything used when checking is using `rgb` and used `parseColorToRGB` to make sure any detected hsl/hex **inside** a string is converted as those 2 only seems to get converted to an `rgb` when being parsed into a `stylesheet`.

For anyone who like to test it
[binaries.zip](https://github.com/darkreader/darkreader/files/5696491/binaries.zip)

Regards,
Gusted

kkoomen:
@Gusted What's your personal estimation of when this PR can be merged?

Gusted:
Hi @kkoomen

Days? Weeks? I honestly don't know I try to improve some of the logic here and their, but it's ready for production, it covers all edge cases I've tested stability and shouldn't break after testing some extreme edge cases. However due to personal things of Alexander, I don't expect it get reviewed this week.

Gusted:
Hi @alexanderby 

When running `npm run test:inject` it will crash the browser, due to something with an infinte stack overflow in `getVariablesMatches`, with an special note to `'should use fallback when nested variables are missing'` test case whereby it seems to be the case the fallbacks are causing this.

Overall I like the simplified version.


Regards.
Gusted

alexanderby:
Thanks. It is unfinished yet. I will also add some logic to rebuild the rules where types become known asynchronously.

jace4511 (approved):

Andre601:
@Gusted @alexanderby Just a heads up. If you didn't already should you consider blocking the above user (jace4511) from this repository as they spammed approvals, issues and prs on other repositories (Perhaps it's an automated account?)

While I did report the account is there no guarantee of a fast handling of this.

Proof:
proof

Gusted:
> @Gusted @alexanderby Just a heads up. If you didn't already should you consider blocking the above user (jace4511) from this repository as they spammed approvals, issues and prs on other repositories (Perhaps it's an automated account?)

This comment actually got filtered by the spam filter, but we are required to follow the [Code of Conduct](https://github.com/darkreader/darkreader/blob/master/CODE_OF_CONDUCT.md#enforcement-guidelines)'s enforcements rules. Which will require some additional steps before reaching 'Ban'-level. I've seen this 'user' already on some PR's which his 'annoying' approving PR's without knowing what it actually does. I will forward this issue to alexander.

kizule:
Hi, when is a release of the new version expected to happen?

alexanderby:
While it fixes many websites, it also breaks or makes some websites slower. So it will be released, when it will be stable.

kizule:
That makes sense, thank you for this information.

PR #4293, opened 2020-11-26
Fix dark text on LinkedIn.com

sd2048:


Gusted:
@Myshor Linkedin fixes are mostly your artwork 😛

Myshor (changes requested):
.global-nav__primary-link-text
First: I did not touch it because it must be different when selected, hovered and in other status. With your proposal you always set one color to all states. I must reject.

The big ammount of buttons/links/etc.
I could think about accept but... It will break averything in light mode. So using light color should be:
color: $(rgba(246, 244, 241, 0.8));

Make the changes and I will approve.

Myshor (commented):
Checked what you want to fix with that big ammount of classes... It's already fixed with my variable ingerention

:root {
    --color-text-low-emphasis: ${rgba(0,0,0,0.6)};
}

And will be included in new release.

sd2048:
Okay, does that make this PR unnecessary?

Myshor:
It seems... I just can't select something like "Not needed" 😉
BTW thanks for approach for the fix. 👍

sd2048:
Got it, thanks.

PR #4298, opened 2020-11-27
Added a fix for Aftonbladet.se

rappos:
Aftonbladet.se is one of the largest newspapers in Sweden. I added a fix for a button on the main page, which can be seen on all sub-pages. There was a contrast error, that had to do with Darkreader inverting black text to white. 

Will probably find more things to correct as I use the website more in the future.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -256,6 +256,15 @@ svg.reviews__icon-topic
 
 ================================
 
+aftonbladet.se
+
+CSS
+._3KEys{
+    color: black
+}
+
+================================
+
 ai2.appinventor.mit.edu
 
 CSS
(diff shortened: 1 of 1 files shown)


Gusted:
Hi @rappos!

The used CSS Selectors, selects both buttons 1 with the yellow one *Which you probably want to targets* and the login button.
The best way to tackle this is to select the *what seems to me* the button that want's you to buy an premium version with 
```CSS
a[href$="plus_plusikon"] > :nth-child(1)
```
On a side-note using the `black` value can be replaced by `${white}` so people with custom settings will see an modifed version of the `black` according to their settings.

Regards,
Gusted

rappos:
Hi @Gusted !
Thanks for the feedback, CSS is not my strong suit, but I think I understood your feedback and fixed the PR. Hopefully, this new one should work!

//Rappos

Gusted:
> Hopefully, this new one should work!

Correct, just for the code-style. Make sure to have an space between the Selector and `{` so `div{` becomes `div {`.

PR #4313, opened 2020-11-29
Fixes for omni.se and worldcubeassociation.org

viktorzenk:
omni.se:
Fixed background color for articles.

worldcubeassociation.org:
Fixed 22 flags being inverted. Pretty janky/hardcoded fix, feel free to find a more elegant solution! Example page with flags visible: https://www.worldcubeassociation.org/results/rankings/333/single?region=world&show=by+region
Fixed event selectors being indiscernible between checked and unchecked. Example: https://www.worldcubeassociation.org/profile/edit?section=preferences (I don't think it's visible anywhere that doesn't require an account).

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5351,6 +5351,21 @@ IGNORE IMAGE ANALYSIS
 
 ================================
 
+omni.se
+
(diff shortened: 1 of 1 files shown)


Gusted:
Make different PR's for different sites.

Issue #4299, opened 2020-11-27
L'internaute - synonymes


mikerouxfr:
**Website**
https://www.linternaute.fr/dictionnaire/fr/definition/tester/

**How to Reproduce**
> 1. Scroll down to the "Synonymes" section
> 2. See that the section is barely readable

**Expected behavior**
The container should gets a darkened background a lengthened text.

**Actual behavior**
Nothing happen to this section because the website's css overrides it (selector higher priority).

**Screenshots**
![image](https://user-images.githubusercontent.com/9786566/100455733-7ed97780-30bf-11eb-8fe2-7769f2a4ee23.png)

**System info:**

- OS: Windows 10 Pro 20H2 (19042.630)
- Browser: Firefox Developer 84.0b4 (64 bits)
- Darkreader Version: 4.9.25

**Additional context**
So far, my suggested snippet to fix looks like this but it doesn't work with Light mode.
```css
================================

linternaute.fr

INVERT
.dico_liste.grid_line > li

CSS
.dico_liste_greyandwhite .dico_liste li a {
    color: var(--darkreader-neutral-background);
}
```

Gusted:
Hi @mikerouxfr!

Seems like dark reader is not detecting the sheet change for some reason, however the fix should encounter this. As for light mode. Dark Reader light mode doesn't apply any fixes as they only should be used for dark mode currently. You can open up a PR for the fix.

Regards,
Gusted

PR #4327, opened 2020-11-30
Fix for linternaute.fr

mikerouxfr:
Synonymes section was not inverted, see #4299

### DIFF
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -1205,6 +1205,18 @@ NO INVERT
 
 ================================
 
+linternaute.fr
+
+INVERT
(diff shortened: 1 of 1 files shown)


Gusted (changes requested):
I think the CSS can be simplified to 
```CSS

CSS
.dico_liste_greyandwhite .dico_liste li {
}
```

mikerouxfr:
Hi!

The original website has banded rows.

- My suggested change displays as:
  image

- The simplified version displays as:
  image


I think we should keep the originally intended design as it improves readability a little.

-----

A note about the padding a the right side of the block: it appears in both versions AND in original website so I think we shouldn't try to fix this.

PR #4337, opened 2020-11-30
Fix for nested lists in ConnectWise Control Access

ScottBeeson:


Gusted:
Hi @ScottBeeson!

Is their an example URL where I can test this fix on?
Also seeing the CSS It seems like it's just reverts it's brightness back to 100%

Regards,
Gusted

ScottBeeson:
No, there is no public URL. You have to have a subscription.

Yes, it only changes the brightness. The elements in question have a brightness of like 20% so when it is inverted to a black background it is unreadable.

PR #4343, opened 2020-12-01
Fix for google search timer fullscreen

Yash-Singh1:
Before:
Screenshot from 2020-12-01 09-51-58
After:
Screenshot from 2020-12-01 09-54-30

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3421,6 +3421,13 @@ a[href*="about/products"]
 
 ================================
 
+google.com/search
+
+INVERT
+img.act-icon-dark-gray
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -3421,6 +3421,13 @@ a[href*="about/products"]
 
 ================================
 
+google.com/search
```
Gusted, line 3424: Hmmm, al existing Google fixes will be lost trough this, please add the INVERT to `www.google.*` around L8543

Yash-Singh1: Done in fd3881c

PR #4372, opened 2020-12-02
Add basic Thunderbird support

### LINKED ISSUE #946: Thunderbird compatibility
If it's not too much to ask, please add Thunderbird 60+ compatibility.

I'm fully aware of dark themes but they leave the message pane unthemed (it's not realistically possible to theme it anyways).

Thank you!

Jakeler:
- Resolves #946 
This PR adds support for making mail content in Thunderbird dark. Works with version 78+, see #946.

### Working
- Dark mail viewer
- Different engines: Filter(+), Static and Dynamic
- Button and UI settings
- Global on/off switch

### Known issues so far
- (controlled through themes: Mail composition/editor is bright)
- DISABLED: Toggle current site does nothing, maybe use sender/subject infos from messages API
- FIXED: Mails are bright for a split second while loading, causing a "flash" every time when switching
- FIXED: Log message: `Security Error: Content at moz-extension://3bb4a850-a13b-495d-a7ab-3aed7cd70655/background/index.html may not load or link to chrome://messagebody/skin/messageBody.css.` Does this break something?
- FIXED: Thunderbird build: I tried to edit the tasks to add another output, but I failed. It looks like a lot of work, almost a rewrite, because of the hardcoded firefox functions/parameters. The build actually does not need many changes, just a patched `manifest.json` with the "messagesModify" permission. Can someone help, please?
To currently build, copy the `manifest-thunderbird.json` to `manifest-firefox.json` and then use the firefox output in TB.
- FIXED: links in the UI do not work, nothing happens on click


src/background/tab-manager.ts
```diff
@@ -124,11 +133,11 @@ export default class TabManager {
...
 
-    registerMailDisplayScript() {
-        // @ts-ignore: not a standard WebExtension API, Thunderbird only
-        chrome.messageDisplayScripts.register({
+    async registerMailDisplayScript() {
+        await (chrome as any).messageDisplayScripts.register({
             js: [
-                {file: '/inject/index.js'}
+                {file: '/inject/fallback.js'},
+                {file: '/inject/index.js'},
```
Yash-Singh1, line 140:
```suggestion
                { file: '/inject/fallback.js' },
                { file: '/inject/index.js' },
```
Bracket spacing

Yash-Singh1 (changes requested):
Bracket spacing

Gusted:
Hi @Jakeler!

Thanks for the basic support!
I've added some improvements, to e.g. fix the security error and build thunderbird files with it's appropriate manifest.json.
I've also added our fallback script to prevent the white flashes as mentioned.
I need to look into if we can possible filter specific messages as it should be technically possible as an internal message URL looks like 
`imap://[email-addres]@[Email-Proivder]:933/fetch%3EUID%3E/INBOX%3E[message number]` Which Decoded from URI looks like `/fetch>UID>/INBOX>[message number]` whereby the message number is different, however I don't know how this looks for other email providers and pop3, I think disabling the sitelist/site toggle for now should do it as temporary fix.

Regards,
Gusted

src/background/tab-manager.ts
Gusted: Running code-style over this reverts it back to normal

Jakeler:
Hi @Gusted,
thank you so much for all your work!
The fallback script was not working for me, because of the `matchMedia` call returns false. I have just fixed it with the isThunderbird check, is that okay?

Good catch with the message number in the URL. Maybe we could get more message info with the [messages API](https://thunderbird-webextensions.readthedocs.io/en/latest/messages.html) to make a filter. But I also think disabling the sitelist/site toggle is fine for now.

Gusted:
> The fallback script was not working for me, because of the `matchMedia` call returns false. I have just fixed it with the isThunderbird check, is that okay?

Hmm, I'm not sure if it should be overriden as people with a light theme and light theme on Thunderbird could now experience the opposite 'dark flashes', but it shouldn't be a too much of a problem. However if you could replace the `isThunderBird` with the actual test as it will now grow the bundle(fallback.js) with including platform.ts into it.

Regards,
Gusted

Gusted:
@Jakeler Do you have a light theme preference?
The mail compositor does show dark for me 
image
I've enabled `Reader's default color` which probably causes to have this color 
image

Jakeler:
I have the default theme enabled, it picks up the color from my system (KDE settings). That makes the UI darker than the builtin dark theme, so I like it better. Except the editor, which is white... with the TB dark theme it looks like your screenshot for me as well. Okay, in any case this is controlled through the themes, so definitely not a issue for darkreader.

Also I see that the `matchMedia` part detects the theme correctly. No white flashing with the actual dark theme. Though the dark flashing happens with the dark theme + darkreader off. I just noticed this also happens in the browser, dark flashing is just not as visible as the white at night. If anything this should be changed in general, so not relevant here for TB. I will revert that check.
BTW: Can't TS use tree shaking to strip out unused code, if only such a small part is imported?

I think the main problems are now solved, it works quite well already!

Gusted:
> BTW: Can't TS use tree shaking to strip out unused code, if only such a small part is imported?

We use a modified version of typescript compiler to have useful compiler errors https://github.com/ezolenko/rollup-plugin-typescript2 it has basic tree shaking support. Might be time to switch the compiler to something like ESBuild 🤷🏽‍♀️

alexanderby:
Do you think the UI makes sense for the mail client?

Jakeler:
@alexanderby I think most of the UI makes still sense, for example the global toggle and filter settings. Only the site list/toggle with URL does not work, we could replace the URL with the sender or subject of the mail. So in general I would just keep the UI.

I found a new bug in the UI though: the links do not work, nothing happens on click.

Gusted:
> @alexanderby I think most of the UI makes still sense, for example the global toggle and filter settings. Only the site list/toggle with URL does not work, we could replace the URL with the sender or subject of the mail. So in general I would just keep the UI.

Yeah, for now just disabling that toggle button will do it.



> I found a new bug in the UI though: the links do not work, nothing happens on click.

We use `<a href="">` So seems like thunderbird currently disallows to open links through this?

Jakeler:
Yes, not sure if it is intentional or a TB bug, these are the errors after clicking:
```
WebExtension context not found! ExtensionParent.jsm:1113
    getContextById resource://gre/modules/ExtensionParent.jsm:1113
    recvAPICall resource://gre/modules/ExtensionParent.jsm:968
    recvAPICall self-hosted:844
    _recv resource://gre/modules/ConduitsChild.jsm:78
    receiveMessage resource://gre/modules/ConduitsParent.jsm:333

TypeError: browser is nullBrowserElementParent.jsm:81:21
    receiveMessage resource://gre/actors/BrowserElementParent.jsm:81
```
Searching for it brings up are a few results, but nothing with meaningful info. 
I remember trying to open URLs in a external browser from a WebExtension in TB version 68 (about a year ago) and it was basically impossible, apparently for security reasons. 
Only option that works AFAIK is to open it in a internal TB tab with `chrome.tabs.create({url: ...})`, like a browser. A solution would be to make a `Link` component, replace all `<a href="">` and do that through `onclick`. What do you think?

src/ui/popup/components/link.tsx
```diff
@@ -0,0 +1,22 @@
...
+function openTab(url: string) {
+    return () => chrome.tabs.create({url});
+}
+
+export default function Link({cls, url}: LinkProps, ...children) {
+    return (
+        <a
+            class={cls}
+            onclick={openTab(url)}
+            rel="noopener noreferrer"
```
Gusted, line 17: This can be removed if `chrome.tabs` is being used. As it won't have any affect.

src/ui/popup/components/link.tsx
```diff
@@ -0,0 +1,22 @@
+import {m} from 'malevic';
+
+interface LinkProps {
+    cls: string;
+    url: string;
+}
+
+function openTab(url: string) {
```
Gusted, line 8: Alternative can be written as 
`const openTab = (url: string) => chrome.tabs.create({url});`
As a one-liner :)

Gusted (changes requested):

src/ui/popup/components/link.tsx
Jakeler: True, good point.

src/ui/popup/components/link.tsx
Jakeler: No, this would create tabs already when the component loads, it must return a function. 
But it can be certainly written shorter, put directly into onclick instead, making it a zero-liner ;)

Gusted:
As stylesheets won't change in mails We can safely disable watching them bring down the CPU Usage to 0% on idle https://github.com/darkreader/darkreader/pull/4372/commits/7313e99fcfaffeb425f5b75e526944a68c7bc210

Jakeler:
Is there anything missing that stops merging this currently? As far as I know it works reliably for everyone now on 78.6+ for reading mails. Maybe a official beta for Thunderbird?

Also do you currently manually build and upload to the addon stores? 
Should be possible to automate with [firefox-addon-action](https://github.com/marketplace/actions/firefox-addon-action) and [chrome-extension-upload-publish](https://github.com/marketplace/actions/chrome-extension-upload-publish).

src/ui/popup/components/header/index.tsx
```diff
@@ -7,6 +7,7 @@ import SystemIcon from '../../main-page/system-icon';
 import WatchIcon from '../../main-page/watch-icon';
 import SiteToggle from '../site-toggle';
 import MoreToggleSettings from './more-toggle-settings';
+import Link from '../link';
```
alexanderby, line 10: Remove Link component, refactoring should not be a part of this PR.

tasks/copy.js
```diff
@@ -19,19 +19,20 @@ function getCwdPath(/** @type {string} */srcPath) {
...
 }
 
-async function patchFirefoxManifest({debug}) {
+async function patchManifest({debug, firefox, thunderbird}) {
     const manifest = await fs.readJson(`${srcDir}/manifest.json`);
-    const patch = await fs.readJson(`${srcDir}/manifest-firefox.json`);
-    const patched = {...manifest, ...patch};
-    const firefoxDir = getDestDir({debug, firefox: true});
-    await fs.writeJson(`${firefoxDir}/manifest.json`, patched, {spaces: 4});
+    const fireFoxPatch = await fs.readJson(`${srcDir}/manifest-firefox.json`);
```
alexanderby, line 24: Use `const patch = ` and load only from a single file.

alexanderby (changes requested):

alexanderby:
If it has no problems, I will publish it when the next release will be ready.

src/ui/popup/components/header/index.tsx
Jakeler: This is not pure refactoring, it fixes the Thunderbird problem where `<a href="">` does not work (as described above). 
Otherwise the onclick arrow function with chrome.tabs API would have to be added everywhere, which seems like a lot of code duplication. Or what do you suggest instead?

alexanderby: Same problem in Firefox, have a look at index.tsx file, it should have a custom click event handler

Jakeler: On normal Firefox the anchor tags work as well. You are referring to `onMoreToggleSettingsClick`, that gets passed with the body state to the header?
So should a helper function be defined in `body.tsx` and passed as parameter to the header, instead of the component import?

alexanderby: Try adding Thunderbird here https://github.com/darkreader/darkreader/blob/df8273af4462b27b59c802b4a1fbd77f536a7f99/src/ui/popup/index.tsx#L44

Jakeler: Oh thanks, I didn't see that. 
It already uses all Firefox fixes though: `isFirefox = userAgent.includes('firefox') || isThunderbird;`. 
It just did not work, because https://github.com/darkreader/darkreader/blob/df8273af4462b27b59c802b4a1fbd77f536a7f99/src/ui/popup/utils/issues.ts#L40 immediately closes the tab again... On Thunderbird the popup closes anyway when switching tabs, so this is not not needed, I have added the isThunderbird check now there and removed the Link component.
Is this okay now?

PR #4379, opened 2020-12-04
Fix for whatsapp home page

Yash-Singh1:
- Fixes logo
- Fixes icons
- Fixes QR (tested)

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -8136,7 +8136,7 @@ web.whatsapp.com
 INVERT
 .selectable-text > span._3PxOr
 span[data-icon="audio-download"]
-.invisible-space > span
+.landing-main .invisible-space > span
 div._1PTz1 > span._2ezGC
 div.landing-header > span > svg
 
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -8136,7 +8136,7 @@ web.whatsapp.com
 INVERT
 .selectable-text > span._3PxOr
 span[data-icon="audio-download"]
-span[dir="ltr"] > ._2ezGC
+.invisible-space > span
 div._1PTz1 > span._2ezGC
 div.landing-header > span > svg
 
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -8136,6 +8136,9 @@ web.whatsapp.com
 INVERT
 .selectable-text > span._3PxOr
 span[data-icon="audio-download"]
+span[dir="ltr"] > ._2ezGC
+div._1PTz1 > span._2ezGC
```
Gusted, line 8140: This inverts the icon inside the QR-Code?

I don't see why, due to the [Error correction math](Reed–Solomon error correction) in the QR-Code standard it won't matter at all if it's inverted or not it should work correctly.

src/config/dynamic-theme-fixes.config
```diff
@@ -8136,6 +8136,9 @@ web.whatsapp.com
 INVERT
 .selectable-text > span._3PxOr
 span[data-icon="audio-download"]
+span[dir="ltr"] > ._2ezGC
```
Gusted, line 8139: Please use
```CSS
.invisible-space > span
```
We can't use a dir="ltr" as e.g. Arabic systems will have dir="rlt" and whatsapp has a known behaviour whereby the selectors are generated again every ~1 week.

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
```diff
@@ -8136,6 +8136,9 @@ web.whatsapp.com
 INVERT
 .selectable-text > span._3PxOr
 span[data-icon="audio-download"]
+.invisible-space > span
```
Gusted, line 8139:
```CSS
.landing-main .invisible-space > span
```
So it will only be applied on the `login` page.

src/config/dynamic-theme-fixes.config
Gusted: Remove this one Line and add under `IGNORE INLINE STYLE`
```CSS
.landing-main div[data-ref] span > svg > path
```

Gusted (commented):
Should be good after 2 little edits 👌🏽

PR #4384, opened 2020-12-04
Fix for daltonmaag.com

Yash-Singh1:


src/config/dynamic-theme-fixes.config
```diff
@@ -1624,6 +1624,13 @@ blockquote {
 
 ================================
 
+daltonmaag.com
+
+INVERT
+#panel-3 > div > div > div > div > img
```
Gusted, line 1630: No elements found on this selector, did you mean to add some extra div's? Alternative you could use
```CSS
#panel-3 div > img
```
Or even shorter
```CSS
#panel-3 img
```

PR #4390, opened 2020-12-05
add forum.xda-developers.com to dark-sites

linsui:


Gusted:
Seems like their new Forum Design came with default Dark Mode, nice.

PR #4400, opened 2020-12-06
Add cache to parseURL

### LINKED ISSUE #1158: DarkReader in Dynamic mode crashes / becomes unresponsive when loading a certain page.
On Chromium, I can repeatedly crash the extension upon loading a certain page: https://downloads-eu.bitwig.com/stable/2.5/Release-Notes-2.5.html

This only seems to happen in Dynamic mode.

I made a video demonstrating this: https://youtu.be/F_NSCgKWNIk

On Firefox 65.0.1 (64-bit) I observed that it just consumed lots of CPU when loading that page, but it did not seem to crash the extension.
Furthermore, on my normal Chromium account, I can't seem to re-enable the extension anymore unless I restart Chromium.

Browser: Chromium v72.0.3626.121 (Official Build) Built on Ubuntu , running on Ubuntu 16.04 (64-bit)

Gusted:
- URL Constructors are expensive in time.
- Adding a cache resolve this problem, because the main $url is 95% the same one.
- Resolves #1158

### DIFF
--- a/src/inject/dynamic-theme/index.ts
+++ b/src/inject/dynamic-theme/index.ts
@@ -20,6 +20,7 @@ import {createAdoptedStyleSheetOverride} from './adopted-style-manger';
 import {isFirefox} from '../../utils/platform';
 import {injectProxy} from './stylesheet-proxy';
 import {parse} from '../../utils/color';
+import {parsedURLCache} from '../../utils/url';
 
 const variables = new Map<string, string>();
 const INSTANCE_ID = generateUID();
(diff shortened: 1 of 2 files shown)


alexanderby (approved):
Hope we are not modifying these parsed URLs somewhere.

Gusted:
parseURL on it's own not, however getAbsoluteURL is being used frequently especially for background-images. However it doesn't seem to break. We definitely should test this cached behaviour on major sites to see if it doesn't actually break edge-cases.

PR #4417, opened 2020-12-07
fixes #1273 for wiki.haskell.org

### LINKED ISSUE #1273: Math formulas on a lot of pages are invisible

hyiltiz:
fixes #1273 for wiki.haskell.org

### DIFF
--- a/src/config/fix_inversion.json
+++ b/src/config/fix_inversion.json
@@ -888,7 +888,10 @@
             "rules": ".qrcode { border: 8px solid black; }"
         },
         {
-            "url": "wikipedia.org",
+            "url": [
+                "wikipedia.org",
+                "wiki.haskell.org"
+            ],
             "invert": [
                 ".mwe-popups-discreet > svg",
                 ".mw-mmv-overlay"


Gusted:
Hi @hyiltiz!

This issue was fixed a very while ago already.

Regards,
Gusted

hyiltiz:
For `wiki.haskell.org`? I see the issue was about wikipedia and wikipedia was fixed, but wiki.haskell.org has exactly the same problem. `wiki.haskell.org` doesn't seem to be mentioned in the project repo.

Gusted:
> `wiki.haskell.org` doesn't seem to be mentioned in the project repo.

Because it was never reported.

PR #4429, opened 2020-12-08
github.com already has dark mode

zwpaper:


alexanderby:
This list includes websites that are always dark.

PR #4433, opened 2020-12-08
Disallow newlines in fixes

### LINKED ISSUE #4425: [Feature Request] Stick with newlines or no newlines between CSS rulesets
**Is your feature request related to a problem? Please describe.**

No

Currently, the Dynamic Theme Fixes, Filter Theme Fixes, and the Static Theme Fixes sometimes have newlines between rulesets or sometimes they don't. I think that there should be one solution that should be used all the time, either stick with newlines or without newlines.

**Additional context**

This will keep the configuration more consistent

Gusted:
- Don't allow multiple newlines in the CSS section.
- Resolves #4425


Yash-Singh1:
@Gusted Should there be a test for this? Also, I think the `code-style` script should remove all newlines between rulesets in configuration.

Gusted:
I can add test cases later to ensure newlines are removed and fail the test if it's detected an multiple newline. I don't think the code-style script should do this, it's purely for the codebase.

Gusted:
https://github.com/darkreader/darkreader/blob/master/tests/config/config.tests.ts#L68 already test if the CSS is formatted the right way. And also with this PR should filter out the double newlines.

Gusted:
Going to add documentation about this later in Contributing.md along with other code-style guidelines which additionally may or may not receive tests.

PR #4440, opened 2020-12-09
Add github.com in dark sites list

hashlash:
GitHub now has a dark theme feature!

https://github.com/githubevents/universe2020/discussions/17

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -270,6 +270,7 @@ gifrun.com
 gifyourgame.com
 gikken.co
 giphy.com
+github.com
 githubuniverse.com
 gitkraken.com
 gitmoji.kaki87.net


Yash-Singh1:
@hashlash This feature is only for sites that by default have dark mode and always will stay in dark mode.

hashlash:
Oh, but what about docs.docker.com #1653? It uses the light theme by default (but there's a toggle to switch it to dark mode)

Gusted:
> Oh, but what about docs.docker.com #1653? It uses the light theme by default (but there's a toggle to switch it to dark mode)

Then we need to remove that entry from the dark sites, as it shouldn't be in their then.

Also we only accept dark-by-default, which github.com isn't.

PR #4451, opened 2020-12-09
add style for github.com pr code review

javasgl:
- green means : addition

- red means: deletion


javasgl:
Results the following：
image

Gusted:
#4291 will do this by correctly giving the right color to styles, both for light/dark theme.

Issue #1327, opened 2019-05-17
Select native site dark theme if available


moham96:
Hi,
Some site already has a dark/black theme to select which usually looks better than the one implemented by darkreader, so it would be nice if darkreader can detect if the website has a dark theme and activate it instead of color inverting.
This can be implemented by having a database of these sites similar to [dark-sites.config](https://github.com/darkreader/darkreader/blob/master/src/config/dark-sites.config)
the database should contain the domain along with how the dark mode can be activate (usually cookie can influence the theme or DOM object that can be clicked.

FranklinYu:
Example:

* https://support.google.com/a/answer/9296621 see the moon button at bottom-right corner
* https://docs.docker.com/install/overview/ see the toggle at right bar

I'm not sure whether this should include sites that only provides dark theme for logged-in users (like Reddit or Twitter) but I prefer not to. Ideas?

FranklinYu:
Actually I hope that websites ultimately use proper CSS for automatic dark mode: #1285 but that's beside the point.

FranklinYu:
Some HTTP endpoint toggling the theme was mentioned in https://github.com/darkreader/darkreader/issues/1837#issuecomment-579782693. So now we have three possible ways to toggle:

1. cookie
2. an endpoint (possibly require some session)
3. clickable DOM object

I assume that DOM object changes more frequently than previous two.

Gusted:
> Some HTTP endpoint toggling the theme was mentioned in [#1837 (comment)](https://github.com/darkreader/darkreader/issues/1837#issuecomment-579782693). So now we have three possible ways to toggle:
> 
> 1. cookie
> 2. an endpoint (possibly require some session)
> 3. clickable DOM object
> 
> I assume that DOM object changes more frequently than previous two.

Or just get the html element and check which color it is (if color = #000 than native dark has been enabled don't need to be darken. Else Darken Native theme hasn't been enabled)

KaKi87:
Well, you still need the cookie or endpoint to enable the native theme, so we might as well use it to detect whether it is enabled or disabled.

Gusted:
But not everyone would want to have native theme. 

> Well, you still need the cookie or endpoint to enable the native theme, so we might as well use it to detect whether it is enabled or disabled.

To detect you can simple just get an element that is static loaded by default and check the styles of it Endpoint or a Cookie will require some debug time to figure out and add them in the code where not everyone is comfortable with for Privacy parts.

[...]

PR #4454, opened 2020-12-09
[NO-JIRA] Exclude github.com

georgegillams:
[Github.com now supports native dark mode](https://github.com/settings/appearance). We should therefore encourage use of this setting instead of applying custom styling.

Let me know if there's a better way to do this, or if you want to wait until the GitHub feature is out of beta.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -269,6 +269,7 @@ gifrun.com
 gifyourgame.com
 gikken.co
 giphy.com
+github.com
 githubuniverse.com
 gitkraken.com
 gitmoji.kaki87.net


Gusted:
4th PR excluding github.com and once again closing because it's not dark-by-default.

georgegillams:
> 4th PR excluding github.com and once again closing because it's not dark-by-default.

Ah - sorry for creating more noise. I should have checked the closed PR's not just open ones 🤦

It would be really nice if DR could let users know when a site has a dark-mode option, as a lot of sites like GitHub, StackOverflow etc seem to be going for the opt-in approach.

Maybe a message could appear asking if the user wants to exclude the site from DR and go to the app settings?

Is this a feature you're working on or interested in?

Gusted:
> Maybe a message could appear asking if the user wants to exclude the site from DR and go to the app settings?
Is this a feature you're working on or interested in?

#1327

PR #3917, opened 2020-10-23
Restructure Info about Dark Reader

Gusted:
- Make README.md less bloated.
- Moved contribute-related info to CONTRIBUTE.md
- Made some changed to ISSUE_TEMPLATE
- Added question template specific for questions.
- Restructed layout of README.md
- Restructed layout of CONTRIBUTING.md
- Give a better image and description for new users what Dark Reader actually is/do.
- Add more text by the sponsoring text.

See README.md at https://github.com/darkreader/darkreader/tree/Documentation-Patch

PR #4456, opened 2020-12-09
Add carbon.now.sh to the list

bukanvalen:
Pretty surprised that it wasn't on the list. So I decided to add it.

By the way, GitHub now has a built-in dark theme. But it's still in beta. Looks pretty slick though. Should we add it to the list? Thanks

image

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -90,6 +90,7 @@ byte.co
 caniuse.com
 captaindisillusion.com
 caracal.club
+carbon.now.sh
 cardsity.app
 carl.gg
 cascadr.co


Gusted:
> By the way, GitHub now has a built-in dark theme. But it's still in beta. Looks pretty slick though. Should we add it to the list? Thanks

Only dark-by-default sites, which doesn't seems to be the case.

Yash-Singh1:
@Gusted Do you mean sites that are always dark or sites that are dark by default but have an option for light mode.

Gusted:
`Dark-by-default` so a fresh new user on that site that prefer a light-system should still see an dark theme. If their is an option to switch back to light mode doesn't matter a lot.

bukanvalen:
@Gusted thank you for the explanation. I was also a bit confused before. I'll look out for any other dark-by-default sites

Gusted:
> @Gusted thank you for the explanation. I was also a bit confused before. I'll look out for any other dark-by-default sites

Yeah I agree it's a bit vague now, we could update the info in #3917 with a better sentence.

@Yash-Singh1 Any ideas maybe?

PR #4459, opened 2020-12-09
Fix bright red background on track.toggl.com.

samhanes-fs:
track.toggl.com has a very pale pink background, which dynamic mode turns into a very jarring red. This forces it to be the default Dark Reader background color.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -7607,6 +7607,15 @@ body {
 
 ================================
 
+track.toggl.com
+
+CSS
+#root, .content-wrapper > * {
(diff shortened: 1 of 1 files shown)


Gusted:
Hi @samhanes-fs!

It seems like this `red background` only occur on the settings, is this right?

Regards,
Gusted

samhanes-fs:
@Gusted I also get the red background on the main Timer page in the header and between the day blocks. Dynamic mode assigns the actual day blocks a slightly lighter gray that looks fine, but the page background behind them is red.

## Before this change
before

## After this change
after

Gusted:
Hmm weird, I only get this red background-color on the settings page however it isn't a big of a deal then.

PR #4472, opened 2020-12-10
Better notifyOfNews

### LINKED ISSUE #6637: [Feature Request] Add option to toggle off news fetching
# Feature Request

DarkReader silently sends a POST request to `darkreader.github.io` which is **redirected** to `darkreader.org` a proprietry website as far as I know.

This poses a privacy issue, For the sake of the argument lets suppose that the owner of `darkreader.org` decides to take nefarious actions and track users without consent,
In that case, I could not know about that unless im told to.

So as far as things currently are, DarkReader could be classified as a spyware, unfortunately.
It collects user's data (including IP which is not anonymous).

## Screenshots
Location in code: `./src/background/newsmaker.ts:56`
`https://github.com/darkreader/darkreader/blob/0097a6d4c32fae10f13bbe8236d89e53a25191a8/src/background/newsmaker.ts#L56`

## Additional Context
I could suggest few options:
* Transparency over `darkreader.org` backend
* Toggle switch option for connecting to `darkreader.org`
+ Bonus: disable news fetching by default and issue a request only upon clicking "Get news!" or "What's new?"

A similar discussion took place on Etcher's github page:
https://github.com/balena-io/etcher/issues/2977

Please enlighten me if there was any mistake in my statements.
I'd like to amphasize that im a big fan of the work done on this project, it is really outstanding and i wish for it to become much more popular, please give users the privacy that they deserve.
 
EDIT: Typo and link to Etcher issue

Gusted:
- Add toggle to enable/disable this. Fixes #6637.
- Only subscribe when it's enabled.
- Add unSubscribe function.
- Make notifyOfNews enabled by standard.


bershanskiy:
Overall, this looks good and works locally. I have two minor nits:
 - make alarm listener into a top-level function (e.g., `Newsmaker.alarmListener`) and remove it in `Newsmaker.unSubscribe`
 - `Newsmaker.subscribe` unconditionally calls `Newsmaker.updateNews`, which creates a network requesrt each time. This might be wasteful, e.g., if extension is being constantly reloaded by `build:watch`.

Feel free to ignore these, I can perform these changes later, if you don't have the time.

bershanskiy:
@Gusted should we merge this now or do you plan to fix the nits I described in above comment?

alexanderby:
Guys, let's use another flag. `notifyOfNews` was for those guys who didn't want to see an icon. At the first iteration there was an icon showing up, some people found it distracting (but I guess there are very few of those). Most of the users have already `false` value stored in their configs.

So please add something like `fetchNews` etc. I think `notifyOfNews` can be removed.

alexanderby:
The date URL query parameter prevented caching the results. I don't know if `no-cache` helps here, it all depends on GitHub server configuration. I remember config requests used to be cached. Then I used a time stamp as a parameter, and a few years later GitHub complained at me it was not good. So I've added this date. I'm not sure if the requests won't be cached without it.

bershanskiy:
I just tested loading the URL and it has header `cache-control: max-age=600`, so could would be cached up to 10 minutes. I feel like it's acceptable, considering we make this request every 4 hours (so cache would expire before we even try to make the next request).

alexanderby:
Hmm, OK, let's try without this argument.

bershanskiy:
I implemented all of my own suggestions and @alexanderby requests, I think it is ready to be merged.

Gusted (commented):
Approving my own PR :) Had a quick look and it seems good to me.

bershanskiy:
I'll wait for 24 hours to give Alexander time to comment. :)

PR #4492, opened 2020-12-13
Improve speed on builds

Gusted:
- Replace `rollup` with `ESBuild.`
- Custom transformer for jest to use ESBuild.
- Custom preprocessor for Karma to use ESBuild.
- Native type checking from typescript compiler.
- Enable incremental to use caching on typescript compiler.
- Everything works like normal.
- *Code-Coverage with ESBuild requires another custom plugin, expected to be complex.* I didn't use it anyway so can be implemented later, if it's important to have this feature.
- Improve speed on using tasks due to wrapping heavy tasks of `bundle-html.js` into a function and call when necessary.
- Removed/reduced necessary timeouts in test cases and bring overall testing with browsers to 0-10 seconds.
- Resolves, some issues with walking away from my PC when building or testing Dark Reader :).
- Overall should be Cleaner & Faster & Customize-able, due to custom plugins support of jest/karma/esbuild.
- Removed 10 dependency replacing the functionality with ESBuild or a custom plugin.


Gusted:
>     * _Code-Coverage with ESBuild requires another custom plugin, expected to be complex._ I didn't use it anyway so can be implemented later, if it's important to have this feature.

I've debugged/hack around for around 3 hours now and my conclusion is that current state of ESBuild's plugin API isn't *currently* capable to add instanbul's code coverage technique, and will require to have rollup dependency back to only use ESBuild to `compile` the code and rollup & ESBuild as `compiler` don't like each other a lot currently. So unless this `code-coverage` is a must now, I can better close this PR.

Gusted:
Their seems to be not treeshaking in place for the debug functions `logWarn` & `logInfo` and I've filled a issue about this https://github.com/evanw/esbuild/issues/598 so hopefully can be at a later point being eliminated for production code.

Gusted:
https://github.com/darkreader/darkreader/pull/4492/commits/11fe4638cb6a157287a5bfd74deece78987ad710 Coverage is finally enabled after a lot of testing and making it portable and thus making it this PR finally having 0 cons.

PR #4501, opened 2020-12-14
Make sure it will pass node without errors

### LINKED ISSUE #4498: [Feature Request] Load as an import without exception being thrown when using Server Side Rendering
**Is your feature request related to a problem? Please describe.**

I'm using DarkReader API inside a Svelte/Sapper project.

This technology stack enables SSR of a TS page before serving to the browser.

Because node javascript runtime for SSR does not contain window, or window.chrome (or possibly others), node throws an error when importing the darkreader package.

This error is not recoverable because darkreader cannot be conditionally included (I'm in typescript), and the entire page and it's imports are required to be loaded.

The following is thrown when trying to run a component for Server Side Rendering.
```
ReferenceError: window is not defined 
```

This can be verified by running the following
```bash
$ node node_modules/darkreader/darkreader.js
```
Which will result in
```
/xxx/node_modules/.pnpm/darkreader@4.9.26/node_modules/darkreader/darkreader.js:4203
    throw e;
    ^

ReferenceError: window is not defined
    at /xxx/node_modules/.pnpm/darkreader@4.9.26/node_modules/darkreader/darkreader.js:244:5
    at /xxx/node_modules/.pnpm/darkreader@4.9.26/node_modules/darkreader/darkreader.js:7:68
    at Object.<anonymous> (/xxx/node_modules/.pnpm/darkreader@4.9.26/node_modules/darkreader/darkreader.js:10:2)
```

Ensure there are no "assumed property is defined" checks inside DarkReader when package is imported.

For example
```typescript
if (!window.chrome) {
    window.chrome = {} as any;
}
if (!chrome.runtime) {
    chrome.runtime = {} as any;
}
```

**Additional context**

My current workaround is open up node_modules/darkreader/darkreader.js and surround the entire method body with a try/catch, and log the error to the console.  This is not ideal if other errors are thrown.

Gusted:
- Add extra conditions to only call certain objects when they are defined.
- Improve support for SSR.
- Resolves #4498

### DIFF
--- a/src/api/chrome.ts
+++ b/src/api/chrome.ts
@@ -1,11 +1,9 @@
 import {readResponseAsDataURL} from '../utils/network';
 import {callFetchMethod} from './fetch';
+import {chromeAPI} from './mock';
 
-if (!window.chrome) {
-    window.chrome = {} as any;
-}
-if (!chrome.runtime) {
-    chrome.runtime = {} as any;
+if (!chromeAPI.runtime) {
+    chromeAPI.runtime = {} as any;
 }
 
 const messageListeners = new Set<(...args) => void>();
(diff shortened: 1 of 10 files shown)


alexanderby (changes requested):
Could you please put your fix for iOS issue #4788 into a separate PR, because I have a feeling that something strange happens here?

Gusted:
> because I have a feeling that something strange happens here?

I don't know what's happening, maybe that specific API wasn't implemented. I only know how to not let it bug out.

Gusted:
Not needed.

PR #4504, opened 2020-12-15
Fix postmarketos.org

huupoke12:
[wiki.postmarketos.org](https://wiki.postmarketos.org/wiki/Main_Page) doesn't have a dark mode.


Gusted:
Hi @huupoke12!

Currently having those selectors into the `dark-sites.config` isn't being merged due to the new changes how URL's work. #373 

Regards,
Gusted

PR #4523, opened 2020-12-17
Fixed "Verified Icon" on Google Translate

vinayakkuradia:
On searching 'translate', google provides mini translator in search results, which has a verified icon (Shield and a Right Tick mark) for some basic phrases or words. Which was black so I added an inversion fix.

Before: 
ub9o2EW

After:
cxa5tDM


Gusted:
#4411 this was already fixed.

PR #4532, opened 2020-12-18
emma.cline.engineer already implements a dark mode

EmmaJCline:
I added a dark mode that supports the prefers-color-scheme and will soon have a toggleable switch in the top right corner to also control it. The dark mode is a similar color scheme to darkreader so thanks! 

This should maintain alphabetical order.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -210,6 +210,7 @@ ellie-app.com
 elybeatmaker.com
 emanuelemilella.com
 emeraldchat.com/app
+emma.cline.engineer
 emuparadise.me
 energized.pro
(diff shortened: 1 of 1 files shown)


Gusted:
Hi @EmmaJCline!

Unfortunately to get into this list is to have a dark mode by default even if the users prefers to have a light theme, so that when dark reader decides to not style the page as it's considered as dark mode it should be expected a dark mode for all kind of users.

Regards,
Gusted 🏳️‍🌈

EmmaJCline:
Ahh thanks for the explanation! That makes complete sense and I appreciate your prompt response.

PR #4541, opened 2020-12-18
Add invert to Facebook logo in Instagram settings

busybox11:


Gusted:
@Myshor could you test out this kne

Myshor:
I don't use Instagram.

Even if I login with facebook to test It would be nice to have some example URL to check which facebook logo should be inverted or not.
I don't want to go through whole instagram to search it as I don't even use it.

So @busybox11 can you share some example URL that will help with checks?

busybox11:
Yeah, I can do that. I left my house like 2 minutes ago, so I'll do it when I'll come back (if I remember to do it lol).

busybox11:
Hey, I'm back.
See the facebook logo on the bottom left of the card.
image
This is the fix

image
This is without the fix

Myshor (commented):
Ok... It works but as I know a little bit facebook family apps it's better to not use classes for selector as they can change very often. ;) I recommend using svg[aria-label*="Facebook"] instead. ;)

Myshor:
Just check if it will work for you 'cause my aria-label was in Polish but I think all languages will include that "Facebook" pattern.

busybox11:
Ookay.

busybox11:
Yup, it works!

PR #4557, opened 2020-12-21

almaceleste:
add rule for openedu.ru to prevent ugly outline in focused button

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5792,6 +5792,15 @@ img[src="images/home_banner.png"]
 
 ================================
 
+openedu.ru
+
+CSS
+button:focus {
+    outline: none;
+}
+
+================================
+
 opengeofiction.net
 openstreetmap.org
 


src/config/dynamic-theme-fixes.config
```diff
@@ -5792,6 +5792,15 @@ img[src="images/home_banner.png"]
 
 ================================
 
+openedu.ru
+
+CSS
+button:focus {
+    outline: none;
```
renestalder, line 5799: If I'm allowed to give an input here (even though I'm not a maintainer of Dark Reader): This solution causes accessibility issues as the outline style is significant for keyboard navigation. The outline shouldn't be removed without adding a sufficient alternative, e.g.

```css
outline: 2px solid transparent; /* Windows High-Contrast Mode fix */
box-shadow: 0 0 0 2px whatevercolourfits; /* Replace default outline with box-shadow */
```

Gusted:
The site has the same outline and removing it isn't something Dark Reader should do and quoting `This solution causes accessibility issues `.

almaceleste:
yep, didn't think about it. 
ok, I'll remove it for myself with userstyle.

PR #4577, opened 2020-12-24
Add fixes for decathon.in

Vedanth-P:
This fixes an issue with all (as far as I can see) product images and many other images on the Decathlon India website. Those elements, when in dark mode, have a dark tint over them, which affects visibility and makes them look greyed out. This fix sets the background color of certain parent elements of those images.
Examples of these images can be seen on the search results at [https://www.decathlon.in/search?query=brake%20pad](https://www.decathlon.in/search?query=brake%20pad)

Different colors ranging from HTML Grey to HTML White were tried, and HTML Silver seemed to be a good middle ground between the still tinted look of Grey and the potentially unappealing contrast of White. There, however, still persists a very slight tint when using Silver. The only way I could get rid of the tint completely was to use White.

Additionally, some text elements on the elements affected by this needed correction, as they were light and could not be read on the HTML Silver backgrounds. An example of these can be seen in the "Advantages" section at [https://www.decathlon.in/p/8325111/bikes-brakes-and-brake-pads/disc-brake-pads-compatible-with-shimano-deore-tektro](https://www.decathlon.in/p/8325111/bikes-brakes-and-brake-pads/disc-brake-pads-compatible-with-shimano-deore-tektro) (for example, the "Compatibility" heading under the "Advantages" section; the image on the left needed correction because of the tinting earlier mentioned, and the text on the right needed correction because of the fix for the images).

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1764,6 +1764,18 @@ INVERT
 
 ================================
 
+decathlon.in
+
+CSS
+.media, .swiper-slide {
+    background-color: silver !important;
+}
+.media-body {
+    color: black;
+}
+
+================================
+
 decathlon.pl
 
 INVERT
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1767,11 +1767,8 @@ INVERT
 decathlon.in
 
 CSS
-.media, .swiper-slide {
-    background-color: silver !important;
-}
-.media-body {
-    color: black;
+img {
+    mix-blend-mode: normal !important;
 }
 
 ================================


src/config/dynamic-theme-fixes.config
```diff
@@ -1764,6 +1764,18 @@ INVERT
 
 ================================
 
+decathlon.in
+
+CSS
+.media, .swiper-slide {
```
Gusted, line 1770: Replace this with:
```CSS
.common_card .card_img img {
    mix-blend-mode: none !important;
}
```

src/config/dynamic-theme-fixes.config
```diff
@@ -1764,6 +1764,18 @@ INVERT
...
+    background-color: silver !important;
+}
+.media-body {
```
Gusted, line 1773: You can remove this fix if you use the above fix ^

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
Vedanth-P: I tried it, but it doesn't work :(. The images on the search results and the "Advantages" section are as they were before the fix. I also tried setting the mix-blend-mode property for the classes img-fluid, media and media-left, but to no avail

Gusted: What about:
```CSS
img {
}
```

Vedanth-P: This does the job.
The images in the "advantages" section [here](https://www.decathlon.in/p/8325111/bikes-brakes-and-brake-pads/disc-brake-pads-compatible-with-shimano-deore-tektro) do not blend in quite as well as they would without dark mode -- is this a concern? This is only an aesthetic problem, and not relating to usability or anything of the sort.

Vedanth-P: [I've committed the change](https://github.com/Vedanth-P/darkreader/commit/e10a78e52064bf6324f286cc408a32bbddbb4d87) to Vedanth-P/darkreader, branch patch-3 -- the same branch that was created when I submitted the pull request. Is that what I'm supposed to do (I'm kinda new to GitHub and Git)?

Gusted: > [I've committed the change](https://github.com/Vedanth-P/darkreader/commit/e10a78e52064bf6324f286cc408a32bbddbb4d87) to Vedanth-P/darkreader, branch patch-3 -- the same branch that was created when I submitted the pull request. Is that what I'm supposed to do (I'm kinda new to GitHub and Git)?

Correcto.

PR #4582, opened 2020-12-24
Add godbolt.org to dark sites

aspenluxxxy:
It has a dark theme by default.


Gusted:
image
Dark theme is based of color-scheme preference.

PR #4588, opened 2020-12-26
Fix for drive.google.com/file

Rakleed:
- Invert file contents.

And if the user is not satisfied with the quality of the inversion, then he can simply disable DarkReader for this site, and his site will remain dark in color, and the document will look like the original.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2177,6 +2177,7 @@ drive.google.com/file
 INVERT
 div[role="menu"] > div[role="menuitem"] > div > div > div
 div[role="menu"] > div[role="menuitem"] > div > div
+div[role="document"]
 
 CSS
 div[role="toolbar"] div[role="button"] > div[class*='-']:not([onclick]):not(:link):not(:visited):not([style*="background-image"]):first-child,


Gusted:
> And if the user is not satisfied with the quality of the inversion, then he can simply disable DarkReader for this site, and his site will remain dark in color, and the document will look like the original.

Uhm, their are also other parts of files that get darkened by Dark Reader and disabling it will not style those parts.

Rakleed:
Other parts of files? Did you mean the site? If you meant that when you turn off DarkReader some parts of the site will be light, then I suggest turning off DarkReader temporarily, for one file, and then turning it back on.

Gusted:
> Other parts of files? Did you mean the site? If you meant that when you turn off DarkReader some parts of the site will be light, then I suggest turning off DarkReader temporarily, for one file, and then turning it back on.

If you disabling the site via the current button will disable it for `drive.google.com` not just the `/file` path. I already can see the issues rumbling in about this.

Rakleed:
I rather meant that you opened some document on drive.google.com/file. If it was incorrectly inverted, then you temporarily disabled the inversion of the site, and when you finished working with the site, you turned it back on or the next time you open the site.

The rule I propose to add is very useful for me, since I often look at PDFs on Google Drive. And I would like them to be inverted by default.

Gusted:
It's a very vague selector. It doesn't make any difference between items. e.g. photo's, some pdf.
As well as not everyone wants this it's best to integrate this into the PDF logic already existent, however `previews` on normal `drive.google.com/drive/` also needs to be affected, but then you get a problem with other files as well that use the same selector.

Rakleed:
It's a shame that there is no good solution to this problem at the moment.

rakleed:
@Gusted is it possible to modify this rule so that it works only when the content in a `div` with the class `aLF-aPX-aPU-awE` ends with `.pdf`? Or is it impossible for Dark Reader to make such a rule?

<details><summary>DevTools Screenshot</summary>

Screenshot 2022-10-08 at 03 06 16

</details>

wan7teen-ping:

https://youtu.be/aIaqr4eq18g

Pada Sab, 8 Okt 2022, 8:09 AM Pavel Moiseenko ***@***.***>
menulis:

Gusted:
> @Gusted is it possible to modify this rule so that it works only when the content in a `div` with the class `aLF-aPX-aPU-awE` ends with `.pdf`? Or is it impossible for Dark Reader to make such a rule?

That would be impossible. But what is possible is to add the previous element into consideration:
image

Which has an PDF icon, not sure if this for every file or just those classified as PDF.

rakleed:
@Gusted only PDF files have this icon, other file types have a different icon. But for starters, it would be nice to add at least a PDF.

And how can you make such a rule if these elements have different parents?

<details><summary>Screenshot</summary>

Screenshot 2022-10-10 at 16 58 16

</details>

Gusted:
> And how can you make such a rule if these elements have different parents?

Oh, I'm sorry but in that case you would need to rely upon `:has` CSS selector. But that's not supported quite well https://caniuse.com/css-has

PR #4597, opened 2020-12-26
Remove blokada.org from already dark websites list

KaKi87:
screenshot-blokada org-2020 12 26-22_53_24


KaKi87:
CC @sprinkledcheese (#2618)

PR #4609, opened 2020-12-28
Fixed tutorialspoint.com

Mertyn:
I fixed some inversion issues on the homepage and learning pages of [tutorialspoint.com](https://tutorialspoint.com/).
There are more pages of the site, that need fixing but this is all I got done so far.

tutorialspoint1

tutorialspoint2

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -7918,6 +7918,18 @@ INVERT
 
 ================================
 
+tutorialspoint.com
+
+INVERT
+i.fa
+i.fab
+i.fal
+img.tp-logo
+button.slick-prev.slick-arrow::before
+button.slick-next.slick-arrow::before
+
+================================
+
 tv.yandex.*
 
 INVERT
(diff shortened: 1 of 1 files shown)


Mertyn:
I don't really understand why the test failed, I did everything exactly as it should be :(

Gusted:
The fix should be above `tv.yandex.*`

Mertyn:
Oh okay

Gusted:
> The darkreader devtools put that there though. So maybe there is an issue there.

No, the `tv.yandex` isn't currently in the devtools as it was recently added.

Mertyn:
> No, the `tv.yandex` isn't currently in the devtools as it was recently added.

Got it :)

Gusted:
> Is there a tool or something you can recommend, that automatically inserts it in the right spot of the config file?

Devtools itself, I test every change locally so most of the new sites are alreaedy in my devtool.

Mertyn:
> Devtools itself, I test every change locally so most of the new sites are alreaedy in my devtool.

Okay, well I'm just fixing it in my browser and then copy the changes over to the config file in the repo. So maybe I gotta improve my way of working with that, so that i always have the most recent config :D

PR #4610, opened 2020-12-28
Fixed developer.mozilla.org documentation pages

Mertyn:
I fixed many icons on the docs, that didnt invert, but there may have been some I haven's seen and fixed yet.
Also I couln't fix the seach icon in the top right, maybe someone else knows how to best fix that.

mdn

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1917,9 +1917,17 @@ developer.mozilla.org
 INVERT
 mark
 .icon-thumbs-down-alt
-span[class*="bc-head-icon"]::before
+.bc-head-icon-edge::before
 .last-modified::before
 .newsletter-hide > svg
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -1917,9 +1917,17 @@ developer.mozilla.org
 INVERT
 mark
 .icon-thumbs-down-alt
-span[class*="bc-head-icon"]::before
```
Gusted, line 1920: This was specially crafted so all icons of any browser was inverted.

Gusted (changes requested):

Mertyn:
I have added the request and while i was at it found some more icons that needed fixing. Additionally the logo had a problem, where it was already inverted on the home page, but not on the docs pages (before my fix), so i fixed that aswell.

PR #4619, opened 2020-12-29
Revert "Improve Arabic translation (ar.config)"

OmarNamis:
Reverts darkreader/darkreader#4554

The translation was already accurate.

We're not translating the tags (such as @set_location) but rather what they are referring to in English:
https://github.com/darkreader/darkreader/blob/master/src/_locales/en.config

To be fair, I did have a typo in there.


Gusted:
> The translation was already accurate.
> 
> We're not translating the tags (such as @set_location) but rather what they are referring to in

Aha. Thanks for the explaination.

OmarNamis:
You're welcome @Gusted , and thank you @HamzaKhairy !

PR #4641, opened 2021-01-01
Support For Google Sheets Add-ons

kennytrytek:
Sheets add-ons run in an iframe, so simple inversion is all that's needed.


Gusted:
Iframes are by default styled trough Dark Reader, no need for an INVERT. If their is an bug with a specific add-on please open up an issue.

PR #4645, opened 2021-01-01
Fix for bitly.com

Rakleed:
- Remove background.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -986,7 +986,7 @@ bitly.com
 
 CSS
 .hero-content {
-    background: none !important;
+    background-image: none !important;
 }
 
 ================================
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -982,6 +982,15 @@ span.code {
 
 ================================
 
+bitly.com
+
+CSS
+.hero-content {
+    background: none !important;
```
Gusted, line 989: Use `background-image`

Rakleed: Why?

Gusted: It should specify the CSS property.

Rakleed: Ok, fixed.

PR #4669, opened 2021-01-04
circleci.com

bjfletcher:
The build log output is already dark (not the whole webpage though), these lines should keep the build output dark while inverting the rest of the page.

### DIFF
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -251,6 +251,14 @@ NO INVERT
 
 ================================
 
+circleci.com
+
+INVERT
+.css-1ibtrcs
+.css-1lflngw
+
+================================
+
 cloud-catcher.squiddev.cc
 
 INVERT


Gusted:
Hi @bjfletcher!

It seems like the CSS Selectors your using to be a bit of obscure and those kind of CSS Selectors are know to change every x time. I mostly would give an better CSS Selector, however there seems to be no demo of circleci nor a public build log. Can you by any chance get a CSS Selector that doesn't rely on the random number an example to this could be:
```CSS
#buildoutput > div > div
```
In this example it doesn't rely on the class names that are 'randomized`.

Regards,
Gusted

bjfletcher:
Hello, Gusted! It's really nice to hear from you and thank you every so much for Dark Reader. It's one of the best things to happen to my life!

Regarding CSS selectors, ah yes, that's really tricky as CircleCI seems to use a system where virtually everything is a `div` and CSS selectors are randomised.

It seems that CircleCI changes the selectors quite infrequently - I've been using these for over a month now. I'm happy to monitor to see how frequent they change, if ever.

I wonder if it's worth talking to friendly CircleCI developers about this. It may be that an easier option for CircleCI developers would be to develop a dark mode of their website than changing their HTML & CSS system.

image

Gusted:
> Regarding CSS selectors, ah yes, that's really tricky as CircleCI seems to use a system where virtually everything is a `div` and CSS selectors are randomised.

Ah that kind of site 😄 

> It seems that CircleCI changes the selectors quite infrequently - I've been using these for over a month now. I'm happy to monitor to see how frequent they change, if ever.

Hmm a month, that must be a record for in the books from my knowledge with other services.

> I wonder if it's worth talking to friendly CircleCI developers about this. It may be that an easier option for CircleCI developers would be to develop a dark mode of their website than changing their HTML & CSS system.

Il expect an response that they are using framework x that does this for them, because framework x has feature y. Standard decisions.

Looking at the screenshot their seems be not a really secure/trustworthy method to handle it other than the css class. I will merge it for now.

Regards,
Gusted

PR #4670, opened 2021-01-04

ycchiam:
Fixing the radio selection where it was not visible after inversion.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2240,6 +2240,15 @@ CSS
 
 ================================
 
+doordash.com
+
+CSS
+.fYPzgJ::after {
+    background-color: white;
+}
+
+================================
+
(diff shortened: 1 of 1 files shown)


Gusted:
Hi @ycchiam!

Where can I observe this `radio selection`?

Regards,
Gusted

ycchiam:
> Hi @ycchiam!
> 
> Where can I observe this `radio selection`?
> 
> Regards,
> Gusted

on food menu, after selecting an item, some of them have the option to be customize. The radio button on the list is having the same color with the background. Just replaced the current pull request with #4687, to follow the description conventions over here.  Sorry for the inconveniences.

Gusted:
> to follow the description conventions over here. Sorry for the inconveniences.

We don't currently have any description conventions as 90% I can find where the fix is made for and check it out without needing to ask anything, don't worry.

PR #4685, opened 2021-01-06
Fixed schoolathome.ca

AndrewDaGuy:
some inverting there and there, this is for my school buds so their eyes wont burn off


Gusted (approved):
LGTM! Thanks

PR #4687, opened 2021-01-06
Fix for doordash.com

ycchiam:
Fixing the issue where some of the radio selection being invisible.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2240,6 +2240,13 @@ CSS
 
 ================================
 
+doordash.com
+
+INVERT
+.fYPzgJ::after
+
+================================
+
 dotaunderlords.gamepedia.com
 dota2.gamepedia.com
 
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2243,7 +2243,7 @@ CSS
 doordash.com
 
 INVERT
-.fYPzgJ::after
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -2240,6 +2240,13 @@ CSS
 
 ================================
 
+doordash.com
+
+INVERT
+.fYPzgJ::after
```
Gusted, line 2246: It seems to me that's already inverted by default 
image

ycchiam: image

The check boxes are fine, the issue only occurs on radio selection, where users are only allowed to select one option.
Example in the screenshot is after the code change.

Gusted: Do you have a example Link to a store who has this as well which item it is?

ycchiam: Try this, almost every item on the menu had the radio button.
https://www.doordash.com/store/super-chicken-tysons-17410/en-US

Gusted: Aha thanks!
I already was afright of this and the CSS selector doesn't seem to be working for me due to being it randomized.
An alternative Selector can be:
```CSS
input[kind="RADIO"][id^="Toggle"]::after
```

ycchiam: Ah, thanks for the suggestion. This is much safer way of changing it.
Will update the branch.

Gusted:
Thanks!

PR #4688, opened 2021-01-06
Add MeWe.com custom CSS

Karmalakas:
By default MeWe.com social network `body` and `html` are `100%` height. That makes DarkReader applied styles look good only on the very top of the page, but if you scroll just a bit, there's a black bar and then you can't separate other elements from the background. Setting height to `auto`, fixes this

Untitled

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5261,6 +5261,16 @@ g > rect
 
 ================================
 
+mewe.com
+
+CSS
+html,
+body {
+    height: auto !important;
+}
+
+================================
+
 midkar.com
 
 CSS
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5265,7 +5265,7 @@ mewe.com
 
 CSS
 body > .ember-view {
-    background-color: #2c2d2d !important;
+    background-color: ${#ebf0f3} !important;
 }
 
 ================================
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5264,9 +5264,8 @@ g > rect
 mewe.com
 
 CSS
-html,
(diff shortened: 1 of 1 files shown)


Gusted:
> By default MeWe.com social network `body` and `html` are `100%` height. That makes DarkReader applied styles look good only on the very top of the page, but if you scroll just a bit, there's a black bar and then you can't separate other elements from the background. Setting height to `auto`, fixes this

This a known issue and setting the `height` is a most no-go especially for 2 root elements. It can affect other sites as well that doesn't introduce this issue and has different layouts that will get messed up trough this. A possible fix is to set the body or html to a fixed color that is the same as the other root element.

Regards,
Gusted

Karmalakas:
> A possible fix is to set the body or html to a fixed color that is the same as the other root element.

Setting body or html color won't work, because of their height. Hence the issue. Changed it to `div` color instead. Used fixed color, because none of 4 DR variables do the job - it's either same as all blocks or a light color. If there're other variables not listed in [Dynamic variables section](https://github.com/darkreader/darkreader#dynamic-variables), please let me know

Gusted:
>  Used fixed color, because none of 4 DR variables do the job - it's either same as all blocks or a light color. If there're other variables not listed in [Dynamic variables section](https://github.com/darkreader/darkreader#dynamic-variables), please let me know.

Hmm that's a bit of a problem, but I think to still respects the user's setting(e.g. brightness) it can be changed to something like `${white}` or `${the original color}` which the original color replaced of course. Dark Reader will take care of it to get it modified.

Regards,
Gusted

Karmalakas:
> it can be changed to something like `${white}` or `${the original color}` which the original color replaced of course. Dark Reader will take care of it to get it modified.

Aha.. :) Didn't know that. Took default light `body` color and the result is even better :) And as you said, it's taken into account when changing brightness. Thanks for the tip

Gusted:
Thanks!

PR #4723, opened 2021-01-10
docs: Update README.md for adding a new usage for the react developer.

Turkyden:
## react-darkreader

> A React hook for adding a dark / night mode to your site inspired by darkreader.

Add a new usage for the react developer. [react-darkreader &rarr;](https://react-darkreader.vercel.app/)

In adition to this, looking forward to other framework such as [Vuejs](), [Angular]() ...

@alexanderby @Gusted

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,6 +258,10 @@ const CSS = await collectCSS();
 const isEnabled = isDarkReaderEnabled();
 ```
 
+... or you can also use the darkreader with other modern frontend framework. Please refer to:
+
+- [react-darkreader](https://github.com/Turkyden/react-darkreader)
+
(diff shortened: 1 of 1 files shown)


Gusted:
Hi @Turkyden!

First off, please don't ping us. We will review the PR when we have time.
We currently don't credit any third-party application that uses Dark Reader, not in a application-way nor api-way.
I'm unsure if we should do this, due to the fact that the API is used by a small percentage of users and by the fact that their also alternatives out their and listen them all will be messy and the `better` ones meaning to choose an repo over another.

Regards,
Gusted

Turkyden:
Thank you !

PR #4737, opened 2021-01-11
Add fixes for Amazon India and LibreTexts

Vedanth-P:
On LibreTexts ([libretexts.org](libretexts.org)) articles, there are sometimes images with transparent backgrounds displayed in the body of the article. Sometimes, diagrammatic images with transparent backgrounds become difficult to see on a dark background. This fix makes the backgrounds of images in the body white, so that the transparent-background images can be seen properly.
[An example page:](https://math.libretexts.org/Bookshelves/Applied_Mathematics/Book%3A_Math_in_Society_(Lippman)/06%3A_Graph_Theory/6.02%3A_Graphs) the right-side image above the line "This type of simplified picture is called a graph" is corrected by this fix.
[Another example page:](https://chem.libretexts.org/Bookshelves/Ancillary_Materials/Reference/Organic_Chemistry_Glossary/Resonance_Energy) the image above the line "The difference in energy between X and X3 is the resonance energy of X."

On the Amazon India website (amazon.in), there are places where an SVG graphic is used to depict the Rupee symbol. In some places, notably the shopping cart, the rupee symbol is black; and hence when the page is inverted with Dark Reader it becomes illegible. This fix inverts those elements.
The rupee symbol can be seen on any search results page, where the graphic is used to show the subtotal of the items in the cart near the top-right corner of the page.
In the cart itself, the prices of individual items are accompanied by this graphic
Before fix
After fix


src/config/dynamic-theme-fixes.config
```diff
@@ -470,6 +470,13 @@ CSS
 
 ================================
 
+amazon.in
+
+INVERT
+.currencyINR
```
Gusted, line 476: Move this fix into `amazon.*`

src/config/dynamic-theme-fixes.config
```diff
@@ -4668,6 +4675,15 @@ CSS
 
 ================================
 
+libretexts.org
+
+CSS
+.mt-content-container img {
```
Gusted, line 4681: What about 
```CSS
INVERT
.internal[alt$=".svg"]
```
It should only target the `math` graphs.

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
Vedanth-P: Done

src/config/dynamic-theme-fixes.config
Vedanth-P: I thought it best to target all images in the text body, as there might be images of other formats with transparent backgrounds. The writing [here](https://chem.libretexts.org/Bookshelves/Ancillary_Materials/Reference/Organic_Chemistry_Glossary/Resonance_Energy) has a PNG image with a transparent background above the line "The difference in energy between X and X3 is the resonance energy of X."
And I used the `.mt-content-container` selector to target only the images in the main text body; I was being conservative and didn't want to break anything I wasn't aware of. Is that unnecessary?

Also, the reason I used CSS instead of `INVERT` is that color-coded images may be adversely affected. LibreTexts hosting academic content, I thought that isn't unlikely.

Gusted: `I was being conservative and didn't want to break anything I wasn't aware of. Is that unnecessary?`

The preference goes to INVERT instead of giving users a white background-color.
We encoutered this problem with Wikipedia trough:
```CSS
.mwe-popups-discreet > svg,
div .thumbimage[src$=".png"],
div .thumbimage img[src$=".png"] {
    background-color: white;
}
.mw-mmv-image .svg,
.fullImageLink [src*=".svg"],
a[href$=".svg"]:hover > img,
a[href*=".gif"]:hover > img {
    background-color: rgba(255, 255, 255, 0.75) !important;
    background-blend-mode: color;
}
```
Something similair can be implemented for libretexts to ensure it will not need a white background-color.

Vedanth-P: Do you mean using a light grey background instead of a white background?

Gusted: > Do you mean using a light grey background instead of a white background?

Only while hovering due to some issues we encountered that still can happen on white background.

Vedanth-P: So just to confirm, what you're suggesting is to leave the images unaltered, and to give them a light background only when they are hovered over. Is that correct?

Gusted: That's correct.

Gusted (approved):
Thank you, looks good.

PR #4767, opened 2021-01-13
A patch

### LINKED ISSUE #4766: colour blindness
Thank you very much for this marvellous tool!
Do you somehow support usability for colour blind people? 
Is is possible to selectively change the colour of visited links? I use the darkreader plugin for firefox. However, I cannot change the colour of vistited links directly in the firefox settings, because it will also affect the background colour.

Many thanks for any help.

Gusted:
- Update colors.
- Add user-agent style into shadow-root.
- Resolves #4766

### DIFF
--- a/src/inject/dynamic-theme/index.ts
+++ b/src/inject/dynamic-theme/index.ts
@@ -141,9 +141,15 @@ function createShadowStaticStyleOverrides(root: ShadowRoot) {
     const inlineStyle = createOrUpdateStyle('darkreader--inline', root);
     inlineStyle.textContent = getInlineOverrideStyle();
     root.insertBefore(inlineStyle, root.firstChild);
+
+    const userAgentStyle = createOrUpdateStyle('darkreader--user-agent');
+    userAgentStyle.textContent = getModifiedUserAgentStyle(filter, isIFrame, filter.styleSystemControls);
+    root.insertBefore(userAgentStyle, inlineStyle.nextSibling);
+
     const overrideStyle = createOrUpdateStyle('darkreader--override', root);
     overrideStyle.textContent = fixes && fixes.css ? replaceCSSTemplates(fixes.css) : '';
-    root.insertBefore(overrideStyle, inlineStyle.nextSibling);
+    root.insertBefore(overrideStyle, userAgentStyle.nextSibling);
+
     shadowRootsWithOverrides.add(root);
 }
 
@@ -466,6 +472,7 @@ export function removeDynamicTheme() {
     }
     shadowRootsWithOverrides.forEach((root) => {
         removeNode(root.querySelector('.darkreader--inline'));
+        removeNode(root.querySelector('.darkreader--user-agent'));
         removeNode(root.querySelector('.darkreader--override'));
     });
     shadowRootsWithOverrides.clear();
--- a/src/inject/dynamic-theme/modify-css.ts
+++ b/src/inject/dynamic-theme/modify-css.ts
@@ -72,9 +72,6 @@ export function getModifiedUserAgentStyle(theme: Theme, isIFrame: boolean, style
     lines.push('a {');
     lines.push(`    color: ${modifyForegroundColor({r: 0, g: 0, b: 238}, theme)};`);
     lines.push('}');
-    lines.push('a:visited {');
-    lines.push(`    color: ${modifyForegroundColor({r: 85, g: 26, b: 139}, theme)};`);
-    lines.push('}');
     lines.push('table {');
     lines.push(`    border-color: ${modifyBorderColor({r: 128, g: 128, b: 128}, theme)};`);
     lines.push('}');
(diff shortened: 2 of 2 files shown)


src/inject/dynamic-theme/index.ts
```diff
@@ -141,9 +141,15 @@ function createShadowStaticStyleOverrides(root: ShadowRoot) {
     const inlineStyle = createOrUpdateStyle('darkreader--inline', root);
     inlineStyle.textContent = getInlineOverrideStyle();
     root.insertBefore(inlineStyle, root.firstChild);
+
+    const userAgentStyle = createOrUpdateStyle('darkreader--user-agent');
+    userAgentStyle.textContent = getModifiedUserAgentStyle(filter, isIFrame, filter.styleSystemControls);
```
alexanderby, line 146: Maybe let's cache the text content of user agent style some way? Some pages can have many custom elements.

src/inject/dynamic-theme/modify-css.ts
```diff
@@ -70,7 +70,7 @@ export function getModifiedUserAgentStyle(theme: Theme, isIFrame: boolean, style
     lines.push(`    color: ${modifyForegroundColor({r: 0, g: 0, b: 0}, theme)};`);
     lines.push('}');
     lines.push('a {');
-    lines.push(`    color: ${modifyForegroundColor({r: 0, g: 64, b: 255}, theme)};`);
+    lines.push(`    color: ${modifyForegroundColor({r: 0, g: 0, b: 238}, theme)};`);
```
alexanderby, line 73: Why this color is changed?

alexanderby (changes requested):

src/inject/dynamic-theme/modify-css.ts
Gusted: The original one introduce green, which is weird.
So looking at [Firefox](https://searchfox.org/mozilla-central/source/modules/libpref/init/StaticPrefList.yaml#1165)'s color for this. It hasn't any green same story for [Chromium](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/dom/text_link_colors.cc;l=42). Was their any reason for the custom color to make it more appropriate blue?

alexanderby: There is a problem with pure blue (240 degrees in HSL). In light mode the difference is barely visible, but in dark mode it looks too dark. Increasing the lightness makes it look violet like a visited link. So the slight hue change to 220 degrees makes it look blue and more contrast. If you have a look at native dark themes for some websites, designers usually slide the hue a little bit, to make it more bright. There's also a compensation in the formula, that slides 240 degrees to 220.
Screen Shot 2021-01-20 at 20 34 08
This way we get more contrast for blue links: your suggestion becomes 220 deg, the previous value becomes 212 deg in HSL. This was an irritating issue in the past.
Screen Shot 2021-01-20 at 20 46 31

alexanderby:
I don't know if this PR actually solves a problem for color blindness. As far as I understand you fix a problem for Shadow DOM. To fix a problem for visited links, use [:visited](https://developer.mozilla.org/en-US/docs/Web/CSS/:visited) pseudo-class. Previously I tried using `a:visited`, but it broke styles for some colors, I don't remember where, so try using just `:visited` selector.

alexanderby:
I've tried to use `:visited` selector. The problem is that pseudo-class selector has higher specificity than tag selector, and many links become violet, what website authors usually don't expect it. So I think the solution is to add more color options for links and visited links, as we did for scrollbars. Auto should not override anything, default should use styles for "a" and "a:visited".

src/inject/dynamic-theme/modify-css.ts
Gusted: Got it 👍🏽

Gusted:
> but it broke styles for some colors, I don't remember where, so try using just `:visited` selector.

Well, the problem is, that it's static. And some sites e.g. GitHub specific override the visited behaviour to not style links like repo's.
image
However, adding this into it I realize it will be styled and not expected behaviour, that's why I had it removed https://github.com/darkreader/darkreader/pull/4767/commits/1c83f0fac119ef612a0a332c27ccbb983a42f081

Also I can pull some browser-specific selectors to get all visited links.


>  So I think the solution is to add more color options for links and visited links, as we did for scrollbars. Auto should not override anything, default should use styles for "a" and "a:visited".

Will do 👍🏽.

Gusted:
Succeeded by #5038

src/inject/dynamic-theme/index.ts
saifalsowaidi: it could be found in another book in page #1772569

hope you can find your answer

`
```suggestion
    userAgentStyle.textContent = getModifiedUserAgentStyle(filter, isIFrame, filter.styleSystemControls);
    
```

saifalsowaidi:
thnx

saifalsowaidi (commented):
g

PR #4772, opened 2021-01-14
Fix for sports.ru

Rakleed:
- Invert logo.

It would have been more correct to use "IGNORE IMAGE ANALYSIS", but it didn't work for me.


Gusted:
> It would have been more correct to use "IGNORE IMAGE ANALYSIS", but it didn't work for me.

It is a `SVG` with a closed shadow-root, which somehow is getting styled by Dark Reader.

PR #3255, opened 2020-07-27
Style Standard Elements

Gusted:
- Add option to style standard elements e.g. button, input, select.
- Resolves #734
- Resolves #3239

Issue #3239, opened 2020-07-26
[BUG] Text boxes and drop-down menus look like Win98 style when addon enabled


romhacks:
### **Describe the bug**

When the dark reader add-on is enabled, drop-down menus and text boxes that have not been stylized by the website change from their normal, rounded look that is the OS default, to squared ones that look to be from Windows 98. See screenshots, the website I used as an example was the AUR page for anaconda.

Add-on disabled:
![Screenshot_20200726_104031](https://user-images.githubusercontent.com/42524580/88482113-3b160c00-cf2d-11ea-82d3-bac2b20bc637.png)

Add-on enabled:
![Screenshot_20200726_104048](https://user-images.githubusercontent.com/42524580/88482119-4b2deb80-cf2d-11ea-86bd-dd7e13824889.png)

You can see how the Keywords text box and the Search By drop-down are changed to look different from how they do originally.

### **Steps to reproduce**

- Go to any website with unstylized drop-down menus or text boxes (I used [AUR's page on anaconda](https://aur.archlinux.org/packages/anaconda/) as an example)
- Enable/Disable Dark reader and look at the difference in the boxes

### **Expected behavior**

The drop-down menus and text boxes should stay the same shape/style, just darker.

### **Actual behavior**

The drop-down menus and text boxes are replaced with differently shaped/styled ones that look outdated.

### **Screenshots**

Screenshots are in the first section.

### **System Information:**
 - OS: Arch Linux with KDE Plasma 5.19.3
 - Browser: Firefox Nightly 80.0a1 (2020-07-25) (though I can confirm the bug exists in Firefox Stable, too.)
 - Dark Reader version: 4.9.15

Gusted:
This is a OS thingy. As the drop down menu's etc are being fetched for each OS different. But we don't touch it just the colors and thst shouldn't affect if it's rounded. Could you make a screenshot of the styles that has been applied to the element trough devtools.

romhacks:
You mean to screenshot DevTools when I have selected the text box in it? (sorry my english is not perfect)

Gusted:
The element itself so the `rounded element`

romhacks:
Apologies if this is not what you are looking for.
DevTools of the text box:
![Screenshot_20200726_111652](https://user-images.githubusercontent.com/42524580/88482830-c85b5f80-cf31-11ea-8968-d36bbed678fa.png)
DevTools of the drop-down menu:
[...]

Issue #734, opened 2018-09-18
Don't style form inputs on Firefox


ascclemens:
Setting `border-color` or `background-color` on any `button` or `input` causes them to stop looking like native components and instead makes them look like some ancient OS.

What buttons look like in Firefox when styled by Dark Reader: ![oh god](https://i.imgur.com/vMgcVud.png)

What they should look like: ![please](https://i.imgur.com/8XdyEg5.png)

Chrome ignores this, but Firefox makes a non-native button instead. Not styling form inputs, or having an option for it, would help make the browser look more integrated with the OS.

rolagg:
> perhaps only on macOS
Windows 10, happens both on Chrome and Firefox.

TheMatten:
Affected on Arch Linux with Gnome 3.30.1 in Firefox too.

PR #4782, opened 2021-01-15
Fix for dark themes

Gitoffthelawn:
This is a fix that needs to be carefully considered before accepting, as it affects global elements.

I have found this fix required when using dark themes in the Windows OS and using Firefox.  Without this fix, HTML input, select, and textarea fields will sometimes be impossible to read, as they will be white on white.

I have been running with this fix for several months without any issues (and with it working successfully to resolve the aforementioned issue), but I have not performed any testing outside of my environments.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -6,6 +6,9 @@ INVERT
 span#closed_text > img[src^="https://www.gstatic.com/images/branding/googlelogo"]
 
 CSS
+input, select, textarea {
+    color: var(--darkreader-selection-text) !important;
+}
 .vimvixen-hint {
     background-color: ${#ffd76e} !important;
     border-color: ${#c59d00} !important;


Gusted:
Hi @Gitoffthelawn!

This is indeed a fix that needs to be looked at carefully, this is a _known_ issue(#734 #3239) Their had been succes by fixing this issue, by explicitly not styling those elements inside Dark Reader(#3255). Did you try this PR out?

    Go to theme page.
    Go to "Font & more".
    Disable Style System Controls.
   
Regards,
Gusted

Gitoffthelawn:
Hello @Gusted!

I had played with that setting a few months ago, but I did not understand fully what it really was supposed to do.  IIRC, I didn't see any immediate differences, so I just set it back to the default.  I only tried it on a few sites at the time, and it's very possible (probable, in fact!) that none of the sites I visited had `input`, `select`, or `textarea` elements.

Is the a good documentation page for that setting, or if not, can you sum up what it really is supposed to do?

Gusted:
Hi @Gitoffthelawn!

The setting disable the static styling of those elements. https://github.com/darkreader/darkreader/blob/master/src/inject/dynamic-theme/modify-css.ts#L65 however if a site custom style those elements, dark reader will of course modify them. However I see that most sites don't style the `color` property and can result in the `initial` color black and making it unreadable. 
If we want to make this change it probably should respect this setting and moving it to the code itself and 'injecting' it into the `*`.

Regards,
Gusted

Gusted:
AFAIK a good case where this styleSystemControl can be observed https://searchfox.org

PR #4783, opened 2021-01-15
Fix dark filter (inversion) mode

Gitoffthelawn:
This is another fix that needs to be carefully considered before accepting, as it affects global elements.

I have found this fix required when using dark themes in the Windows OS and using Firefox.  Without this fix, HTML input, select, and textarea fields will sometimes be impossible to read, as they will be white on white.

### DIFF
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -27,6 +27,9 @@ REMOVE BG
 .compatibility-with-darkreader-below-4-3-3
 
 CSS
+input, select, textarea {
+    color: var(--darkreader-selection-text);
+}
 .compatibility-with-darkreader-below-4-3-3 {
     background: white !important;
 }


Gusted:
Hi @Gitoffthelawn!

Unfortunately their are 2 problems with this.
1. The variable doesn't exist in Filter(+).
2. I've checked on Linux with Firefox and remembered that those elements are fully styled by User-Agent and so far I see is not touched in Filter(+).

Regards,
Gusted

Gitoffthelawn:
Hello @Gusted!

1.  Ah, that *is* an issue.  I rarely use Filter+, so that would make sense why I've never encountered the issue.  I literally only use Filter+ on 1 site at this time.  What could be a solution?
2. I've heard the expression "fully styled by User-Agent" before, and I've never 100% understood exactly what that means.  Every time I think I understand it, I find a case that breaks the rule.  What does that really mean?  Also, does it apply to Firefox on Windows?

You might be the perfect person to talk with more about dark theming the viewport within Firefox.  When I'm also using a dark Windows OS theme, I find many rendering issues, almost all of them revolving around the `html`, `body`, `input`, `select`, and `textarea` HTML elements.  I've developed a whole bunch of workarounds, but at this point many of my workarounds have workarounds because there are so many cases!

If this conversation would interest you, perhaps it would be good for me to open an issue about it so that others will be more likely to see it and perhaps contribute.  What do you think?

Gusted:
> If this conversation would interest you, perhaps it would be good for me to open an issue about it so that others will be more likely to see it and perhaps contribute. What do you think?

Feel free the to open up an issue.

>     1. Ah, that _is_ an issue.  I rarely use Filter+, so that would make sense why I've never encountered the issue.  I literally only use Filter+ on 1 site at this time.  What could be a solution?

Not sure, a static color should be appropriate for this case.

>     2\. I've heard the expression "fully styled by User-Agent" before, and I've never 100% understood exactly what that means.  Every time I think I understand it, I find a case that breaks the rule.  What does that really mean?  Also, does it apply to Firefox on Windows?

It is a weird expression. So assuming you know what [a user agent](https://developer.mozilla.org/en-US/docs/Glossary/user_agent) is. Let's take for example Firefox, it has to style the HTML elements. Those elements include `html`, `body`, `input`, `select`, and `textarea`. However due to styling styles by different operating system's(OS's) those elements can differ how they look between OS's. Example:
For a Linux machine an input looks like this 
image
It has it's unique round border whereby(IIRC) Windows has no round borders, but an outline and so has every OS their own unique twist whereby even certain values can be manipulated on OS level. So if you hear the expression "Style by the User-Agent" it means in other words "It's styled by the browser based on which OS it's running" and even thereby has the browser also a certain input, an element can look different on the same OS between browsers.

Regard,
Gusted

Gitoffthelawn:
@Gusted wrote:
> So if you hear the expression "Style by the User-Agent" it means in other words "It's styled by the browser based on which OS it's running" and even thereby has the browser also a certain input, an element can look different on the same OS between browsers.

Thanks Gusted.  It's the use of the word *"fully"* that confuses me.  For example, using a custom dark theme for Windows, drop-down input fields look great within the OS and native non-browser software.  But in Firefox, such controls sometimes become illegible (light on light or dark on dark).  I want the browser to fully style them, but sites override the browser with colors that cause illegibility.

I have hoped that the web would eventually become OS-agnostic so that a website would render identically regardless of the OS.  This would make coding, testing, debugging, and documentation all more efficient.

A few questions for you, if you have time:

1. Do you know if there is a way to tell Firefox "ignore as much as possible about the underlying OS", and just have it focus on whether the user wants a dark or light theme?
2. Alternatively, is there a way to instruct Firefox to present websites as they would a specific OS?  For example, use Linux-like controls regardless of the OS.
3. Is there a way to tell Firefox to *always* style HTML elements like `select` and `input`?
4. Is there a way to tell Dark Reader to *always* style HTML elements like `select` and `input`?

As a bit of an aside, it's interesting that web browsers even try to emulate subtle OS characteristics, as the browsers typically render all elements manually.  IOW, even though a drop-down list may look "Windowsish", it's only because Firefox is using CSS to make `select` elements look close to what Windows provides by default.  There are so many custom controls on websites these days, that it seems like a better idea to just have standard "web controls" instead of browsers creating OS "lookalike" controls.

Thinking aloud, I wonder if it makes any sense to have Dark Reader include its own CSS for controls (`select`, et al.) that work wonderfully on any OS with either dark or light themes.  I mention this because in order to workaround many of the issues I've uncovered dark theming different websites, I have had to create my own `select` controls for drop-downs, checkboxes, radio buttons, et al.  When a site has issues with being dark themed, I'll apply my custom controls, and often that results in a good looking dark theme.

Gusted:
>     1. Do you know if there is a way to tell Firefox "ignore as much as possible about the underlying OS", and just have it focus on whether the user wants a dark or light theme?

Nope. Why? -> Answer 2.

>     2\. Alternatively, is there a way to instruct Firefox to present websites as they would a specific OS?  For example, use Linux-like controls regardless of the OS.

Not really. I had a look at some of the code responsible but they don't have an `override` mechanism or something.



>     4\. Is there a way to tell Dark Reader to _always_ style HTML elements like `select` and `input`?

Enabled by default.

>     3. Is there a way to tell Firefox to _always_ style HTML elements like `select` and `input`?

Seems to be always. [This file](https://searchfox.org/mozilla-central/source/layout/style/res/forms.css) is always injected as User-Agent style. 



> As a bit of an aside, it's interesting that web browsers even try to emulate subtle OS characteristics, as the browsers typically render all elements manually. IOW, even though a drop-down list may look "Windowsish", it's only because Firefox is using CSS to make `select` elements look close to what Windows provides by default. There are so many custom controls on websites these days, that it seems like a better idea to just have standard "web controls" instead of browsers creating OS "lookalike" controls.

Pretty sure they want to follow the OS's unwritten style guides, so for Users it will feel like any other application.
For more info how/why they style certain things the [`appearance` CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/appearance) is what you're looking for.



> Thinking aloud, I wonder if it makes any sense to have Dark Reader include its own CSS for controls (`select`, et al.) that work wonderfully on any OS with either dark or light themes.

It is possible and seeing a lot of issues with them it should be a good idea. Probably I will stick to the Linux/Mac styles of how certain styles look like.

Regards,
Gusted

Gusted:
Btw, I think we should convert it into an issue.

PR #4807, opened 2021-01-19
Added fix for app.kognity.com

AnthonyWritesBadCode:
This website requires a paid login, as a result it may not be possible to visually inspect the changes. Instead, here are some screenshots.

First, without darkreader enabled:
image

Without any additional styling, the page has a greenish background color and it's really difficult to view graphs:
image

The proposed code makes all images appear on white backgrounds so that they are more easily visible and it corrects the background color:
image

Inverting the images looks great with graphs and diagrams but also causes normal images to be inverted. There is no class/id to distinguish between graphs and photos:
image
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -617,6 +617,23 @@ CSS
 
 ================================
 
+app.kognity.com
+
+INVERT
+img.KogCalculator
+// .content-image-figure > img
+
+CSS
+body, .KogDashboard-insideLoader {
+    background: none var(--darkreader-neutral-background) !important;
+
+}
+img {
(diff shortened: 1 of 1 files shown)


AnthonyWritesBadCode (commented):
Removed a comment

AnthonyWritesBadCode (commented):
Fixed images so that pngs are inverted and all other image types are displayed on a white background

Graphs:
image

Photos:
image

AnthonyWritesBadCode (commented):
Fixed SVGs being displayed on a black background instead of just being inverted

AnthonyWritesBadCode:
Forgot to mention this earlier:
The SVG icon for questions where calculators are allowed is inverted
image

PR #4813, opened 2021-01-19
darkmode sidebar fix for testudo.umd.edu

benjaminnow:


Gusted (changes requested):
Append `!important;` to the CSS properties.

Gusted (approved):

PR #4836, opened 2021-01-21
Updated app.kognity.com fix

AnthonyWritesBadCode:
When viewing a topic overview, the modal, the subtopic items, the "Previous Subtopic" button and the "Next Subtopic" button all had weird greenish colors.
This website required a paid login, so screenshots are provided instead.

Previously:
image

Fixed
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -633,12 +633,15 @@ img.KogCalculator
 .content-image-figure > img[src*="png"]
 
 CSS
-body, .KogDashboard-insideLoader {
+body, .KogDashboard-insideLoader, .subtopic-modal-body, .subtopic-modal-footer {
     background: none var(--darkreader-neutral-background) !important;
 
 }
 img:not([src*="png"]):not([src*="svg"]) {
     background-color: white !important;
+}
+.ng-scope > .final-node-level:hover, button.KogButton:hover {
+    background-color: #1B1D1E !important;
 }
 
 ================================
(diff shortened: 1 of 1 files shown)


AnthonyWritesBadCode (commented):
Fixed another greenish area

AnthonyWritesBadCode:
Fixed another greenish area on a popover:

Before:
image

After:
image

Gusted:
Hi @AnthonyWritesBadCode!

LGTM, however it breaks in light mode.

Regards,
Gusted

AnthonyWritesBadCode:
Hi @Gusted

I could be wrong, but aren't there various sites that don't perform great in light mode? Even this page has some parts of the text that aren't easily readable.
Also, if you don't mind me asking, what is the point of the light mode?
Finally, is there a fix for this, like maybe being able to write different code for dark and light modes?

Thanks,
Anthony

Gusted:
> I could be wrong, but aren't there various sites that don't perform great in light mode?

That shouldn't be the case and should be fixed.

> Also, if you don't mind me asking, what is the point of the light mode?

People use it, a lot of people actually do. Personally I don't use it. Only in cases when it was daylight and didn't really needed dark mode and let my eyes rest from dark mode.



> Finally, is there a fix for this, like maybe being able to write different code for dark and light modes?

1. Use `${color}` which will be modified/inverted trough Dark Reader and only applied in dark mode.
2. Use the variables. https://github.com/darkreader/darkreader#dynamic-variables

Regards,
Gusted

> Edit: Sorry if this came across as rude. That definitely wasn't my intention

No problem.

PR #4837, opened 2021-01-21
fix for gitlab.com

linsui:
Fix the comment action icons.

sshot-001


linsui:
Oh, wait, this revert too much.

PR #4843, opened 2021-01-21
Fix unreliable startup on thunderbird

Gusted:
- Ensure settings are being set.

@alexanderby This should fix that problem you were talking about. It indeed seems to be an issue with the storage. Thanks for the tip!

### DIFF
--- a/src/background/utils/extension-api.ts
+++ b/src/background/utils/extension-api.ts
@@ -1,5 +1,7 @@
 import {isPDF} from '../../utils/url';
 import {isFirefox, isEdge} from '../../utils/platform';
+import {DEFAULT_SETTINGS} from '../../defaults';
+import {logWarn} from '../../inject/utils/log';
 
 declare const browser: {
     commands: {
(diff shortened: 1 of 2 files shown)


Gusted:
[build-thunderbird.zip](https://github.com/darkreader/darkreader/files/5851666/build-thunderbird.zip)

src/background/utils/extension-api.ts
```diff
@@ -40,16 +40,24 @@ export function canInjectScript(url: string) {
 let isWriting = false;
 
 export async function readSyncStorage<T extends {[key: string]: any}>(defaults: T): Promise<T> {
-    return new Promise<T>((resolve) => {
+    return new Promise<T>((resolve, reject) => {
         chrome.storage.sync.get(defaults, (sync: T) => {
+            if (chrome.runtime.lastError) {
```
alexanderby, line 45: What if there will be `chrome.runtime.lastError`? It will throw an exception and the extension won't start.

Maybe log an error and return defaults?

src/background/utils/extension-api.ts
```diff
@@ -40,16 +40,24 @@ export function canInjectScript(url: string) {
...
             resolve(sync);
         });
     });
 }
 
 export async function readLocalStorage<T extends {[key: string]: any}>(defaults: T): Promise<T> {
-    return new Promise<T>((resolve) => {
+    return new Promise<T>((resolve, reject) => {
         chrome.storage.local.get(defaults, (local: T) => {
+            if (chrome.runtime.lastError) {
```
alexanderby, line 57: Same here.

alexanderby (changes requested):

alexanderby (approved):

Gusted:
Lets not push a new release until further confirmation or the issue you talked about also stopped for you.

ElhemEnohpi:
I installed from build-thunderbird.zip above, and it seems to work fine for me. I've also been using the earlier builds, and have never encountered a problem where they didn't start up.

alexanderby:
I will push an update then.

PR #4844, opened 2021-01-22
add fix for rule34.xxx

CesiumCs:
### Heads up!
this fix is for a site that is not safe for work.

their header uses a 15px green image to color it??? why on earth is beyond me but simply removing it seems to fix it
also changed the navbar so it wouldnt be a dark green, but match the rest of the background instead


AndrewDaGuy:
I'm pretty sure NSFW websites won't get fixed/merged :/

PR #4845, opened 2021-01-22
Fix for reddit

SmaltParrot50:
Fixed annoyance where when using dark reader with the native reddit light theme would cause some visual disturbances.
Note: This time with the proper updated dynamic-theme-fixes.config, not sure how to delete or edit my other post

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -6744,10 +6744,15 @@ IGNORE IMAGE ANALYSIS
 
 reddit.com
 
-INVERT
-#COIN_PURCHASE_DROPDOWN_ID > div > span
-
 CSS
+#SHORTCUT_FOCUSABLE_DIV > div:nth-child(3) > header > div > div._3dnbqz69WJTFCss8wl7Wlk > a > svg._1bWuGs_1sq4Pqy099x_yy- > g {
+    fill: ${black} !important;
+}
+
+#COIN_PURCHASE_DROPDOWN_ID > div{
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -6744,10 +6744,15 @@ IGNORE IMAGE ANALYSIS
 
 reddit.com
 
-INVERT
-#COIN_PURCHASE_DROPDOWN_ID > div > span
-
 CSS
+#SHORTCUT_FOCUSABLE_DIV > div:nth-child(3) > header > div > div._3dnbqz69WJTFCss8wl7Wlk > a > svg._1bWuGs_1sq4Pqy099x_yy- > g {
```
Gusted, line 6748: `div._3dnbqz69WJTFCss8wl7Wlk`
`svg._1bWuGs_1sq4Pqy099x_yy-`
Needs to be replaced, with something else. The `_[...]` is random generated.

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
SqueezeingDynamicProcess: It seems consistent for me across multiple webbrowsers and login-states. It doesn't change for me at all, maybe it's region depended?

Gusted: It's generated on production build by the front-end software Reddit uses. If they release an new 'production' build to reddit.com the CSS could break as the CSS is then invalid.

PR #4860, opened 2021-01-23
Add engine + mode into html attribute

### LINKED ISSUE #4835: [Feature Request] Implement Darkreader support via CSS body class?
I'm a full time webdeveloper and I use Darkreader almost everywhere, even on my own sites while developing them and I love it.
There are however minor css issues that pop up that I could fix for myself locally in the extension, but I'd rather fix it for every visitor who uses darkreader on these sites.

Would it be possible for the extension to just add something like a "darkreader-activated" body class to pages? That way I can easily hook in for users that use this extension and fix my sites up properly for them, because my bosses probably won't give me the time or budget to do a full manual darkmode..

Gusted:
- Add attribute into `HTML` element to specify engine + mode.
- This enables to specify the mode inside the dynamic fixes to solve certain problems.
- Let's users solve certain dark-mode CSS problems for their sites locally.
- Resolves #4835


KaKi87:
From what I'm seeing, the attribute key is dynamic, why ? That makes it difficult to select in CSS

Gusted:
> From what I'm seeing, the attribute key is dynamic, why ? That makes it difficult to select in CSS

I think I've made a little mistake in the code, it should be `data-darkreader="'Engine'-'Filter'"` so it could become `data-darkreader="dynamic-dark"` and could easily be targeted with CSS.

KaKi87:
I think the JS misses the dash between engine and mode, and perhaps you forgot to update the test suite ^^

alexanderby (changes requested):
I would suggest to use two attributes: `data-darkreader-mode` (dynamic, static, filter) and `data-darkreader-scheme` (dark, dimmed). So it will be possible to use it in sites fixes (e.g. `[data-darkreader-scheme="dark"] body { color: black; }`) or maybe it will be simpler for those, who would like to use it somehow in their website's stylesheets.

src/inject/dynamic-theme/index.ts
```diff
@@ -406,6 +406,8 @@ export function createOrUpdateDynamicTheme(filterConfig: FilterConfig, dynamicTh
         if (isAnotherDarkReaderInstanceActive()) {
             return;
         }
+        document.documentElement.setAttribute('data-darkreader-mode', filter.engine);
```
alexanderby, line 409: `filter.engine` will return `dynamicTheme`, which is a bit odd, so let's use just `'dynamic'`.

src/inject/index.ts
```diff
@@ -11,14 +11,14 @@ function onMessage({type, data}) {
         case 'add-static-theme': {
             const css = data;
             removeDynamicTheme();
-            createOrUpdateStyle(css);
+            createOrUpdateStyle(css, 'static');
```
alexanderby, line 14: `case 'add-css-filter':` should use `'filter'` too, so you will have to split the condition.

alexanderby (changes requested):
I don't remember if I mentioned, when 5.0 will be about to be released, "engine" will become "mode" with "dynamic", "static" and "filter" values ("svgFilter" as a filter option); "mode" will become "scheme" with "dark" and "dimmed" values (or maybe "light", but people often get confused when it's light mode or when the extension is off).

Gusted:
> I don't remember if I mentioned, when 5.0 will be about to be released, "engine" will become "mode" with "dynamic", "static" and "filter" values ("svgFilter" as a filter option); "mode" will become "scheme" with "dark" and "dimmed" values (or maybe "light", but people often get confused when it's light mode or when the extension is off).

That sounds a lot more logical that naming. Just like renaming the current `Filter` variable within the source code to `theme` so it will be separated.

Thereby dimmed is also pretty neat as with the current settings it will be more dimmed than when it's off.

PR #4861, opened 2021-01-23
Update dark-sites.config

alfawal:
Added random dark-themed sites

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -73,13 +73,16 @@ blackhatworld.com
 blackle.com
 blacklemag.com
 blackmod.net
+blastpremier.com
 blink.sh
 blizzard.com
 block.wingysam.xyz
 blog.aractus.com
+blog.counter-strike.net
 blog.dota2.com
 blog.ja-ke.tech
 blox.link
+blueprintjs.com
 bogleech.com
 bondoer.fr
 bonsaihd.live
@@ -87,6 +90,7 @@ botboy.snaz.in
 bouncingdvdlogo.com
 breaksforeyes.app
 brew.sh
+bri.io
 broadcasthe.net
 browserslist.vercel.app
 brunotome.dev
@@ -213,6 +217,7 @@ ellie-app.com
 elybeatmaker.com
 emanuelemilella.com
 emeraldchat.com/app
+emkc.org
 emuparadise.me
 energized.pro
 enlightenment.org
@@ -234,6 +239,7 @@ femto.dev
 femto.pw
 filterblade.xyz
 find.lolpros.gg
+fireship.io
 fisp.io
 flixxo.com
 flooxer.com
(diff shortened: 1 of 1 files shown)


src/config/dark-sites.config
```diff
@@ -73,20 +73,24 @@ blackhatworld.com
...
+blastpremier.com
 blink.sh
 blizzard.com
 block.wingysam.xyz
 blog.aractus.com
+blog.counter-strike.net
 blog.dota2.com
 blog.ja-ke.tech
 blox.link
+blueprintjs.com
```
Gusted, line 85: https://blueprintjs.com/docs/

src/config/dark-sites.config
```diff
@@ -313,6 +319,7 @@ hellofromhe.re
 hermes.ksdfg.me
 hidive.com
 himovies.to
+hltv.org
```
Gusted, line 322: image

src/config/dark-sites.config
```diff
@@ -398,6 +406,7 @@ marekmaskarinec.github.io
 maroon.jonah.pw
 marte.dev
 matteotiscia.com
+mau.fi
```
Gusted, line 409: image

Gusted (changes requested):

src/config/dark-sites.config
alfawal: They have a dark theme option, I thought it was applied by default since I tried visiting the site from different browsers.

src/config/dark-sites.config
alfawal: Their dark theme is based on the time zone, I'm not sure if you guys count this or not.
2021-01-24 09_44_32-Window

src/config/dark-sites.config
alfawal: This one is weird, tried visiting the site from 5 other browsers and it was always dark themed!
Turns out that it depends on your browser's appearance option, if it was on dark then it should been displayed as dark themed site.

src/config/dark-sites.config
Gusted: > They have a dark theme option, I thought it was applied by default since I tried visiting the site from different browsers.

They don't, I had `privacy.resistFingerprinting` to `true` to prefer a light-colour-scheme and will automatically prefer light colour scheme.

src/config/dark-sites.config
Gusted: We don't count that.

src/config/dark-sites.config
Gusted: > Turns out that it depends on your browser's appearance option, if it was on dark then it should been displayed as dark themed site.

And the sites inside dark-sites.config needs to be by default regardless of appearance option have a dark theme.

Gusted (changes requested):
Remove those 3 sites.

PR #4865, opened 2021-01-24
Add Site to dark-sites.config

SethFalco:
Just adds a website which is dark mode by default to the dark-sites.config.


SethFalco:
Closed because it uses system theme, so now all pages are dark by default for users that prefer light theme on their system.

PR #4902, opened 2021-01-28
Fix transparent image backgrounds for canvas.usask.ca in dark mode

smac89:
Questions with transparent images and dark text were really difficult to read in dark mode.

This makes the background of the image lighter so that the dark text is readable

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1380,6 +1380,15 @@ IGNORE IMAGE ANALYSIS
 
 ================================
 
+canvas.usask.ca
+
+CSS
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -1380,6 +1380,15 @@ IGNORE IMAGE ANALYSIS
 
 ================================
 
+canvas.usask.ca
+
+CSS
+#questions .text img {
+    background-color: currentColor;
```
Gusted, line 1387: Add `!important;`

Gusted (changes requested):

PR #4967, opened 2021-02-03
Update dark-sites.config

gansel51:
Adding ball systems to dark-sites.config


Gusted:
Loading page is white:
image
And if an user has `privacy.resistFingerprinting` enabled in Firefox it will blatantly give a white screen.

gansel51 (commented):
@Gusted once you click in the whole site is dark. I have problems using darkreader on it after that initial page.

PR #4969, opened 2021-02-03
Npm v7

Gusted:


Gusted:
https://github.blog/2021-02-02-npm-7-is-now-generally-available/

```
Please note, npm 7 is now published as latest to the npm registry and will be the default version installed 
when you run npm install --global npm. If you want to install npm 6, please run npm install --global npm@6.
```
And it so happens that it someone got updated automatically. Merging so I don't have to discard the package-lock.json whole time.

PR #5003, opened 2021-02-10
Added some jira cloud styling

samuelgfeller:
Been using these changes for a while, and they help a lot for clarity while we're waiting for an [official dark theme](https://jira.atlassian.com/browse/JRACLOUD-63150)

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -848,9 +848,37 @@ body, .body, table, tbody, tr, td, .corner_bottom {
 
 atlassian.net
 
+INVERT
+img[src*="https://github.githubassets.com/favicon.ico"].smart-link-icon
+.sc-jTzLTM /* Pre-loading comment avatar */
+
 CSS
 #jira-issue-header, #jira-issue-header-actions {
     background-color: transparent !important;
+}
+span.code /* Inline code span */ {
+    background-color: rgba(240,246,252,0.15) !important; 
+}
+.ghx-parent-group, /* Kanban board parent issue */
+.ghx-issue /* Kanban board issue*/
+{
+/* Important needed for .ghx-issue*/
+    background-color: rgba(119, 183, 255, 0.05) !important; 
+}
+span.loader-wrapper a, /* Smart links */
+.css-1ruu48h, /* Normal issue action button below title */
+.css-tmtnt9, /* Button attached to arrow down btn */
+.css-14v1uf5, /* Arrow down btn */ 
+.css-wvfva4 /* Three dots btn */
+{
+    background-color: rgba(240,246,252,0.1); 
+}
+.css-1yfnrso, /* Task in modal box */
+.aui-flag, .aui-message /* Flash message */ {
+    box-shadow: 0 0 20px rgba(240,246,252,0.15) !important;
+}
+.css-1ua1xqz /* Kanban background when issue open in modal */ {
+    background-color: rgba(65, 65, 83, 0.6);
 }
 
 ================================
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -848,10 +848,38 @@ body, .body, table, tbody, tr, td, .corner_bottom {
...
+{
+/* Important needed for .ghx-issue*/
+    background-color: rgba(119, 183, 255, 0.05) !important; 
+}
+span.loader-wrapper a, /* Smart links */
+.css-1ruu48h, /* Normal issue action button below title */
+.css-tmtnt9, /* Button attached to arrow down btn */
+.css-14v1uf5, /* Arrow down btn */ 
+.css-wvfva4 /* Three dots btn */
+{
```
Gusted, line 873: Move the `{` on the same line after `css-wvfva4`

src/config/dynamic-theme-fixes.config
```diff
@@ -848,10 +848,38 @@ body, .body, table, tbody, tr, td, .corner_bottom {
...
+
 CSS
 #jira-issue-header, #jira-issue-header-actions {
     background-color: transparent !important;
 }
+span.code /* Inline code span */ {
+    background-color: rgba(240,246,252,0.15) !important; 
+}
+.ghx-parent-group, /* Kanban board parent issue */
+.ghx-issue /* Kanban board issue*/
```
Gusted, line 863: Move the `{` on the same line after `ghx-issue`

Gusted (changes requested):
LGTM.

Just make sure to remove the comments as those won't be necessary.
Additionally make sure to add `!important;`

samuelgfeller:
> Just make sure to remove the comments as those won't be necessary.
I quite disagree with that. If someone (including me) is interested in changing it further or just differently classes like `.css-tmtnt9`, `.css-14v1uf5` and `.css-wvfva4` are absolutely unusable as they don't give context. They would have to find each on their own and that is a pain with jira because most classes change after every session so they are not stable. 
It took me some time to find those classes that don't change (and I tested it for a few weeks).

But if it's the rules.. I removed them :)

src/config/dynamic-theme-fixes.config
samuelgfeller: Done

Gusted:
> I quite disagree with that. If someone (including me) is interested in changing it further or just differently classes like `.css-tmtnt9`, `.css-14v1uf5` and `.css-wvfva4` are absolutely impossible to locate as they don't give context. They would have to find each on their own and that is a pain with jira because most classes change after every session so they are not stable.
> It took me some time to find those classes that don't change (and I tested it for a few weeks).

That's correct. On the other hand shipping a lot of CSS Selectors with comments would increase the file size even more. However if you'd want any context you can add comment inside this PR to those selectors with a little description. So anyone who might want to change it could lookup this PR(should be easy with `git blame`) and see the context of why it was merged.

Regards,
Gusted

Gusted (approved):

samuelgfeller:
> That's correct. On the other hand shipping a lot of CSS Selectors with comments would increase the file size even more. However if you'd want any context you can add comment inside this PR to those selectors with a little description. So anyone who might want to change it could lookup this PR(should be easy with git blame) and see the context of why it was merged.

True although there are simple solutions to this as well in my opinion to enable the option for comments. I think a minority would even make comments so if the users who want to make short comments on weird selectors like `.css-14v1uf5` are permitted to do so, the small file size increase is more than worth it in balance to the added clarity of the file for future contributors in my opinion.   

I would even suggest making it as an official suggestion to everyone and right before adding the CSS to the HTML something like the following would remove any comments. 
```js
// Regex from https://stackoverflow.com/a/59094308/9013718
let output = css.replace(/\/\*[\s\S]*?\*\/|\/\/.*/g, '');
```
But I never coded an extension and have frankly no idea about it. I'm sure it's not that simple and I don't expect that to go further. Just presenting my thoughts :)

PR #5005, opened 2021-02-10
Do not invert any popups on Gurushots.com

Karmalakas:
Currently inverted popups look like this:
image

Original popups look like this:
image

Non inverted popups would look like this:
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4119,6 +4119,9 @@ CSS
     background-color: var(--darkreader-neutral-text) !important;
 }
 
+NO INVERT
+iframe.ab-modal-interactions *
+
 ================================
 
 habr.com


Karmalakas:
Although I'm not quite sure why clicking _Apply_ on dev tools the change disappears, but it's still applied 😕

Gusted:
> Although I'm not quite sure why clicking _Apply_ on dev tools the change disappears, but it's still applied confused
> 
> I guess that's why tests failed.
> I took this [rule from the docs](https://github.com/darkreader/darkreader#fixes-for-filter-and-filter-modes)
> 
> > If an inverted element contains images or other content that becomes wrongly displayed, `NO INVERT` rule can be used.
> 
> Any tips @Gusted?
> Thank you

`Fixes for Filter and Filter+ modes` It isn't meant for Dynamic nor does it do anything.

Karmalakas:
> `Fixes for Filter and Filter+ modes` It isn't meant for Dynamic nor does it do anything.

But it does :/ After applying the view changes
Anyway, any advice how to stop inverting this whole popup?

Gusted:
> But it does :/ After applying the view changes

It might be related that Dark reader is modifying the page again.


> Anyway, any advice how to stop inverting this whole popup?

Not sure, without any context 🤷🏽‍♀️. If you go to `darkreader--fallback` inside the head element, due it have any textContent?

Karmalakas:
Only this
```
<style class="darkreader darkreader--fallback" media="screen"></style>
```

Gusted:
> Only this
> 
> ```
> <style class="darkreader darkreader--fallback" media="screen"></style>
> ```

Hmm, not sure what else could causes this. As fallback is the only one capable of doing such mess.

Karmalakas:
So there's no way to stop inverting, except writing rules for every element inside this popup?

Gusted:
> So there's no way to stop inverting

Their multiple ways, however without context IDK.

Karmalakas:
What do you mean by context? There's a page gurushots.com and from time to time it shows offers and other info popups with overlay, which are messed up. I've added screenshots too. Not sure what more I can provide. Just tell me and I'll try to explain or add more screenshots

Gusted:
> There's a page gurushots.com and from time to time it shows offers and other info popups with overlay, which are messed up.

This what I meant by context. Why the pop-up shows up.

Karmalakas:
So what are the ways to fix them?

Gusted:
> So what are the ways to fix them?

I'm not sure what's happening and after 5 minutes going forth and back I didn't receive such popup.

Karmalakas:
These popups are pretty rare. Only when you use some of the tools, sometimes it suggests to get more.

Karmalakas:
Here's another example with DR enabled
image

And that's how popup code looks in inspector with DR enabled
image

erkinalp:
Why is a transparent background not enough?

Karmalakas:
Which top div? `.demo` inside the frame or the frame holder itself - `.ab-show`?

erkinalp:
Both.

Karmalakas:
Doesn't help
image

Karmalakas:
Why was it closed when there's still no solution? Or did I miss it?

Gusted:
Housekeeping the Pull requests at the moment. Either they are getting closed when there is not activity and I couldn't reproduce the behavior. Or I patch up the fix and merge it.

Gusted:
And because this fix is not a valid fix due to `NO INVERT` not even being something. And you said it fixed something, it's something else.

Karmalakas:
To reproduce this, you have to be a registered member and I think maybe even need to participate in challenges. Not sure about the latter though. Please advise how to approach this issue and I'll try and update the PR. IMO closing this, because that's simply not how it's supposed to be and not offering an alternative is a bit like telling _We don't care, do whatever you want, but that's not how extension works_ :(

Gusted:
> lease advise how to approach this issue and I'll try and update the PR

Like I said, the PR is most likely not solving this as seeing from the screenshot it's more likely another issue that needs some depth-in knowledge about Dark Reader's logic(because this is most likely caused by our `darkreader--fallback` element).

> IMO closing this, because that's simply not how it's supposed

Like I said, I'm simply housekeeping the PR's as It was getting messy for me to oversee everything and a lot where either already solved or need a bit of a patch.


> and not offering an alternative is a bit like telling _We don't care, do whatever you want, but that's not how extension works_ :(

This has nothing to do with the extension. We had almost 100 outstanding PR's from which I now could reduced to 60 which more manage-able. I do care but this PR was nonsense and doesn't even work with Dark Reader's logic.

Karmalakas:
Should I just one the issue then and keep reminding from time to time for someone to take a look at it? Because now, from my point of view, it looks like extension needs some new feature or something that isn't there to support `NO INVERT` for Dynamic themes.

And as I mentioned previously, it works when "Apply" is clicked, but changes simply disappear and on the next load again the same issue.

Gusted:
> Should I just open the issue

Yes please.



> keep reminding from time to time for someone to take a look at it?

Please no, I periodically review certain issues. And it's either "Oh this is fix-able and Il be working on it" or it's "What here is going on is just out of our scope, so Il leave it for now to take with a fresh mind a look on it later on". 



> it looks like extension needs some new feature or something that isn't there to support `NO INVERT` for Dynamic themes.

There is no INVERT within Dynamic theme and such support is thereby bit pointless. Like I said this is more likely a issue with dark reader and not fixable with CSS. Because...

> And as I mentioned previously, it works when "Apply" is clicked, but changes simply disappear and on the next load again the same issue.

Of this behavior. In this case dark reader darkens the style again and this time succeeds. So the bug is, why is it in certain cases just not succeeding.

PR #5025, opened 2021-02-12
Update dark-sites.config

benfoster04:
Adding itsfoss.com: Already has darkmode switch located at bottom left


Gusted:
As from our README.md
`If a website is already dark (has all pages dark by default),`
Which `itsfoss.com` unfortunately isn't we cannot accept it.

PR #5038, opened 2021-02-13
Add option to change Link's color

### LINKED ISSUE #4766: colour blindness

### LINKED ISSUE #5036: [Feature Request] Replace with title
# Feature Request

## Is your feature request related to a problem? Please describe.
Was reading Wikipedia, reason I use this extension is to reduce the glare I get from the screen. The blue hyperlinks don't help.

## Describe the solution you'd like
Overriding default hyperlink colors to replace with another set of colors. (Instead of Blue/Purple.. to perhaps Yellow/Orange)

## Additional context
I know there's probably other extensions for this, but considering how Dark Reader does invert colors already, I bet it can be done.

Gusted:
- Add option under Colors UI to change Link's color.
- Style visited links accordingly.
- Resolves #4766
- Resolves #5036


src/inject/dynamic-theme/modify-css.ts
```diff
@@ -85,9 +85,26 @@ export function getModifiedUserAgentStyle(theme: Theme, isIFrame: boolean, style
...
-    lines.push('a {');
-    lines.push(`    color: ${modifyForegroundColor({r: 0, g: 64, b: 255}, theme)};`);
-    lines.push('}');
+    let linkColor: string;
+    let visitedLinkColor: string;
+    if (theme.linkColor === 'auto') {
+        linkColor = modifyForegroundColor({r: 0, g: 64, b: 255}, theme);
+    } else {
+        const hsl = rgbToHSL(parse(theme.linkColor));
+        const lighten = (lighter: number) => ({...hsl, l: clamp(hsl.l + lighter, 0, 1)});
```
alexanderby, line 94: Great use of a function, but it is called just once, so maybe unnecessary.

src/inject/dynamic-theme/modify-css.ts
```diff
@@ -85,9 +85,26 @@ export function getModifiedUserAgentStyle(theme: Theme, isIFrame: boolean, style
...
+        linkColor = hslToString(hsl);
+        visitedLinkColor = hslToString(lighten(0.2));
```
alexanderby, line 96: I thought you will add another color picker for visited links, but let's try a single one. Visited links usually differ by color (usually violet), so let's try to do lightness +0.2, saturation -0.3 and hue +45 (or between 30 and 45 if it gets too red). If hue surpasses 360, it should start from zero (subtract 360).

alexanderby (changes requested):

src/definitions.d.ts
```diff
@@ -54,6 +54,7 @@ export interface Theme {
     lightSchemeTextColor: string;
     scrollbarColor: '' | 'auto' | string;
     selectionColor: '' | 'auto' | string;
+    linkColor: '' | 'auto' | string;
```
alexanderby, line 57: I think it should not have `''` (default).

src/inject/dynamic-theme/modify-css.ts
```diff
@@ -85,9 +85,29 @@ export function getModifiedUserAgentStyle(theme: Theme, isIFrame: boolean, style
...
+        linkColor = hslToString(hsl);
+        const h: number = (hsl.h > 315 ? hsl.h - 360 : hsl.h) + 45;
+        visitedLinkColor = hslToString({
+            l: clamp(hsl.l + 0.2, 0, 1),
+            s: clamp(hsl.s - 0.3, 0, 1),
+            h});
+    }
+    if (linkColor) {
+        lines.push('a {');
+        lines.push(`    color: ${linkColor};`);
```
alexanderby, line 103: I don't see any effect in most cases, maybe let's use `!important` if it's not `auto`?

src/inject/dynamic-theme/modify-css.ts
```diff
@@ -85,9 +85,29 @@ export function getModifiedUserAgentStyle(theme: Theme, isIFrame: boolean, style
...
+        lines.push('}');
+    }
+    if (visitedLinkColor) {
+        lines.push('a:visited {');
+        lines.push(`    color: ${visitedLinkColor};`);
```
alexanderby, line 108: Same here

src/ui/popup/theme/controls/link.tsx
```diff
@@ -0,0 +1,26 @@
...
+export default function LinkColorEditor(props: LinkColorEditorProps) {
+    return (
+        <ThemeControl label="Links">
+            <ColorDropDown
+                value={props.value}
+                colorSuggestion={'#3391ff'}
+                onChange={props.onChange}
+                onReset={props.onReset}
+                hasAutoOption
+                hasDefaultOption
```
alexanderby, line 22: Remove `hasDefaultOption`

alexanderby (changes requested):

src/inject/dynamic-theme/modify-css.ts
```diff
@@ -85,9 +85,29 @@ export function getModifiedUserAgentStyle(theme: Theme, isIFrame: boolean, style
...
+        linkColor = hslToString(hsl);
+        const h: number = (hsl.h > 315 ? hsl.h - 360 : hsl.h) + 45;
+        visitedLinkColor = hslToString({
+            l: clamp(hsl.l + 0.2, 0, 1),
+            s: clamp(hsl.s - 0.3, 0, 1),
+            h});
+    }
+    if (linkColor) {
+        lines.push('a {');
+        lines.push(`    color: ${linkColor} !important;`);
```
alexanderby, line 103: It should not be !important, when the value is "auto". Otherwise it forces all links to look blue.

src/inject/dynamic-theme/modify-css.ts
```diff
@@ -85,9 +85,29 @@ export function getModifiedUserAgentStyle(theme: Theme, isIFrame: boolean, style
...
+    let linkColor: string;
+    let visitedLinkColor: string;
+    if (theme.linkColor === 'auto') {
+        linkColor = modifyForegroundColor({r: 0, g: 64, b: 255}, theme);
+    } else {
+        const hsl = rgbToHSL(parse(theme.linkColor));
+        linkColor = hslToString(hsl);
+        const h: number = (hsl.h > 315 ? hsl.h - 360 : hsl.h) + 45;
+        visitedLinkColor = hslToString({
+            l: clamp(hsl.l + 0.2, 0, 1),
```
alexanderby, line 97: It got a bit dull. Let's leave lightness unchanged.

alexanderby (changes requested):

src/inject/dynamic-theme/modify-css.ts
Gusted: Woops I forgot that you mentioned that.

src/inject/dynamic-theme/modify-css.ts
```diff
@@ -85,9 +85,30 @@ export function getModifiedUserAgentStyle(theme: Theme, isIFrame: boolean, style
     lines.push(`    border-color: ${modifyBorderColor({r: 76, g: 76, b: 76}, theme)};`);
     lines.push(`    color: ${modifyForegroundColor({r: 0, g: 0, b: 0}, theme)};`);
     lines.push('}');
-    lines.push('a {');
-    lines.push(`    color: ${modifyForegroundColor({r: 0, g: 64, b: 255}, theme)};`);
-    lines.push('}');
+    let linkColor: string;
```
alexanderby, line 88: Let's put your whole change into a separate function like it's done for selection color.

alexanderby (changes requested):
You know, it doesn't seem like it has much sense to try guess a visited link color.
Screen Shot 2021-03-30 at 23 21 33
Let's add a Visited Link color picker.

It's late night, so you can do it later, cause I will review it tomorrow anyway 🥱🌙

Gusted:
> It's late night, so you can do it later, cause I will review it tomorrow anyway yawning_facecrescent_moon

Going to prepare for school day. Have a good sleep 😄.

xcasxcursex:
I'd really like to see this feature, is there anything I could do to assist?


I'm trying to get a low-contrast, low-colour-temperature theme, for low eye strain....but there's blue all over the place. 

In my current attempts at solving this problem, I'm using hue rotation for the links, and this works quite well for visited links, it keeps the logic intended by the site. Perhaps some logic like this might work for DR:

Get default link color from the page
Calculate required hue rotation to match target color
Apply this same hue rotation to all links

I've only been doing this statically (I assume default blue links and rotate 180deg to get the colour I want) but so far so good. It does fail in reality because it doesn't interact with DR very well (it styles text that DR styles, breaking DR's style), which is why I'd love to help with this CR.


A heads-up that once this issue is fixed it is likely to expose a lot of other requests. For example, this text box I'm typing in right now uses the same blue that the links do, as it's outline when active, and obviously once the links are fixed this is the last blue thing on the page - so this fix will make other problems more apparent (sites tend to assume blue links and then style other elements on the site to match the links). But I will be happy to help with any site-specific fixes that arise from this.

Gusted:
> I'd really like to see this feature, is there anything I could do to assist?

It's currently sitting here to get pick up, because if I just merge this as now it would probably break something big which then causes five or ten duplicated issues made about it and the rest would be history. I'm pretty careful with feature requests, because they tend to break something and it should be decided if Dark Reader wants to support this feature or of this is out of it's scope and for this specific PR I haven't been able to make my mind up.

xcasxcursex:
I understand. For me this is a #6032 type 2 situation, but I do understand it's good to be cautious about it, so no pressure. I'll stay subscribed here and there in case you could use my help.

xcasxcursex:
So I guess this isn't happening now and we're stuck with blue links forever?

Gusted:
You don't need Dark Reader to do this. You can make your modification via [Stylus](https://github.com/openstyles/stylus).

xcasxcursex:
> You don't need Dark Reader to do this. You can make your modification via [Stylus](https://github.com/openstyles/stylus).

Sadly, it's not quite so simple... As per the above:

> It does fail in reality because it doesn't interact with DR very well (it styles text that DR styles, breaking DR's style), which is why I'd love to help with this CR.

Plus, there's that other thing I mentioned, where once you 'fix' the blue links it leaves behind other 'broken' blue stuff which clashes with your DR theme....

Take for example this search engine result, where the hue rotation away from blue links actually created *more* blue links including some text which DR had already themed:

image

image

So, it's more of a violentmonkey kind of thing than stylus, but even then, it's really a job only DR can do properly. My violentmonkey scripts for dealing with this attempt to read DR variables.

joatmon3:
I tried using DR for a while but this issue (inability to distinguish un/visited links) was one of the reasons I had to move to another option (the other major issue was stability). If this were addressed, I'd definitely consider switching back to using DR, for whatever that is worth.

lilydjwg:
I have made a working visited link highlight for dark reader using stylus:

```css
@scope ([data-darkreader-scheme="dark"]) {
  a:active, a:visited:active {
    color: #ff6666;
  }
  a:visited {
    color: #ffadff;
  }
}
```

This works only for pages that dark reader is active, so that it doesn't conflict with existing dark themes a page may have.

PR #5041, opened 2021-02-13
Fix crossword puzzles on nytimes.com

jonafato:
Most of the board can simply be inverted. Blacked out squares aren't
properly inverted, so additional custom CSS targets those squares
specifically.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -6271,6 +6271,7 @@ nytimes.com
 
 INVERT
 .svelte-1v1dl99
+#xwd-board
 
 CSS
 .css-oylsik,
@@ -6288,6 +6289,9 @@ a[data-testid] > svg {
 }
 .svelte-15nnlbj {
     font-weight: bold !important;
+}
+#xwd-board rect[class^="Cell-block--"] {
+    fill: ${black} !important;
 }
 
 ================================


Gusted:
Hi @jonafato!

Do you have an example link of the crossword puzzle?

Regards,
Gusted

jonafato:
@Gusted Sorry, should have included that originally. [This one](https://www.nytimes.com/crosswords/game/daily/2012/08/11) is currently free to play from the archive as well as [today's mini puzzle](https://www.nytimes.com/crosswords/game/mini).

There's probably some improvement to be had on the blocked squares in the light scheme, but without some change the dark scheme leaves the puzzle completely light and makes no distinction between blocked out squares and regular squares. Happy to make changes if necessary!

Screenshots below:

## Addon disabled
image

## Light scheme (without this change)
image

## Dark scheme (without this change)
image

## Light scheme (with this change)
image

## Dark scheme (with this change)
image

src/config/dynamic-theme-fixes.config
```diff
@@ -6289,6 +6290,9 @@ a[data-testid] > svg {
 .svelte-15nnlbj {
     font-weight: bold !important;
 }
+#xwd-board rect[class^="Cell-block--"] {
+    fill: var(--darkreader-neutral-background) !important;
```
Gusted, line 6294: What about `    fill: ${black} !important;`

jonafato: I tried that at first. Based on this note in the readme

>`CSS` rule adds custom CSS to a web page. `!important` keyword should be specified for each CSS property to prevent overrides by other stylesheets. Dynamic mode supports `${COLOR}` template, where `COLOR` is a color value before the inversion (`white` will become `black` in dark mode).

I expected it to invert to white when using the dark scheme. This didn't happen, and the blocked squares are black under both light and dark schemes (same behavior happens if I use `${white}`). I'm not sure if I've just misread that or if there's some issue with the NY Times puzzle that interferes with the extension. I'm fine with black squares for both versions, but it may be low contrast for some folks in dark mode.

Happy to go with whatever approach you prefer here, just wanted to raise that in case the behavior isn't what's expected.

Gusted: > I expected it to invert to white when using the dark scheme. This didn't happen, and the blocked squares are black under both light and dark schemes (same behavior happens if I use ${white}). 

So. Using the `${COLOR}`, only will works in dark mode and will not affect light mode. Your correct it should invert to White, however this fix also inverts the whole board and thus all colors of the cells and makes the white into black.

jonafato: Here's a screenshot showing that the black squares stay black in dark mode using `${black}`.

image

If that's acceptable, works for me. I've just pushed an amended commit with that change.

Gusted (approved):

jonafato:
@Gusted Thanks for walking me through my first contribution here!

Gusted:
> @Gusted Thanks for walking me through my first contribution here!

Happy to be part of it :D. And let the near future meet us again.

PR #5044, opened 2021-02-13
Fix background on redbubble.com

sithladyraven:
Fixes various background issues on redbubble.com


Gusted:
redbubble.com would be fixed next version when #4291 is merged and thus making this fix redundant.

PR #5059, opened 2021-02-15
Improve media query detection

### LINKED ISSUE #3779: [Broken Website] microsoftstore.com.cn
**Website**
https://www.microsoftstore.com.cn/

**How to Reproduce**

> 1. Go to the site

**Expected behavior**
Dark background
**Actual behavior**
Not dark background

**System info:**
 - OS: Windows
 - Browser: Edge
 - Darkreader Version: 4.9.20

### LINKED ISSUE #5058: [Broken Website] bestreviewguide completely white
# Broken Website

## Website
https://www.bestreviewguide.com/en/reviews/electronics/best-headphones/audio-technica-ath-sr5bt-review/

## How to Reproduce
Just visit the homepage

## Expected behavior
Dark

## Actual behavior
Bright

## Screenshots
Screenshot_20210215_164257-1

## System Information

- OS: Linux
- Browser: Firefox 85
- Dark Reader Version: 4.9.27

Gusted:
- Split queries between 'and'.
- Make sure to detect `@media print, screen and (min-width: 64em)`.
- Make sure it doesn't had any other conditions.
- Resolves #3779
- Resolves #5058


Gusted:
[binaries.zip](https://github.com/darkreader/darkreader/files/5984010/binaries.zip)

src/inject/dynamic-theme/css-rules.ts
```diff
@@ -7,7 +7,14 @@ export function iterateCSSRules(rules: CSSRuleList, iterate: (rule: CSSStyleRule
     forEach(rules, (rule) => {
         if (rule instanceof CSSMediaRule) {
             const media = Array.from(rule.media);
-            if (media.includes('screen') || media.includes('all') || !(media.includes('print') || media.includes('speech'))) {
+            media.map((mediaRule) => {
```
alexanderby, line 10: This `.map()` looks strange, it doesn't actually return anything. `map()` should be used to return another array, and it is not recommended to push or change anything in outer scope, because it is hard to understand.

src/inject/dynamic-theme/css-rules.ts
```diff
@@ -7,7 +7,14 @@ export function iterateCSSRules(rules: CSSRuleList, iterate: (rule: CSSStyleRule
...
+                if (mediaRule.includes(' and ')) {
+                    mediaRule.split(' and ').forEach((splittedRule) => media.push(splittedRule));
+                } else {
+                    return mediaRule;
+                }
+            });
+            if (media.includes('screen') || media.includes('all') || !((media.includes('print') && (media.length !== 1 && !matchMedia(rule.media.mediaText).matches)) || media.includes('speech'))) {
```
alexanderby, line 17: I would suggest to make this condition more readable and utilise `some()` method. The last condition is hard to understand, I'm not sure matchMedia should be there:
```typescript
const isScreenOrAll = media.some((m) => m.startsWith('screen') || m.startsWith('all'));
const isPrintOrSpeech = ...
if (isScreenOrAll || !isPrintOrSpeech) ...
```

alexanderby (changes requested):

src/inject/dynamic-theme/css-rules.ts
```diff
@@ -6,8 +6,17 @@ import {logWarn} from '../utils/log';
 export function iterateCSSRules(rules: CSSRuleList, iterate: (rule: CSSStyleRule) => void) {
     forEach(rules, (rule) => {
         if (rule instanceof CSSMediaRule) {
-            const media = Array.from(rule.media);
-            if (media.includes('screen') || media.includes('all') || !(media.includes('print') || media.includes('speech'))) {
+            const media: string[] = [];
+            Array.from(rule.media).forEach((mediaRule) => {
```
alexanderby, line 10: The idea was that if there's `media.some()` used below, you will not have to loop through this array, search for ` and ` etc.

But OK, let's optimise it a bit. Use `let isScreenOrAll = false; let isPrintOrSpeech = false;`. Then loop through the records. If any starts with screen, all etc, then assign `true` for a corresponding boolean.

After that remove `media.some()`.

src/inject/dynamic-theme/css-rules.ts
```diff
@@ -6,8 +6,17 @@ import {logWarn} from '../utils/log';
...
+                if (mediaRule.includes(' and ')) {
+                    mediaRule.split(' and ').forEach((splittedRule) => media.push(splittedRule));
```
alexanderby, line 12: Not sure, but I think it cannot be `(min-width: 10rem) and screen`. If it can, let's use `includes()` instead of `startsWith()`.

src/inject/dynamic-theme/css-rules.ts
```diff
@@ -6,8 +6,17 @@ import {logWarn} from '../utils/log';
...
+                } else {
+                    return media.push(mediaRule);
```
alexanderby, line 14: This is a void function, it should not return anything.

alexanderby (changes requested):

src/inject/dynamic-theme/css-rules.ts
alexanderby: `print, screen and (max-width: 64em)` will return an array `['print', 'screen and (max-width: 64em)']`. Just tried `print, screen and (max-width: 64em)`, it returned `['print', 'not all']`, simply saying it is not possible. So I think you can use `startsWith()` for sure.

Gusted: For the current sites that has this problem they use the same ways as this so I don't see an problem as well by using `startWith()`

tests/inject/dynamic/media-query.tests.ts
```diff
@@ -116,4 +116,24 @@ describe('MEDIA QUERIES', () => {
...
+            '    h1 { background: green; }',
+            '    h1 strong { color: orange; }',
+            '</style>',
+            '<style class="testcase-style-2" media="print, screen and (max-width: 9999999px)">',
+            '    h1 { background: gray; }',
+            '    h1 strong { color: red; }',
+            '</style>',
+            '<h1>Some test foor...... <strong>Oh uhm removing styles :(</strong>!</h1>',
+        );
+        debugger;
```
alexanderby, line 132: debugger;

tests/inject/dynamic/media-query.tests.ts
```diff
@@ -116,4 +116,24 @@ describe('MEDIA QUERIES', () => {
...
+        container.innerHTML = multiline(
+            '<style class="testcase-style">',
+            '    h1 { background: green; }',
+            '    h1 strong { color: orange; }',
+            '</style>',
+            '<style class="testcase-style-2" media="print, screen and (max-width: 9999999px)">',
+            '    h1 { background: gray; }',
+            '    h1 strong { color: red; }',
+            '</style>',
+            '<h1>Some test foor...... <strong>Oh uhm removing styles :(</strong>!</h1>',
```
alexanderby, line 130: Oh uh media query

alexanderby (changes requested):

tests/inject/dynamic/media-query.tests.ts
Gusted: Going to add an eslint rule later I forget this so many times within tests.

alexanderby (approved):

PR #5092, opened 2021-02-19
Update dark-sites.config

lollilol:
added lollilol.xyz and plex.doorcraft.de

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -396,6 +396,7 @@ listen.moe
 lolesports.com
 lollilol.cf
 lollilol.ml
+lollilol.xyz
 lookmovie.ag
 lordofthemanor.io
 lospec.com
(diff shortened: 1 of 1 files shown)


src/config/dark-sites.config
```diff
@@ -518,6 +519,7 @@ playcode.io/$
 playerdb.co
 pleroma.site
 pleroma.social
+plex.doorcraft.de
```
Gusted, line 522: Please remove this entry.
image

Gusted (changes requested):

src/config/dark-sites.config
lollilol: lmao, its dark if you login (google for plex webinterface)
do you only want sites on the list that are dark without having to login or sites that are restricted to only a small userbase?

PR #5093, opened 2021-02-19
Add feedly.com to dynamic-theme-fixes.config

SebastianRasch:
Better background for articles in list view when selected or mouse-over (when you don't want to use the site's own dark mode to get automatic mode switching which feedly.com doesn't support)

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3274,6 +3274,18 @@ CSS
 
 ================================
 
+feedly.com
+
+CSS
+.entry.u0:hover {
+    background-color: rgb(35, 35, 35) !important;
+}
+.entry--selected.u0, .entry--selected.u0:hover, button.secondary:hover {
+    background-color: rgb(50, 50, 50) !important;
+}
+
+================================
+
 feynmanlectures.caltech.edu
 
 INVERT
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3280,8 +3280,9 @@ CSS
 .entry.u0:hover {
     background-color: ${#ccc} !important;
 }
-.entry--selected.u0,
-.entry--selected.u0:hover,
+.entry--selected.u0, .entry--selected.u0:hover,
+.entry--selected.u4,
+.entry--selected.u5,
 button.secondary:hover {
     background-color: ${#ccc} !important;
 }
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3278,10 +3278,12 @@ feedly.com
 
 CSS
 .entry.u0:hover {
-    background-color: rgb(35, 35, 35) !important;
+    background-color: ${#ccc} !important;
 }
-.entry--selected.u0, .entry--selected.u0:hover, button.secondary:hover {
-    background-color: rgb(50, 50, 50) !important;
+.entry--selected.u0,
+.entry--selected.u0:hover,
+button.secondary:hover {
+    background-color: ${#ccc} !important;
 }
 
 ================================
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3278,7 +3278,8 @@ feedly.com
 
 CSS
 .entry.u0:hover,
-.entry--selected.u0, .entry--selected.u0:hover,
+.entry--selected.u0,
+.entry--selected.u0:hover,
 .entry--selected.u4,
 .entry--selected.u5,
 button.secondary:hover {
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3277,9 +3277,7 @@ CSS
 feedly.com
 
 CSS
-.entry.u0:hover {
-    background-color: ${#ccc} !important;
-}
+.entry.u0:hover,
 .entry--selected.u0, .entry--selected.u0:hover,
 .entry--selected.u4,
 .entry--selected.u5,


Gusted:
Hi @SebastianRasch!

Could you share an example URL where I can observe this change?

Regards,

Gusted

SebastianRasch:
Hi @Gusted!

Difficult to share a link to this specific list view because you need an account first. It's free but if you don't want to go through the hustle I can send you screenshots instead?

https://feedly.com/i/subscription/feed%2Fhttp%3A%2F%2Fnewsfeed.zeit.de%2Fsport%2Findex

src/config/dynamic-theme-fixes.config
```diff
@@ -3274,6 +3274,18 @@ CSS
 
 ================================
 
+feedly.com
+
+CSS
+.entry.u0:hover {
```
Gusted, line 3280: It seems to be that this `hover` over an entry is provided by feedly and thus adding this fix would cause extra functionality than feedly has which isn't the purpose of this `dynamic-theme-fixes.config` file.

src/config/dynamic-theme-fixes.config
```diff
@@ -3274,6 +3274,18 @@ CSS
...
+    background-color: rgb(35, 35, 35) !important;
+}
+.entry--selected.u0, .entry--selected.u0:hover, button.secondary:hover {
```
Gusted, line 3283: Please format it to
```CSS
.entry--selected, 
.entry--selected:hover, 
button.secondary:hover {
```
Please notice that the .u0 is redundant as AFAIK I have on every entry an `.u4` instead and doesn't specify any further.

src/config/dynamic-theme-fixes.config
```diff
@@ -3274,6 +3274,18 @@ CSS
...
+    background-color: rgb(50, 50, 50) !important;
```
Gusted, line 3284: Please change it to `background-color: ${#cdcdcd} !important;`

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
SebastianRasch: On the normal white mode you can see the background changing at mouse-over. Also, on the built-in dark mode that Feedly offers you see the background changing at mouse-over. Only when using Dark Reader, the background change is not visible at mouse-over. Hope that makes sense. This is not an extra functionality and hence needs to be fixed.

src/config/dynamic-theme-fixes.config
SebastianRasch: Actually, if I leave .u0 off it is not the same experience for me. Somehow, if you don't use .u0, the colours are being added up if you hover on a selected article so it becomes even brighter then just selected or just hover. That's two different colours then and that's exactly what we don't want to achieve.

src/config/dynamic-theme-fixes.config
SebastianRasch: OK! I used ccc instead of cdcdcd to make it a tiny bit brighter; cdcdcd was a too dark compared with the normal background (unselected/unhovered)

src/config/dynamic-theme-fixes.config
SebastianRasch: You mean that it has a different colour then `--selected`? Yeah I forgot that that was CSS I injected myself. I changed it to the same colour now.

src/config/dynamic-theme-fixes.config
```diff
@@ -3274,6 +3274,20 @@ CSS
 
 ================================
 
+feedly.com
+
+CSS
+.entry.u0:hover {
```
Gusted, line 3280: Making this selector using the same fix as the fix under this selector a little improvement can be made and moving this selector to the other selectors.

src/config/dynamic-theme-fixes.config
SebastianRasch: OK I found the differences, it showed `.u4` for you because you were in the "Magazine View". The "Cards View" has `.u5` and the "Article View" doesn't use an background at all. I use "Titles-Only View", which uses `.u0`

Change this in Settings > General.

src/config/dynamic-theme-fixes.config
SebastianRasch: You're absolutely right. How about this?

Please also read my changes to my answers above, I found out why it shows `.u4` for you.

src/config/dynamic-theme-fixes.config
```diff
@@ -3274,6 +3274,19 @@ CSS
 
 ================================
 
+feedly.com
+
+CSS
+.entry.u0:hover,
+.entry--selected.u0, .entry--selected.u0:hover,
```
Gusted, line 3281: Every entry on a new line ;)

src/config/dynamic-theme-fixes.config
Gusted: > Please also read my changes to my answers above, I found out why it shows `.u4` for you.

I've read it and the changes now looks 100% to me and ready to be merged.

src/config/dynamic-theme-fixes.config
SebastianRasch: Yes boss! ;) Any reason for this?

src/config/dynamic-theme-fixes.config
SebastianRasch: Awesome! Thanks for your help man!

src/config/dynamic-theme-fixes.config
Gusted: Code-style :) Just an unified format which I've been trying to 'enforce'. I really should get around and make some documentation on it + some linting rules. So it won't get a mess one day.

Gusted (approved):
LGTM

SebastianRasch:
Hey @Gusted, when is this coming to Dark Reader for Safari? The changes are already live in Dark Reader for Edge but not on Safari.
By the way are you planning an editor for Safari as well? That'd be really nice. 🙏🏻

Gusted:
Hi @SebastianRasch!

I don't maintain nor know the Safari's version of Dark Reader. I'm simply the main maintainer. Any info about Safari should be going to @alexanderby or more preferably on the mail darkreaderapp@gmail.com

SebastianRasch:
OK thanks! But generally speaking, shouldn't the rules be applied to all versions of Dark Reader? Or do you have to suggest rules for the Safari version separately?

Gusted:
> I don't maintain nor know the Safari's version of Dark Reader.

It should be? I only know that it exist and is hopefully switching soon to the webExtension variant which then yes, have the rules applied.

PR #5094, opened 2021-02-19
Added github.com since it has a dark mode now

tsndr:


Gusted:
#4454 #4440 #4429

tsndr:
Oh sorry, didn't see those. Makes sense.

PR #5095, opened 2021-02-19
medium.com - Removed underline from links

tsndr:
Since it isn't necessary and medium doesn't use underlines by default so why should it in dark mode.


Gusted:
Their seems like a specific issue with medium.com, however checking that issue again it doesn't seem to occur and. Correctly this fix can be removed. Thank you!

tsndr:
Thank you for merging my PR :)

PR #5100, opened 2021-02-20
Updated sprite icon URL for Google Classroom and Drive file preview

kohmy2036:
Both sites are now using `v-sprite35.svg`, tested on both sites.


Gusted (approved):
LGTM and let's hope this time it won't change back 😄

PR #5107, opened 2021-02-22
Fix for Microsoft Edge Translate extension

AhmedElghrbawy:
- Fix wrong text color.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -19,7 +19,7 @@ a[href="https://coinmarketcap.com/"] > svg[width="94"][height="16"] > path {
     fill: var(--darkreader-neutral-text) !important;
 }
 #edge-translate-panel-body {
-   color:  var(--darkreader-neutral-text) !important;
+    color: var(--darkreader-neutral-text) !important;
 }
 
 IGNORE INLINE STYLE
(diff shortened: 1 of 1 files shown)


Gusted:
Hi @AhmedElghrbawy!

Could you share an link where I can download this extension?

Regards,
Gusted

AhmedElghrbawy:
[Edge Translate](https://microsoftedge.microsoft.com/addons/detail/edge-translate/bfdogplmndidlpjfhoijckpakkdjkkil)

Gusted:
Hi @AhmedElghrbawy!

I'm not sure, but it does show for me it's correctly and the texts gives me a white color.

Regards,
Gusted

AhmedElghrbawy:
Hey @Gusted!

I tried setting everything back to default for both extensions to see if there is something wrong with my configuration, yet the problem still exists (at least for me).

Here is how to produce the problem:

* Select a word or a phrase, right click and translate that word.
* This works well for the first time you translate something, but for the following times **in the same Edge tab**, the translated text would be displayed in wrong colors (black text on black background).

Screenshot_3

Increasing the contrast as I always like to configure dark reader, would make the translated text unreadable at all.

If this displays correctly for you, then maybe there is something wrong with my setup making this is my own problem.



Thanks, and sorry for bothering.

Gusted (approved):
Ah I see! Thank you

PR #5113, opened 2021-02-23
Remove unused plugin-commonjs

jatin-code777:
`@rollup/plugin-commonjs` is not required. This PR removes it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,6 @@
       "version": "4.9.28",
       "license": "MIT",
       "devDependencies": {
-        "@rollup/plugin-commonjs": "17.1.0",
         "@rollup/plugin-node-resolve": "11.2.0",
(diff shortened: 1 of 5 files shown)


Gusted:
@alexanderby Could you chop in here I remember the switch from gulp to npm scripts when this was introduced however it indeed looks it's not needed. Where I actually tought it was needed for malevic to be injected properly.

alexanderby:
Hi @jatin-code777! Could you please run `npm install`, then run `npm run build` and check if it runs without errors and `build/` folder contains JS and HTML files?

jatin-code777:
Hi, @alexanderby. I had checked this, but I just checked it again and it works without errors. I even `diff`ed the generated build folders before and after making this change and they are the same. `npm run test` also passes.

alexanderby:
Great, thank you.

PR #5118, opened 2021-02-23
Add minor fixes for trip101.com (dynamic filter)

kenrickbuduan:
Trip101 is an up-and-coming travel website based in Singapore. However, there are some weird issues on the website when using Dark Reader. Some logos and images don't look nice on dynamic filter, hence the optimization.


Gusted (approved):
Thank you!

PR #5135, opened 2021-02-25
Update dark-sites.config

gordcodes:
add ign.com, hulu.com, netflix.com

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -328,6 +328,7 @@ hoppscotch.io
 hostedtalk.net
 hotstar.com
 htmlpasta.com
+hulu.com
 humblebundle.com/$
 humblebundle.com/accessibility
 humblebundle.com/charities
@@ -340,6 +341,7 @@ ibookpile.net
 identinator.com
 ieuan.tech
 ieuan.xyz
+ign.com
 illicoweb.videotron.com
 imgbox.com
 imgops.com
@@ -444,6 +446,7 @@ nanhu.ca
 nchristopher.me
 neal.fun/size-of-space
 neovim.io
+netflix.com
 neundex.com
 newgrounds.com
 nexusmods.com
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -774,6 +774,7 @@ yandexdataschool.ru/$
 yasio.pl
 yohoho.io
 you.dj
+youtube.com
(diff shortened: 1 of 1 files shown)


gordcodes (commented):
Add youtube.com

src/config/dark-sites.config
```diff
@@ -328,6 +328,7 @@ hoppscotch.io
 hostedtalk.net
 hotstar.com
 htmlpasta.com
+hulu.com
```
Gusted, line 331: image
Doesn't seem that dark to me.

src/config/dark-sites.config
```diff
@@ -340,6 +341,7 @@ ibookpile.net
 identinator.com
 ieuan.tech
 ieuan.xyz
+ign.com
```
Gusted, line 344: Not that dark either
image

src/config/dark-sites.config
```diff
@@ -444,6 +446,7 @@ nanhu.ca
 nchristopher.me
 neal.fun/size-of-space
 neovim.io
+netflix.com
```
Gusted, line 449: Already been listed under www.netflix.com

src/config/dark-sites.config
```diff
@@ -771,6 +774,7 @@ yandexdataschool.ru/$
 yasio.pl
 yohoho.io
 you.dj
+youtube.com
```
Gusted, line 777: Won't need to open youtube to know this one isn't dark either.

src/config/dark-sites.config
gordcodes: If you have a hulu account and log in the entire site is dark
image

src/config/dark-sites.config
gordcodes: I guess IGN International hasn't rolled out dark mode based on theme yet
image

src/config/dark-sites.config
gordcodes: Looks like Youtube US may only have themes and hasn't rolled it out internationally yet also. And this is with the dark mode add-on disabled.
image

gordcodes:
@Gusted  can you actually check youtube? Do you actually not have a dark mode in Europe?

Gusted:
>  Do you actually not have a dark mode in Europe?

We have an dark-mode, but just not by default, which is a must to get an site into this list.

PR #5140, opened 2021-02-25
Added dark themed websites

alfawal:


src/config/dark-sites.config
```diff
@@ -61,6 +61,7 @@ benleggiero.me
 bequiet.com
 bestblackhatforum.com
 beta.destinyitemmanager.com
+betterdiscordlibrary.com
```
Gusted, line 64: image
not dark

src/config/dark-sites.config
```diff
@@ -161,6 +162,7 @@ deepweblinks.net
 del.dog
 deltarune.com
 demonforums.net
+ders.eba.gov.tr/ders/proxy/*/liveMiddleware/eba.html
```
Gusted, line 165: I really have no idea what this is, but probably won't work due to `*` in the middle of the string.

Gusted (changes requested):

src/config/dark-sites.config
alfawal: I can send a screenshot of it here but I have to remove it anyway because `.../proxy/VERSION/...` changes frequently except if you have something that place-holds any value (if this feature doesn't exist I would recommend making it).

src/config/dark-sites.config
alfawal: This is so weird again, I think this depends on the Windows theme preferences.
Do you have any tips that help me determine if it was a dark site by default or not?

src/config/dark-sites.config
Gusted: >  (if this feature doesn't exist I would recommend making it).

#2517  :)

src/config/dark-sites.config
Gusted: Personally I use Firefox + Private Windows + `privacy.resistFingerprinting`(to force white-theme-preference) to check sites.

src/config/dark-sites.config
alfawal: > but probably won't work due to * in the middle of the string.

If #373 is resolved then `ders.eba.gov.tr/ders/proxy/*/liveMiddleware/eba.html` should work!
Here's a screenshot of it, should I add it as it was before? if not, just merge it :)
2021-02-26 09_50_58-Window

src/config/dark-sites.config
alfawal: Okay thanks!

fsrrt (approved):

Gusted (approved):

PR #5144, opened 2021-02-26
add twitter to dark-sites.config

gordcodes:
add twitter.com

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -701,6 +701,7 @@ tuner.ninja
 tusharsadhwani.dev
 twelvesmith.com
 twist.moe
+twitter.com
 tycrek.com
 tynker.com/ide
 typing.works


Panquesito7:
I don't think [Twitter](https://twitter.com) is valid as you can toggle dark mode in the [settings](https://twitter.com/settings/display).

gordcodes:
You don't have to be logged into twitter to get dark mode. It automatically detects if you have your computer set to dark theme.
image

alfawal:
They don't count that as a valid dark-themed website.
I did commit something similar before and they let me delete it before merging to master.

Gusted:
> I did commit something similar before and they let me delete it before merging to master.

delete -> remove, I think that better suits.
And your totally right we don't count such sites into list.

gordcodes:
@Gusted  what sites do count as fitting the global list then, sites that are built as dark theme without theme detection? Why force that requirement on the global list if there's an API built to detect if a user wants light or dark theme?

Gusted:
As stated in [Read me](https://github.com/darkreader/darkreader/blob/master/README.md)
`If a website is already dark (has all pages dark by default), you can add it to the dark-sites.config file`
With the big notices to `has all pages dark by default`.

PR #5151, opened 2021-02-28
Fix dle.rae.es

Elaborendum:
I don't know much about this, so this is mostly the result of trial and error.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2481,31 +2481,26 @@ div.opineo-side-slider-widget > button
 
 dle.rae.es
 
-INVERT
-.bloque_publi DE
-.spr_fundacionlacaixa
-.spr_dle90x110
-.spr_logo-enclave-rae
-.spr_google-play-badge-100x39
-.spr_badge-appstore-lrg-96x28
-.spr_edtri
-.spr_ay
-.spr_cita
-.spr_unidrae
-.spr_consultas
-.spr_pnwe2
-img[src="/images/logos/ASALE2.png"]
-img[src="/images/logos/rae.png"]
-img[src="/images/LibroDeEstilo_300.jpg"]
-img[src="/app/doc/es/img/dle.jpg"]
-
 CSS
-.container {
-    background-color: ${#f2f2f2} !important
+.bloque_publi DE,
+.spr_fundacionlacaixa,
+.spr_dle90x110,
+.spr_logo-enclave-rae,
+.spr_google-play-badge-100x39,
+.spr_badge-appstore-lrg-96x28,
+.spr_edtri,
+.spr_ay,
+.spr_cita,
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -2479,6 +2479,36 @@ div.opineo-side-slider-widget > button
...
+.spr_unidrae
+.spr_consultas
+.spr_pnwe2
+img[src="/images/logos/ASALE2.png"]
+img[src="/images/logos/rae.png"]
+img[src="/images/LibroDeEstilo_300.jpg"]
+img[src="/app/doc/es/img/dle.jpg"]
+
+CSS
+.container {
```
Gusted, line 2503: Why making it darker than it is?

src/config/dynamic-theme-fixes.config
```diff
@@ -2479,6 +2479,36 @@ div.opineo-side-slider-widget > button
 
 ================================
 
+dle.rae.es
+
+INVERT
```
Gusted, line 2484: That are a lot of inverts but lgtm.

src/config/dynamic-theme-fixes.config
Elaborendum: Most images are just an object inside a white square (they don't have transparency even if they should), so when Dark Reader darkens the website, most images stick out because of their white backgrounds.

So I inverted them, but their new backgrounds were a darker shade of black when compared to the website's. Meaning, they still sticked out, in a different way, so I darkened the website.

Is there a better way to deal with cases like this?

Gusted: > Is there a better way to deal with cases like this?

We invert them, like you do now. However that's it, we could apply a better optimized filter to it e.g.

```CSS
invert(93.7%) hue-rotate(180deg) contrast(90.6%)
```
Which seems to be the sweat spot for this site, so I suggest wrap all those css selectors into a big group(each selector on a new line) and apply `filter: invert(93.7%) hue-rotate(180deg) contrast(90.6%) !important` onto it.

Elaborendum: Ok, I did it, I think that's it.

src/config/dynamic-theme-fixes.config
```diff
@@ -2477,6 +2477,31 @@ div.f-grid-3.logo-column > a > img
...
+.spr_consultas,
+.spr_pnwe2,
+img[src="/images/logos/ASALE2.png"],
+img[src="/images/logos/rae.png"],
+img[src="/images/LibroDeEstilo_300.jpg"],
+img[src="/app/doc/es/img/dle.jpg"] {
+	filter: invert(93.7%) hue-rotate(180deg) contrast(90.6%) !important
+}
+
+
```
Gusted, line 2504: Remove this new line

PR #5171, opened 2021-03-03
Add fix for heise.de

va1entin:
This inverts the dark "heise online" logo in the top left corner on german tech news site https://heise.de

The dark "heise+" logo right next to it is still too dark. I didn't create a fix for it yet, because it's not as straightforward and I am not a subscriber of heise+ anyway :smiley:


Gusted (approved):
Thank you!

PR #5174, opened 2021-03-03
Fixed math expressions in cubawiki.com.ar

luisbustamante097:
The math expressions where black and it couldn't be distinguished with the background 
- Example: https://www.cubawiki.com.ar/index.php/Final_14/06/2019_(Probabilidad_y_Estad%C3%ADstica)
So I fixed


Gusted (approved):
Thank you!

PR #5183, opened 2021-03-05
Fix for BestBuy.com radio button selections not being visible

MaleNurse:
Radio button selections in checkout were not visible. Setting background color of that particular type input back to `initial` fixed the issue.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1125,6 +1125,15 @@ body {
 
 ================================
 
+bestbuy.com
+
+CSS
+input[type="radio" i] {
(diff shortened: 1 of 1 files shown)


Gusted:
Hey @MaleNurse!

I cannot find a quick example of a radio button on bestbuy, could you link an URL?

Regards,
Gusted

MaleNurse:
@Gusted 
Yeah, not many of them. If you add something to your cart and then go to your cart you'll see it; so no permanent URL to share, unfortunately, but here's a screenshot with the change applied:

Capture

Gusted (approved):
Thank you!

PR #5200, opened 2021-03-08
Fix for privacytools.io

AndrewDaGuy:
Changes:
- Inverted oeck logo


AndrewDaGuy:
I guess #1285 would fix almost everything, so I'm going to remove the fix from dad9335 and reopen it.

PR #5207, opened 2021-03-08
Fix dynamic mode article background color on hs.fi

ristomatti:
Changes brown article element background color on article listing pages to `var(--darkreader-neutral-background)`. A different background color has apparently been added to business sections. This is a recent design change on the site.

Preferably I would like to retain a different background color but I didn't find any documentation on other background color variables to use. Or would this be a case to use a hardcoded value?

URL: https://www.hs.fi

**Before:**
image

**After:**
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4325,6 +4325,15 @@ INVERT
 
 ================================
 
+hs.fi
+
+CSS
+article, article > article {
+  background-color: var(--darkreader-neutral-background) !important;
+}
+
+================================
+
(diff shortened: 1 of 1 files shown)


Gusted (approved):
Thank you!

Gusted:
> Preferably I would like to retain a different background color but I didn't find any documentation on other background color variables to use. Or would this be a case to use a hardcoded value?

`background-colorL ${#ccc} !important;` however the result can be quite bad depending on the original color.

ristomatti:
That was fast, thanks a lot!

PR #5208, opened 2021-03-08
Color inversions for 'move' menu elements

blameitonjoe:
When clicking the move icon (between the star and cloud icons) at the top, a small menu window ui lets you choose the destination. Some of the following icon elements were not inverted and I'm adding them to INVERT:
- Open folder in new window
- Window closing "x"
- Create new folder
- Select folder right chevron (hover)
- Shared with me (at root)
- Starred (at root)

Also added to CSS a bright color for the scrollbar thumb for this little window.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2624,6 +2624,7 @@ img[src$="googlelogo_dark_clr_74x24px.svg"]
 body[itemtype*="PresentationObject"] #docs-titlebar-share-client-button .scb-button-icon
 g.punch-filmstrip-indicator > image
 .docs-gm .docos-icon-overflow-three-dots-size
+.de-fe-cb-he .de-Cf-f,.de-ik-db-Gi, .de-fe-Gf-Df-Ld-u .de-ik-Ld-Df,.de-ik-pg-je,.de-ik-Rh-je,.de-ik-pg,.de-ik-Rh,.de-ik-kk-n-af
 
 CSS
 .docs-preview-palette-item {
@@ -2656,6 +2657,9 @@ CSS
 }
 .docs-title-input-label:not([style*="pointer-events: auto"]) > #docs-title-input-label-inner {
     visibility: hidden !important;
+}
+.de-Zd-Kc ::-webkit-scrollbar-thumb {
+    background-color: rgb(255 255 255 / 20%);
 }
 
 ================================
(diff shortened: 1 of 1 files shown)


Gusted (commented):
I think we can re-use the fix from `drive.google.com/picker`.

```CSS
INVERT
div[role="menu"] div[role="menuitem"]:not([class*=" "]) > div > div > div
div[role="button"][tabindex="0"]

CSS
div[role="menu"] div[role="menuitem"][class*=" "] > div > div > div,
div[role="button"][aria-disabled="true"] > div {
    filter: invert(50%) !important;
}

IGNORE INLINE STYLE
svg[class=""]
svg[class=""] *
```
Which does an good job on it while maintaining not to rely on obscure CSS Selectors.

blameitonjoe:
Absolutely, @Gusted. This was my first ever contribution and tried to help on something that might benefit drive users. Sorry to waste your time.

rakleed:
@Gusted can I just remove the rules separately for `drive.google.com/picker` and add them to `drive.google.com`?

Gusted:
> can I just remove the rules separately for `drive.google.com/picker` and add them to `drive.google.com`?

I'm pretty sure I've set it to `/picker` for a reason, because due to conflicting CSS selectors. But give it a try and make sure you don't break anything, have a special note for the context menu's that you get when you right click on files/folders etc.

rakleed:
The rule `INVERT div[role="button"][tabindex="0"]` breaks a lot. If I change it to `INVERT div[role="dialog"] div[role="button"][tabindex="0"]`, then everything will be displayed normally in the main interface, but for some reason it will stop working on https://drive.google.com/file. Also, if I add the rules `IGNORE INLINE STYLE svg[class=""]`, and `IGNORE INLINE STYLE svg[class=""] *` at the same time, they will also break the file properties icon in the main interface.

And in general, I did not understand what the rules are for `INVERT div[role="menu"] div[role="menuitem"]:not([class*=" "]) > div > div > div`, `IGNORE INLINE STYLE svg[class=""]`, `IGNORE INLINE STYLE svg[class=""] *`, so if I add the rules `INVERT div[role="dialog"] div[role="button"][tabindex="0"]`, `CSS div[role="menu"] div[role="menuitem"][class*=" "] > div > div > div, div[role="button"][aria-disabled="true"] > div { filter: invert(50%) !important; }` for https://drive.google.com, and do not touch the rules for https://drive.google.com/picker, then everything will be displayed correctly.

Gusted:
Rakleed fixed this one with #5835

PR #5210, opened 2021-03-09
Fix for pcdiga.com

OrangSquid:
Fix the white squares that appear on product descriptions
Fix black text at the top of the page


Gusted (approved):
Thank you!

PR #5214, opened 2021-03-09
Fix for wikipedia.org

OrangSquid:
Fix for Wikipedia's new design where Wikipedia's name appears dark at the top of the page

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -9752,6 +9752,8 @@ span[class^="ts-"]
 img[src$="Rename_icon_Cyr.svg.png"]
 img[alt="Aksara Sunda.png"]
 img[alt="Aksara kawi name.png"]
+img[src*="/static/images/mobile/copyright/wikipedia-wordmark"]
+img[src*="/static/images/mobile/copyright/wikipedia-tagline"]
 
 CSS
 .mwe-popups-discreet > svg,


src/config/dynamic-theme-fixes.config
```diff
@@ -9752,6 +9752,8 @@ span[class^="ts-"]
 img[src$="Rename_icon_Cyr.svg.png"]
 img[alt="Aksara Sunda.png"]
 img[alt="Aksara kawi name.png"]
+img[src*="/static/images/mobile/copyright/wikipedia-wordmark"]
```
Gusted, line 9755: It seems like `.header .branding-box > a > span > img` on line 9743 have the same effect.

src/config/dynamic-theme-fixes.config
```diff
@@ -9752,6 +9752,8 @@ span[class^="ts-"]
...
+img[src*="/static/images/mobile/copyright/wikipedia-tagline"]
```
Gusted, line 9756: Where can I find this element?

OrangSquid: https://wikipedia.org/static/images/mobile/copyright/wikipedia-wordmark-fr.svg
https://wikipedia.org/static/images/mobile/copyright/wikipedia-tagline-pt.svg
It's these two elements. For me only the portuguese version of Wikipedia has the new design. These two elemets were not inverted before I applied this fix. I made it a substring since the language indicator at the end varies.

Gusted: Doesn't seems to harm or break anything so good to be merged

Gusted (approved):
Thank you!

PR #5224, opened 2021-03-10
Changed file links to "local"

AndrewDaGuy:
Changed it so clicking the links would "locally" go to the file instead of going to darkreader/darkreader :]


Gusted (commented):
Pretty sure this can break services/websites that uses our readme to display information about dark reader e.g. npmjs.com/package/darkreader

PR #5234, opened 2021-03-11
Fix dynamic theme for iconify.design

Snazzah:
Darkreader shouldn't affect emojis with color within [iconify.design](https://iconify.design/).

Before fix

After fix


Gusted (commented):
Not sure if their are any other elements inside this element, otherwise
```CSS
svg.iconify path
svg.iconify g
```
Can be formatted to 
```CSS
svg.iconify *
```
The same thing for the other 2 selectors.

Gusted (approved):
Thank you

PR #5235, opened 2021-03-11
Made it use the `question` template

AndrewDaGuy:
Made it use the question template

Example:
-  https://github.com/darkreader/darkreader/issues/new?assignees=&labels=bug&template=bug_report.md&title=%5BBUG%5D+Replace+with+title (Bug report)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 ## Questions
 
 Most questions can be answered by reading the [help page](https://darkreader.org/help/).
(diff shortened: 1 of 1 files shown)


Gusted:
Think it's better to have an dedicated questions template.

AndrewDaGuy:
> Think it's better to have an dedicated questions template.

Yep, instead of having something like a bug label on a question

AndrewDaGuy:
Oops, forget this pr was a thinig

PR #5242, opened 2021-03-11
Fix for microsoftedge.microsoft.com

rakleed:
- Invert the separator between reviews and category.
- Invert the letter in the profile logo in user reviews.

Example of site page: https://microsoftedge.microsoft.com/addons/detail/dark-reader/ifoakfbpdcdoeenechcleahebpibofpc

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5867,6 +5867,14 @@ body > .ember-view.dialog_wrapper {
 
 ================================
 
+microsoftedge.microsoft.com
+
+INVERT
+svg[viewBox="0 0 16 16"]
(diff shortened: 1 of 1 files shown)


AndrewDaGuy:
> Invert the letter in the profile logo in user reviews.

Won't that break other profile pictures?

rakleed:
> > Invert the letter in the profile logo in user reviews.
> 
> Won't that break other profile pictures?

@AndrewDaGuy I looked at about a hundred reviews, and it looks like Microsoft is automatically hiding the images and displaying the first letter of the name instead. So that shouldn't be a problem.

Gusted (approved):
Thank you!

PR #5251, opened 2021-03-12
Fix for wikipedia.org

### LINKED ISSUE #5250: [Broken Website] wikipedia.org
# Broken Website

## Website
https://en.wikipedia.org/wiki/Unified_Modeling_Language#Diagrams

## How to Reproduce
just go there

## Expected behavior
The first diagram (Hierarchy of UML 2.2 Diagrams, shown as a class diagram) should be readable

## Actual behavior
In Dynamic and Static themes black lines on dark background are unreadable, only Filter themes work. In addition, when mouse pointer hovers above the diagram, Dynamic theme changes diagram background colour which probably is wrong too because it doesn't change when Dark Reader is off.

## System Information

- OS: windows 10, linux
- Browser: chrome 87, firefox 86
- Dark Reader Version: 4.9.29

## Additional context
Some of the diagrams below are similarly affected, also here https://en.wikipedia.org/wiki/IDEF1X#IDEF1X_building_blocks Could be related to #5222 - looks similar.

AndrewDaGuy:
- Partially _resolves_ #5250

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -9866,7 +9866,7 @@ img[alt="Aksara Sunda.png"]
 img[alt="Aksara kawi name.png"]
 img[src*="/static/images/mobile/copyright/wikipedia-wordmark"]
 img[src*="/static/images/mobile/copyright/wikipedia-tagline"]
-[src="//upload.wikimedia.org/wikipedia/commons/thumb/e/ed/UML_diagrams_overview.svg/600px-UML_diagrams_overview.svg.png"]
+img[src*="UML_diagrams_overview.svg"]
 
 CSS
 .mwe-popups-discreet > svg,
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -9866,6 +9866,7 @@ img[alt="Aksara Sunda.png"]
 img[alt="Aksara kawi name.png"]
 img[src*="/static/images/mobile/copyright/wikipedia-wordmark"]
 img[src*="/static/images/mobile/copyright/wikipedia-tagline"]
+[src="//upload.wikimedia.org/wikipedia/commons/thumb/e/ed/UML_diagrams_overview.svg/600px-UML_diagrams_overview.svg.png"]
```
Gusted, line 9869: Shorting it to something like
```CSS
img[src*="UML_diagrams_overview.svg"]
```
Should be cleaner.

AndrewDaGuy: Done 👍 c3954f3fa022d060ee8873b234c69b8ed2ec5de6

PR #5252, opened 2021-03-12
Fix for suckless.org

AndrewDaGuy:


src/config/dynamic-theme-fixes.config
```diff
@@ -8276,6 +8276,22 @@ body {
 
 ================================
 
+suckless.org
+*.suckless.org
```
Gusted, line 8280: This is redundant.

AndrewDaGuy:
Why do i keep accidentally deleting the same branch over and over

src/config/dynamic-theme-fixes.config
```diff
@@ -8276,6 +8276,21 @@ body {
 
 ================================
 
+*.suckless.org
```
Gusted, line 8279: Change it to the right place and I meant the *.suckless.org is redunant.

AndrewDaGuy: 50fb573 Done

PR #5287, opened 2021-03-18
invert invisible icons for onscreen keyboard

isaec:
Edulastic has an onscreen keyboard for entering math equations. It uses images to show the icons, and these images were not being inverted in dark mode - grey image on black background was basically unusable. This should invert those buttons, to make it easy to read them. 

Before:
image
This commit:
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2998,6 +2998,13 @@ CSS
 
 ================================
 
+edulastic.com
+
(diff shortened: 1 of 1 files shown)


AndrewDaGuy (approved):
Seems fine, Andrew approved :+1:

isaec:
This is so cool (and my first ever pr), thank you!

Gusted (approved):

PR #5290, opened 2021-03-18
Fix for wikipedia.org

rakleed:
- Remove obsolete rule.

Since the icon "Go to the section" was changed (https://github.com/darkreader/darkreader/pull/4794), and also on https://ru.wikipedia.org/wiki/Симметрия ​the spelling of the word with accents breaks down


Gusted (approved):
Good observation.

PR #5294, opened 2021-03-19
Add manen.me to dark sites

manen:


Gusted:
White to my eyes
image

manen:
It uses media queries, if that's not enough I'll change it to dark-only

Gusted:
> It uses media queries, if that's not enough I'll change it to dark-only

The list is exclusively for dark-only sites.

manen:
Okay, it's dark now

Gusted (approved):

PR #5295, opened 2021-03-19
Fix for what-if.xkcd.com

andryandrew:
Removed non-dark-mode-friendly background image on https://what-if.xkcd.com .
Right now it actually gets inverted but with a too light background. Is it possible to invert it completely?
For example, under Filter it looks OK.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -9937,6 +9937,12 @@ INVERT
 .archive-image
 .logo
 
+CSS
+.entry {
+    background-image: none !important;
+    background-color: var(--darkreader-neutral-background);
+}
+
 ================================
 
 whatsapp.com


Gusted:
Not sure what happend, but they do get styled correctly: 
image

andryandrew:
Mmm I see it as incorrect (too light, readability of the website is very low). IMHO, the checkered background should be fully inverted like in does in "filter" mode, or removed if it's impossible. But that's certainly my opinion, so feel free to close the issue if it's outside the scope of the dynamic-theme-fixes :)

Gusted:
> Mmm I see it as incorrect (too light, readability of the website is very low). IMHO, the checkered background should be fully inverted like in does in "filter" mode

Correct. However it still should by default on 4.9.29 be inverted without any fixes. And as forcing such inversion is currently not an option within Dark Reader.

PR #5301, opened 2021-03-20
Added scord.eu

hex3928:
It's completely dark by default


AndrewDaGuy (changes requested):
Can you preserve alphabetical order? Something like this:
```CSS
...
saq.com
scord.eu
sdp.com
...
````

PR #5310, opened 2021-03-21
Sala del Cembalo

erkinalp:


AndrewDaGuy (changes requested):
Can you have one selector per line? Something like this:
```CSS
blah1,
.blah2,
.blah3 {
/* some css here */
}
```

Gusted (approved):

Issue #1384, opened 2019-06-18
Unable to interact with check boxes on Protonmail


geeknik:
While using Firefox 68 + DarkReader 4.7.12, I'm unable to toggle check boxes on Protonmail. Instead of selecting the item like normally, it just acts like a click and opens the email. I'm also unable to star any emails.

![Screenshot_2019-06-18 (17) Inbox geeknik protonmail ch ProtonMail](https://user-images.githubusercontent.com/466878/59704153-16475500-91c1-11e9-8431-0ff574c6cbf9.png)

geeknik:
I reset the Protonmail theme to defaults under Settings->Appearance and it fixed the interaction issue.

ShalokShalom:
I have tried this too and mine are still locking not very smooth. 
Same as in your screenshot. 
Text reads fine now, thanks.

mikerouxfr:
Same issue even after using the reset to default trick.

I guess it is because **DarkReader** use `background-color` instead of `background` in the injected inline CSS.

-----

Here is the custom set rule I used to fix the issue:
```css
================================

mail.protonmail.com

CSS
.customCheckbox-input + .customCheckbox-mask {
    background: #262829;
}
```

Click your **DarkReader** shortcut and find the `🛠 Dev Tools` (name may differ in other locales; Then, insert the rule somewhere in the file (below beta.protonmail.com for example).

Gusted:
I've merged a PR(#4065) to fix this for all users who uses protonmail so it will become a permanent change as the devtools will be overridden after an update.

aha999:
Hi! this is still an issue with protonmail, what can we do ? Protonmail doesn't have a dark theme AFAIK

ShalokShalom:
This is the current situation. 

![Screenshot_20210327_142222](https://user-images.githubusercontent.com/6344099/112722117-e5549200-8f07-11eb-8b52-1c075fcf3a35.png)

Its possible to interact, the mark is barely visible.

slmb3:
#5343 will make the checkboxes appear usable again. If you don't want to wait, delete the `mail.protonmail.com` block in Dark Reader > Dev tools for immediate fix.

slmb3:
@aha999 @ShalokShalom what browser are you both using?

ShalokShalom:
Firefox desktop

slmb3:
I'm not sure what happened, but I put back in the `mail.protonmail.com` fix, and it's working again :/

PR #5343, opened 2021-03-27
Revert "Fix for mail.protonmail.com (#4065)"

slmb3:
This reverts commit 93087bc039c4db6352631a7882718e4ad87ce98e.
Fixes https://github.com/darkreader/darkreader/issues/1384

image
The dynamic-theme-fix makes the non-checked checkbox look a bit better, but also causes the checked checkboxes to appear non-checked. Reverting this makes the non-checked checkboxes look a bit weird, but allows the checked ones to be usable. I tried messing with it a bit to get non-checked to look better and have checked still work, but couldn't easily figure out how to not break it due to some inline CSS stuff. 

So if someone else wants, they can follow up with a fix that makes non-checked look better while not  breaking checked.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5769,15 +5769,6 @@ IGNORE INLINE STYLE
 
 ================================
 
-mail.protonmail.com
-
-CSS
-.customCheckbox-input:not(:checked) + .customCheckbox-mask {
-    background: none !important;
-}
-
-================================
-
 manjaro.org
 
 CSS


Gusted:
Hi @slmb3!

Could you give some instruction how this bug is produced in the first place.
For me it just shows fine and removing that fix just breaks it more.

This is with the current fix on non-beta:
image

slmb3:
@Gusted that's odd. Maybe it's a browser thing? I have the bug on firefox. Are you using chrome?

Gusted:
> @Gusted that's odd. Maybe it's a browser thing? I have the bug on firefox. Are you using chrome? Also can you add a snip if what it looks like when you remove the fix?

Nope, as well on Firefox.


With current fix: 
image

Without current fix:
image

slmb3:
Weird, suddenly it's working for me again. I'll keep an eye on it, thanks.

PR #5356, opened 2021-03-29
fix domain for iubh-fernstudium.de

va1entin:
German university IUBH changed their name and domain to IU - this change updates the domain to make the dynamic theme work again.


Gusted (approved):
Thank you.

Issue #5373, opened 2021-03-30
[Broken Website] Sephora.com


monktr:
# Broken Website
Sephora
## Website

[sephora.com](https://www.sephora.com/)
## How to Reproduce

Go to the site.
See that logo and some icons are staying dark on the dark background.
## Expected behavior

Logos/icons become inverted to be visible on dark background.
## Actual behavior

Logos/icons do not become inverted and are not visible on dark background.
## Screenshots

![image](https://user-images.githubusercontent.com/62554193/112951084-fcae9d80-9108-11eb-81ba-03ce656d5177.png)
## System Information

- OS:  macOS Big Sur 11.2.3
- Browser:  Chrome version 89.0.4389.90
- Dark Reader Version:  4.9.29

## Additional context

I have already created a fix for this and will be submitting a pull request immediately after posting this issue.

PR #5374, opened 2021-03-30
Fix for Sephora.com

monktr:
Inverts icons and logo to make visible
Resolves issue #5373

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -8043,6 +8043,18 @@ CSS
 
 ================================
 
+sephora.com
+
+INVERT
+img[alt="Sephora"]
+svg.css-9uy14h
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -8043,6 +8043,18 @@ CSS
 
 ================================
 
+sephora.com
+
+INVERT
+img[alt="Sephora"]
+svg.css-9uy14h
```
Gusted, line 8050: `-9uy14h`
and all the other random nonsense can break within weeks/months. I consider to have something else as an identifier that isn't random.

monktr: I have submitted an updated pull request with descriptive names instead of the random nonsense.
Pull request #5403

PR #5421, opened 2021-04-01
Add the Godot game engine (godotengine.org)

OnurKader:
[Godot](https://godotengine.org), _it's temporarily changed for April Fools_

[Godot Asset Library](https://godotengine.org/asset-library/asset) is an exception, it needs to be inverted.

Would `inversion-fixes.config` need to be edited for it?

EDIT: The default inversion for the asset library works great

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -303,6 +303,7 @@ glslsandbox.com
 go.dailynow.co
 goaudio.cc
 god.freevar.com
+godotengine.org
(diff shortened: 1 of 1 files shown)


Gusted:
image

OnurKader:
Seems like they have two modes, it appears dark on Firefox but light on QuteBrowser by default, after enabling dark mode it switched to the dark CSS. My apologies, only tested on a private window (Firefox).

Issue #5477, opened 2021-04-05
[Broken Website] Comixology.com


EthanBarczak:
# Broken Website

## Website
[Comixology.com](https://www.comixology.com)

## How to Reproduce

1. Go to comixology.com.
2. Notice the logo is dark on the dark background
3. Go to the unlimited tab
4. Notice the illegible words on the white image.

## Expected behavior

The logo to become inverted and visible. Also for the text to be readable in the unlimited tab.

## Actual behavior

The logo is still darkened and hard to read and the text in the unlimited tab should be darkened because of the light image behind it.

## Screenshots

![](https://i.imgur.com/VxEJCuz.png)

![](https://i.imgur.com/jWiVaSA.png)

## System Information

- OS:  Windows 10 Home
- Browser:   Chrome 89
- Dark Reader Version:  4.9.31

## Additional context

I have a fix already lined up for this site that I will be dropping as soon as I drop this issue. A poor web design led to them making an image the background for the unlimited tab and any type of inversion leads to the comics and banner being inverted as well, so that is why I decided to make the text dark.

Gusted:
#5478

PR #5478, opened 2021-04-05
added fix for comixology.com

EthanBarczak:
inverts the logo to make it visible, also fixes the text on the unlimited tab
resolves issue #5477


Gusted (approved):
Thank you! Looks good.

PR #5509, opened 2021-04-08
Adding mypodficacademia.com to dark-sites.config

dr-off:
This is a website I run and it uses its dark theme by default.


Gusted (approved):
Thank you!

PR #5519, opened 2021-04-09
Update dark-sites.config

benfoster04:
Adding royal.uk due to Duke of Edinburgh death announcement. With darkreader enabled, the bottom section of the page which sits outside of the viewport is grey and causes a visual mismatch between the black of the page.


Gusted:
The site is _temporarily_ a black theme, thereby it's a visual mismatch, nothing major.

PR #5526, opened 2021-04-09
add teddit.net tot dark-sites

TomJansen:


AndrewDaGuy:
Are you sure its dark by default?
image

TomJansen:
Oops sorry, it has a prefers-color-scheme, but it is not dark by default

PR #5527, opened 2021-04-10
Fix code blocks on developer.apple.com

wangy412:
Before the code blocks' background was white but the text was also why so you couldn't see the text.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2603,7 +2603,6 @@ pre, .code-listing {
     background-color: var(--darkreader-neutral-background) !important;
 }
 
-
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -2599,6 +2599,9 @@ CSS
 .disabled {
     color: ${gray} !important;
 }
+pre, .code-listing {
```
Gusted, line 2602:
```suggestion
pre,
.code-listing {
```
Code-Style.

Gusted (commented):
See suggestion.

Gusted (approved):
Thank you!

PR #5561, opened 2021-04-15
Added Blur Busters, GitHub, Material UI and Twitch to dark sites exceptions

nobelharvards:
[Blur Busters](https://blurbusters.com), GitHub, [Material UI](https://material-ui.com) and [Twitch](https://www.twitch.tv) are browser/OS theme aware and will automatically switch accordingly.

Can be confirmed on Windows 10 by toggling OS wide theme while visiting those sites in a private tab. Twitch only checks on page load and may need a refresh.

image


Gusted:
And all those kinds of site aren't allowed within this site list.

PR #5572, opened 2021-04-17
Fix svg text for 1843 articles on economist.com

pranavmangal:
Fixes some svg text for articles from the [1843 magazine](https://www.economist.com/1843) on The Economist.

Before:
before

After:
after


AndrewDaGuy (approved):
LGTM, Andrew Approved c:

Gusted (approved):
Thank you!

PR #5576, opened 2021-04-17
Don't request resources when blocked

### LINKED ISSUE #5556: [BUG] Background image that should be blocked by uBO is appearing
# Bug Report

## Describe the bug
I honestly am not sure at this point what is going on here.  On the surface, it appears that Dark Reader is showing an image that should have been blocked by uBlock Origin.  This needs more investigation to determine what is going on.

I've seen this issue before, but it was not easy to replicate.  Finally, I've found a place where I can easily replicate it.  See STR.

## Steps to reproduce
1. Install uBlock Origin (uBO) v1.34.0 in Firefox
2. Install Dark Reader v4.9.31 in Firefox
3. Enable Dark Reader v5 beta mode in v4.9.31 and be sure Dynamic dark mode is enabled
4. Go to https://sourceforge.net/
5. Notice the background image that looks like this:
DR1
6. Add uBO filter: `||fsdn.com/con/$important` and save filters in uBO
7. Go back to the Firefox tab with https://sourceforge.net/
8. Hold down the `Shift` key, and press Firefox's refresh button to refresh the page while bypassing the browser's cache
9. Notice that the background mentioned above is still present
10. Use Dark Reader's drop-down menu to disable Dark Reader
11. Notice that the background mentioned above disappears
12. Use Dark Reader's drop-down menu to re-enable Dark Reader
13. Notice that the background mentioned above re-appears

## Expected behavior
The background image to no longer appear regardless of whether or not Dark Reader is enabled.

## Actual behavior
The background image only appears when Dark Reader is enabled.

## Screenshots
DR1

## System Information
- OS: Windows
- Browser: Firefox 87.0
- Dark Reader Version: v4.9.31

## Additional context
Normally, this would just be a simple case of using the Dynamic Theme Editor to tweak the dynamic theme.  But in this case, I don't currently understand why the background image is being shown at all given that it should be blocked by uBO.  It's like Dark Reader is bypassing uBO filtering, but this requires further investigation before that conclusion can be made.

Gusted:
- When a content blocker(ex UBlock origin) blocked a resource we set for such URL an empty value so we won't "bypass" the blockade. 
- Resolves #5556

### DIFF
--- a/src/background/tab-manager.ts
+++ b/src/background/tab-manager.ts
@@ -78,6 +78,15 @@ export default class TabManager {
 
         const fileLoader = createFileLoader();
 
+        const onWebRequestError = (responseDetails: chrome.webRequest.WebResponseErrorDetails) => {
+            if (responseDetails.error.toLowerCase().includes('abort')) {
+                fileLoader.set({url: responseDetails.url, responseType: 'data-url', value: ''});
+                fileLoader.set({url: responseDetails.url, responseType: 'text', value: ''});
+            }
+        };
+
+        !isThunderbird && chrome.webRequest.onErrorOccurred.addListener(onWebRequestError, {urls: ['<all_urls>']});
+
         chrome.runtime.onMessage.addListener(async ({type, data, id}: Message, sender) => {
             if (type === 'fetch') {
                 const {url, responseType, mimeType} = data;
--- a/src/background/utils/network.ts
+++ b/src/background/utils/network.ts
@@ -98,6 +98,7 @@ interface FetchRequestParameters {
     url: string;
     responseType: 'data-url' | 'text';
     mimeType?: string;
+    value?: string;
 }
 
 export function createFileLoader() {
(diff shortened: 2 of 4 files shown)


Gusted:
[binaries.zip](https://github.com/darkreader/darkreader/files/6329406/binaries.zip)
For anyone who wants to try-out.

ntnguyen1234:
I have checked the testing xpi for firefox and so far it resolves the sourceforge issue

alexanderby:
While the fix seems simple, I need to check what happens after adding one more permission, maybe the extension will be blocked for all users until they accept the new permission. If so, it should be added as optional permission with some user's flag (e.g. *Hide ad-blocked images*, I don't know).

And another problem, maybe you've heard, Manifest v3 will come out by the end of this year, and maybe this API will be broken. Same story for your another great PR with white flash fix.

Gusted:
>  If so, it should be added as optional permission with some user's flag (e.g. _Hide ad-blocked images_, I don't know).

Sounds good in my ears.

Yep, I've head about the ~~stupid proposed~~ Manifest v3. Currently just Chrome canary has it? Under a flag right, might give it a shot and check it out what got broken/needs changes. Anyway as It's a non-standard it can change any moment and checking it out now could be different than checking it out next week. 
While it's just the web request that get replaced. The other trick with declerativeContentscripts stays 
image
Might as well just give them the faster execution to Firefox if their isn't another workaround.

lodenrogue (approved):

Gusted:
Hmm I still would love to see this in some form of another, but currently (i) requires a new *invasive* permission (ii) is not MV3-compatible.

PR #5618, opened 2021-04-22
Ubereats map and logo fix

roberestarkk:
Some inversions for UberEats to display the map on the order tracking and checkout pages, and to site-wide invert the Logo because it looks nicer that way.

The reason for the long-form css selector chaining is to avoid inadvertently applying inversions on other pages, so I tried to make the selectors too specific to match other pages.

I initially tried to do something like *.ubereats.com/*/orders to narrow down the changes to only a subset of pages, but had no luck. Perhaps a feature suggestion for next version if I'm not the only person crazy enough to want to apply styles like this.

Love the plugin, long may it reign!

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -139,6 +139,7 @@ IGNORE INLINE STYLE
 
 INVERT
 img[alt*='Home']
+body > #root > #wrapper > #main-content h1 + div
 body > #root > #wrapper > div[class^='c']
 body > #root > #wrapper > div[class^='c'] > header
 body > #root > #wrapper > div[class^='c'] > footer
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -135,6 +135,17 @@ IGNORE INLINE STYLE
 
 ================================
 
+*.ubereats.com
+
+INVERT
+img[alt*='Home']
+body > #root > #wrapper > div[class^='c']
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -135,6 +135,18 @@ IGNORE INLINE STYLE
 
 ================================
 
+*.ubereats.com
+
+INVERT
+img[alt*='Home']
+body > #root > #wrapper > #main-content h1 + div
+body > #root > #wrapper > div[class^='c']
```
Gusted, line 143: I think if you remove `body > #root` here and just leave `#wrapper` as control you should be good to go.

roberestarkk: Quite possibly, yes!    
But I didn't want to repeatedly comb through the whole functionality of the site to try to whittle down the selector to the minimum required (particularly since my food was arriving!).    
All it'd take is for one part of one page under one circumstance to be similar enough inside it's #wrapper, and then the selector is interfering with parts that work just fine.    

If you're asking me why I didn't, I preferred to err on the most cautious side from the beginning, and I ran out of momentum when my food arrived. 
If you're asking me to do it and see what happens because the selectors for fixes should be as minimal and efficient as possible as a matter of standard practice/policy, I can do that no worries, but it'll have to wait until I order something from Uber again to test it properly.

Gusted: Got it! Just a minor issue nothing breaking 😄

Gusted (approved):

PR #2310, opened 2020-04-17
Filter fix

Gusted:
As of the chromium version `81.0.4035.0` 
To comfirm yourself [this](https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Win_x64/733497/) Version is for windows 64 bit to download `81.0.4034.0` where the patch was not into the source code of chromium,
[The issue](https://bugs.chromium.org/p/chromium/issues/detail?id=501582)
[The fix that has been made in chromium](https://chromium-review.googlesource.com/c/chromium/src/+/1979258)

Some major credits to @alexanderby To pointing out the issue and letting me know setting the background-color to white should fix this issue. (@sierratango88 Pointing this also) I should pay closer attention to comments next time...

Closes #2256

PR #5637, opened 2021-04-25
Fix background on nextjs.org

alexalekseyenko:
Extend dark background down the entire length of the page in Filter and Filter+ modes.

### DIFF
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -1512,6 +1512,15 @@ INVERT
 
 ================================
 
+nextjs.org
+
+CSS
+body.vsc-initialized {
+    height: auto;
+}
+
+================================
+
 nodejs.org
 
 INVERT


Gusted:
Where can I observe this change?

alexalekseyenko:
It's showing up for me on most of their pages other the landing page and Showcase. For instance, the blog:
https://nextjs.org/blog

Also, my fix is wrong anyway; the `vsc-initialized` class is coming from another extension (https://github.com/igrigorik/videospeed in case you're curious). Sorry, it was late 😅

Now that we're on the topic, though, I've noticed this happening on quite a few pages; the background looks fine at first, but as soon as you start scrolling past the initial viewport, the background has that gray color that makes everything hard to read. Another example of this is:
https://www.kubeflow.org/docs/

Invariably there's some element near the top with a `height: 100%;` or `height: 100vh;` which seems to be causing it. For instance, for the above page, try running this in the console on the above page:
`document.querySelectorAll('.container-fluid.td-outer')[0].style.height = 'auto'`

I haven't had a chance to dig into the Dark Reader source, so I'm not sure what the root cause is, but I'm assuming changing the `height` property on every top-level element is probably not what I should be doing. Do you know why this might be happening?

This seems to only happen with the Filter/Filter+ modes. I know that the Dynamic mode is supposed to be the official one, but I've found it to be a bit sluggish and this issue with the viewport-sized dark background is actually the only glaring problem I've had with Filter/Filter+.

Gusted:
> but I'm assuming changing the height property on every top-level element is probably not what I should be doing. Do you know why this might be happening?

Yup, most likely a bug within browsers itself as I couldn't see why it was changed colors other than some bug.  Filter/Filter+ relies on the `filter:` CSS Property https://developer.mozilla.org/en-US/docs/Web/CSS/filter

> Now that we're on the topic, though, I've noticed this happening on quite a few pages; the background looks fine at first, but as soon as you start scrolling past the initial viewport, the background has that gray color that makes everything hard to read. Another example of this is:
https://www.kubeflow.org/docs/

I see, I cannot find an explanation for it so I'm going to have a little fiddle around with this. btw are you using Firefox or Chromium? Otherwise Il need 2 bug reports for 2 browsers. As  I can see it happening within Firefox.

Gusted:
Seems like I overlook some things, so we effectively invert the `html` element, however on those site the `html` is set to a fixed height and the `body` element is longer. Than such gray/black will come as the `html` element seems to get extra styled with the filter and all other content get a "normal" filter.

alexalekseyenko:
It seems this functionality was added in #2310 but I'm not sure it's doing what we want. I can understand why we might set `background` for the `html` element (as the CSS filter can't invert a color that isn't there), but if we apply the color matrix to the background color before setting it, then when it gets inverted again by the CSS filter, we end up getting something untoward, like `rgb(128, 128, 127)`. Could we do this instead?
https://github.com/darkreader/darkreader/pull/5927

PR #5644, opened 2021-04-26
Fix embedded code editor background under dash.cloudflare.net

Razboy20:
CSS fix for the Monaco editor under Cloudflare's built-in worker editor and changes colors to match theming (from a previous bright yellow).

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2485,7 +2485,7 @@ a[data-testid^="link-homepage"]
 
 CSS
 .monaco-editor-background, .monaco-editor .margin {
-    background-color: rgb(24 27 29) !important;
+    background-color: ${rgb(226 229 231)} !important;
 }
 
 ================================
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2483,6 +2483,11 @@ dash.cloudflare.com
 INVERT
 a[data-testid^="link-homepage"]
 
+CSS
+.monaco-editor-background, .monaco-editor .margin {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -2483,6 +2483,11 @@ dash.cloudflare.com
 INVERT
 a[data-testid^="link-homepage"]
 
+CSS
+.monaco-editor-background, .monaco-editor .margin {
+    background-color: rgb(24 27 29) !important;
```
Gusted, line 2488:
```suggestion
    background-color: var(--darkreader-neutral-background) !important;
```

Razboy20: @Gusted --darkreader-neutral-background does not match the theme of the website sadly (background becomes way too dark in comparison)

Gusted: It's recommended to still use this, because when users changes their colors due to brightness/contrast/sepia etc. the variable will change according to it, whereby this is static and will be out of place.

Razboy20: Ah, is there a way to mix/manipulate the --darkreader-neutral-background color, similar to scss color modules?

Gusted: Not really, it's controlled by the user settings, however you could switch to `${originial color}` eg `${white} ` will turned to be black-ish.

Razboy20: Ah, that gave me an idea. I flipped the rgb color, and set that as the original, and it looks good now!

Gusted (approved):

PR #5674, opened 2021-04-28
Add seaofthieves.fandom.com to dark-sites.config

Abragus:
seaofthieves.fandom.com is already in dark mode:

Sea of thieves Wiki


Gusted (approved):
Thank you!

PR #5709, opened 2021-05-02
properly invert overview icons on collegeboard apclassroom

isaec:
before and afters below - not sure if this page can be easily accessed.

image

image

image

image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -747,6 +747,14 @@ IGNORE IMAGE ANALYSIS
 
 ================================
 
+apclassroom.collegeboard.org
+
+INVERT
+a.overview-icon svg
+button.Zoom--expander path
+
+================================
+
 api.kde.org
 
 INVERT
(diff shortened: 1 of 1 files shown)


isaec:
with the better selector, the magnify icon now looks like:
image

src/config/dynamic-theme-fixes.config
```diff
@@ -747,6 +747,14 @@ IGNORE IMAGE ANALYSIS
 
 ================================
 
+apclassroom.collegeboard.org
+
+INVERT
+a.overview-icon svg
+button.Zoom--expander svg g > *:first-child
```
Gusted, line 754:
```suggestion
button.Zoom--expander svg g > :first-child
```
In the world of CSS no need to specify the `*` in this case.

Gusted (approved):
Thank you!

isaec:
yep, thank you! Is that sort of selector the right way to only select the first path of an svg otherwise?

Gusted:
> yep, thank you! Is that sort of selector the right way to only select the first path of an svg otherwise?

Yup! That's the correct way to specify that.

isaec:
> > yep, thank you! Is that sort of selector the right way to only select the first path of an svg otherwise?
> 
> Yup! That's the correct way to specify that.

got it, thanks again - i'm pretty new to webdev.

PR #5718, opened 2021-05-03
joemonster.org hard fixes

FadeMind:
Hard workaround messy CSS

https://joemonster.org


Gusted (commented):
Seems more like this has to be fixed within dark reader

Gusted:
As had some good amount of work onto the variables implementation code - I came across this PR again, but I don't remember what's broken as I currently don't see anything to be fixed. @FadeMind is the site good for you?

FadeMind:
Default (without this PR fix) dark reader look:




It is not good. Some hard coded are ugly. Thats why I made hardcoded PR. 

CSS on this site is nightmare. 

This PR fix mostly all major issues visuals.

Gusted:
This could be me but I don't see most of the issues that's are being shown, other than some of the more yellow elements not being read-able.

PR #5719, opened 2021-05-03
Detecting dark pages

Gusted:
@alexanderby :) The hackery done in your precious `variables.ts` I'm sorry.

So yeah, it will detect dark pages and disable accordingly, as this is just a prototype a few things have to be adressed.
1. 'main element' so as for now we'll check again the `html` and `body` element as those are 99% used as main elements and thus pretty accurate, problem 1%. Sites like reddit.com have some nested element which is then used as `main` element. Il propose some new selector or section? inside the fixes config called `MAIN ELEMENT` which then can be used to fix sites like reddit by having
```CSS
reddit.com

MAIN ELEMENT
body > div > div
```
2. False postives, as for now theirs 1 false positive known, duckduckgo.com search results pages. By default they inject a `no-theme` which for dark-preffered users will be a dark color and dark reader will think 'OH yeah dark page!' well it is removed in matter of split seconds, so another section!
```CSS
duckduckgo.com

IGNORE POSSIBLE DARK SELECTOR
.no-theme body
```
3. Detecting native website support, as we are now able to detect* dark pages we also know when to leave the site when they have native website support also it is possible to watch for attribute changes on the `main element` and to still watch stylesheets changes but don't modify them, just to check if the 'darkness' is gone and thus solving another outstanding problem.
4. Force, what if user X still want dark reader on page x well have to come op with such logic whereby the pop-up knows that the site is detected dark and we can send dark page request(with some additonial argument like `force: true`?) to enable(new list? to remember this) dark reader and disable the automatic dark detector thingy.

- Resolves? TODO as this solution can resolve multiple fundamental-different problems.


alexanderby:
Nice try. Maybe let's have a look at how Disqus detects a dark theme?

Currently it looks too complex, making some work for each variable seems like an expensive task.

I tried to measure, for some URLs (for some reason other than home page) it (`getConnectionMessage()` method) takes about 200ms, for sites like Gmail 200ms for many of its iframes. So maybe that's one of the reasons why it loads slow?

Gusted:
> Nice try. Maybe let's have a look at how Disqus detects a dark theme?

You mean Disqus.com? Not sure why but they will fall with reddit.com under the special cases and need such new section to define which element to assign as 'main' element.



> Currently it looks too complex, making some work for each variable seems like an expensive task.

It is complex, it might require a refactor of how we collect styles and then distribute them among callback(one for stylesheet-modifier and another to check if any of the styles is considered dark) so it will look more clean but still have the same functionality

. 

> I tried to measure, for some URLs (for some reason other than home page) it (`getConnectionMessage()` method) takes about 200ms, for sites like Gmail 200ms for each of its iframes. So maybe that's one of the reasons why it loads slow?

That sounds like #2517 need that "is it possible in the list" functionality we talked about to migrate such issues.

alexanderby:
I wish it will be as simple as possible, since it is a side feature. These auto-detected websites should be stored in some list, maybe later merged with Dark Sites (let's not remove it on this stage). So a user will be able to toggle the style if it was falsy-positive.

Anyway the optimisation with 2nd level and 3rd+ level domains and Sets should be useful, because it uses hash codes internally and comparison is very fast, unlike what is done by regular expressions.

Gusted:
> These auto-detected websites should be stored in some list, maybe later merged with Dark Sites (let's not remove it on this stage). So a user will be able to toggle the style if it was falsy-positive.

Just to be on the same level a new file under src/config? or an memory list. As I tend to have more memory list than actual hardcoded list.

> Anyway the optimisation with 2nd level and 3rd+ level domains and Sets should be useful, because it uses hash codes internally and comparison is very fast, unlike what is done by regular expressions.

It should be very fast check before doing the regex to just "Is it possible that's this in the list?" and if it's true you go to the regex because sitelist can contain regex or $ ^ etc which cannot be accounted for in such list.

swrobel:
This is definitely cool! It would still be great to have this merged as well, for a really simple solution that uses a standard tag: https://github.com/darkreader/darkreader/pull/3902

Gusted:
https://github.com/darkreader/darkreader/pull/5719/commits/fc2a8e6527f7b3bcb50173230f99cd31538077bd fixes the duckduckgo.com problem

Mahdi-Noor-Hasan:
It would be really great. Waiting for this freature.

src/inject/dynamic-theme/is-it-dark.ts
```diff
@@ -0,0 +1,71 @@
...
+            backgroundColor = tryParseColor(variableStore.getRawValue(backgroundValue));
+        } else {
+            backgroundColor = tryParseColor(backgroundValue);
+        }
+
+        if (!backgroundColor) {
+            return false;
+        }
+
+        const backgroundColorHSLA = rgbToHSL(backgroundColor);
```
Artoria2e5, line 47: This likely matters little, but I would prefer to have a better lightness algo than HSL. sRGB relative luminance (Y) would be a good choice in this case (as it's simple to calculate), although the even simpler luma (Y') will work fine too.

src/inject/dynamic-theme/is-it-dark.ts
```diff
@@ -0,0 +1,71 @@
...
+async function waitTillBody(): Promise<void> {
+    return new Promise<void>((resolve) => {
+        if (!bodyObserver) {
+            resolve();
+        }
+        bodyCallbacks.add(resolve);
+    });
+}
+
+export function IsItDark(variableStore: VariablesStore, selectorText: string, styleRule: CSSStyleDeclaration): boolean | Promise<boolean> {
```
Artoria2e5, line 20: Probably better to have three classes of output, dark, medium, and light.

src/inject/dynamic-theme/is-it-dark.ts
Gusted: Not really sure if we should switch to sRGB. We do already uses it for our image algorithm. But all modifications to CSS is being done to HSL and really would like to keep that in par. If we switch to sRGB it might be better to switch all HSL-luminance algorithms to sRGB-luminance so it will be in par and avoid weird situations.

src/inject/dynamic-theme/is-it-dark.ts
Gusted: I'm not sure. What should DR do on the "medium" output?

Artoria2e5: Do nothing, because the page does not make enough sense.

Gusted: That is the behavior of `dark` output. So Ill prefer to stay with a simple boolean as output.

Gusted:
> Possibly dumb question: is there some way you could take an actual snapshot of the page (sort of like a frame buffer), and then do some quick checks to see if most of the pixels are on the darker side, i.e. brightness below a certain value? If that's even possible, can it be done in a performant way? If that would work, you could bypass having to figure what the main element is and all that stuff that can break or be a pain to manage.

That was already proposed somewhere, but that will introduce a lot of problems like that if the site isn't dark we will have a long time of white content which isn't nice. Thereby it will just take longer to load DR onto the site if we don't do it from the start.

Ambient-Impact:
> > Possibly dumb question: is there some way you could take an actual snapshot of the page (sort of like a frame buffer), and then do some quick checks to see if most of the pixels are on the darker side, i.e. brightness below a certain value? If that's even possible, can it be done in a performant way? If that would work, you could bypass having to figure what the main element is and all that stuff that can break or be a pain to manage.
> 
> That was already proposed somewhere, but that will introduce a lot of problems like that if the site isn't dark we will have a long time of white content which isn't nice. Thereby it will just take longer to load DR onto the site if we don't do it from the start.

Fair enough!

emtiu:
Suggestion: The Dark Reader icon in the toolbar could reflect the current state of this detection. Currently, it displays two states:
1. Dark Reader inactive (face in the light)
2. Dark Reader active (face in the dark)

It would be very useful if it reflected additional states:
1. Dark Reader inactive
2. Dark Reader active, and:
    1. working normally,
    2. manually disabled for the current page
    3. automatically disabled by dark-detection

Gusted:
Superseeded by https://github.com/darkreader/darkreader/pull/7995

PR #5726, opened 2021-05-04
Add debounce to shortcut

### LINKED ISSUE #5725: [BUG] Toggle current site is performed multiple times if key is pressed for a longer time.
# Bug Report

## Describe the bug
On pressing the shortcut keys for toggle current site, if the key is accidentally pressed for a longer time, toggle happens multiple times. 

## Steps to reproduce
When the assigned shortcut key is pressed for a longer time, it happens. I checked and this behaviour is consistent with all other shortcuts as well.

Step:
- Assign a shortcut in  [shortcuts tab](chrome://extensions/shortcuts)
- Press that shortcut for a longer time
 
## Expected behaviour
- It should only toggle once when I press the shortcut once, either on key up or key down.

## Actual behaviour
- Toggles multiple time

## Screenshots
 n/a

## System Information

- OS: macOS
- Browser: Chrome 90
- Dark Reader Version: 4.9.32

## Additional context
 n/a

Gusted:
- Resolves #5725


src/background/extension.ts
```diff
@@ -148,7 +149,7 @@ export class Extension {
             // Fix for Firefox Android
             return;
         }
-        chrome.commands.onCommand.addListener(async (command) => {
+        chrome.commands.onCommand.addListener(debounce(500, async (command) => {
```
lodenrogue, line 152: Is this addListener similar to the EventEmitter addListener? If so, should we move each event into it's own encapsulated function? It seems this is a catch-all function for any kind of event.

Gusted: > Is this addListener similar to the EventEmitter addListener?

EventEmitter is pure Node.js way of Events and doesn't correspond with this kind of events that are executed from internal code.



> If so, should we move each event into it's own encapsulated function? It seems this is a catch-all function for any kind of event.

I'm not sure why because if we will do this with moving to such encapsulated function more know as block-scoped functions we will our ability to call private functions `this.` whereby having the () => format won't do this.

lodenrogue (approved):

PR #5731, opened 2021-05-05
Update Browser and DR versions

AndrewDaGuy:
Changes:
- Chrome 89 -> 90
- Firefox 87 -> 88
- Dark Reader 4.9.30 -> 4.9.32


Gusted:
I don't think we should bump it each time it's rather an indication of the format

PR #5734, opened 2021-05-05
Fix for nicehash.com

AndrewDaGuy:
Changes:
- Inverted hash power image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -7652,6 +7652,7 @@ nicehash.com
 
 INVERT
 .chart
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -7652,6 +7652,7 @@ nicehash.com
 
 INVERT
 .chart
+img[src*="hashpower@2x.5817a6d2.png"]
```
Gusted, line 7655:
```suggestion
img[src*="hashpower@2x"]
```

Gusted (approved):

PR #5735, opened 2021-05-05
Fix for killedbygoogle.com

AndrewDaGuy:
rip google cardboard 😔


Gusted (changes requested):
Like Alexander says this config should extensively grow especially due to this kind of selectors, find a common one and use it.

```css
img[src^="Itematoms__Icon"]
```

PR #5740, opened 2021-05-05
Made contributing.md more readable part 2

AndrewDaGuy:
didn't notice  "variety of ways" was was shown two times


Gusted (approved):
Good catch!

PR #5746, opened 2021-05-06
kreato.cf is dark

kreat0:


Gusted (approved):

src/config/dark-sites.config
```diff
@@ -410,6 +410,7 @@ kiwifarms.net
 knockout.chat
 kodenames.io
 krunker.io
+kreato.cf
```
Gusted, line 413: Put this entry before krunker.io

Gusted (approved):

PR #5773, opened 2021-05-09
Make search icon visible on habr.com

jtraub:
Search icon is invisible on habr.com for some reason which makes it impossible to open search bar on the site (you need to click on the icon to toggle search bar)
image

Proposed change makes it visible and clickable
image


src/config/dynamic-theme-fixes.config
```diff
@@ -5163,6 +5163,9 @@ CSS
 html {
     text-shadow: none !important;
 }
+.main-navbar .icon-svg{
```
Gusted, line 5166:
```suggestion
.main-navbar .icon-svg {
```

jtraub: @Gusted fixed

Gusted (approved):

PR #5775, opened 2021-05-09
d3ward.github.io is dark

FadeMind:
https://d3ward.github.io/toolz/src/adblock.html

* dark via switch icon on right top


Gusted:
Not dark-by-default

PR #5777, opened 2021-05-09
Removed white stripes on LearnOpenGL hover element

CourtneyAbraham:
There are white stripes on the hover element on LearnOpenGL, this just sets the background image to none and opacity to 1.
this way it's actually possible to read what's there

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -6279,7 +6279,7 @@ learnopengl.com
 CSS
 #hover {
     background-image: none !important;
-    opacity: 1;
+    opacity: 1 !important;
 }
 
 ================================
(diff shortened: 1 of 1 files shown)


Gusted:
Hi!

In which behavior does this hover element shows up? It seems to be hidden by default.

Regards,
Gusted

CourtneyAbraham:
[Here is a picture of the behaviour without what I've added](https://prnt.sc/12qp9yu)
[And here is a picture with what I've added enabled](https://prnt.sc/12qpcjj)

Both times I am hovering over the region I've highlighted red

src/config/dynamic-theme-fixes.config
```diff
@@ -6274,6 +6274,16 @@ INVERT
 
 ================================
 
+learnopengl.com
+
+CSS
+#hover {
+    background-image: none !important;
+    opacity: 1;
```
Gusted, line 6282:
```suggestion
    opacity: 1 !important;
```

Gusted (approved):

PR #5790, opened 2021-05-11
Fix for cdc.gov

rakleed:
- Invert logo.

Example of site page: https://www.cdc.gov/ticks/removing_a_tick.html


rakleed:
I specified a path with an attribute, since when I specify only the `.cdc-logo` class, the logo breaks on the [main page](https://www.cdc.gov/).

Gusted (approved):
Thank you!

PR #5801, opened 2021-05-12
Asana: Fix white bg on PortfolioItemRow class

JamesDBartlett:
Applies the same fix as was applied to the `InboxExpandableThread` class on the line above this (line 1029 currently).
Selecting a Portfolio Item causes a modal window to appear, which blurs everything else behind it, so there's no need to apply the same fix to the `PortfolioItemRow--selected` class in this case.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1026,14 +1026,12 @@ INVERT
 .DatePickerCalendarDate--today .DatePickerCalendarDate-button::after
 
 CSS
-.InboxExpandableThread.InboxExpandableThread--selected {
-    background: var(--darkreader-neutral-background) !important;
-}
+.InboxExpandableThread.InboxExpandableThread--selected,
 .PortfolioItemRow {
     background: var(--darkreader-neutral-background) !important;
 }
 .PortfolioItemRow:hover {
-    background: var(--darkreader-neutral-hover) !important;
+    background: ${#FFF} !important;
 }
 
 ================================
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -1029,6 +1029,12 @@ CSS
 .InboxExpandableThread.InboxExpandableThread--selected {
     background: var(--darkreader-neutral-background) !important;
 }
+.PortfolioItemRow {
+    background: var(--darkreader-neutral-background) !important;
+}
+.PortfolioItemRow:hover {
```
Gusted, line 1035:
```CSS
var(--darkreader-neutral-hover)
```
This isn't a variable.

src/config/dynamic-theme-fixes.config
```diff
@@ -1029,6 +1029,12 @@ CSS
 .InboxExpandableThread.InboxExpandableThread--selected {
```
Gusted, line 1029:
```suggestion
.InboxExpandableThread.InboxExpandableThread--selected,
.PortfolioItemRow  {
```

src/config/dynamic-theme-fixes.config
```diff
@@ -1029,6 +1029,12 @@ CSS
...
     background: var(--darkreader-neutral-background) !important;
 }
+.PortfolioItemRow {
```
Gusted, line 1032: With the suggestion above you don't need this declaration anymore.

src/config/dynamic-theme-fixes.config
JamesDBartlett: Hmm. If I remove this line, the hover-highlighting feature stops working, so it must be doing something. Any ideas?

Gusted: I don't have a clue, it should be ignored CSS as it's invalid variable. So I suggest setting it to something like `${#333}` or anything else that's appropriate.

Gusted (approved):
Thank you!

JamesDBartlett:
Glad I could help!

PR #5802, opened 2021-05-13
 Added roosterteeth.fandom.com to dark sites exceptions and corrected spelling mistake "sponser" to "sponsor" in README.md.

nobelharvards:
[roosterteeth.com](https://roosterteeth.com) is already dark, but not subdomains like [devices.roosterteeth.com](https://devices.roosterteeth.com) or [store.roosterteeth.com](https://store.roosterteeth.com).

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -1,4 +1,3 @@
-^roosterteeth.com
 0bin.net
 0x00sec.org
 12bytes.org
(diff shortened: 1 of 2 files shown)


src/config/dark-sites.config
```diff
@@ -1,3 +1,4 @@
+^roosterteeth.com
```
Gusted, line 1: We actually don't allow allow `^` `$` as of this moment they aren't alphabetic ordered.

PR #5821, opened 2021-05-14
account.xiaomi.com bg profile fix

FadeMind:
https://account.xiaomi.com/ (you need an account to login there)
Fix: removing white background profile image section. I removed sensitive details from images.


Gusted (approved):
Seems good.

PR #5823, opened 2021-05-14
Fix for dev.to

rakleed:
- Fix logo.


Gusted (approved):
Tricky, but assuming they won't change there logo's soon, good to go.

PR #5827, opened 2021-05-15
Add wormhole.app and www.wickeditor.com/editor/

Elaborendum:
I was thinking of adding snapdrop.net too, but I don't know if it's right. It changes according to the system theme, but its dark mode's background is erased by DR, plus, there are other instances too, what to do?


Gusted:
>  It changes according to the system theme,

Big no.

PR #5835, opened 2021-05-16
Fix for drive.google.com

rakleed:
- Invert icons in file picker in the main interface.

Example of site page: https://drive.google.com/drive

Discussion: https://github.com/darkreader/darkreader/pull/5208


Gusted (approved):
Seems good, nothing seems broken.

PR #5849, opened 2021-05-18
Fix for developer.mozilla.org

rakleed:
- Invert footer logo.
- Correctly invert icons in the browser compatibility table.

I tried to make the rules shorter, but I couldn't. Example of site page: https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme


Gusted:
> I tried to make the rules shorter, but I couldn't. Example of site page:

At least it's working now and some 100 lines of rules.

Gusted (approved):

PR #5862, opened 2021-05-20
Fix for flickr.com logo

datygra:
Fix for the flickr logo (svg) in the header/nav bar


Gusted (approved):
Seems good, thank you!

PR #5866, opened 2021-05-21
Fix vultr.com

jackreeds:
The site has a couple of `svg` elements which aren't supposed to be modified. And a white patch.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -165,8 +165,13 @@ CSS
 }
 
 IGNORE INLINE STYLE
-svg
-svg *
+.icon-ui*
+.icon-user*
+.icon-sm*
+.svg-logo*
+.svg-illustration*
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -157,6 +157,19 @@ body > #root > #wrapper > div[class^='c'] > #main-content > div:first-child
...
 
+*.vultr.com
+
+CSS
+.svg-banner-shape {
+    fill: var(--darkreader-neutral-background);
+}
+
+IGNORE INLINE STYLE
+svg
```
Gusted, line 168: These are too fague and will target all current SVG's and future one's.

jackreeds: Alright, I'll expand them a bit them.

jackreeds: Can you go through it again?

PR #5868, opened 2021-05-21
Add vultr.com

jackreeds:
The site has a couple of `svg` elements which aren't supposed to be modified. And a white patch.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -157,6 +157,24 @@ body > #root > #wrapper > div[class^='c'] > #main-content > div:first-child
 
 ================================
 
+*.vultr.com
+
+CSS
+.svg-banner-shape {
+    fill: var(--darkreader-neutral-background);
+}
+
+IGNORE INLINE STYLE
+.icon-ui*
+.icon-user*
+.icon-sm*
+.svg-logo*
+.svg-illustration*
+.svg-product*
+.svg-shape*
+
+================================
+
 01net.com
 
 CSS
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -157,6 +157,24 @@ body > #root > #wrapper > div[class^='c'] > #main-content > div:first-child
...
 
+*.vultr.com
+
+CSS
+.svg-banner-shape {
+    fill: var(--darkreader-neutral-background);
+}
+
+IGNORE INLINE STYLE
+.icon-ui*
```
Gusted, line 168: Or I've missed something important in the syntax of CSS, but this shouldn't work and especially not with our matching system. And will probably error dark reader into something like:
'.icon-ui*' is not a valid selector

jackreeds: I intended to tell darkreader to not touch `.icon-ui` at all, so I'm selecting all its children. What if it is changed to `.icon-ui *` (with a space in-between)?

src/config/dynamic-theme-fixes.config
```diff
@@ -157,6 +157,24 @@ body > #root > #wrapper > div[class^='c'] > #main-content > div:first-child
 
 ================================
 
+*.vultr.com
+
+CSS
+.svg-banner-shape {
+    fill: var(--darkreader-neutral-background);
```
Gusted, line 164: don't forget the `!important;`

jackreeds: Hmm, maybe it is a good idea to tell the auto-formatter to do this too.

Gusted: > maybe it is a good idea to tell the auto-formatter to do this too.

Not really, certain fixes will break when you add `!important;` and it's hard to migrate this false-positives

Gusted (approved):

PR #5874, opened 2021-05-22
Fix for google.com

pnaf:
Fixes menu button on mobile view.

Before:
png

After:
png

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -12772,7 +12772,7 @@ a[href*="about/products"][title]
 .JOmIqc
 .hLcKi
 #EcMbV
-#xRRFWe
+#wrap a[href="/mobile/"]
 .vk-sf-b
 .vk-t-btn
 .ChZgtd div::before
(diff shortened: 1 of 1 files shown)


Gusted:
Hi!

Could you share a link whereby I can observe this behavior?

pnaf:
Hi!

I can observe this behavior on https://www.google.com/ using a Firefox for Android user agent (e.g., `Mozilla/5.0 (Android 11; Mobile; rv:88.0) Gecko/88.0 Firefox/88.0`).

src/config/dynamic-theme-fixes.config
```diff
@@ -12772,6 +12772,7 @@ a[href*="about/products"][title]
 .JOmIqc
 .hLcKi
 #EcMbV
+#xRRFWe
```
Gusted, line 12775: I think a better fix is something along the lines:

```suggestion
#wrap a[href="/mobile/"]
```

So it doesn't rely on the obscure selector

Gusted: Reference:

image

pnaf: Sounds good to me!

Gusted (approved):

PR #5879, opened 2021-05-24
Add orama-interactive.itch.io/pixelorama, remove searx.nixnet.services

Elaborendum:
searx.nixnet.services doesn't seem to be dark anymore.

I've got a question about orama-interactive.itch.io. The main site is already dark, but few of it's subpages aren't, should I just add the whole bunch of dark subpages (I've only added one) or add the main domain?

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -528,6 +528,7 @@ open.spotify.com
 openbase.io
 opendota.com
 openemu.org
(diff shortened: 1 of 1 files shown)


Gusted:
> The main site is already dark, but few of it's subpages aren't, should I just add the whole bunch of dark subpages (I've only added one) or add the main domain?

Not worth it to make the list more complicated then it is.

PR #5880, opened 2021-05-24
Invert the OSM map on https://stm.info homepage

austinhuang0131:
Title. Also it's just a static map like OSM.


Gusted (approved):
Thank you.

PR #5883, opened 2021-05-26
Additional Support for Github Enterprise

n-v-o:
Related to pull #2868, not all Github Enterprise domains end in .com
I changed github.*.com to github.*.* . This should get any enterprise version of github.

Additionally, on Github Enterprise Server 3.0.6 at least, there are a couple of subtle differences in their CSS that make the overrides not work, which causes the colors to not display on days one has contributed. I've corrected those:
1. The classes is just `day` instead of `ContributionCalendar-day`
2. Attribute change: `data-Count` replaces  `data-level`

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4771,7 +4771,7 @@ a[href="https://github.com"]
 ================================
 
 github.com
-github.*.com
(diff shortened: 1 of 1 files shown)


Gusted (approved):
Seems good! Thank you!

PR #5885, opened 2021-05-26
Partial rix for alt.hololive.tv

AndrewDaGuy:
Things-not-fixed-yet-list:
- The corners of the lower part is white
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -607,7 +607,7 @@ INVERT
 alt.hololive.tv
 
 INVERT
-[src*="bg"]
+video[src*="bg.mp4"]
 video[src*="first.mp4"]
 img[src*="logo_2line_gradient_color.svg"]
 svg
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -604,6 +604,16 @@ INVERT
 
 ================================
 
+alt.hololive.tv
+
+INVERT
```
Gusted, line 609: Very loosy selectors. Unless your absolute sere that it won't mess up other things, try to "hardened" it a bit.

src/config/dynamic-theme-fixes.config
```diff
@@ -604,6 +604,16 @@ INVERT
...
+video[src*="bg.mp4"]
```
Gusted, line 610: Or there is being something appended to this otherwise you can also replace the `*` to `$`

AndrewDaGuy: There's something appended to it `?20210215165645` (Seems like a upload date or something `? 2021 02 15 165645`)

Gusted: Got it!

PR #5903, opened 2021-05-29
Fix for reuters.com

rakleed:
- Invert logo.


Gusted (approved):

arturstat:
The logo is not being inverted anymore.
We need to change the path:
```
reuters.com

INVERT
path[d^="M46.938 9.77h8"]
```

arturstat:
We need to be more specific:
```
reuters.com

INVERT
div[class*="site-header-module__logo-container__"] path[d^="M46.938 9.77h8"]
```

PR #5906, opened 2021-05-29

derflasher:
I provide teltarif in darkmode for dynamic filtering.
Convince yourself:
https://www.teltarif.de/

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -10725,8 +10725,8 @@ CSS
 body,
 .ttColBack,
 #tthdrbox {
-    background: ${white} !important;
-    color: ${black} !important;
+    background: var(--darkreader-neutral-background);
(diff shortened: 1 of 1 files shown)


Gusted (commented):
I don't really like this fix nor how dark reader is handling this website. It seems to have some bugs which is preferred to be fix within the source code.

derflasher:
Why don't you like the fix?

Gusted:
> Why don't you like the fix?

Because the site is special. It also uses the light themed background-image. Which is common to have removed in the fix so the body will be once again dark. After that the fix is using `${}` whereby `var(--darkreader-neutral-background)` and `var(--darkreader-neutral-color)` is preferred.

derflasher:
> Because the site is special.

That's why we need special handling.

Gusted (approved):
Good for now. Il be looking into this special site soon.

PR #5923, opened 2021-05-31
Fix article title on getpocket.com

orionlee:
Fix getpocket.com 's article titles in lists, e.g. https://getpocket.com/explore

Current: has `text-shadow` that makes it hard to read

Fix: remove the `text-shadow`

Note:
- The specific CSS selector that needs to be overridden is `.c16ptaiz > a .title span`
- In the fix , I use a slightly simpler and more generic one: `a .title span` , following 
  - the simplifying selector guideline
  - make the fix more resilient to the site's style changes.

---

Before the fix:

image

After the fix:

image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4779,6 +4779,15 @@ login-container {
 
 ================================
 
+getpocket.com
+
+CSS
+a .title span {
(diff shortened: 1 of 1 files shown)


Gusted (approved):
Perfection 👍🏽

PR #5926, opened 2021-05-31
Faster UI render

### LINKED ISSUE #5035: [BUG] Settings load verry slow
# Bug Report

## Describe the bug
The loading of the settings window takes a very long time. The old design doesn't take quite as long, but it also takes a very long time for a display to appear.

## Steps to reproduce

Click on the Icon in the Taskbar

## Expected behavior
The settings are displayed immediately

## Actual behavior
Settings take a few seconds to be displayed.

## Screenshots
https://anzah.cloud/index.php/s/as2DAgQHR349wHy

## System Information

- OS: MacOS Big Sure 11.2.1
- Browser: Firefox 85.0.2
- Dark Reader Version: 4.9.27

## Additional context
When I use the add-on in Safari, the behavior does not occur.

Gusted:
- Resolves #5035

Well this is only 1 of the problems, it should do the job for now to make the rendering for malevic a bit easier(might consider adding this into malevic)

Other pitfall is the immense recursive calling of `execute` that over time is taking time.

Fast machine(Firefox) (reduces by ~66ms):

Before:
image

After:
image

Specially slow machine(Firefox) (reduces by ~473ms):

Before:
image

After:
image

### DIFF
--- a/src/ui/controls/color-picker/hsb-picker.tsx
+++ b/src/ui/controls/color-picker/hsb-picker.tsx
@@ -27,6 +27,15 @@ interface HSBPickerState {
     sbTouchStartHandler: (e: TouchEvent) => void;
 }
 
+const HSBPickerDefaults: HSBPickerState = {
+    wasPrevHidden: true,
+    hueCanvasRendered: false,
+    activeHSB: null,
+    activeChangeHandler: null,
+    hueTouchStartHandler: null,
+    sbTouchStartHandler: null
+};
+
 function rgbToHSB({r, g, b}: RGBA) {
     const min = Math.min(r, g, b);
     const max = Math.max(r, g, b);
@@ -105,7 +114,7 @@ function renderSB(hue: number, canvas: HTMLCanvasElement) {
 
 export default function HSBPicker(props: HSBPickerProps) {
     const context = getContext();
-    const store = context.store as HSBPickerState;
+    const store = context.getStore(HSBPickerDefaults) as HSBPickerState;
     store.activeChangeHandler = props.onChange;
 
     const prevColor = context.prev && context.prev.props.color;
@@ -122,16 +131,14 @@ export default function HSBPicker(props: HSBPickerProps) {
 
     function onSBCanvasRender(canvas: HTMLCanvasElement) {
         if (isElementHidden(canvas)) {
-            store.wasPrevHidden = true;
             return;
         }
         const hue = activeHSB.h;
         const prevHue = prevColor && rgbToHSB(parse(prevColor)).h;
-        if (!store.wasPrevHidden && hue === prevHue) {
-            return;
+        if (store.wasPrevHidden || hue === prevHue) {
+            store.wasPrevHidden = false;
+            renderSB(hue, canvas);
         }
-        store.wasPrevHidden = false;
-        renderSB(hue, canvas);
     }
 
     function onHueCanvasRender(canvas: HTMLCanvasElement) {
--- a/src/ui/controls/color-picker/hsb-picker.tsx
+++ b/src/ui/controls/color-picker/hsb-picker.tsx
@@ -135,7 +135,7 @@ export default function HSBPicker(props: HSBPickerProps) {
         }
         const hue = activeHSB.h;
         const prevHue = prevColor && rgbToHSB(parse(prevColor)).h;
-        if (store.wasPrevHidden || hue === prevHue) {
+        if (store.wasPrevHidden || hue !== prevHue) {
             store.wasPrevHidden = false;
             renderSB(hue, canvas);
         }
--- a/src/ui/controls/color-picker/hsb-picker.tsx
+++ b/src/ui/controls/color-picker/hsb-picker.tsx
@@ -19,7 +19,7 @@ interface HSBPickerProps {
 }
 
 interface HSBPickerState {
-    prevHidden: boolean;
+    wasPrevHidden: boolean;
     hueCanvasRendered: boolean;
     activeHSB: HSB;
     activeChangeHandler: (color: string) => void;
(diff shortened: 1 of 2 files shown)


src/ui/controls/color-picker/hsb-picker.tsx
```diff
@@ -18,6 +19,8 @@ interface HSBPickerProps {
 }
 
 interface HSBPickerState {
+    prevHidden: boolean;
```
alexanderby, line 22: Used as `wasPrevHidden` in code

src/ui/controls/color-picker/hsb-picker.tsx
```diff
@@ -118,15 +121,24 @@ export default function HSBPicker(props: HSBPickerProps) {
     }
 
     function onSBCanvasRender(canvas: HTMLCanvasElement) {
+        if (isElementHidden(canvas)) {
+            context.store.wasPrevHidden = true;
```
alexanderby, line 125: Use `store` instead of `context.store` everywhere, there's `const store = context.store` in the beginning. This object reference is always the same for the same virtual node.

src/ui/controls/color-picker/hsb-picker.tsx
```diff
@@ -118,15 +121,24 @@ export default function HSBPicker(props: HSBPickerProps) {
...
+            return;
+        }
         const hue = activeHSB.h;
         const prevHue = prevColor && rgbToHSB(parse(prevColor)).h;
-        if (hue === prevHue) {
+        if (!context.store.wasPrevHidden && hue === prevHue) {
```
alexanderby, line 130: Note sure, but maybe it should be `||` instead of `&&`?

alexanderby (changes requested):

src/ui/controls/color-picker/hsb-picker.tsx
Gusted: Both ways works the same way, but I can see a very very unlikely situation where `||` is suitable.

Gusted: Actually - I was wrong, and having `||` wouldn't render the hue on the first time.

Gusted: Because technically `hue === prevHue` is correct, because it wasn't rendered yet.

src/ui/controls/color-picker/hsb-picker.tsx
```diff
@@ -118,15 +121,24 @@ export default function HSBPicker(props: HSBPickerProps) {
...
 
     function onSBCanvasRender(canvas: HTMLCanvasElement) {
+        if (isElementHidden(canvas)) {
+            store.wasPrevHidden = true;
+            return;
+        }
         const hue = activeHSB.h;
         const prevHue = prevColor && rgbToHSB(parse(prevColor)).h;
-        if (hue === prevHue) {
+        if (!store.wasPrevHidden && hue === prevHue) {
```
alexanderby, line 130: Maybe it will be better if you either assign a default value `true` if it is `null` or `undefined`? Or maybe it's better to rename it into `wasPrevVisible` and invert the logic everywhere? So the default value will be falsy (undefined).

alexanderby (changes requested):

src/ui/controls/color-picker/hsb-picker.tsx
Gusted: > aybe it will be better if you either assign a default value true if it is null or undefined? 

It's never null or defined if you reach this, because on start it will never be visible and then it will assign `true`

alexanderby: Initially `store` object has no such property `wasPrevHidden`, so it is `undefined`.

So you could
```javascript
if (store.wasPrevHidden == null) {
  store.wasPrevHidden = true;
}
```
Oh wait, haha, I forgot how to use a library I created. There is a `context.getStore(defaults)` method, so it can be `const store = context.getStore({wasPrevHidden: true})`.

So later on this stored property will make more sense, if you understand what I mean. The initial value will be correct and this should correctly replace `&&` with `||` (if I don't miss anything again).

alexanderby (changes requested):

src/ui/controls/color-picker/hsb-picker.tsx
Gusted: Had to invert the logic and remove the `return` statement and should be good now.

src/ui/controls/color-picker/hsb-picker.tsx
```diff
@@ -18,12 +19,23 @@ interface HSBPickerProps {
...
 interface HSBPickerState {
+    wasPrevHidden: boolean;
+    hueCanvasRendered: boolean;
     activeHSB: HSB;
     activeChangeHandler: (color: string) => void;
     hueTouchStartHandler: (e: TouchEvent) => void;
     sbTouchStartHandler: (e: TouchEvent) => void;
 }
 
+const HSBPickerDefaults: HSBPickerState = {
```
alexanderby, line 30: Nice decision to reuse the object for each re-rendering. But you know, naming conventions, it should start with lowercase, particularly the whole acronym should we lowercase in the beginning.

src/ui/controls/color-picker/hsb-picker.tsx
```diff
@@ -118,15 +130,22 @@ export default function HSBPicker(props: HSBPickerProps) {
...
     function onSBCanvasRender(canvas: HTMLCanvasElement) {
+        if (isElementHidden(canvas)) {
+            return;
+        }
         const hue = activeHSB.h;
         const prevHue = prevColor && rgbToHSB(parse(prevColor)).h;
-        if (hue === prevHue) {
-            return;
+        if (store.wasPrevHidden || hue !== prevHue) {
+            store.wasPrevHidden = false;
```
alexanderby, line 139: Not sure if this assignment should be moved out of if...else. Just check that it works.

src/ui/controls/color-picker/hsb-picker.tsx
```diff
@@ -118,15 +130,22 @@ export default function HSBPicker(props: HSBPickerProps) {
...
+            renderSB(hue, canvas);
         }
-        renderSB(hue, canvas);
     }
 
-    function onHueCanvasCreate(canvas: HTMLCanvasElement) {
+    function onHueCanvasRender(canvas: HTMLCanvasElement) {
+        if (isElementHidden(canvas) || store.hueCanvasRendered) {
```
alexanderby, line 145: Replacing vice-versa should be more efficient (checking for the flag before calling the function), a micro optimisation.

alexanderby (changes requested):

alexanderby (approved):
Good

PR #5927, opened 2021-06-01
Fix gray inverted background in Filter mode

alexalekseyenko:
When Filter mode is selected, some pages have a gray background beyond the initial viewport. For instance:
https://nextjs.org/blog

This happens because `background` is set to an already-inverted color on the `html` element, and is therefore inverted for a second time when the CSS filter is applied.

This commit also fixes a bug in how the presence of [Chromium Issue 501582](https://bugs.chromium.org/p/chromium/issues/detail?id=501582) is being detected by `hasChromiumIssue501582()`:
https://github.com/darkreader/darkreader/blob/dc5edc707a57e2a7fda38899e76cd77412818172/src/generators/css-filter.ts#L25
If the issue was resolved in '81.0.4035.0', and it looks like [it was](https://chromium-review.googlesource.com/c/chromium/src/+/1979258), then `hasChromiumIssue501582()` should return `true` when:
```js
 compareChromeVersions(chromiumVersion, '81.0.4035.0') <= 0
```

### DIFF
--- a/src/generators/css-filter.ts
+++ b/src/generators/css-filter.ts
@@ -79,11 +79,10 @@ export function cssFilterStyleSheetTemplate(filterValue: string, reverseFilterVa
 
     if (!frameURL) {
         const light = [255, 255, 255];
-        // If browser not affected by Chromium Issue 501582, set light background on html
-        // else apply color matrix first to invert light color before setting as background
-        const bgColor = hasPatchForChromiumIssue501582() && config.mode === FilterMode.dark ?
-            light :
-            applyColorMatrix(light, createFilterMatrix(config)).map(Math.round);
+        // If browser affected by Chromium Issue 501582, set dark background on html
+        const bgColor = !hasPatchForChromiumIssue501582() && config.mode === FilterMode.dark ?
+            applyColorMatrix(light, createFilterMatrix(config)).map(Math.round) :
+            light;
         lines.push('');
         lines.push('/* Page background */');
         lines.push('html {');
--- a/src/generators/css-filter.ts
+++ b/src/generators/css-filter.ts
@@ -22,7 +22,7 @@ export enum FilterMode {
 export function hasPatchForChromiumIssue501582() {
     return Boolean(
         isChromium &&
-        compareChromeVersions(chromiumVersion, '81.0.4035.0') <= 0
+        compareChromeVersions(chromiumVersion, '81.0.4035.0') >= 0
     );
 }
 
@@ -79,10 +79,11 @@ export function cssFilterStyleSheetTemplate(filterValue: string, reverseFilterVa
 
     if (!frameURL) {
         const light = [255, 255, 255];
-        // If browser affected by Chromium Issue 501582, set html element's background dark
+        // If browser not affected by Chromium Issue 501582, set light background on html
+        // else apply color matrix first to invert light color before setting as background
         const bgColor = hasPatchForChromiumIssue501582() && config.mode === FilterMode.dark ?
-            applyColorMatrix(light, createFilterMatrix(config)).map(Math.round) :
-            light;
+            light :
+            applyColorMatrix(light, createFilterMatrix(config)).map(Math.round);
         lines.push('');
         lines.push('/* Page background */');
         lines.push('html {');
--- a/src/generators/css-filter.ts
+++ b/src/generators/css-filter.ts
@@ -22,7 +22,7 @@ export enum FilterMode {
 export function hasChromiumIssue501582() {
     return Boolean(
         isChromium &&
-        compareChromeVersions(chromiumVersion, '81.0.4035.0') >= 0
+        compareChromeVersions(chromiumVersion, '81.0.4035.0') <= 0
     );
 }
 
@@ -80,15 +80,7 @@ export function cssFilterStyleSheetTemplate(filterValue: string, reverseFilterVa
     if (!frameURL) {
         // If user has the chrome issue the colors should be the other way around as of the rootcolors will affect the whole background color of the page
         const rootColors = hasChromiumIssue501582() && config.mode === FilterMode.dark ? [0, 0, 0] : [255, 255, 255];
-        const [r, g, b] = applyColorMatrix(rootColors, createFilterMatrix(config));
-        const bgColor = {
-            r: Math.round(r),
-            g: Math.round(g),
-            b: Math.round(b),
-            toString() {
-                return `rgb(${this.r},${this.g},${this.b})`;
-            },
-        };
+        const bgColor = `rgb(${rootColors.join(',')})`;
         lines.push('');
         lines.push('/* Page background */');
         lines.push('html {');
--- a/src/generators/css-filter.ts
+++ b/src/generators/css-filter.ts
@@ -78,10 +78,11 @@ export function cssFilterStyleSheetTemplate(filterValue: string, reverseFilterVa
     });
 
     if (!frameURL) {
+        const light = [255, 255, 255];
         // If browser affected by Chromium Issue 501582, set html element's background dark
         const bgColor = hasPatchForChromiumIssue501582() && config.mode === FilterMode.dark ?
-            applyColorMatrix([0, 0, 0], createFilterMatrix(config)).map(Math.round) :
-            [255, 255, 255];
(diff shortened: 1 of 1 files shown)


src/generators/css-filter.ts
```diff
@@ -22,7 +21,7 @@ export enum FilterMode {
 export function hasChromiumIssue501582() {
```
Gusted, line 21: Please adjust the comment that is above this function, as I actually now realizes I didn't got the function name right as it was being used for "hasPatchForChromiumIssue501582" was a better name.

src/generators/css-filter.ts
```diff
@@ -79,16 +78,7 @@ export function cssFilterStyleSheetTemplate(filterValue: string, reverseFilterVa
...
-        const [r, g, b] = applyColorMatrix(rootColors, createFilterMatrix(config));
-        const bgColor = {
-            r: Math.round(r),
-            g: Math.round(g),
-            b: Math.round(b),
-            toString() {
-                return `rgb(${this.r},${this.g},${this.b})`;
-            },
-        };
+        const bgColor = hasChromiumIssue501582() && config.mode === FilterMode.dark ? 'black' : 'white';
```
Gusted, line 81: I don't recall anything on why we needed colorMatrix to be applied over these values. It doesn't seem to matter?

@alexanderby could you enlighten this a bit?

Gusted (changes requested):
Actually seems good to me, just make sure to have that comment changed.

src/generators/css-filter.ts
alexanderby: The issue is that after some Chrome version the filter is not applied to the transparent background. So we add a color to make it not transparent. Then the user wants to apply his preferred brightness, contrast and sepia to this color.

alexanderby: I think I guess, the filter used to not apply to the background of root element at all, even if some color was not assigned, maybe they've changed it in a way that it is now applied.

alexanderby: My understanding of the issue was wrong last year.

So, if there is an issue 501582 (version before 81), the old logic should work. If not, just put a white color, it should be inverted by filter in dark mode. For ancient browsers we just don't care.

alexanderby (changes requested):

alexalekseyenko:
Thank you for the prompt and helpful feedback. This makes more sense now.

I just pushed some changes incorporating your suggestions. Full disclosure: I haven't tested it in Chromium <= 81.0.4035.0 but I assume those users will let us know if it's an issue.

Gusted:
>  I haven't tested it in Chromium <= 81.0.4035.0 but I assume those users will let us know if it's an issue.

Angry user here, it doesn't work in the not having the patch for the issue versions:
image

src/generators/css-filter.ts
```diff
@@ -19,10 +19,10 @@ export enum FilterMode {
  * Bug report: https://bugs.chromium.org/p/chromium/issues/detail?id=501582
  * Patch: https://chromium-review.googlesource.com/c/chromium/src/+/1979258
  */
-export function hasChromiumIssue501582() {
+export function hasPatchForChromiumIssue501582() {
     return Boolean(
         isChromium &&
-        compareChromeVersions(chromiumVersion, '81.0.4035.0') >= 0
+        compareChromeVersions(chromiumVersion, '81.0.4035.0') <= 0
```
Gusted, line 25: `compareChromeVersions` will return `0` if it's the same. Will return `1` if(this case) `chromiumVersion` is bigger and `-1` if `chromiumVersion` is less.

So you'd want to still invert this comparison to  `>=` as if chromiumVersion is bigger or the same(so `-1` or `0` as return values) return true.

alexalekseyenko: Thank you for catching that!

alexalekseyenko:
Does this produce the expected behavior?

Gusted:
Yes!

image
image

src/generators/css-filter.ts
alexalekseyenko: I think this might have been happening because the "dark" background was being set as:
https://github.com/darkreader/darkreader/blob/fb9b87a0efc8e0c6e7d3ad4a131859561b9b8439/src/generators/css-filter.ts#L83
whereas the inverse or `rgb(0, 0, 0)` really isn't very dark.

I went ahead and changed that, but I think I see now that you intended for `hasPatchForChromiumIssue501582()` to mean that Chromium has patched the issue and not that we should. In that case, I'll go ahead and invert the comparison as well as the logic below:
https://github.com/darkreader/darkreader/blob/fb9b87a0efc8e0c6e7d3ad4a131859561b9b8439/src/generators/css-filter.ts#L81-L84

alexalekseyenko:
Thank you for catching that!

src/generators/css-filter.ts
```diff
@@ -78,21 +78,15 @@ export function cssFilterStyleSheetTemplate(filterValue: string, reverseFilterVa
...
-            r: Math.round(r),
-            g: Math.round(g),
-            b: Math.round(b),
-            toString() {
-                return `rgb(${this.r},${this.g},${this.b})`;
-            },
-        };
+        const light = [255, 255, 255];
+        // If browser affected by Chromium Issue 501582, set dark background on html
+        const bgColor = !hasPatchForChromiumIssue501582() && config.mode === FilterMode.dark ?
```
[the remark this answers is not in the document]
alexanderby: Here's how it used to be before the fix https://github.com/darkreader/darkreader/pull/2310/files#diff-b3bf9845410975de4e9089094d063eedac8f2a83d9019e4d38723aeeb670a34eL66

So we had a bit false assumption a year ago. Strange that nobody noticed during the whole year.

alexanderby (changes requested):

src/generators/css-filter.ts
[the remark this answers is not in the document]
alexalekseyenko: @Gusted 
How does it look in the legacy Chromium build? I'm running this patch in Chromium 90 and it seems to have fixed the issue for me. My interpretation of the logic was that it meant to "manually" set a dark `background` on `html` in the affected versions, or light otherwise, relying on the injected CSS to do the rest.

For what it's worth, I've been seeing the gray background on many pages for quite a while and I think I remember finding at least one open issue about it. I'm not sure how many people use Filter vs Dynamic mode, so hard to say what the impact would have been.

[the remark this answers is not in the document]
Gusted: > For what it's worth, I've been seeing the gray background on many pages for quite a while and I think I remember finding at least one open issue about it. I'm not sure how many people use Filter vs Dynamic mode, so hard to say what the impact would have been.

A lot people I come across that uses Filter are very frustrated with the performance of Dynamic and don't care about all these fancy colors being darkened etc and just want some dark background with light color with some good options to change brightness etc.



> How does it look in the legacy Chromium build?

Looking good with the latest commits 
image




>  My interpretation of the logic was that it meant to "manually" set a dark `background` on `html` in the affected versions, or light otherwise, relying on the injected CSS to do the rest.

Like Alexander said we had a good false assumption about this whole bug. 

So just everyone on the same page:

We have this example HTML as demo

```HTML
<html style="filter: filter(100%) hue-rotate(180deg)">
<body style="background-color: red">
<p> here some text lalala</p>
</body>
</html>
```
Let's take this legacy version of Chromium:
image
You see, the body isn't being effected and the actually red color. 
Wow the bug! Only HTML is being affected by this filter and not it's siblings.

New version of Chromium:
image
You see, now the red is being affected by the filter! The intended behavior.

As at first it wasn't being affected we had a simple black background-color, because it wouldn't be affecting anything otherwise. However in the new version we had to set a white background-color, because it now WAS affecting other elements and thus causing it to be a white background-color. 

Yeah we are relying on other CSS, but because of mixed behavior across versions we are setting this background-color on newer versions to white-ish.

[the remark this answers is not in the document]
alexalekseyenko: Yeah, I find myself in that camp too—the improved performance is worth it. Though, other than this inverted background issue, I haven't really had much trouble.

Is this working as intended or should we make some more changes?

alexalekseyenko:
Bump.

Anything else need to be done here? I've been running this patch for a couple of weeks now with no issue.

Gusted (approved):
It's working.

Issue #5924, opened 2021-05-31
[Broken Website] Google Drive File Preview Icons


RuthlessRuler:
# Broken Website
## Website
drive.google.com

## How to Reproduce

  - Go to google drive site
  - Open PDF/MP3/Video/etc
  - icons to download/print/3 dots & etc are all black thus submerged with background and not visible.
  -  Weird Font Rendering in Open with google

## Expected behaviour
The icons should be in white colour to be visible to the users.

## Actual behaviour
The Icons are too getting inverted to black colour thus not visible

## Screenshots
Before:
![image](https://user-images.githubusercontent.com/20885012/120157654-e01a0900-c210-11eb-8160-b89b5c450ace.png)
After:
![image](https://user-images.githubusercontent.com/20885012/120157484-b4971e80-c210-11eb-9dd3-14487473afd9.png)

## System Information

- OS: Windows & Linux
- Browser: Firefox 88
- Dark Reader Version:  4.9.33

Gusted:
@rakleed it seems like the fix for drive.google.com is causing this #5835. 
Specific 
```CSS
div[role="dialog"] div[role="button"][tabindex="0"]
```

RuthlessRuler:
Okay.
I would also like to add that i was doing some additional testing, so if Open A file from Google Drive, then click on 3 dots and click open in new Tab, the icons are actually visible there properly. (i.e the Link is drive.google.com/file/***)

Google Drive Preview: 3 Dot menu Preview Pane:
![image](https://user-images.githubusercontent.com/20885012/120192488-d99f8780-c238-11eb-9691-385ec6b92c42.png)

Google Drive File Link: 3 dot menu
![image](https://user-images.githubusercontent.com/20885012/120192922-69ddcc80-c239-11eb-9f84-8ef0ef141b27.png)
Dark Reader Off:
![image](https://user-images.githubusercontent.com/20885012/120192083-5f6f0300-c238-11eb-98bf-22295b6f1766.png)
Dark Reader On:
![image](https://user-images.githubusercontent.com/20885012/120192184-7e6d9500-c238-11eb-9d39-bf162a2bc2e6.png)

rakleed:
[...]

PR #5930, opened 2021-06-01
Fix for drive.google.com

rakleed:
- Resolves https://github.com/darkreader/darkreader/issues/5924.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3434,6 +3434,7 @@ INVERT
 img[src$="google_gsuite"]
 div[role="menuitem"] svg
 div[role="dialog"][style^="width: 350px;"] div[role="button"][tabindex="0"]
+div[role="dialog"] ~ div[role="menu"] > div[role="menuitem"] > div > div:not([style*="background-image"])
 
 CSS
 span[data-type="spelling"] {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -3433,7 +3433,7 @@ drive.google.com
 INVERT
 img[src$="google_gsuite"]
 div[role="menuitem"] svg
-div[role="dialog"] div[role="button"][tabindex="0"]
+div[role="dialog"][style^="width: 350px;"] div[role="button"][tabindex="0"]
```
Gusted, line 3436: Does this play well with different screen resolutions?

rakleed: Yes, I checked.

Gusted: Great. Il take a look to fix the other problem that was mentioned.

Gusted:
@rakleed 

```CSS
div[role="dialog"] ~ div[role="menu"] > div[role="menuitem"] > div > div:not([style*="background-image"])
```

It matches if any dialog is displayed(like pdf preview) then it will `generallty select the role=menu sibling`(https://developer.mozilla.org/en-US/docs/Web/CSS/General_sibling_combinator/General_sibling_combinator) and then it has the same fix as drive.google.com/file I've tested it for 10 minutes across various pages and it should for in this exact case and any other dialog's that has this toolbar(hopefully).

Gusted (approved):

PR #5933, opened 2021-06-01
Fix for qubes-os.org

AndrewDaGuy:


Gusted (changes requested):
Unless unvisible(like the grey logo's) let's not add them. 
image

Gusted (approved):

PR #5367, opened 2021-03-30
Big fix for wikipedia.org

AndrewDaGuy:
Yes, another big ol fix for wikipedia

Note: *Most* Only Images/Icons that cannot be / barely visible seen in dark mode are inverted to minimize file size

To-do:

---
- [x] https://en.wikipedia.org/wiki/Little-o_notation
- [x] https://en.wikipedia.org/wiki/Dutch_language
- [x] https://en.wikipedia.org/wiki/Japan
- [x] https://en.wikipedia.org/wiki/Geometry
- [x] https://en.wikipedia.org/wiki/Uwu_(emoticon)
- [x] https://en.wikipedia.org/wiki/Facial_expression
- [x] https://en.wikipedia.org/wiki/Hangul
- [x] https://en.wikipedia.org/wiki/List_of_National_Historic_Landmarks_in_Idaho
- [x] https://en.wikipedia.org/wiki/Horizon
- [x] https://en.wikipedia.org/wiki/Vertebrate
- [x] https://en.wikipedia.org/wiki/Email
- [x] https://en.wikipedia.org/wiki/Discord_(software)
- [x] https://en.wikipedia.org/wiki/Fandom
- [x] https://en.wikipedia.org/wiki/Hexagon
- [x] https://en.wikipedia.org/wiki/Tumblr
- [x] https://en.wikipedia.org/wiki/Katakana
- [x] https://en.wikipedia.org/wiki/List_of_emoticons
- [x] https://en.wikipedia.org/wiki/Cheese
- [x] https://en.wikipedia.org/wiki/Pixel
- [x] https://en.wikipedia.org/wiki/Gaudo_culture
- [x] https://en.wikipedia.org/wiki/Jam
- [x] https://en.wikipedia.org/wiki/Torus
- [x] https://en.wikipedia.org/wiki/Character_(symbol)
- [x] https://en.wikipedia.org/wiki/Portal:Food
- [x] https://en.wikipedia.org/wiki/U.S._Army_Esports
- [x] https://en.wikipedia.org/wiki/Vantablack
- [x] https://en.wikipedia.org/wiki/Sufganiyah
- [x] https://en.wikipedia.or/wiki/National_Register_of_Historic_Places_listings_in_Teton_County,_Idaho
- [x] https://en.wikipedia.org/wiki/Ames_Community_School_District
- [x] https://en.wikipedia.org/wiki/Pan_dulce_(sweet_bread)
- [x] https://en.wikipedia.org/wiki/System_of_linear_equation
[...]

PR #5939, opened 2021-06-02
Reverted wikipedia fixes to commit 01a5e67

AndrewDaGuy:
Reverts the wikipedia fixes to commit (01a5e67fa293428d42bdf4f12460b2b284f8192c)

This commit was before the shortening of images that messed up other images in #5367 :c


Gusted:
I wasn't meant to merge this. As I rather wanted to add a comment to say that the fix should be removed all together as it will keep on going and the list is now horribly long for specific images.

PR #5953, opened 2021-06-03
Changed jira selectors and added panel styling

samuelgfeller:
Also removed transparent header I have no idea who made this and why, but it looks ugly. Way better if the header actually has a background colour.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1227,6 +1227,9 @@ code:first-of-type {
 .ak-editor-panel[data-panel-type="warning"] {
     background-color: ${rgb(255, 250, 230)} !important;
 }
+.cCWaFY {
+    background-image: none !important;
+}
 
 IGNORE INLINE STYLE
 .ak-editor-panel
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1191,8 +1191,7 @@ img[src^="https://latexmath.bolo-app.com/render/"]
 
 CSS
 span.code {
-    background-color: rgba(240,246,252,0.15) !important; 
-    padding: 2px 3px !important;                                                                
+    background-color: rgba(240,246,252,0.15) !important;                                                
 }
 .ghx-parent-group,
 .ghx-issue {
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1227,7 +1227,7 @@ code:first-of-type {
 .ak-editor-panel[data-panel-type="warning"] {
     background-color: ${rgb(238, 229, 187)} !important;
 }
-.cCWaFY {
+.cCWaFY, .dFpipQ {
     background-image: none !important;
 }
 
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1216,18 +1216,16 @@ code:first-of-type {
     background-image: linear-gradient(to right, rgba(255,255,255,0.03),rgba(255,255,255,0.03)calc(1ch + 16px),transparent calc(1ch + 16px),transparent) !important;
 }
 .ak-editor-panel[data-panel-type="note"] {
-    background-color: rgba(49, 24, 63, 0.71) !important;
-    color: rgba(229, 210, 255, 0.8) !important; 
+    background-color: ${rgb(234, 230, 255)} !important;
 }
 .ak-editor-panel[data-panel-type="info"] {
-    background-color: rgba(6, 52, 77, 0.99) !important;
-    color: rgba(200, 230, 255, 0.95) !important;
+    background-color: ${rgb(222, 235, 255)} !important;
 }
 .ak-editor-panel[data-panel-type="error"] {
-    background-color: rgba(66, 13, 0, 0.70) !important;
+    background-color: ${rgb(255, 235, 230)} !important;
 }
 .ak-editor-panel[data-panel-type="warning"] {
-    background-color: rgba(66, 53, 0, 0.79) !important;
+    background-color: ${rgb(255, 250, 230)} !important;
 }
 
 IGNORE INLINE STYLE
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1216,16 +1216,16 @@ code:first-of-type {
     background-image: linear-gradient(to right, rgba(255,255,255,0.03),rgba(255,255,255,0.03)calc(1ch + 16px),transparent calc(1ch + 16px),transparent) !important;
 }
 .ak-editor-panel[data-panel-type="note"] {
-    background-color: ${rgb(234, 230, 255)} !important;
+    background-color: ${rgb(217, 200, 228)} !important;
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -1190,30 +1190,46 @@ img[src*="https://github.githubassets.com/favicon.ico"].smart-link-icon
 img[src^="https://latexmath.bolo-app.com/render/"]
 
 CSS
-#jira-issue-header, #jira-issue-header-actions {
-    background-color: transparent !important;
-}
 span.code {
     background-color: rgba(240,246,252,0.15) !important; 
+    padding: 2px 3px !important;
```
Gusted, line 1195: Uhm what? We don't do cosmetic CSS like padding.

src/config/dynamic-theme-fixes.config
```diff
@@ -1190,30 +1190,46 @@ img[src*="https://github.githubassets.com/favicon.ico"].smart-link-icon
...
+}
+.ak-editor-panel[data-panel-type="info"] {
+    background-color: rgba(6, 52, 77, 0.99) !important;
+    color: rgba(200, 230, 255, 0.95) !important;
+}
+.ak-editor-panel[data-panel-type="error"] {
+    background-color: rgba(66, 13, 0, 0.70) !important;
+}
+.ak-editor-panel[data-panel-type="warning"] {
+    background-color: rgba(66, 53, 0, 0.79) !important;
```
Gusted, line 1231: Please use it like this for every other CSS that's being added:
`${orignialcolor}`
Example:
Originialcolor = red 
then 
```CSS
background-color: ${red}
```
Dark Reader will automatically darken the color for you and take in account the user's settings.

samuelgfeller: Ah alright.. I lowered the vibrancy of red and orange a bit more than how Dark Reader automatically darkened to be able to read a bit better but I guess it's not me that can decide on how bright certain colors should be. 

How do you mean it automatically darkens the color ? Probably I misinterpreted it but I can't get it work for the info panel and note panel. 
* If I do nothing: only warning and error are highlighted https://i.imgur.com/HSauCEG.png 
* If I do like you mentioned its the same as with nothing, although I probably did it wrongly:   https://i.imgur.com/ySfVd9G.png
* With the changes I pushed: https://i.imgur.com/tkOIiPh.png

Naturally, I pressed "Apply" after each change. What am I doing wrong?

src/config/dynamic-theme-fixes.config
samuelgfeller: Removed it in commit 73eef25

src/config/dynamic-theme-fixes.config
Gusted: For the info one you used `()` instead of `{}` simple syntax mistake. And for the note... I'm not sure you could use the originial color in rgb and use that.

Don't forget to add `!important;`

src/config/dynamic-theme-fixes.config
```diff
@@ -1190,30 +1190,45 @@ img[src*="https://github.githubassets.com/favicon.ico"].smart-link-icon
 img[src^="https://latexmath.bolo-app.com/render/"]
 
 CSS
-#jira-issue-header, #jira-issue-header-actions {
-    background-color: transparent !important;
-}
 span.code {
-    background-color: rgba(240,246,252,0.15) !important; 
+    background-color: rgba(240,246,252,0.15) !important;
```
Gusted, line 1194:
```suggestion
    background-color: rgba(240,246,252,0.15) !important;
```
accidental tab?

src/config/dynamic-theme-fixes.config
samuelgfeller: It seems to have been the `!important`. I changed the values with the orignial rgb colors and it resulted in the following

result

samuelgfeller:
7f72785 this commit is to remove this preloading image (bottom left)

avatar

src/config/dynamic-theme-fixes.config
samuelgfeller: You sure though that its not better to take the "custom" colors I made? How dark reader processes the original colors automatically is very dark and we can barely tell a difference on not so bright screens. Here are the two side to side https://imgur.com/a/JIu3nhl

Gusted: > You sure though that its not better to take the "custom" colors I made?

Eek no. People who tend to have a good amount of customization just take my higher contrast and slightly lower brightness as example I tend to not allow hardcoded values because no matter what they won't change according to the users setting. 

You always can change the color so dark reader will process it differently.

samuelgfeller:
> Now I'm not sure how the html structure is here, but could it be changed to not depend on the this selector as they could change any time and are generated by the bundler, meaning the next time jira gets an update the selector will change.

That's true and so is the majority of my selectors. I even [asked Jira](https://community.atlassian.com/t5/Jira-Software-questions/Prevent-CSS-selectors-from-changing/qaq-p/1618215) if they could set some permanent selectors. 

Maybe there is a technique I am not aware of, but I couldn't find a better way to locate this div via the permanent elements in the structure. All parents are generated too :( 

[Here](https://jsfiddle.net/078nL2pt/) is the HTML code. 

image

src/config/dynamic-theme-fixes.config
samuelgfeller: > I tend to not allow hardcoded values because no matter what they won't change according to the users setting.

Alright, I understand. 

> You always can change the color so dark reader will process it differently.

I tried to darken the original color but that resulted in a totally wrong generated color. Even though the original was purple, the generated was blueish, so the process is more complicated than just darken the original color. For me right now a big project started, so unfortunately I haven't more time to dedicate fiddling with colors to find a good one. 

What we could do though is take the darkreader formula that darkens the color and reverse my hardcoded colors to get the "bright equivalent" color.

Gusted:
> Maybe there is a technique I am not aware of, but I couldn't find a better way to locate this div via the permanent elements in the structure. All parents are generated too :(
> 
> [Here](https://jsfiddle.net/078nL2pt/) is the HTML code.

**** f*ck. That's some pain in the ass. That will just be the old normal regular CSS updates.  Yeah that's no luck.

src/config/dynamic-theme-fixes.config
Gusted: > What we could do though is take the darkreader formula that darkens the color and reverse my hardcoded colors to get the "bright equivalent" color.

If you take the the correct color and put that in the `${}`, get the result of that color from the dev-tools and use that as the value for the fix. It should be the exact same color after you put it in `${}`

samuelgfeller:
Jira: 
> Unfortunately, the obfuscated css selectors are a deep part of our styling pipelines that are there to avoid unintended styling consequences of similar human curated class names.

They do it on purpose too. So annoying.

src/config/dynamic-theme-fixes.config
samuelgfeller: Oh I didn't know this alright, I'll do so.

samuelgfeller:
Okay I'm done with my changes now, you can merge if you agree to everything.

Gusted (approved):

PR #5956, opened 2021-06-04
Update dark-sites.config

wmartinmimi:
added android-car-4.github.io
website dark on default


wmartinmimi:
it is basically the same as martinmimigames.github.io

Issue #5502, opened 2021-04-07
[Performance] Memory leak


leonard-vlaicu:
# Performance Issue

## Website
https://azure.microsoft.com/en-us/services/

## How to Reproduce
  I have 8GB of RAM. When I tested this, only 50% was in use.
  Dark Reader is ON.
  After entering the site, the RAM is continously getting eaten until to 98-99% in use.
  
## System Information
- OS: Windows
- Browser: Firefox 87
- Dark Reader Version: 4.9.31

- Do you have this performance issue with Dark Reader disabled? No
- Does it happen in other browsers? Haven't tested
- Do you have hardware acceleration enabled? Yes

![Screenshot (2)](https://user-images.githubusercontent.com/56519529/113855896-3c305600-97a9-11eb-929c-b96b7c85d52a.png)
![Screenshot (4)](https://user-images.githubusercontent.com/56519529/113855888-3aff2900-97a9-11eb-9578-848beefcd43c.png)

ikenna8317:
Same with me. No matter the browser Dark reader consumes too much system resources. I had to uninstall the extension.

Zerossoul:
I am noticing this on Safari as well. It's sad because I love how fast Safari is in big sur over Chrome, but there are no other dark mode extensions that work well in Safari. Props to Dark Reader for atleast having it.

jersten:
I'm also having this issue. I'm using LibreWolf, a modified version of firefox aimed at being much more secure. This extension causes my computer to freeze up for about 7 minutes at a time, and I have 16Gb of RAM which should be more than enough. I'd say this needs to be fixed asap. I had to uninstall Darkreader.

Gusted:
@alexanderby Do we by any chance store images within the content script? I know the background-script has a cache for fetched elements. But the numbers shows that we are somehow holding onto big images that seems to be coming from the urlToImage function and GC ain't cleaning it up.

alexanderby:
There's `URL.createObjectURL`. For inverting the images they are put into SVG with a filter. I think data URLs are used because image URLs don't work in Blob URLs as far as I remember.

Gusted:
Hmm, yeah. Pretty weird, might get a minimal case and send it as a bug report if it's confirmed that the GC isn't cleaning up as their was around 2gb of data: url's within the memory and the analyzeImage function was being pretty heavy.

alexanderby:
[...]

PR #5963, opened 2021-06-05
Use raw(base64'ed) svg as data: url

### LINKED ISSUE #5957: [Broken Website] Mozilla.org and Disroot upload
# Broken Website
https://www.mozilla.org/en-US/firefox/new/
https://disroot.org/en/services/upload

## How to Reproduce
Just go to these sites. Firefox one misses some elements, and disroot one misses huge photo at the top.

## Screenshots
firefox
disroot

## System Information
Windows 10/64bit
Firefox 89.0
Darkreader 4.9.33

## Additional context
I didn't want to open 2 issues thinking it may bother you, sorry for that if it is a problem.

Gusted:
- Resolves #5957

mozilla.org disallows blob: URL's as image-src, this also resolves some pain in managing memory of using blob's.

### DIFF
--- a/src/inject/dynamic-theme/image.ts
+++ b/src/inject/dynamic-theme/image.ts
@@ -139,10 +139,8 @@ function analyzeImage(image: HTMLImageElement) {
     };
 }
 
-const objectURLs = new Set<string>();
-
-export function getFilteredImageDataURL({dataURL, width, height}: ImageDetails, filter: FilterConfig) {
-    const matrix = getSVGFilterMatrixValue(filter);
+export function getFilteredImageDataURL({dataURL, width, height}: ImageDetails, theme: FilterConfig): string {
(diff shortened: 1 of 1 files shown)


Gusted:
I'm not going to like this change, because this will make debugging harder with the devtools: https://bugzilla.mozilla.org/show_bug.cgi?id=1714747

Gusted:
@alexanderby I actually tought this will also just solve the memory issue with #5502 but it seems like it's something otherwise. For some reason the reported site had: 
image

a lot of data64 jpeg images which without dark reader didn't exist.

alexanderby (approved):
I remember the reason for using Object URLs (blobs) was slow performance, especially in Firefox, when a large text was put into a style. But since textContent was replaced with CSSOM (insertRule etc), things can be different. So if you see no performance issues, let it be.

PR #5965, opened 2021-06-05
Fix red background on myanimelist.net

EFHIII:
The background on the left and right of the main page of https://myanimelist.net/ are manually set to `#D20202` which is very bright and presumably only there to be annoying.
Here, I change it to the neutral background color which looks much nicer. I used the same selector that `#D20202` was under;
```css
body:not(.ownlist) {
}
```

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -7963,6 +7963,15 @@ INVERT
 
 ================================
 
+myanimelist.net
+
+CSS
+body:not(.ownlist) {
+    background-color: var(--darkreader-neutral-background) !important;
+}
+
+================================
+
 mybrandnewlogo.com
 
 IGNORE INLINE STYLE


Gusted:
Hi @EFHIII!

I'm not sure what you get, but it shows for me that the background is by default `#FFF` and darkreader does it job by darkening it.

EFHIII:
> Hi @EFHIII!
> 
> I'm not sure what you get, but it shows for me that the background is by default `#FFF` and darkreader does it job by darkening it.

image

Also do be sure it wasn't something else I have on chrome, I tested in Explorer, which is stock:
image

Gusted:
> , I tested in Explorer, which is stock:

Internet Explorer is the last thing you want to be comparing against. But that's now without dark reader and it shows a red background-color. 

I'm not sure what you are doing but there is something going on here that shouldn't be happening.

EFHIII:
> Internet Explorer is the last thing you want to be comparing against. But that's now without dark reader and it shows a red background-color.
I was just using IE to make sure the red background was what happens normally (without Dark Reader or anything else) since you said you get white, as a sanity check. It's the same thing in other browsers. 

I'm not doing anything special. The background is just red sometimes. I haven't tested too much, but for one reason or another, sometimes the red version of the site gets served and other times it doesn't. The easiest way to force it to serve you the red version is probably to use a VPN or proxy from somewhere that happens to get served the red version. When the white version is served, Dark Reader does it's job, as you said.

Gusted:
> The background is just red sometimes. I haven't tested too much, but for one reason or another, sometimes the red version of the site gets served and other times it doesn't.

If this is the case myanimelist.net should fix it up. But in this case as I normally don't allow patches that fix broken behavior of the site, seems reasonable enough.

Gusted (approved):

PR #5976, opened 2021-06-06
[static-theme] Preserve transparency when applying transformation matrix

gabrielfin:
`FADE BG` and `FADE TEXT` in static mode had no transparency, because `applyColorMatrix()` dropped the alpha value.

~~I hope this way of fixing it is ok. This adds an `undefined` alpha value to colors that don't have it, but since `rgb()` has a check for it, I suppose it's ok.
Should I have used `push()` instead of `concat()`?~~

EDIT: I changed it to push the alpha value instead of concatenating, since an undefined alpha created issues in other places

### DIFF
--- a/src/generators/static-theme.ts
+++ b/src/generators/static-theme.ts
@@ -62,7 +62,7 @@ export default function createStaticStylesheet(config: FilterConfig, url: string
     const theme = Object.entries(srcTheme).reduce((t, [prop, color]) => {
         const [r, g, b, a] = color;
         t[prop] = applyColorMatrix([r, g, b], createFilterMatrix({...config, mode: 0}));
-        if(a !== undefined) {
+        if (a !== undefined) {
             t[prop].push(a);
         }
         return t;
--- a/src/generators/static-theme.ts
+++ b/src/generators/static-theme.ts
@@ -60,7 +60,11 @@ function mix(color1: number[], color2: number[], t: number) {
 export default function createStaticStylesheet(config: FilterConfig, url: string, frameURL: string, staticThemes: StaticTheme[]) {
     const srcTheme = config.mode === 1 ? darkTheme : lightTheme;
     const theme = Object.entries(srcTheme).reduce((t, [prop, color]) => {
+        const [r, g, b, a] = color;
         t[prop] = applyColorMatrix(color, createFilterMatrix({...config, mode: 0}));
+        if(a !== undefined) {
+            t[prop].push(a);
+        }
         return t;
     }, {} as ThemeColors);
 
--- a/src/generators/static-theme.ts
+++ b/src/generators/static-theme.ts
@@ -61,7 +61,7 @@ export default function createStaticStylesheet(config: FilterConfig, url: string
     const srcTheme = config.mode === 1 ? darkTheme : lightTheme;
     const theme = Object.entries(srcTheme).reduce((t, [prop, color]) => {
         const [r, g, b, a] = color;
-        t[prop] = applyColorMatrix(color, createFilterMatrix({...config, mode: 0}));
+        t[prop] = applyColorMatrix([r, g, b], createFilterMatrix({...config, mode: 0}));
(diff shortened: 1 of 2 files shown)


gabrielfin:
Any news on this? Sorry about the persistence, but I would like to submit some fixes to static theme, and this bug is blocking me.

Gusted:
I know partially how these matrixes work, most work hasn't been done with me. I cannot really judge if this is a good change or not.  From what I see is that alpha is now just being pushed onto the matrix again without any transforming. Which is bad because that isn't what this function is about.

We currently do this just outside the function to push the alpha back on into the appropriate function. This is how we do it at dynamic theme:
https://github.com/darkreader/darkreader/blob/a39d76cfacab1e97bf1affafa99aad972e9fd73e/src/generators/modify-colors.ts#L69-L73

And such change is something I can allow :)  

Also, I'm not sure but now you allow the alpha value to be a input. But that shouldn't fix anything because at this moment I don't see any function taking advantage of this.

Regards,
Gusted

src/generators/static-theme.ts
```diff
@@ -60,7 +60,11 @@ function mix(color1: number[], color2: number[], t: number) {
 export default function createStaticStylesheet(config: FilterConfig, url: string, frameURL: string, staticThemes: StaticTheme[]) {
     const srcTheme = config.mode === 1 ? darkTheme : lightTheme;
     const theme = Object.entries(srcTheme).reduce((t, [prop, color]) => {
+        const [r, g, b, a] = color;
         t[prop] = applyColorMatrix(color, createFilterMatrix({...config, mode: 0}));
+        if(a !== undefined) {
+            t[prop].push(a);
+        }
```
Gusted, line 67:
```suggestion
        a && t[prop].push(a);
```

Welcome to modern javascript whereby it will be automatically checked for truthiness 😄

Gusted (commented):
This change seems good. Just to ensure it will work now and in the future let's add some unit tests. That part can be leaved for now I can add them.

src/generators/static-theme.ts
gabrielfin: Alpha value could be zero I guess? If at some point someone wants to represent the color "transparent".

Gusted: That's correct... I'm not sure how dynamic is handling as in there case it seems `a` is always defined.

Gusted: As if it's defined, the check could be removed altogether.

Gusted: Nevermind about what I said. This code is good to go. Il add unit tests for this one later in another PR.

Gusted (approved):
Thank you! Sorry for the long wait, thought it was something more in-depth about something I didn't have any knowledge. Good catch 🎉

src/generators/static-theme.ts
```diff
@@ -60,7 +60,11 @@ function mix(color1: number[], color2: number[], t: number) {
 export default function createStaticStylesheet(config: FilterConfig, url: string, frameURL: string, staticThemes: StaticTheme[]) {
     const srcTheme = config.mode === 1 ? darkTheme : lightTheme;
     const theme = Object.entries(srcTheme).reduce((t, [prop, color]) => {
+        const [r, g, b, a] = color;
         t[prop] = applyColorMatrix(color, createFilterMatrix({...config, mode: 0}));
```
Gusted, line 64:
```suggestion
        t[prop] = applyColorMatrix([r, g, b], createFilterMatrix({...config, mode: 0}));
```

src/generators/static-theme.ts
```diff
@@ -60,7 +60,11 @@ function mix(color1: number[], color2: number[], t: number) {
...
+        if(a !== undefined) {
```
Gusted, line 65:
```suggestion
        if (a !== undefined) {
```

Gusted (commented):
Let me fix these tests for you.

PR #5977, opened 2021-06-06
add jabodent.com & josephojo.com to dark-sites.config

okikio:


Gusted:
Doesn't match criteria

```
If a website is already dark, that regardless of system's preferred color-scheme has all pages dark by default. You can add it to the dark-sites.config file (please, preserve the alphabetical order).
```

Issue #5988, opened 2021-06-08
[Feature Request] Make mode switching shortcut per-site by default (or with option)


phoenixeliot:
# Feature Request

## Is your feature request related to a problem? Please describe
I often use a keyboard shortcut to switch between modes on a given site, as many sites work better in one mode or another. However, by default the keyboard shortcut switches the *default* mode for all sites, unless I open the extension with my mouse, and click "only for <website>", *then* switch the mode. This is much slower than just using a keyboard shortcut.

## Describe the solution you'd like
I would like the default behavior (or perhaps with an option) of the "theme generation mode" shortcut to switch the "only for this site" toggle on for the current site, so the shortcut only changes the theme for the current site by default.

phoenixeliot:
Ah, huh. Good to know. Well, I'm working on a fix that will at least be usable for me, so I'll see what I can do for that.

phoenixeliot:
One idea I have, that I'll see how well it pans out: Create one default theme for each filter mode, and have this shortcut switch between those themes

phoenixeliot:
I'm curious, also: How much do y'all expect people to create different themes? I'm struggling to think of a use case beyond this one right here, for having multiple themes to switch between. I'm not really sure what a "theme" would logically encapsulate from a UX perspective.

phoenixeliot:
(If you already have a discussion on how the whole theme thing works from a UX perspective I'd love to read that)

phoenixeliot:
Added a PR with my good-enough-for-me fix. Let me know if you think something like it would be feasible to merge.

PR #5989, opened 2021-06-09
Add shortcut for changing theme engine per-site

phoenixeliot:
This addresses my feature request, #5988.

Submitting as a draft for now because at this point I've just gotten it good enough for me personally to use.

The way it works right now:
- There is a new keyboard shortcut for the extension, "Theme mode for current site"
- If you execute the shortcut, it will create a site-specific theme if there is none, and set the theme engine to be the next one in the list for that site theme

Things of note:
- I don't know how this project's translation system works, or if I need to update the other locale files with anything. This currently fails a unit test about the locale files.
- It looks like even before my changes, the 'change theme generation mode' shortcut always only changes the global setting. If you have a per-site theme (eg "Only on <x.com>" set to true), then the original shortcut would do nothing visible until you turn "only on ..." off, or visit a different page that uses the global theme. How the original shortcut behaves might merit changes if this PR might be merged into the main repo.
- This probably doesn't behave as expected if you have user-created themes for multiple sites. It will change the theme's engine instead of just changing the engine per-site. I'll defer fixing this until contributors have chimed in about whether they think any of this is feasible to have merged in.

### DIFF
--- a/src/_locales/en.config
+++ b/src/_locales/en.config
@@ -140,6 +140,9 @@ Dynamic
 @theme_generation_mode
 Theme generation mode
 
+@theme_generation_mode_current_site
+Theme generation mode for current site
+
 @custom_browser_theme_on
 Custom
 
--- a/src/background/extension.ts
+++ b/src/background/extension.ts
@@ -165,12 +165,38 @@ export class Extension {
                     this.toggleURL(url);
                 }
             }
+            // If they use the normal shortcut, it only changes the global setting (even if this site has its own theme)
             if (command === 'switchEngine') {
                 console.log('Switch Engine command entered');
                 const engines = Object.values(ThemeEngines);
-                const index = engines.indexOf(this.user.settings.theme.engine);
-                const next = index === engines.length - 1 ? engines[0] : engines[index + 1];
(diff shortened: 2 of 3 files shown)


Gusted:
>     * I don't know how this project's translation system works, or if I need to update the other locale files with anything. This currently fails a unit test about the locale files.

Yeah we require that have all other config files as well to include this new translation.



>     * How the original shortcut behaves might merit changes if this PR might be merged into the main repo.

The were made back in the day and probs before we had some more advanced sitelist system and never really got updated.



>     * It will change the theme's engine instead of just changing the engine per-site. I'll defer fixing this until contributors have chimed in about whether they think any of this is feasible to have merged in.

It's quite a mess around with the different presets etc and to detect all of them and know which one to change.

Gusted:
> Also, I have a question for you, @Gusted. The last page of open PRs is all your pending/open PRs. I don't know if you want me to flag/ping you on those or if they are all still valid and pending work you eventually intend to merge/finish. The oldest PR dated back about 2 1/2 years from May 2020. Could you please give me guidance on those in particular?

Feel free to ping me on it, it gives me an excuse to actually take a look at it again 😅.

PR #5996, opened 2021-06-09
Removing redundancy github.*.* includes github.com

n-v-o:
I should have seen this when i submitted my last PR but

`github.*.*` includes `github.com`, so no reason to have both. I just eliminated the redundancy.


Gusted:
> github._._ includes github.com, so no reason to have both. I just eliminated the redundancy.

That isn't how the matching work.

The specific *.* means you will have something like github.gov.uk as TLD which github.com isn't.

n-v-o:
Oh right. My bad. Overthinking it

PR #5997, opened 2021-06-09
Add 2 sites to global dark sites list

Elaborendum:


src/config/dark-sites.config
```diff
@@ -740,6 +741,7 @@ telugucz.com
 terminal.sexy
 textfiles.com
 tf2mart.net
+tgw1916.net
```
Gusted, line 744: Not dark

Gusted: You shouldn't disable dark reader on a site if it's not doing it's job. That's like if a feature has a bug that you remove the whole feature from the application...

Elaborendum: Yeah, in hindsight it doesn't make much sense. I will just add the page that's actually fully dark and fix the rest later in another PR.

Gusted (approved):

Issue #5947, opened 2021-06-03
[BUG] Jira cloud hangs when typing /panel


samuelgfeller:
# Bug Report

## Describe the bug

The new Jira cloud issue view has a functionality where it creates a rectangular box highlighted with a background color and a logo on the left side. It generates when typing `/panel` in the issue description.   
After clicking on a panel suggestion (warning, info, error etc.) the tab freezes until the browser alerts that the page didn't respond.   

## Steps to reproduce

* Go to a Jira issue with the new view (@Gusted please [contact me](mailto:contact@samuel-gfeller.ch), so I can give you the credentials to connect you to my jira workspace and let me know here when you wrote the email)
* Enable the dark reader extension
* Start to type `/panel` in the description view [like shown in this video](https://live.staticflickr.com/video/51222069128/fa1ac95d15/1080p.mp4?s=eyJpIjo1MTIyMjA2OTEyOCwiZSI6MTYyMjcwNjY5NSwicyI6ImY4N2VkODk0ZWRhZWMzMzBiOTIyMmNiY2MwMGY0Yjk4OGI4MjMwYTkiLCJ2IjoxfQ)
* Select a panel from the dropdown 

## Expected behavior

Panel creates and tab not freezing

## Actual behavior

Tab stops responding and after some time the browser alerts that the tab didn't respond.

## Screenshots

[This video](https://live.staticflickr.com/video/51222069128/fa1ac95d15/1080p.mp4?s=eyJpIjo1MTIyMjA2OTEyOCwiZSI6MTYyMjcwNjY5NSwicyI6ImY4N2VkODk0ZWRhZWMzMzBiOTIyMmNiY2MwMGY0Yjk4OGI4MjMwYTkiLCJ2IjoxfQ)

## System Information

- OS:  Windows 10
- Browser:  Chrome  91.0.4472.77 / Microsoft Edge 91.0.864.37
- Dark Reader Version:  4.9.33

## Additional context

Note: Bug only appears in edit mode. If panel was created without dark reader on and description is saved and only then dark reader is enabled, it can display the panels.

Gusted:
>     * Go to a Jira issue with the new view (@Gusted please [contact me](mailto:contact@samuel-gfeller.ch), so I can give you the credentials to connect you to my jira workspace and let me know here when you wrote the email)

:) Done

samuelgfeller:
Probably linked with #1595

Gusted:
[...]

PR #5999, opened 2021-06-10
Parent div to ignore inline style

samuelgfeller:
As discussed in #5947 

Also, I removed the avatar preloading thing as it changes very frequently (multiple times a day)


Gusted (approved):
Looking good

Issue #777, opened 2018-10-16
Some questions about Firefox's default black theme


1608637229:
I set up a black squat theme on Firefox, but every time I restart my browser, Firefox's main page will change back to bright. I think this is caused by this extension, I hope you can fix it, thank you!
    In addition, in the search engine of Baidu(www.baidu.com), it only works on the first page, and all following pages are changed back to bright.

1608637229:
My settings are incorrect, I changed the theme to custom.

Issue #740, opened 2018-09-21
Dark Reader theme uses Firefox light theme as a base, causing some issues with the new tab.


TheV360:
I usually use the default Firefox dark theme for most of my browsing. On the new tab page, it looks like this:

![Neat, the default Firefox dark theme also darkens the new tab page!](https://user-images.githubusercontent.com/25177071/45897682-d1af9a00-bda5-11e8-869d-0cec2d2c7010.png)

However, when I enable the Dark Reader browser theme, the new tab page isn't dimmed, and it actually gets set to its light theme counterpart.

![Oh no, the page is bright and hurting my eyes!](https://user-images.githubusercontent.com/25177071/45897636-adec5400-bda5-11e8-8f42-a4dc5f7212cb.png)

It seems that Dark Reader uses the Firefox light theme as a base for its own theme. I think that it should choose which theme to use as a base based on whether Dark Reader is on or not. This is a very minor issue, though. The only difference between themes is a slight change in brightness and text color, but it makes Dark Reader feel more consistent with the rest of the browser.

I am using version 62.0 of Firefox, and version 4.6.8 of Dark Reader.

TheV360:
Issue with Firefox, not Dark Reader. Firefox's new tab page only changes to the dark theme when the default dark theme is selected. Not even Firefox Color can change the new tab page theme.

PR #6007, opened 2021-06-11
Fix dynamic for ups.com

Gitoffthelawn:


src/config/dynamic-theme-fixes.config
```diff
@@ -11868,6 +11868,16 @@ CSS
 
 ================================
 
+ups.com
+
+CSS
+#ups-alertsWrap,
+#ups-skipNav {
+    background: #740 !important;
```
Gusted, line 11876: Hmm hardcoded values 👀

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
Gitoffthelawn: I use hard coded colors locally because they are fast work well for me, but I agree 100% that they may not be ideal for the community.

I use CSS overlays locally because I can create the fixes faster using my tools than using the Dark Reader UI.  I wanted to share with our community, so I converted several of them to Dark Reader syntax/format.

If you let me know a better way to handle this case, I'll make edits here and also use whatever knowledge you teach me for all future submissions.

Gusted: So for this kind of values you simply can wrap it into a `${#740}` and then check the elements again and copy the background-color value. Wrap that again into a `${pasted_value}` and use that instead. It will on your settings default to `#740` and for other users they will be slightly different to match with the user's settings.

Gitoffthelawn: Thanks.  I think I'm following what you are saying.  Here's what I think you are saying... please correct any errors on my part (I corrected what I think may be one small error on your part).

1. Instead of using my custom CSS overlay to verify the code is working, use Dark Reader's Dev Tools.
2. In Dark Reader's Dev Tools, assign a value of `${#740}` to the `background` property of the element.
3. Go to Firefox's Dev Tools, and see what value gets assigned to that element.
4. Copy that value to the clipboard; let's call it `${copied_value}`
5. In Dark Reader's Dev Tools, replace `${#740}` with `$(copied_value}`.
6. In Dark Reader's Dev Tools, hit the `Apply` button.
7. If things don't looks good, adjust the color (or use a different Dark Reader supported technique).
8. Once everything looks good, copy the entire section to GitHub and create a pull request.

Did I get it right?

It's a fair amount of time/energy just to fix one site, but I'll continue to do my best to create pull requests for each site I think may have enough general interest to be helpful to others.

Gusted: > Did I get it right?

Yeah you did.

It's a very special case. As most of the time what I do is get the originial value of such property(not the one dark reader is applying currently) and throw that into `${}` and 9/10 it's perfection or I use a color that looks like it. e.g. in this cause `${orange}`, however in such case it's a bit longer.

Gitoffthelawn: Thanks.  Glad I got it right. :)   I'll copy our dialogue to a text file for reference for future contributions.

I actually had a hunch that hardcoded values were a no-no, but I thought I saw quite a few items in the file that use them, so I thought maybe it was okay.  It might be worthwhile scanning through it for hardcoded values?

BTW, are there exceptions for `#FFF` and `#000` (and maybe `#777`)?  IIRC, there were multiple existing fixes that used those values, and if I'm not mistaken, you recently accepted another one of my pull requests that used one of those values.

Gusted: > so I thought maybe it was okay. It might be worthwhile scanning through it for hardcoded values?

Or they an exception which on a occasion happens when something got broken like a feature etc(qr code border). Or they where merged into the config 1+ year ago. 

> BTW, are there exceptions for #FFF and #000 (and maybe #777)? IIRC, there were multiple existing fixes that used those values, and if I'm not mistaken, you recently accepted another one of my pull requests that used one of those values.

With #FFF #000 you can just use var(--darkreader-neutral-background) and var(--darkreader-neutral-color) as they are more natural overall. 

Hmm that must be by mistake, lately got more random "merged" when I submit feedback.

Gitoffthelawn: Thanks for the additional details.  Noted and appreciated.

Gusted (approved):
Thank you!

src/config/dynamic-theme-fixes.config
```diff
@@ -11868,6 +11868,15 @@ CSS
...
 
+ups.com
+
+CSS
+#ups-alertsWrap,
+#ups-skipNav {
+    background: ${#FE9} !important;
+}
+
+================================
```
Gusted, line 11879: Add an extra new line after line.

Gitoffthelawn:
Based on your helpful feedback, I pushed an improvement to my patch in my branch for this change (patch-5) (https://github.com/Gitoffthelawn/darkreader/tree/patch-5).

I see a failed check in this repo though... what's the story there?

Do I need to do anything else?

Gitoffthelawn (commented):
Looks good.

src/config/dynamic-theme-fixes.config
Gusted, line 11879:
```suggestion
================================

```

Gusted (approved):

PR #6008, opened 2021-06-11
Dynamic fix for websemantics.uk

Gitoffthelawn:


Gusted:
Could you share a URL where I can observe this behavior?

Gitoffthelawn:
I can't find which specific page needed that code, but I found a bunch of pages that benefit from some related new code I wrote.  I'll submit that soon.  It's always possible the new code will fix the page(s) I found before that needed fixing, so I think I'll replace the code submitted above with the new code, and if I (or someone else) comes across a page that requires fixing `#output *`, we can always add this original code too.

In the meantime, I found a page of their site that requires that an element not be styled by Dynamic Mode at all:

https://websemantics.uk/tools/wcag-colour-contrast-checker/

It's the `#exampleText` element.  How can one make Dynamic Mode completely ignore a specific CSS selector?

PR #6023, opened 2021-06-13
Remove unnecessary background page HTML file

bershanskiy:
This just removes an unecessary HTML file from extension source and
build. If needed, browser will generate a background page automatically.

This change is a tiny refactor which just removes an extra file. It does not change anything in terms of functonality of the extension because browsers automatically generate a placeholder background page if needed. E.g., if you open backgound page in Chrome dev tools, you'll see this:
```
<!DOCTYPE html>
<html><head></head><body>
<script src="background/index.js"></script>
</body></html>
```
Which is pretty much identical to current contents of [src/background/index.html](https://github.com/darkreader/darkreader/blob/master/src/background/index.html)

# Tests
Tested manually (built extension, loaded it, and pressed various buttons to make sure everything still works).


Gusted:
Hi @bershanskiy!

Does this fix any issues? Or have some value, I'd prefer not to modify the manifest.json as the extensions stores don't like that and most likely will reject your application, some other extensions got in trouble with that.

bershanskiy:
Thanks for prompt reply @Gusted!

> Does this fix any issues?

No, it just simplifies things a bit and removes some technical debt.

My long-term goal is to make Dark Reader background non-persistent and therefore (1) more efficient for use on mobile devices and (2) compatiable with service worker contexts of Manifest V3. This PR just advances goal 2 a bit, but it's not crucial.

> I'd prefer not to modify the manifest.json as the extensions stores don't like that and most likely will reject your application, some other extensions got in trouble with that.

I know that extension stores have issues with changing extensions changing permissions, but I don't think I have ever heard of extensions being rejected because changing background value. I'm pretty sure that would be a bug in CWS review process. Either way, this change is not critical and I can work towards non-persistent background without it.

Gusted:
> My long-term goal is to make Dark Reader background non-persistent

You don't have to remove the index.html for that one, actually I don't have a clue why we have persistent at the moment as the background-script is event driven(which shouldn't be impacted persistent?) 

> (2) compatiable with service worker contexts of Manifest V3.

Even though I still think this whole manifest v3 is a big joke from Google towards developers we certainly have to embrace for it.

> more efficient for use on mobile devices 

That's another issue that won't be fixed by removing persistent background.
It's the content_scripts you want to be looking at for performance improvements.

> but I don't think I have ever heard of extensions being rejected because changing background value. 

IIRC Firefox had some things to do with it.

bershanskiy:
> You don't have to remove the index.html for that one

Yes, this commit is not necessary for non-persistent background. You can just change `"persistent"` to `false` and that's it. Sorry if my comment wasn't clear.

> actually I don't have a clue why we have persistent at the moment as the background-script is event driven(which shouldn't be impacted persistent?)

There are a few gotchas there. For example, Chrome persists background context for as long as it has at least one open port and you might have to use ports for comminication with DevTools tab because of undocumented aspects of Chrome APIs... So you might need to establish a port connection when you user opens DevTools page and make sure to shut it down when DevTools are closed.

> > more efficient for use on mobile devices
>
> That's another issue that won't be fixed by removing persistent background.

What do you mean? When I disable Dark Reader on Dark Reader pop-up page and open Chrome task manager, I see Dark Reader process consuming ~30 MB of RAM. That's non-negligiable amount for a resource-constrained device, especially considering Dark Reader is "disabled".

> It's the content_scripts you want to be looking at for performance improvements.

OK, I'll take a look.

PR #6034, opened 2021-06-15
Fix dynamic for plumbingforums.com

Gitoffthelawn:
Fixes the search page too

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -8909,6 +8909,23 @@ a[title="Google apps"]
 
 ================================
 
+plumbingforums.com
+
+INVERT
+.node-icon::before
+
+CSS
+div,
+.node-icon {
+    background-color: var(--darkreader-neutral-background) !important;
+}
+.input:focus,
+.input.is-focused::placeholder {
+    background: var(--darkreader-neutral-background) !important;
+}
+
+================================
+
 plus.google.com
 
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -8909,6 +8909,23 @@ a[title="Google apps"]
...
+.node-icon::before
+
+CSS
+div,
+.node-icon {
+    background-color: var(--darkreader-neutral-background) !important;
+}
+.input:focus,
+.input.is-focused::placeholder {
+    background: var(--darkreader-neutral-background) !important;
```
Gusted, line 8924: Any particular reason why you used `background` property? Otherwise you can merge the selector with the CSS declaration above.

Gitoffthelawn: Yes, for the input fields, it's a `background-image` that causes the problematic background, so we need to use either `background` or `background-image`, but not `background-color`.  And since we are replacing the image with a color, and not an image, we have to use `background` and not `background-image`.

The alternative is:
```
input:focus,
.input.is-focused::placeholder {
}
```

and then add those same selectors to the other ruleset.

The code I pushed will have better performance and less memory use than this alternative.

Gusted: Got it 👍🏽

Gusted (approved):

PR #6036, opened 2021-06-15
Math symbols visiblity on oalevelsolutions.com

mobi2002:
The math symbols on oalevelsolutions.com are not correctly shown with dark reader dynamic mode as they are transparent images which need to be inverted.
example link: http://oalevelsolutions.com/past-papers-solutions/cambridge-international-examinations/as-a-level-mathematics-9709/pure-mathematics-p1-9709-01/year-2020-february-march-p1-9709-12/cie_20_fm_9709_12_q_3/

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -8381,6 +8381,13 @@ CSS
 
 ================================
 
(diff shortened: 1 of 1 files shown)


Gusted (approved):
Thank you!

PR #6038, opened 2021-06-15
Tutanota 2FA QR Code fix

### LINKED ISSUE #5420: [Broken Website] Tutanota 2FA QR Code doesn't appear properly
## Website
mail.tutanota.com

## How to Reproduce
Go to settings in the Tutanota account and try to add 2FA.

## Expected behavior
A proper QR code supposed to show up.

## Actual behavior
It shows a white square without any black lines when dark reader extension is enabled.

## System Information
Firefox 87 
Windows 10 64 bit
Dark reader version 4.9.29

mobi2002:
Inverting the qr code makes it visible.
Fixes #5420

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -7147,6 +7147,13 @@ CSS
 
 ================================
 
+mail.tutanota.com
+
+INVERT
+div.mb > div > svg > path
+
+================================
+
 makeuseof.com
 
 IGNORE IMAGE ANALYSIS
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -7149,8 +7149,10 @@ CSS
 
 mail.tutanota.com
 
-INVERT
-div.mb > div > svg > path
+CSS
+div.mb > div > svg > path {
+    fill: black !important;
+}
 
 ================================
 


src/config/dynamic-theme-fixes.config
```diff
@@ -7147,6 +7147,13 @@ CSS
 
 ================================
 
+mail.tutanota.com
+
+INVERT
```
Gusted, line 7152:
```suggestion
IGNORE INLINE STYLE
```

mobi2002: Ignore inline style doesn't work as this specific selector by itself doesn't appear to have inline style. The QR code remains invisible with this.
Another solution could be to hard code the fill value for this selector. Hard code because the qr should be black on white regardless of the theme. The white part already exists with dark reader enabled.

`CSS
div.mb > div > svg > path {
    fill: black !important;
}`

Contributor guidelines say to not hard code color codes but wrapping it as ${black} makes dark reader make it white when dark mode is enabled.
 
Please tell me if I should try something else.

Gusted: > Ignore inline style doesn't work as this specific selector by itself doesn't appear to have inline style. The QR code remains invisible with this.

You sure? You might wanna change the selector to `div.mb > div > svg *` as INVERT is a possible solution, but not something that should be used in this situation.

Gusted (approved):
That should also work.

PR #6048, opened 2021-06-17
oalevelsolutions.com math symbol fix

mobi2002:
This was fixed in #6036 but apparently elsewhere on the site the selector is a bit different so here is a solution that should work on all of them.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -8423,7 +8423,7 @@ CSS
 oalevelsolutions.com
 
 INVERT
-* > p > span > img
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -8423,7 +8423,7 @@ CSS
 oalevelsolutions.com
 
 INVERT
-.WordSection1 > p > span > img
+* > p > span > img
```
Gusted, line 8426:
```suggestion
p > span > img
```

No need to specify `* >`

Gusted (approved):

PR #6062, opened 2021-06-18
Google docs and slides use user defined text selection background color

nash-pillai:


src/config/dynamic-theme-fixes.config
```diff
@@ -3486,6 +3486,10 @@ CSS
 .kix-cursor-caret {
     border-color: ${black} !important;
 }
+.kix-selection-overlay {
+    background-color: var(--darkreader-selection-background) !important;
+    border: none !important;
```
Gusted, line 3491: What about styling the border as well? They seems to do it on purpose.

nash-pillai: I didn't think it mattered if I removed the border since it was only 1 pixel. I will replace `none` with `var(--darkreader-selection-background)` and that should make the border visible and with the correct color.

Gusted (approved):
Seems good, thank you for your contribution!

PR #6063, opened 2021-06-18
poeditor.com fix

maximmax42:


arche-dev (changes requested):
```css
.chart img
```

Make it one selector per line, something like:

```css
.someselector
h1
```

Gusted:
> ```css
> .chart img
> ```
> 
> Make it one selector per line, something like:
> 
> ```css
> .someselector
> h1
> ```

I'm not sure what we are talking about but `.chart img` looks good into me. As I don't assume you want to invert every `img` which I won't be merging.

maximmax42:
Well, there's only one img inside chart class that needs to be inverted. I can change it to `.chart > img` but .chart only has one layer of child objects anyway, so `>` would be redundant.

Gusted:
In that case I don't see @arche-dev relevance.

Gusted (approved):
Thank you for your contribution!

arche-dev:
> I'm not sure what we are talking about but `.chart img` looks good into me. As I don't assume you want to invert every `img` which I won't be merging.

Ah, makes sense

PR #6072, opened 2021-06-19
Add fixes for Google Classroom

Vedanth-P:
On any Google Classroom, in the "Classwork" tab, there is a list, and each item of the list is some material or assignment. A list item gets highlighted when is it clicked or hovered on.
When an item is clicked (and thus highlighted) and the item above it hovered on, they both normally get the same color (in this example, the lower one has been clicked and the higher hovered over);
Two highlighted in light mode

but with Dark Reader, the lower one goes completely blank.
Two highlighted in dark mode - 1

With some experimentation, I came to the conclusion that giving the two highlighted nodes different colors solves the problem; I still don't know why though.
This change leaves the opacity of the hovered-over list item at 100%; all the remaining list items are given an opacity of 99%. This creates the required color difference without drastically changing the look to the user.
Two highlighted in dark mode - 2

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2311,6 +2311,9 @@ div[style="bottom: 0px;"] > div[style^="opacity:"] div[role="button"] > div:not(
 div[role="menu"] > div[role="menuitem"] > div > div:not([style*="background-image"]) {
     background-image: url('//ssl.gstatic.com/docs/common/viewer/v3/v-sprite35.svg') !important;
 }
+li[guidedhelpid="classworkTopicListGh"]:not(hover) > div {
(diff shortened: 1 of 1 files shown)


Gusted:
> With some experimentation, I came to the conclusion that giving the two highlighted nodes different colors solves the problem; I still don't know why though.
This change leaves the opacity of the hovered-over list item at 100%; all the remaining list items are given an opacity of 99%. This creates the required color difference without drastically changing the look to the user.

Great we are on the same page. How in earth does this work - yes the fix works as intended, but in theory it shouldn't. I see some related code towards these elements and also does some check for ctrl/alt. So I guess they are doing something fishy. Anyway great finding!

Gusted (approved):

PR #6078, opened 2021-06-20
Ensure proper destroy of manager 

### LINKED ISSUE #6075: Styles are cached
Dark Reader apparently now caches styles. That makes it impossible to fix the broken style when visiting websites based on mdBook (most of which are consistently broken all the time). The only solution is to turn off Dark Reader, refresh the page, and make sure the styles are fully loaded before toggling it for the first time.

Thanks to this caching behavior it is even more inconvenient to fix broken webpages.

# Broken Website
## Website
https://poga.github.io/lua53-notes
Most other websites based on mdBook 

## How to Reproduce
visit site

## Expected behavior
Syntax highlighting exists, header border isn't way too bright, the entire page doesn't share one background color

## Actual behavior
The exact opposite

## Screenshots
Broken:
image

Non-broken:
image

## System Information
- OS: Microsoft Windows 10 Home build 19041
- Browser: Google Chrome 91.0.4472.77 (Official Build) (64-bit) (cohort: Stable)
- Dark Reader Version: 4.9.33

## Additional context
Possibly similar problem to #3540?

### LINKED ISSUE #6135: [Broken Website] Quillbot lot of buttons and text missing
# Broken Website
## Website

https://quillbot.com/premium

## Expected behavior

Buttons and text to be visible

## Actual behavior

Not visible buttons

## Screenshots

image

image

## System Information

- OS:  Windows 10
- Browser:  Edge stable
- Dark Reader Version:  4.9.33

Gusted:
- Ensure that `LoadingEnd` is called, to check if the `fallbackStyle` can be removed.
- Ensure that we reject the promise of `linkLoading`
- Ensure that when dark reader is toggled that the map/set is cleared.
- Add a bunch of debugging code to make it easier next time to debug these kinds of issues.
- Error on loading styles that doesn't have a `href`

- Resolves #6075
- Resolves #6135

### DIFF
--- a/src/inject/dynamic-theme/index.ts
+++ b/src/inject/dynamic-theme/index.ts
@@ -204,7 +204,7 @@ function createDynamicStyleOverrides() {
 }
 
 let loadingStylesCounter = 0;
-const loadingStyles = new Set();
+const loadingStyles = new Set<number>();
 
 function createManager(element: StyleElement) {
     const loadingStyleId = ++loadingStylesCounter;
@@ -224,6 +224,7 @@ function createManager(element: StyleElement) {
     function loadingEnd() {
         loadingStyles.delete(loadingStyleId);
         logInfo(`Removed loadingStyle ${loadingStyleId}, now awaiting: ${loadingStyles.size}`);
+        logInfo(`To-do to be loaded`, loadingStyles);
         if (loadingStyles.size === 0 && isDOMReady()) {
             cleanFallbackStyle();
         }
@@ -269,7 +270,7 @@ function onDOMReady() {
         cleanFallbackStyle();
         return;
     }
-    logWarn(`DOM is ready, but still have ${loadingStyles} styles being loaded.`);
+    logWarn(`DOM is ready, but still have styles being loaded.`, loadingStyles);
 }
 
 let documentVisibilityListener: () => void = null;
--- a/src/inject/dynamic-theme/index.ts
+++ b/src/inject/dynamic-theme/index.ts
@@ -269,7 +269,7 @@ function onDOMReady() {
         cleanFallbackStyle();
         return;
     }
-    logWarn(loadingStyles);
+    logWarn(`DOM is ready, but still have ${loadingStyles} styles being loaded.`);
 }
 
 let documentVisibilityListener: () => void = null;
--- a/src/inject/dynamic-theme/index.ts
+++ b/src/inject/dynamic-theme/index.ts
@@ -2,7 +2,7 @@ import {overrideInlineStyle, getInlineOverrideStyle, watchForInlineStyles, stopW
 import {changeMetaThemeColorWhenAvailable, restoreMetaThemeColor} from './meta-theme-color';
 import {getModifiedUserAgentStyle, getModifiedFallbackStyle, cleanModificationCache, parseColorWithCache, getSelectionColor} from './modify-css';
 import type {StyleElement, StyleManager} from './style-manager';
-import {manageStyle, getManageableStyles, storedRejectorsForLoadingLinks} from './style-manager';
+import {manageStyle, getManageableStyles, cleanLoadingLinks} from './style-manager';
 import {watchForStyleChanges, stopWatchingForStyleChanges} from './watch';
 import {forEach, push, toArray} from '../../utils/array';
 import {removeNode, watchForNodePosition, iterateShadowHosts, isDOMReady, addDOMReadyListener, removeDOMReadyListener} from '../utils/dom';
@@ -465,7 +465,7 @@ export function removeDynamicTheme() {
     shadowRootsWithOverrides.clear();
     forEach(styleManagers.keys(), (el) => removeManager(el));
     loadingStyles.clear();
-    storedRejectorsForLoadingLinks.clear();
+    cleanLoadingLinks();
     forEach(document.querySelectorAll('.darkreader'), removeNode);
 
     adoptedStyleManagers.forEach((manager) => {
--- a/src/inject/dynamic-theme/index.ts
+++ b/src/inject/dynamic-theme/index.ts
@@ -2,8 +2,7 @@ import {overrideInlineStyle, getInlineOverrideStyle, watchForInlineStyles, stopW
 import {changeMetaThemeColorWhenAvailable, restoreMetaThemeColor} from './meta-theme-color';
 import {getModifiedUserAgentStyle, getModifiedFallbackStyle, cleanModificationCache, parseColorWithCache, getSelectionColor} from './modify-css';
 import type {StyleElement, StyleManager} from './style-manager';
-import {loadingLinkStyles} from './style-manager';
-import {manageStyle, getManageableStyles} from './style-manager';
+import {manageStyle, getManageableStyles, loadingLinkStyles} from './style-manager';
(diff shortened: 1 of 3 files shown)


Gusted:
If someone want's to test it out:
[binaries.zip](https://github.com/darkreader/darkreader/files/6682344/binaries.zip)

Gusted:
@Karmalakas could you try the binary above? The other issue very resembles your current situation at gurushots.

arche-dev:
> If someone want's to test it out:
> [binaries.zip](https://github.com/darkreader/darkreader/files/6682344/binaries.zip)

Syntax highlighting works, so other broken things also should be working :>

Gusted:
> so other broken things also should be working :>

Could be, at least more situations should be covered now and not result in a fallback style being displayed.

src/inject/dynamic-theme/index.ts
```diff
@@ -265,7 +267,9 @@ const cancelRendering = function () {
 function onDOMReady() {
     if (loadingStyles.size === 0) {
         cleanFallbackStyle();
+        return;
     }
+    logWarn(loadingStyles);
```
alexanderby, line 272: Maybe more clear message should be here?

src/inject/dynamic-theme/style-manager.ts
```diff
@@ -81,6 +81,9 @@ document.addEventListener('__darkreader__inlineScriptsAllowed', () => {
     canOptimizeUsingProxy = true;
 });
 
+export let loadingLinkStyle = 0;
```
alexanderby, line 84: Looks like it shouldn't be exported

src/inject/dynamic-theme/style-manager.ts
```diff
@@ -145,6 +148,7 @@ export function manageStyle(element: StyleElement, {update, loadingStart, loadin
 
     let isLoadingRules = false;
     let wasLoadingError = false;
+    let loadingLinkID = 0;
```
alexanderby, line 151: Let's call it counter instead of ID. It becomes ID when you pass it into a function where the reference to the counter is lost.

src/inject/dynamic-theme/style-manager.ts
```diff
@@ -162,7 +166,9 @@ export function manageStyle(element: StyleElement, {update, loadingStart, loadin
                 isStillLoadingError(accessError)
             ) {
                 try {
-                    await linkLoading(element);
+                    loadingLinkID = ++loadingLinkStyle;
```
alexanderby, line 169: Hmm, looks like you have to use `const` here, so each style has its own ID. Also there is a naming convention to use `Id` instead of `ID` in JS and other similar languages, e.g. `document.getElementById`.

src/inject/dynamic-theme/style-manager.ts
```diff
@@ -422,6 +428,11 @@ export function manageStyle(element: StyleElement, {update, loadingStart, loadin
         pause();
         removeNode(corsCopy);
         removeNode(syncStyle);
+        loadingEnd();
+        if (loadingLinkID && loadingLinkStyles.has(loadingLinkID)) {
```
alexanderby, line 432: I think you need a boolean here, e.g. `isLinkLoading` instead of `loadingLinkID`. Or maybe don't need at all.

src/inject/dynamic-theme/style-manager.ts
```diff
@@ -422,6 +428,11 @@ export function manageStyle(element: StyleElement, {update, loadingStart, loadin
...
+            loadingLinkStyles.get(loadingLinkID)();
```
alexanderby, line 433: Looks a bit ( )( ). Try putting a callback into another variable.

src/inject/dynamic-theme/style-manager.ts
```diff
@@ -465,20 +476,23 @@ export function manageStyle(element: StyleElement, {update, loadingStart, loadin
     };
 }
 
-async function linkLoading(link: HTMLLinkElement) {
+async function linkLoading(link: HTMLLinkElement, loadingID: number) {
```
alexanderby, line 479: Use `Id` for `ID`

src/inject/dynamic-theme/style-manager.ts
```diff
@@ -465,20 +476,23 @@ export function manageStyle(element: StyleElement, {update, loadingStart, loadin
...
             cleanUp();
+            logInfo(`Linkelement ${loadingID} has been loaded`);
             resolve();
         };
         const onError = () => {
             cleanUp();
-            reject(`Link loading failed ${link.href}`);
+            reject(`Linkelement ${loadingID} couldn't be loaded. ${link.href}`);
         };
+        loadingLinkStyles.set(loadingID, reject);
```
alexanderby, line 495: As far as you put rejectors here, give `loadingLinkStyles` a proper name. Also, take a look if you can use `link` element itself as a key in a WeakSet, Set, WeakMap or Map instead of numeric IDs.

alexanderby (changes requested):

Karmalakas:
> @Karmalakas could you try the binary above? The other issue very resembles your current situation at gurushots.

This didn't help. Same issue persists
image

Will keep monitoring still, because this kind of popup is not the main issue

src/inject/dynamic-theme/style-manager.ts
Gusted: Good catch 😄

Gusted: Good catch 😄

src/inject/dynamic-theme/index.ts
```diff
@@ -459,6 +464,8 @@ export function removeDynamicTheme() {
     });
     shadowRootsWithOverrides.clear();
     forEach(styleManagers.keys(), (el) => removeManager(el));
+    loadingStyles.clear();
+    storedRejectorsForLoadingLinks.clear();
```
alexanderby, line 468: I feel like a bit of Encapsulation (OOP) should be here (even though it's one line): This file should know as few details as possible, so you should import some method for cleaning up something.

src/inject/dynamic-theme/style-manager.ts
```diff
@@ -81,6 +81,9 @@ document.addEventListener('__darkreader__inlineScriptsAllowed', () => {
     canOptimizeUsingProxy = true;
 });
 
+let loadingLinkCounter = 0;
+export const storedRejectorsForLoadingLinks = new Map<number, (reason?: any) => void>();
```
alexanderby, line 85: "stored" part can be removed

src/inject/dynamic-theme/style-manager.ts
```diff
@@ -422,6 +427,13 @@ export function manageStyle(element: StyleElement, {update, loadingStart, loadin
         pause();
         removeNode(corsCopy);
         removeNode(syncStyle);
+        loadingEnd();
+        let reject: (reason?: any) => void;
+        if (storedRejectorsForLoadingLinks.has(loadingLinkId)) {
+            reject = storedRejectorsForLoadingLinks.get(loadingLinkId);
+            storedRejectorsForLoadingLinks.delete(loadingLinkId);
+        }
+        reject && reject();
```
alexanderby, line 436: You can put "reject" variable and it's call into the "if" scope.

alexanderby (changes requested):

alexanderby (approved):

Issue #6032, opened 2021-06-14
Turn Dark Reader into what's sussposed to be. 


Gusted:
**@Gusted allows @Gusted to not follow the issue template** 

As a response to, more like a trigger to finally make this post/issue:

> for me it's not about looking good, it's about health and pain management. aesthetics take a distant back seat to that, which is why i'm so happy that dark reader is (at least at this moment) behaving in exactly the way i need

_Originally posted by @esotericist in https://github.com/darkreader/darkreader/issues/4659#issuecomment-860955570_

~~I think~~ I'm certain, that I've failed to understand what dark reader means to certain people since I began hacking around dark reader and now actively maintains it. And I'm sorry for that.   

I'm not even sure where to start, so let's start with everyone who is going to comment on this and reading this,
to understand that there are 2 types of users in the dark reader user base. _I'm going to refer them in the text a lot_

1) The most popular one, people who don't want some white pixels burning into their screen and/or are #dark-mode life or for some reasons they are vampires and think that screens can give Vitamin D. _I belong in this type_ 

2) The less popular one, but the one most benefiting from dark reader. People with health issues, people that without an extension like Dark Reader will have a less if not horrible experience from visiting webpages, that due to these white-ish pixels on a screen can affect these kinds of people's more than a normal person.  

Dark reader is officially listed as an Accessibility(Firefox as Appearance but they don't have this category) extension, however honestly I've never let the focus onto making dark reader better for accessibility reasons, it was always "more feature's" that doesn't improve accessibility by a bit. I want to bring the focus to improve accessibility on web pages with dark reader and help/improve those people that truly benefit from dark reader. Dark reader shouldn't be all about having features that even a smaller % of the user bases(than second type users) uses. 

So, what do I want to do? 
I've added the "help wanted" label and will probably pin this after refactoring my text a thousand times. As I like to have this been seen by a good amount of people.

- People that belong to the second type, please please submit feedback on either this issue(more general) or in a new issue(more specific use-cases) about how dark reader can be improved for you. Let your voice be heard. 
- As I tend to see people that belong in the second type have a good amount of configurations with the brightness, contrast etc. I will try my best to make those configurations more use-able and as an example by going to affect images on webpages by those configurations(see: #509 & #689)
[...]

PR #6080, opened 2021-06-20
Fix Dynamic Mode for downdetector.com on all ~50 global sites

Gitoffthelawn:


src/config/dynamic-theme-fixes.config
```diff
@@ -3551,6 +3551,20 @@ CSS
...
 
+downdetector.*
+allestoringen.*
+xn--allestrungen-9ib.*
+
+CSS
+.img-fluid {
+    filter: invert(0.9) hue-rotate(180deg) brightness(0.8) !important;
+}
+#map-card {
```
Gusted, line 3562: I don't really like the usage of custom filters as they can go out of sync by user's settings. 
Also `.mapboxgl-canvas` is another one to be included for INVERT.

Gitoffthelawn: Thanks.  Which user settings can mess up custom filters?

Initially, I thought I agreed with you that applying a real complex custom filter could possibly create a big issue for some people, but after a little more thought, I'm haven't come up with real-world situations where this would be the case.  IIRC, the resultant image should always be the same if DR is working correctly.

The only issue I can think of off-hand is maybe when someone is, for some reason, using Dark Reader to force a light dynamic theme.  I'm not sure how many people actually do that... do you have any idea?  Even then, I'm not sure if it's an issue.

Regarding this specific filter (no longer speaking in generalities), just inverting, rotating the hue, and lowering the brightness a touch I think would work for just about everyone (if not everyone), especially the main target audience (people who want a dark theme without excessive contrast).  I'm interested in learning about cases where a relatively simple filter like this will cause an issue, though.  I've used them in code outside of DR, and I want to ensure there isn't a case I'm missing.

Gusted: > Thanks. Which user settings can mess up custom filters?

Rather meant that certain setting the user touch brightness/contrast/sepia/grayscale, will be reflected on the page and also be reflected into the `INVERT` section. However with hardcoded filter's like these, they won't. And just stay static while the page is changing.

> I'm haven't come up with real-world situations where this would be the case. IIRC, the resultant image should always be the same if DR is working correctly.

I can think of one. Let's take the canvas as a example, as I have on every site with a few exceptions a very high contrast(141%), due to partially how my display is not really reflecting contrast as it should so it's more in balanced. I really notice hard coded values:

if you use the `INVERT` section
image

using the hardcoded value:
image


^ I've just noticed it doesn't have filter/hue-rotate which I really dislike especially in this situation.

So using the hardcoded value & `filter(1) hue-rotate(180deg)`
image

It's notice-able that's out of sync with the site, and every person encountering this will probably complain and a few of them will open up a bug report of this. 

> I'm not sure how many people actually do that... do you have any idea? Even then, I'm not sure if it's an issue.

Not really a lot, that I can think of... only users that have automation settings and don't disable dark reader at day but rather use the light theme.

> just inverting, rotating the hue, and lowering the brightness a touch I think would work for just about everyone (if not everyone),

Correct, should be good to go... however for people with custom filter settings, they will be notice-able that it isn't a correct fix. 

> especially the main target audience (people who want a dark theme without excessive contrast).

Yeah, I thought that was the main audience too, but I've failed to realize that a good amount of users also use it for other reasons, See #6032

> I've used them in code outside of DR, and I want to ensure there isn't a case I'm missing.

If something doesn't allow variables to play in and simply force such hardcoded value and you know that the site won't be changing, that is good and shouldn't be a problem. However I really try with these fixes to not break user experience, because they are out-of-sync. Il allow certain hardcoded values, but really rare edge cases(e.g. edges of QR codes so they will be recognized).

Gitoffthelawn: Thank you for all your feedback and insight.

Using this site as an example, how to you recommend handling those logos?  When I use the DR dynamic dark mode without any custom settings, some of them are virtually invisible because they are too dark, while others are too bright.

Gusted:
Closing.

PR #6082, opened 2021-06-20
Updated dark-sites (papermc.io)

Bungeefan:


benfoster04:
Note regarding Paper's [javadocs](https://papermc.io/javadocs) page, this is not dark by default at all, nor does it offer a dark mode.

Gusted:
#6804

PR #6085, opened 2021-06-20
Proper cleanup of inline style

Gusted:
When you remove a orginial inlineStyle darkreader's inline style wouldn't be removed. This is now fixed.

```html
<span style="color: red;">Hi</span> // Original HTML
<span data-darkreader-inline-color style="color: red; --darkreader-inline-color:#ff1a1a;">Hi</span> // Modified by dark reader
<span data-darkreader-inline-color style="--darkreader-inline-color:#ff1a1a;">Hi</span> // Modified by site to remove color: red
<span style="--darkreader-inline-color:#ff1a1a;">Hi</span> // Dark reader removes the attribute
```
This was the case, but you be better of removing the `--darkreader-inline-color` as well ->

```html
<span style="color: red;">Hi</span> // Original HTML
<span data-darkreader-inline-color style="color: red; --darkreader-inline-color:#ff1a1a;">Hi</span> // Modified by dark reader
<span data-darkreader-inline-color style="--darkreader-inline-color:#ff1a1a;">Hi</span> // Modified by site to remove color: red
<span>Hi</span> // Dark reader removes the attribute & inline CSS prop.
```

### DIFF
--- a/src/inject/dynamic-theme/inline-style.ts
+++ b/src/inject/dynamic-theme/inline-style.ts
@@ -85,8 +85,8 @@ const overrides: Overrides = {
 };
 
 const overridesList = Object.values(overrides);
-const customPropsList = {};
-overridesList.forEach(({cssProp, customProp}) => customPropsList[customProp] = cssProp);
+const normalizedPropList = {};
+overridesList.forEach(({cssProp, customProp}) => normalizedPropList[customProp] = cssProp);
 
 const INLINE_STYLE_ATTRS = ['style', 'fill', 'stop-color', 'stroke', 'bgcolor', 'color'];
 export const INLINE_STYLE_SELECTOR = INLINE_STYLE_ATTRS.map((attr) => `[${attr}]`).join(', ');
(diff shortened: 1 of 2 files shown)


src/inject/dynamic-theme/inline-style.ts
```diff
@@ -85,6 +85,8 @@ const overrides: Overrides = {
 };
 
 const overridesList = Object.values(overrides);
+const customPropsList = {};
```
alexanderby, line 88: Maybe it should has some other name? E.g. it looks like a map of Dark Reader's custom properties and real CSS properties.

src/inject/dynamic-theme/inline-style.ts
```diff
@@ -317,6 +319,11 @@ export function overrideInlineStyle(element: HTMLElement, theme: FilterConfig, i
         }
         if (overrides.hasOwnProperty(property)) {
             setCustomProp(property, property, value);
+        } else {
+            const customProp = customPropsList[property];
```
alexanderby, line 323: As far as I understand `customProp` holds real CSS prop here, while `property` is actually a custom property.

src/inject/dynamic-theme/inline-style.ts
```diff
@@ -317,6 +319,11 @@ export function overrideInlineStyle(element: HTMLElement, theme: FilterConfig, i
...
+            if (customProp && (!element.style[customProp] && !element.hasAttribute(customProp))) {
```
alexanderby, line 324: `style` property acts a little bit different, e.g. `style.backgroundColor` is used to get `background-color`. So you can use `style.getPropertyValue`. Also, what is `hasAttribute` meant to do?

alexanderby (changes requested):

src/inject/dynamic-theme/inline-style.ts
Gusted: > Also, what is hasAttribute meant to do?

The original value can be either in either in a attribute(e.g. `fill="red"``) or as style(`style="fill: red"`)

alexanderby (approved):

PR #6090, opened 2021-06-21
okonto.pl

FadeMind:
SSO login service from Ringier Axel Springer. Will be used in many services.
Inverted main logos on main domain and service domain.


Gusted (approved):
Seems good.

PR #6098, opened 2021-06-22
Fix shadows on Google

Vedanth-P:
Fixes shadows

Before this, there were two issues:
- The shadows around the search bar were lighter than the background.
- Without Dark Reader, the shadows around the search bar change become darker when when hovered over. This behaviour wasn't accurate with Dark Reader.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -13297,9 +13297,9 @@ CSS
 .minidiv .sfbg,
 #searchform.minidiv,
 .minidiv .emcav .RNNXgb .S6VXfe,
-.sbfc .RNNXgb,
+.CvDJxb .RNNXgb,
 .aajZCb {
-    box-shadow:0 1px 6px 0 rgba(0, 0, 0, 0.28) !important;
+    box-shadow:0 1px 6px 0 rgba(0,0,0,.28) !important;
 }
 .RNNXgb:hover {
     box-shadow:0 1px 6px 2px rgba(0,0,0,.28) !important;
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -13294,9 +13294,15 @@ img.act-icon-dark-gray
 [src="/chrome/static/images/google-translate/screen-english.png"]
 
 CSS
-.RNNXgb,
+.minidiv .sfbg,
+#searchform.minidiv,
+.minidiv .emcav .RNNXgb .S6VXfe,
+.sbfc .RNNXgb,
 .aajZCb {
-    box-shadow: 0 0 2px 0 ${rgba(0,0,0,0.16)}, 0 0 0 1px ${rgba(0,0,0,0.08)} !important;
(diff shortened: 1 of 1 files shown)


Vedanth-P:
I've also fixed some bad shadows on Google Classroom; and I will fix any more bad shadows I find in the next few days on Google Classroom. Should I submit those changes as a separate commit within this pull request, or start a new pull request for those?

src/config/dynamic-theme-fixes.config
```diff
@@ -13294,9 +13294,15 @@ img.act-icon-dark-gray
...
 
 CSS
-.RNNXgb,
+.minidiv .sfbg,
+#searchform.minidiv,
+.minidiv .emcav .RNNXgb .S6VXfe,
+.CvDJxb .RNNXgb,
 .aajZCb {
-    box-shadow: 0 0 2px 0 ${rgba(0,0,0,0.16)}, 0 0 0 1px ${rgba(0,0,0,0.08)} !important;
+    box-shadow:0 1px 6px 0 rgba(0,0,0,.28) !important;
```
Gusted, line 13302: I don't think this is a good idea. The lighter box-shadow makes it more notice-able for these elements. Thereby making shadows darker than the original element makes it the shadow itself fade in the background and not distinguishable. Also please preserve the `${}` behavior.

src/config/dynamic-theme-fixes.config
```diff
@@ -13294,9 +13294,15 @@ img.act-icon-dark-gray
...
+}
+.RNNXgb:hover {
+    box-shadow:0 1px 6px 2px rgba(0,0,0,.28) !important;
```
Gusted, line 13305: Same thing, don't think making it darker is a good idea. It's with my settings(slightly higher contrast) already un-noticeable.

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
Vedanth-P: Well, how about stronger shadows? 2px blur radius and 3px spread along with using the `${}` (sorry I didn't do that the first time) makes it much more visible even on darker backgrounds:
`box-shadow:0 1px 2px 3px ${rgba(255,255,255,.28)} !important;`
Perhaps we could even try increasing opacity if this doesn't suffice?

And I think having a light colored shadow somewhat deteriorates the experience, as it no longer has the appeal of a normal shadow.

src/config/dynamic-theme-fixes.config
Vedanth-P: Here, perhaps stronger shadows could be with 2px blur and 5px spread
`box-shadow:0 1px 2px 5px ${rgba(255,255,255,.28)} !important;`

Gusted: image
It's the same un-noticeable dark for me.

Vedanth-P: I see... So I'll close this pull request, then?

Gusted: Pretty much, I don't really see any fixes from this PR.

PR #6101, opened 2021-06-22
Fix for techdirt.com

neonr4in:
CSS fix for techdirt.com

https://www.techdirt.com/

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -11093,6 +11093,33 @@ h3 {
 
 ================================
 
+techdirt.com
+
+CSS
+body,
+div#contents,
+div#subnav,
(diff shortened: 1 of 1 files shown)


Gusted (commented):
What's this supposed to fix?

Without any fixes this site looks good.

neonr4in:
Does it? It looks like this for me (already tried on multiple profiles with default configuration, all with same results)

techdirt_no_fix

As for the fix, it should look like this

techdirt_with_fix

Gusted:
This is without any of the fixes:
image

If there's something wrong on your side it should be turned into a bug report, on both firefox and chromium they look like they should without any additional fixes.

neonr4in:
All right, something went awry on my end and now it is fixed

Sorry for wasting your time

PR #6102, opened 2021-06-22
Don't force layout change & Prevent Unstyled content

### LINKED ISSUE #835: [Chrome] Dynamic only works after toggling twice on https://docs.python.org

Gusted:
- Check the comments of the code.
- Pretty much we don't want to cause at any time a forced start of the layout, it can and it will causes white flashes.
- This happens because we allow to go do our thing when the DOM is `interactive`, however as noted in the official specifications " but still loading subresources," it will still load stylesheets, so when we force a layout change with non-loaded stylesheets, certain elements will be not styled(and not by the fallback style) and will give you a nice white flash on certain areas.
- Resolves some observed changes of render engines.
- Resolves #835


Gusted:
@alexanderby These are some interesting changes, especially the fix for #835. 

Unless we want to wrap the function that handles the building of rules(`setRule`, `modFilters.forEach`) etc into another function that we can call when we the other styles are loaded. I don't see another efficient way to do this. 

We could make the iterateCSSRules function an asynchronous function and resolve the promise, once all rules has been loaded and iterated over by the `iterate` function. That way we don't need to rebuild the sheet. But could result in longer time that the fallbackStyle is displayed - but technically shouldn't be a problem, because otherwise the site wouldn't be loaded as well.

Gusted:
> We could make the iterateCSSRules function an asynchronous function and resolve the promise, once all rules has been loaded and iterated over by the `iterate` function. That way we don't need to rebuild the sheet. But could result in longer time that the fallbackStyle is displayed - but technically shouldn't be a problem, because otherwise the site wouldn't be loaded as well.

And that's now the case, because rebuilding is much much more expensive than just give that modifiySheet a little wait on the iterateCSSRules.

src/inject/utils/dom.ts
```diff
@@ -199,12 +199,32 @@ export function removeDOMReadyListener(listener: () => void) {
     readyStateListeners.delete(listener);
 }
 
+// `interactive` can and will be fired when their are still stylesheets loading.
+// We use certain actions that can cause a forced layout change, which is bad.
+export function isCompleteDomReady() {
```
alexanderby, line 204: Something is wrong with the name. Is ready state "Complete" (true)? Or is DOM ready (true, but DOM is ready when state is "interactive" too)?

src/inject/utils/dom.ts
```diff
@@ -199,12 +199,32 @@ export function removeDOMReadyListener(listener: () => void) {
...
+    return document.readyState === 'complete';
+}
+
+export const completeDomListeners = new Set<() => void>();
```
alexanderby, line 208: Should it be exported? Use `DOM` in the middle of the name and `dom` in the beginning. Also the word order seems wrong, "complete" at the beginning sounds like a verb.

src/inject/dynamic-theme/inline-style.ts
```diff
@@ -293,12 +293,24 @@ export function overrideInlineStyle(element: HTMLElement, theme: FilterConfig, i
...
+                // getBoundingClientRect forces a layout change. And when it so happens that.
+                // The DOM is not in the `complete` readystate. It will cause the layout to be drawn.
+                // And it will cause a layout of unstyled content which results in white flashes.
+                // Therefor the check if the DOM is at the `complete` readystate.
+                const handleSVGElement = () => {
+                    const {width, height} = element.getBoundingClientRect();
+                    const isBg = (width > SMALL_SVG_LIMIT || height > SMALL_SVG_LIMIT);
+                    setCustomProp('fill', isBg ? 'background-color' : 'color', value);
+                };
+                if (!isCompleteDomReady()) {
```
alexanderby, line 306: `if` is preferred to be without a `!`

src/inject/dynamic-theme/css-rules.ts
```diff
@@ -1,32 +1,53 @@
...
+                iterate(rule);
+            } else if (rule instanceof CSSImportRule) {
+                const iterateCSSImportRule = () => {
+                    try {
+                        iterateCSSRules(rule.styleSheet.cssRules, iterate);
+                    } catch (err) {
+                        logWarn(err);
+                    }
+                };
+                if (!isCompleteDomReady()) {
```
alexanderby, line 29: Usage without `!` is preferred.

src/inject/dynamic-theme/css-rules.ts
```diff
@@ -1,32 +1,53 @@
...
+import {addDOMCompleteListener, isCompleteDomReady} from '../utils/dom';
 import {logWarn} from '../utils/log';
 
-export function iterateCSSRules(rules: CSSRuleList, iterate: (rule: CSSStyleRule) => void) {
-    forEach(rules, (rule) => {
-        if (rule instanceof CSSMediaRule) {
-            const media = Array.from(rule.media);
-            const isScreenOrAll = media.some((m) => m.startsWith('screen') || m.startsWith('all'));
-            const isPrintOrSpeech = media.some((m) => m.startsWith('print') || m.startsWith('speech'));
+export async function iterateCSSRules(rules: CSSRuleList, iterate: (rule: CSSStyleRule) => void) {
```
alexanderby, line 7: As far as I understand, you should leave the function `iterateCSSRules` synchronous. I don't see `await iterateCSSRules` anywhere

alexanderby (changes requested):

src/inject/utils/dom.ts
Gusted: I really should take a variable naming course.

src/inject/dynamic-theme/css-rules.ts
Gusted: So, here is where the "fix" actually relies. We are adding a listener inside this function, to wait for when the `document.readyState === 'complete'`, but this only happens when we have a LinkStyle that we cannot access it's styles, this is because it's still loading.

Now you will think, yeah so? just have the callback execute the `iterate` function.

Correct, that should do the job, however. in `stylesheet-modifer.ts` we would have `iterateCSSRules(rules, (rule) => {` and after the last rule is being iterated(so not the ones that we are waiting for in the callback for `complete` readyState). And it would use the modRules to modify the `darkreader--sync` sheet. So now when it happens that the callback would be fired, it would add the modRules, but because due to the fact that it treated as a synchronous function it won't have any effect because the `buildOverride()` etc is already fired and executed. 

It's a complicated issue that can be paired with "race conditions", because you will execute a callback after the `modifySheet` function has already finished modifying the sheet. But when we will treat is as an asynchronous function, it will await the last `iterate` of the loaded CSSImportRule and thus will be included in the whole modifying process.

alexanderby: I tried observing and here is what I think:

Making this function asynchronous, and using awaits makes some parts of the code being executed after another in an unpredictable way. This can cause problems when reading variables or other cases. I wish the execution flow will be straightforward.

I found out that the problem actually happens (unstyled content on docs.python.org) when error happens in
```typescript
try {
                        iterateCSSRules(rule.styleSheet.cssRules, iterate);
                    } catch (err) {
                        logWarn(err);
                    }
```

So I think what you can do here is revert the functions to being synchronous, and when error happens with reading rules of CSSImportRule, rerun building the sheet override when loading ends (ideally loading of the link, not a document). One of the ways to achieve this is:
```typescript
        function buildOverrides() {
            const force = forceRenderStyle;
            forceRenderStyle = false;
            sheetModifier.modifySheet({
                ...
            });
            isOverrideEmpty = syncStyle.sheet.cssRules.length === 0;
            if (sheetModifier.hasNotLoadedImports()) {
                // "update" function schedules rebuilding the style
                // ideally to wait for link loading, because some sites put links any time,
                // but it can be complicated, so waiting for document completion can do the trick
                addReadyStateCompleteListener(() => update());
            }
        }
```

alexanderby (changes requested):

alexanderby:
By the way, to wait for link loading there's already a `linkLoading` function that can be used like `await linkLoading(element)` or `linkLoading(element).then(() => update())`.

Gusted:
> By the way, to wait for link loading there's already a `linkLoading` function that can be used like `await linkLoading(element)` or `linkLoading(element).then(() => update())`.

That's the problem. I thought so too and tried many fixes with this logic, however upon stumbling the Chromium codebase and had to compile it for other various reasons it became clear that they are cutting some edges here. 

The `<link>` is loaded according to chromium, but also according to them any sub-import(`@import`) can still be set for loading etc. So such approach doesn't help unfortunately, which is actually unintentional behavior in my eyes from Chromium.

Gusted:
> can still be set for loading etc.

And as far as I see, the `readyState === 'complete'` doesn't even ensure that's loaded but cover a good 99.99% of the cases for now.

src/inject/dynamic-theme/css-rules.ts
```diff
@@ -1,29 +1,30 @@
 import {forEach} from '../../utils/array';
 import {isSafari} from '../../utils/platform';
 import {parseURL, getAbsoluteURL} from '../../utils/url';
-import {logWarn} from '../utils/log';
+import {logInfo, logWarn} from '../utils/log';
 
-export function iterateCSSRules(rules: CSSRuleList, iterate: (rule: CSSStyleRule) => void) {
+export function iterateCSSRules(rules: CSSRuleList, iterate: (rule: CSSStyleRule) => void, hasNonLoadedLinks: () => void = () => void 0) {
```
alexanderby, line 6: `hasNonLoadedLinks` (`is`, `was`, `has`) means it will return boolean, but in your case it doesn't return anything. Try renaming it into what it actually does, e.g. `onMediaRuleError`.

Another minor problem I see is `() => void = () => void 0`. Maybe you could add `()` around each part? It is possible to replace `void 0` with `undefined` (it is immutable in modern browsers) or `Function.prototype`, or just strip off the assignment and check for the argument in the code.

alexanderby (changes requested):

alexanderby (approved):
By the way, maybe it's possible to do the "return" if there's an import error, and document is still not loaded.

Gusted:
> By the way, maybe it's possible to do the "return" if there's an import error, and document is still not loaded.

You mean breaking the loop? Not sure, because if such Import is half way in a CSSRuleList it will mean only the first portion will get modified by dark reader and later on the other half will be modified and if it so happens that such CSS like `.main-app { background-color: white }` exist it won't get catched by us until the `dom.readyState === 'complete'`.

Gusted:
After a good year not actually understanding how in earth the #835 could happen and tried to fix the issue on some false assumptions. It can be finally closed and thereby also have a improvements towards FOUS 😄.

PR #6112, opened 2021-06-23
Fix dynamic mode for wordnik.com

Gitoffthelawn:
For `img[alt="wordnik logo"]`, I think it would be better to apply a `filter` that includes `brightness(0.8)`, as the bright white is too intense for a dark theme (IMO).  But I understand the trepidation (and reasons) for not wanting filters with static values.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -13055,6 +13055,14 @@ INVERT
 
 ================================
 
+wordnik.com
+
+INVERT
+img[alt="wordnik logo"]
+.footer-logo
+
+================================
+
 worldcubeassociation.org
 
 CSS


Gusted:
> I think it would be better to apply a `filter` that includes `brightness(0.8)`, as the bright white is too intense for a dark theme (IMO)

I'm not sure but we could include a new dynamic variable that will save the user's filter settings. so you could use like
```css
.logo {
    filter: var(--darkreader-filter) brightness(0.8) !important;
}
```
But that could override the existing brightness levels etc.

Gusted (approved):

Gitoffthelawn:
> > I think it would be better to apply a `filter` that includes `brightness(0.8)`, as the bright white is too intense for a dark theme (IMO)
> 
> I'm not sure but we could include a new dynamic variable that will save the user's filter settings. so you could use like
> 
> ```css
> .logo {
>     filter: var(--darkreader-filter) brightness(0.8) !important;
> }
> ```
> 
> But that could override the existing brightness levels etc.

Hmmm... that's an interesting idea.

Thinking off the top of my head, I wonder if we could also design it so that it's somehow compatible with `calc()` so that you can do something like:

```
.logo {
    filter: var(--darkreader-filter) brightness(calc(0.8 * var(--darkreader-filter-brightness)) !important;
}
```

That's just a first thought for the idea, not actual design!

Gusted:
Something like that could work as a design choice.

PR #6113, opened 2021-06-23
Avoid persistent ports between UI and background

bershanskiy:
Use message passing instead of persistent ports for communication
between UI and background context. This simplifies logic a bit and
avoids unnecessary numerous callbacks and Port message event listeners.


src/background/messenger.ts
```diff
@@ -19,36 +19,45 @@ export interface ExtensionAdapter {
...
+    private changeListenerCount: number;
 
     constructor(adapter: ExtensionAdapter) {
-        this.reporters = new Set();
         this.adapter = adapter;
-        chrome.runtime.onConnect.addListener((port) => {
-            if (port.name === 'ui') {
-                port.onMessage.addListener(async (message) => await this.onUIMessage(port, message));
+        this.changeListenerCount = 0;
+        chrome.runtime.onMessage.addListener((message: Message, sender: any, sendResponse: (any) => void) => {
```
Gusted, line 28:
```suggestion
        chrome.runtime.onMessage.addListener((message: Message, _: any, sendResponse: (any) => void) => {
```

We don't use `sender` as value. < IDE's will go complain about unused values.

src/background/messenger.ts
```diff
@@ -19,36 +19,45 @@ export interface ExtensionAdapter {
...
+                    'apply-dev-dynamic-theme-fixes',
+                    'apply-dev-inversion-fixes',
+                    'apply-dev-static-themes'
+                ].includes(message.type));
             }
         });
     }
 
-    private async onUIMessage(port: chrome.runtime.Port, {type, id, data}: Message) {
+    private async onUIMessage({type, id, data}: Message, sendResponse: (any) => void) {
```
Gusted, line 43:
```suggestion
    private async onUIMessage({type, data}: Message, sendResponse: (any) => void) {
```
Pretty must what ESLint says. It's unused.

src/background/messenger.ts
```diff
@@ -19,36 +19,45 @@ export interface ExtensionAdapter {
...
         this.adapter = adapter;
-        chrome.runtime.onConnect.addListener((port) => {
-            if (port.name === 'ui') {
-                port.onMessage.addListener(async (message) => await this.onUIMessage(port, message));
+        this.changeListenerCount = 0;
+        chrome.runtime.onMessage.addListener((message: Message, sender: any, sendResponse: (any) => void) => {
+            if (message.name === 'ui') {
+                this.onUIMessage(message, sendResponse);
                 this.adapter.onPopupOpen();
+                return ([
```
Gusted, line 32: The type defintion of `ExtensionMessageEvent` don't expect any return value:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6f27c44c9a0040d4d86f3a17738fbf39fd616ed2/types/chrome/index.d.ts#L6574

Any reason why this is included?

src/ui/connect/connector.ts
```diff
@@ -35,73 +23,80 @@ export default class Connector implements ExtensionActions {
         return await this.sendRequest<TabInfo>({type: 'get-active-tab-info'}, ({data}, resolve) => resolve(data));
     }
 
+    private onChangesReceived = ({name, type, data}: Message) => {
+        if (name === 'background' && type === 'changes')
+            this.changeSubscribers.forEach(callback => callback(data))
```
Gusted, line 28:
```suggestion
            this.changeSubscribers.forEach((callback) => callback(data));
```
Code-style ;)

src/ui/connect/connector.ts
```diff
@@ -35,73 +23,80 @@ export default class Connector implements ExtensionActions {
...
 
     resetDevStaticThemes() {
-        this.port.postMessage({type: 'reset-dev-static-themes'});
+        chrome.runtime.sendMessage({name: 'ui', type: 'reset-dev-static-themes'});
     }
 
     disconnect() {
-        this.port.disconnect();
+        if (this.changeSubscribers.size > 0) {
+            this.changeSubscribers = new Set();
```
Gusted, line 97:
```suggestion
            this.changeSubscribers.clear();
```
It's better to clear a Set than creating a new one.

Gusted (changes requested):
Pretty neat change 👏🏽. I supposed a persistent port doesn't work with non-persistent backgrounds? _I should read that sentence and pretty much figure out that the answer is a "no"_.

The overall structure should be good to use `chrome.runtime`, just a few minor things to be changed.

src/background/messenger.ts
bershanskiy: This is crucial for letting `onUIMessage` to use `sendResponse`. Docs say [this](https://developer.chrome.com/docs/extensions/reference/runtime/#event-onMessage):

> This function becomes invalid when the event listener returns, **unless you return true** from the event listener to indicate you wish to send a response asynchronously (this will keep the message channel open to the other end until `sendResponse` is called).

bershanskiy: Alternatively, we could move all logic directly into `onUIMessage` (since it's just a few LoC) and write just `chrome.runtime.onMessage.addListener(this.onUIMessage)`. Then `return` statement would be unnecessary.

Gusted: Heh, that's good to know. That's worth opening up a issue in the type definitions repo.

bershanskiy:
Thanks for speedy review, I have fixed all lints and nits. Does it look better now?

Gusted:
> Thanks for speedy review, I have fixed all lints and nits. Does it look better now?

Yes code-wise. However Firefox seems to error on something? It's really weird due to the error message being vague.
image

Chromium doesn't seems to have a problem.

src/background/messenger.ts
Gusted: That sounds like a better idea as it might come off confusing why to return something - Especially on this behavior with chromium

Gusted:
This is a bunner, only chrome has this `responseCallback` option implemented. Firefox has no such options, and thus will error out now. By any chance we could have a workaround for firefox?

Gusted:
Possible workaround for Firefox would be to go back and rely on getting a ID. And make the sendResponse function a function that fires a `runtime.sendMessage` call with the specified ID and have a listener for that and lookup in a Map the ID and execute the `executor` function with the appropriate arguments.

bershanskiy:
I was programming based on offcial docs on [MDN](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage), which basically say that it's OK to pass response callback to an async function as long as listener returns `true`. I created the following [demo extension](https://github.com/darkreader/darkreader/files/6704970/test.zip) which does exactly that and works fine in Firefox (as expected).

I think Firefox gets tripped up by some combination of private methods, layered classes or some other abstractions. I'll keep looking for a workaround, but I'd prefer to make this work-around active only for Firefox and let browsers without this bug run good code path.

bershanskiy:
Does build script support conditional compilation like? I'd like to create separate versions of `sendRequest` for Chrome and Firefox.

Gusted:
> I think Firefox gets tripped up by some combination of private methods, layered classes or some other abstractions. I'll keep looking for a workaround, but I'd prefer to make this work-around active only for Firefox and let browsers without this bug run good code path.

That could be it - None the less Firefox is not liking that specific code for some reason.



> Does build script support conditional compilation like? I'd like to create separate versions of `sendRequest` for Chrome and Firefox.

Import the `isFirefox` variable from `utils/platform.ts` that should do it.

bershanskiy:
I have added a work-around for that Firefox bug and now it works as expected. Also, since Firefox has a race between promisses already submitted to the loop and context shutdowns, I also made one optimization Chrome-specific (it allows background to avoid sending messages when they are not needed).

src/ui/connect/connector.ts
```diff
@@ -1,107 +1,127 @@
...
-        this.port = chrome.runtime.connect({name: 'ui'});
+        this.changeSubscribers = new Set();
     }
 
-    private getRequestId() {
-        return ++this.counter;
+    private async sendRequest<T>(request: Message) {
+        return new Promise<T>((resolve, reject) => {
+            chrome.runtime.sendMessage({name: 'ui', ...request}, ({data, error}) => {
+                if (error)
```
Gusted, line 14: Braces... Why isn't ESLint erroring on this?

src/ui/connect/connector.ts
```diff
@@ -1,107 +1,127 @@
...
+                    reject(error);
+                else
+                    resolve(data);
+            });
+        });
     }
 
-    private async sendRequest<T>(request: Message, executor: (response: Message, resolve: (data?: T) => void, reject: (error: Error) => void) => void) {
-        const id = this.getRequestId();
+    private async poorMansSendRequest<T>(name: string) {
```
Gusted, line 22: `poorMansSendRequest` 😄 what should the naming represents?

src/ui/connect/connector.ts
```diff
@@ -1,107 +1,127 @@
...
         return new Promise<T>((resolve, reject) => {
-            const listener = ({id: responseId, ...response}: Message) => {
-                if (responseId === id) {
-                    executor(response, resolve, reject);
-                    this.port.onMessage.removeListener(listener);
+            const dataPort = chrome.runtime.connect({name});
+            dataPort.onDisconnect.addListener(() => reject());
+            dataPort.onMessage.addListener(({data, error}) => {
+                if (error)
```
Gusted, line 27: Braces :D

Gusted (commented):
A lot of braces are missing 👀 you can run `npm run code-style` to fix them automatically. Workaround is working for Firefox.

src/ui/connect/connector.ts
bershanskiy: Neither `npm run lint` nor `npm run code-style` does anything. I fixed this and similar issues manually.

src/ui/connect/connector.ts
bershanskiy: I renamed it to `firefoxSendRequestWithResponse` to match actual use of the function.

src/ui/connect/connector.ts
Gusted: Good to know, more reason to clean up the eslint config.

src/background/messenger.ts
```diff
@@ -19,36 +20,58 @@ export interface ExtensionAdapter {
...
+                    'get-data',
+                    'get-active-tab-info'
+                ].includes(message.type));
             }
         });
+
+        // This is a work-around for Firefox bug which does not let to responding to onMessage handler above.
+        if (isFirefox) {
+            chrome.runtime.onConnect.addListener((port) => {
+                let promise = null;
```
Gusted, line 43: `promise` will be override anyway, not really a need to initialize it, but let's add a type definition.
```suggestion
                let promise: Promise<(data) => void>;
```

Gusted (changes requested):
Other than that little minor thing it seems good for me. @alexanderby could you please have a look as well?

src/background/messenger.ts
bershanskiy: I don't think `Promise<(data) => void>` is correct. Since this promise takes on value `Promise<ExtensionData>` or `Promise<TabInfo>`, I changed it to `Promise<ExtensionData | TabInfo>`. (It compiles and works as expected.)

Gusted: 👍🏽

alexanderby:
Hi @bershanskiy! Could you please explain what advantage do you see in using the chrome.runtime messages versus ports? I used ports for faster messages delivery, though I haven't measured. And also they allow delivering messages to exact tabs.

Why is there a Firefox specific code?

bershanskiy:
> Could you please explain what advantage do you see in using the chrome.runtime messages versus ports?

Persistent ports do not allow extension background to go inactive (be entirely discarded from memory). They require background to stay alive and consume RAM even when Dark Reader is set to "OFF" on all sites. This PR is one from series aimed at making background non-persistent for browsers which support this (Chromium and derivatives).

> I used ports for faster messages delivery, though I haven't measured.

I'll check this. Perhaps, it was just a performance issue of a specific browser or a temporary implementation? At high level, both persistent ports and one-time messages should do similar things so they should perform similarly.

> And also they allow delivering messages to exact tabs.

`chrome.tabs.sendMessage` allows delivers message to a specific tab or even a specific frame within that tab. [Here is MDN page for reference.](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/sendMessage)

> Why is there a Firefox specific code?

In short, Firefox has a bug which makes `chrome.runtime.onMessage` ([MDN reference](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage)) not respect `sendResponse` and within the regular code path. Promise-based equivalent of this code still does not work on Firefox (and on Chrome it is officially not supported). Please note that normally modern Firefox supports both `sendResponse` and Promise-style responses, but just has a bug which manifests in this particular case. I find reports of this Firefox bug since at least 2017 but no one bothered to fix it yet.

If needed, I could bring back the entirety of old code path for Firefox (since Firefox does not support non-persistent backgrounds yet) and make the new code path apply only to Chrome.

bershanskiy:
> I used ports for faster messages delivery, though I haven't measured.

I measured Chrome round-trip time for sending a small message from extension page to background and didn't notice any advantage of using either method over the other. Here are preliminarry results:
 - As expected, when non-persistent background context is cold (fully stopped, does not have a process in task manager), here is a large latency, over 10 ms, for both persistent ports and one-time messages. This is expected because main process has to create a new process for background, shove all the relevant info via IPC, and only then start executing code.
 - When backgound context is active and runs the code path for the first time, one round-trip takes 1-2ms.
 - When background context is active and runs the code path repeatedly, one round-trip takes less than 1ms with both methods.
 - The variance is very large. The variance can change after extension reloads.

I used `window.performance.now()` to get the time, which might have extra noise added to it to thwart timing attacks. I haven't ruled that out that yet.

Gusted:
> I used `window.performance.now()` to get the time, which might have extra noise added to it to thwart timing attacks. I haven't ruled that out that yet.

Firefox has such option https://developer.mozilla.org/en-US/docs/Web/API/Performance/now#reduced_time_precision

Chromium has very little noise added, on the floating pointer, so the amount of `ms` should be accurate.

src/ui/connect/connector.ts
```diff
@@ -1,107 +1,133 @@
...
+                chrome.runtime.sendMessage({name: 'ui', type: 'subscribe-to-changes'});
             }
-        });
-        this.port.postMessage({type: 'subscribe-to-changes', id});
+        }
     }
 
     enable() {
-        this.port.postMessage({type: 'enable'});
+        chrome.runtime.sendMessage({name: 'ui', type: 'enable'});
```
alexanderby, line 69: Let's move `chrome.runtime.sendMessage({name: 'ui', ...` to some method, e.g. `sendMessage(type: string, data?: any)`

src/ui/connect/connector.ts
```diff
@@ -1,107 +1,133 @@
...
-        this.port.postMessage({type: 'reset-dev-static-themes'});
+        chrome.runtime.sendMessage({name: 'ui', type: 'reset-dev-static-themes'});
     }
 
     disconnect() {
-        this.port.disconnect();
+        if (this.changeSubscribers.size > 0) {
+            this.changeSubscribers.clear();
+            chrome.runtime.onMessage.removeListener(this.onChangesReceived);
+            if (isChromium) {
```
alexanderby, line 128: Let's replace with `!isFirefox`

src/ui/connect/connector.ts
```diff
@@ -1,107 +1,133 @@
...
+
     subscribeToChanges(callback: (data: ExtensionData) => void) {
-        const id = this.getRequestId();
-        this.port.onMessage.addListener(({id: responseId, data}: Message) => {
-            if (responseId === id) {
-                callback(data);
+        this.changeSubscribers.add(callback);
+        if (this.changeSubscribers.size === 1) {
+            chrome.runtime.onMessage.addListener(this.onChangesReceived);
+            if (isChromium) {
```
alexanderby, line 62: Let's replace all `isChromium` with `!isFirefox`. By the way, how it is supposed to work for Firefox?

alexanderby (changes requested):

src/background/messenger.ts
```diff
@@ -106,6 +129,8 @@ export default class Messenger {
     }
 
     reportChanges(data: ExtensionData) {
-        this.reporters.forEach((report) => report(data));
+        if (this.changeListenerCount > 0 || !isChromium) {
```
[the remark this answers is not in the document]
bershanskiy: On Chromium (determined as `!isFirefox`) connector sends `subscribe-to-changes` and `unsubscribe-from-chages` which inform messenger to increment/decrement `this.changeListenerCount`. So when there is at least one active subscriber, `changeListenerCount > 0` is `true`. On Firefox (determined by `isFirefox`) we don't use `this.changeListenerCount` and just broadcast change notifications unconditionally.

This optimization is needed on Chromium to rule out unnecessary change notification messages which could cause unnecessary onMessage event handler work. On FIrefox, I'm hesitant to use this optimization because Firefox is very racy.

src/ui/connect/connector.ts
bershanskiy: Firefox does not use `changeListenerCount` and just broadcasts update notifications unconditionally.

src/ui/connect/connector.ts
```diff
@@ -1,107 +1,133 @@
...
-            this.port.postMessage({...request, id});
+                dataPort.disconnect();
+            });
         });
     }
 
     async getData() {
-        return await this.sendRequest<ExtensionData>({type: 'get-data'}, ({data}, resolve) => resolve(data));
+        if (isFirefox) {
+            return this.firefoxSendRequestWithResponse<ExtensionData>('ui-get-data');
```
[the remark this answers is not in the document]
bershanskiy: I added `await`, but may be it's better to just remove it on the other return statement? That way there is no extra promise unwrapping and re-wrapping.

src/ui/connect/connector.ts
```diff
@@ -1,107 +1,137 @@
...
+        return await this.sendRequest<TabInfo>({type: 'get-active-tab-info'});
+    }
+
+    private onChangesReceived = ({from, type, data}: Message) => {
+        if (from === 'background' && type === 'changes') {
+            this.changeSubscribers.forEach((callback) => callback(data));
+        }
+    };
+
+    private sendMessage(data?: any) {
```
alexanderby, line 58: I think `?` should be skipped here, it sends something anyway.

alexanderby (changes requested):
Some minor issues. There's `await connector.getData()` in UI code, so `await` is still needed.

src/background/messenger.ts
```diff
@@ -19,36 +20,59 @@ export interface ExtensionAdapter {
...
+    private changeListenerCount: number;
 
     constructor(adapter: ExtensionAdapter) {
-        this.reporters = new Set();
         this.adapter = adapter;
-        chrome.runtime.onConnect.addListener((port) => {
-            if (port.name === 'ui') {
-                port.onMessage.addListener(async (message) => await this.onUIMessage(port, message));
+        this.changeListenerCount = 0;
+        chrome.runtime.onMessage.addListener((message: Message, _: any, sendResponse: (any) => void) => {
```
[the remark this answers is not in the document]
bershanskiy: I used `(any) => void` because it's basically `(any: any) => void`, but I agree that `(response: any) => void` might be easier to read.

src/ui/connect/connector.ts
bershanskiy: Done!

alexanderby (approved):
I'm a bit scared of these changes, but let's test them well :)

Gusted:
> I'm a bit scared of these changes, but let's test them well :)

Consider that we've just had a release. We got a good ~1 month to test it.

bershanskiy:
I found a regression, investigating.

bershanskiy:
I found the cause, will file a PR soon. I'm not sure, but a portion of patch might have been lost during a merge conflict resolution. Good thing we have some time to test and fix all regressions.

PR #6138, opened 2021-06-25
Avoid persistent ports in content scripts

bershanskiy:
Use message passing for communication between content scripts and
background.

This PR replaces persistent ports with non-persistent messages without changing the overall logic.


src/inject/index.ts
```diff
@@ -42,16 +42,18 @@ function onMessage({type, data}) {
...
-    cleanDynamicThemeCache();
-    colorSchemeWatcher.disconnect();
+chrome.runtime.onMessage.addListener(onMessage);
+chrome.runtime.sendMessage({type: 'tab'});
+
+window.addEventListener('unload', () => {
+    chrome.runtime.sendMessage({type: 'tab-unload'});
 });
 
+// TODO: figure out if this needs to call
```
bershanskiy, line 57: The original code included:
```
port.onDisconnect.addListener(() => {
    logWarn('disconnect');
    cleanDynamicThemeCache();
    colorSchemeWatcher.disconnect();
});
```
As far as I see, this code would not run during normal extension operation, since extension background page never called `disconnect()`. This code probably could run when extension is disabled or uninstalled because then background page is torn down (and its ports are closed), but I'm not sure how useful that is.

src/inject/index.ts
```diff
@@ -42,16 +42,18 @@ function onMessage({type, data}) {
...
-const port = chrome.runtime.connect({name: 'tab'});
-port.onMessage.addListener(onMessage);
-port.onDisconnect.addListener(() => {
-    logWarn('disconnect');
-    cleanDynamicThemeCache();
-    colorSchemeWatcher.disconnect();
+chrome.runtime.onMessage.addListener(onMessage);
+chrome.runtime.sendMessage({type: 'tab'});
+
+window.addEventListener('unload', () => {
```
Gusted, line 53:
```suggestion
addEventListener('unload', () => {
```

Gusted (commented):
Going the right way.

src/inject/index.ts
bershanskiy: Makes sense. I'll look into whether it's possible to notify extension about incoming reload to clean cache before reload or clean cache from the last version right upon reload.

src/background/tab-manager.ts
```diff
@@ -20,59 +20,61 @@ interface TabManagerOptions {
...
+                case 'tab-unload':
+                    if (!sender.tab) {
+                        console.error('Unexpected message', message, sender);
+                        break;
+                    }
+
+                    const framesForDeletion = this.tabs.get(sender.tab.id);
+                    if (framesForDeletion) {
+                        framesForDeletion.delete(sender.frameId);
+                    }
```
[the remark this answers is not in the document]
bershanskiy: Shouldn't linter warn about this kind of style preferences?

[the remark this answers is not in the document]
Gusted: It should - but I haven' touched the linter configuration in a long time and have to update some preferences to detect instances like this and about the braces.

src/background/index.ts
```diff
@@ -43,6 +45,7 @@ if (WATCH) {
                     break;
                 }
                 case 'reload:full': {
+                    tabs.forEach((tabId) => chrome.tabs.sendMessage(tabId, {type: 'reload'}));
```
Gusted, line 48: It feels like we can reuse the `queryTabs` function from tab-manager.ts 

https://github.com/darkreader/darkreader/blob/02d6f7dd684e0cd6805c37a11c1908a42874abe3/src/background/tab-manager.ts#L6-L10

So we can avoid having them in a set. Especially because the `unload` event can be interpreted by the browser in many ways and have the chance of not being called at all, because it could be killed by an OOM of the operating system and many more reasons.

src/inject/index.ts
bershanskiy: I have added this feature back in: https://github.com/darkreader/darkreader/pull/6138/commits/3063c315d25e6a8eadf1d0ceefc81d6df08827cc

Gusted (changes requested):

src/background/index.ts
bershanskiy: I changed it to use `chrome.tabs.query` directly because there isn't much bennefit in wrapping it into a Promise just to unwrap that with a `then`. I initially stored tab ids explicitly because `chrome.tabs.query` just returns all tabs whuch exist without regard to whether there are any content scripts, which could be less efficient.

Does this look better now?

Gusted: Synchronous vs asynchronous can have a lot of benefit and ensure that certain things has been done if you rely on certain actions. But in this case I don't see any relevance so that's fine.

> just returns all tabs whuch exist without regard to whether there are any content scripts, which could be less efficient.

It's a debug-copy anyway not something to be used in production so I won't even care if it has 100 tabs whereby 1 tab is actually something with dark reader injected. 

It looks better now.

src/inject/index.ts
```diff
@@ -47,11 +53,11 @@ const colorSchemeWatcher = watchForColorSchemeChange(({isDark}) => {
...
-    colorSchemeWatcher.disconnect();
-});
+chrome.runtime.onMessage.addListener(onMessage);
+chrome.runtime.sendMessage({type: 'tab'});
+
+function onUnload() {
+    chrome.runtime.sendMessage({type: 'tab-unload'});
+}
 
+addEventListener('unload', onUnload);
```
Gusted, line 63: As I just said I don't like to use of `unload` event in anyway as it's unpredictable and kind of a black magic use case within browsers(especially cross-versions). And I cannot accept any code with this kind of event because browsers will not use [back forward cache](https://developers.google.com/web/updates/2018/07/page-lifecycle-api#back-forward-cache) which I know will create a tsunami of issues opening about this behavior from dark reader. Also see: https://developers.google.com/web/updates/2018/07/page-lifecycle-api#the-beforeunload-event.

Gusted (changes requested):
As noted in the other comment, I cannot accept any PR with such event code. So I propose the following solution.

We only use the `unload` on Iframe's(It seems like they don't affect the disabling of forward/backward cache) as I cannot seem to find any web API related to get reliably all IFrames from a tabId. And for non-iFrames we can add a listener to `chrome.tabs.onRemoved` and deal with them trough that.

Gusted:
Hmm, I actually found https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webNavigation/getAllFrames in such way we can use this to get all FrameID's when needed(it's only needed at the `sendMessage` function), it will give some overhead, but I don't really mind that. We also then can change the `this.tabs` to a `Set<number>()` type as we don't use it as a key/value anymore. And remove the `unload` code altogether.

bershanskiy:
I'll switch to `chrome.tabs.onRemoved`. As far as I see, it will be sufficient because it will belete all top-level windows when their tabs are closed. It should be fine to try and fail to deliver a message to a sub-frame which was closed.

I would like to avoid `webNavigation.getAllFrames()` because it requires a very powerful permission `webNavigation` which goes against least-priviledge principle and raises extra warnings during install. Also, when existing user installs update with a new permission, browser automatically disables extension, which will annoy existing users.

bershanskiy:
@Gusted I switched away from `unload` event to `pagehide`, `freeze`, and `resume`. These are actually more informative and do not interfere with the BFcache. This extra info will be helpful later to tick only non-frozen frames to send messages to and avoid awaking frozen tabs. Does this look better?

Gusted:
> This extra info will be helpful later to tick only non-frozen frames to send messages to and avoid awaking frozen tabs. Does this look better?

That certainly looks better 😃

bershanskiy:
Is there anything else I can do to advance this PR? Also, I did not change `this.tabs` to `Set<number>` type because `getMessage()` still uses the frame URL.

Gusted:
@alexanderby seems to have great interest in this kind of PR's 😄

src/background/tab-manager.ts
```diff
@@ -20,59 +21,76 @@ interface TabManagerOptions {
...
                 }
+                frames.set(frameId, {url: senderURL, state: 'normal'});
+            }
 
-                const tabId = port.sender.tab.id;
-                const {frameId} = port.sender;
-                const senderURL = port.sender.url;
-                const tabURL = this.getTabURL(port.sender.tab);
+            switch (message.type) {
+                case 'frame': {
```
alexanderby, line 47: Hmmm, maybe `frame-connect`? *as far as I understand from usage*

src/background/tab-manager.ts
```diff
@@ -155,15 +170,19 @@ export default class TabManager {
...
-            .filter((tab) => this.ports.has(tab.id))
+            .filter((tab) => this.tabs.has(tab.id))
             .forEach((tab) => {
-                const framesPorts = this.ports.get(tab.id);
-                framesPorts.forEach(({url, port}, frameId) => {
+                const frames = this.tabs.get(tab.id);
+                frames.forEach(({url, state}, frameId) => {
+                    if (state === 'frozen') {
+                        // TODO: avoid sending messages to frozen tabs for performance reasons.
+                        logInfo('Sending message to a frozen tab.');
```
alexanderby, line 179: Hmmm, what if the frame resumed, will it stay white? Should something be sent to it?

src/inject/index.ts
```diff
@@ -47,11 +55,23 @@ const colorSchemeWatcher = watchForColorSchemeChange(({isDark}) => {
...
-port.onMessage.addListener(onMessage);
-port.onDisconnect.addListener(() => {
-    logWarn('disconnect');
-    cleanDynamicThemeCache();
-    colorSchemeWatcher.disconnect();
-});
+chrome.runtime.onMessage.addListener(onMessage);
+chrome.runtime.sendMessage({type: 'frame'});
+
+function onPagehide(e) {
```
alexanderby, line 61: onPage**H**ide

alexanderby (approved):
Some minor fixes. Let's check that no frames stay white, after reactivating.

src/background/tab-manager.ts
bershanskiy: With this patch, this code will just log a message about tab/frame being frozen and then will continue and send the message anyway. This behavior matches exactly what extension does before this patch. 

Also, this is most likely suboptimal (because it wakes up the frozen frame), so I left a comment and plan to file a follow-up PR.

After this patch, there are still two issues which will require follow-up patches:
 - Imagine a frame was frozen (which background knows about because it receives a message) and then unloaded (which happens without any callbacks for performance reasons) and background tries to send a message to it. Can we determine if frame dies like that or just ignore existence of such "ghost" frames?
 - Avoid sending excessive messages to frozen tabs all at once. E.g., imagine I have lots of frozen tabs folded to tray or behind other windows and then I toggle global on/off switch. Background probably should keep a record of messages addressed to frozen frames and then actually deliver them once frame wakes up.

alexanderby: Maybe there should be some error, if message is sent to a not existent frame.

Great, looking forward to other patches.

alexanderby (approved):

PR #6143, opened 2021-06-27
Fix for wiki.teamfortress.com

piousdeer:
Test page: https://wiki.teamfortress.com/wiki/Thunder_Mountain_(Capture_the_Flag)
Before:
image
After:
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -12888,7 +12888,8 @@ INVERT
 .mw-wiki-logo
 
 CSS
-html, body {
+html,
+body {
     background-color: var(--darkreader-neutral-background) !important;
 }
 body,
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -12898,7 +12898,8 @@ body,
 #footer {
     background-image: none !important;
 }
-div.vectorTabs ul li.selected *, #searchInput::placeholder {
+div.vectorTabs ul li.selected a,
+#searchInput::placeholder {
     color: var(--darkreader-neutral-text) !important;
 }
 
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -12891,7 +12891,11 @@ CSS
 html, body {
     background-color: var(--darkreader-neutral-background) !important;
 }
-body, #mw-page-base, #p-logo, #content, #footer {
+body,
+#mw-page-base,
+#p-logo,
+#content,
+#footer {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -12882,6 +12882,24 @@ body {
...
 
+wiki.teamfortress.com
+
+INVERT
+.mw-wiki-logo
+
+CSS
+html, body {
+    background-color: var(--darkreader-neutral-background) !important;
+}
```
Gusted, line 12893: I don't see any need for this.

src/config/dynamic-theme-fixes.config
```diff
@@ -12882,6 +12882,24 @@ body {
...
+body, #mw-page-base, #p-logo, #content, #footer {
```
Gusted, line 12894:
```suggestion
body,
#mw-page-base,
#p-logo,
#content,
#footer {
```

src/config/dynamic-theme-fixes.config
```diff
@@ -12882,6 +12882,24 @@ body {
...
+    background-image: none !important;
+}
+div.vectorTabs ul li.selected *, #searchInput::placeholder {
```
Gusted, line 12897: Not sure, but the current modified one by dark reader is in par with the theme of the site and should behave like these beige-ish colors.

piousdeer: These elements are unreadable unless you change the color.

src/config/dynamic-theme-fixes.config
piousdeer: This is to prevent this from happening.
image

Gusted: That's expected to happen on site that don't set a background-color on their `HTML` element.

But now we will be overriding the body element as well which has a correct background-color.

src/config/dynamic-theme-fixes.config
Gusted:
```suggestion
div.vectorTabs ul li.selected a,
#searchInput::placeholder {
```

src/config/dynamic-theme-fixes.config
piousdeer: What do you suggest? Remove this rule and leave this as is?

Gusted: Let the rule as-is.
```suggestion
html,
body {
    background-color: var(--darkreader-neutral-background) !important;
}
```

Gusted (approved):

PR #6148, opened 2021-06-28
radareu.cz

FadeMind:
http://www.radareu.cz/

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -9494,6 +9494,13 @@ INVERT
 
 ================================
 
+radareu.cz
+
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -9494,6 +9494,13 @@ INVERT
 
 ================================
 
+radareu.cz
+
+INVERT
+#radarMap
```
Gusted, line 9500:
```suggestion
#radarMap .leaflet-map-pane
```

These div only contains the panels etc and shouldn't revert all the other UI elements.

Gusted (approved):

PR #6154, opened 2021-06-28
Ensure a non-empty tab's URL value

### LINKED ISSUE #6151: [Feature Request] Proper treatment of `about:blank` pages
# Feature Request

## Is your feature request related to a problem? Please describe

When a window is created in JS with no URL it gets a `window.location` of `about:blank`:

    win = window.open("", "_blank");
    win.location.href // => "about:blank"

These windows are currently mistreated by darkreader in that there is no way that I see to turn it on/off for such a specific page.  Even more than that, the darkreader dropdown on these pages look like something went bad somewhere: there is the top on/auto/off at the top, but clicking the buttons does nothing, and there is nothing else below them (none of the regular UI).

It makes semi-sense to just ignore such pages, since they seem to come from nowhere and it's still good to avoid the jarring sea of bright whiteness if you happen to run into one of them, *however* there are valid uses of such windows, but there's no way currently to disable darkreader for them other than turning it off globally.  (See below for further context.)

## Describe the solution you'd like

For such `about:blank` windows, it makes sense to use their `window.origin` to decide how to treat them.  This way a blank window that was opened from `https://foo.bar/` would be treated as *that*, so if I disable darkreader for `foo.bar`, it would also be disabled on `about:blank` windows it spawns.

## Additional context

What I ran into is an issue with my home-cooked music player, which can use such windows for visualizations.  (Why?  Because having multiple monitors all showing changing colors in sync is a very nice thing...)  My code does something like:

    win = window.open("", "_blank");
    win.document.body.style.background = "#fcc";

but that doesn't work since darkreader gets in its way, and there is no way to turn it off locally.  But if you look at `win.origin` you'll see where it originated.

Gusted:
- Ensure that we have a "default value" for tab.url so legit pages will still handle correctly.
- Because of this behavior only happens on `about:blank`-like pages we simply use that as value.
- Resolves #6151

### DIFF
--- a/src/background/tab-manager.ts
+++ b/src/background/tab-manager.ts
@@ -52,7 +52,7 @@ export default class TabManager {
                 const tabId = port.sender.tab.id;
                 const {frameId} = port.sender;
                 const senderURL = port.sender.url;
-                const tabURL = this.getTabURLOrAboutBlank(port.sender.tab.url);
+                const tabURL = this.getTabURL(port.sender.tab);
 
                 let framesPorts: Map<number, PortInfo>;
                 if (this.ports.has(tabId)) {
@@ -121,11 +121,11 @@ export default class TabManager {
         });
     }
 
-    getTabURLOrAboutBlank(tabURL: string): string {
+    getTabURL(tab: chrome.tabs.Tab): string {
         // It can happen in cases whereby the tab.url is empty.
         // Luckily this only and will only happen on `about:blank`-like pages.
         // Due to this we can safely use `about:blank` as fallback value.
-        return tabURL || 'about:blank';
+        return tab.url || 'about:blank';
     }
 
     async updateContentScript(options: {runOnProtectedPages: boolean}) {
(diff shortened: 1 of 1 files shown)


src/background/tab-manager.ts
```diff
@@ -121,6 +121,13 @@ export default class TabManager {
         });
     }
 
+    // It can happen in cases whereby the tab.url is empty.
+    // Luckily this only and will only happen on `about:blank`-like pages.
+    // Due to this we can safely use `about:blank` as fallback value.
+    getTabURLOrAboutBlank(tabURL: string): string {
```
alexanderby, line 127:
```typescript
getTabURL(tab: chrome.tabs.Tab) {
    // It can happen...
    ...
```

alexanderby (changes requested):

alexanderby (changes requested):
I meant if you rename the function to `this.getTabURL` and make the tab itself as an argument, the code would be shorter and easier to understand. The function name will correspond to what you are actually doing.

alexanderby (approved):

PR #6163, opened 2021-06-30
Fix dawn.com

mobi2002:
Invert logo on dawn.com


src/config/dynamic-theme-fixes.config
```diff
@@ -2919,6 +2919,13 @@ CSS
 
 ================================
 
+dawn.com
+
+INVERT
+div.mb-3 > a > img
```
Gusted, line 2925:
```suggestion
img[alt="Dawn Logo"]
```

Using this selector makes it a little bit clearer what the CSS means. Thereby the `div.mb-3` seems to be the class for the header on every page and thus could result in other images being inverted.

mobi2002: Thanks for correcting me, I am still a beginner to css. Maybe I should have created an issue instead...

Gusted (approved):

PR #6168, opened 2021-07-01
Add CSS patch styles to sheet.best

felipecock:
Added compatibility styling patch to fix the default dark theming in the website [sheet.best](https://sheet.best).

#  Admin screen (sample)
## Original
sheet best_admin1(Laptop HD) original

## Default Dark Reader theming
sheet best_admin1(Laptop HD) default-styled

## New (fixed) Dark Reader theming
sheet best_admin1(Laptop HD) corrected

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -10258,6 +10258,30 @@ CSS
 
 ================================
 
+sheet.best
+
+CSS
+.MuiButton-containedPrimary,
+.MuiButton-containedPrimary * {
+    background-color: #207245 !important;
+}
+.MuiPaper-root {
+    background-color: #fff1 !important;
+}
+.MuiTouchRipple-root,
+.MuiIconButton-label > *,
+.MuiPaper-root * {
+    background-color: #FFF0 !important;
+}
+.MuiLinearProgress-colorPrimary {
+    background-color: #113c24 !important;
+}
+.MuiLinearProgress-barColorPrimary {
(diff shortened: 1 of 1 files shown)


Gusted:
Hmm looking at the screenshots it rather seems like a bug with dark reader itself that shouldn't be covered up by CSS Patches. Is their by any chance on all elements a rule being applied(background-color and color property) that forces that kind of "layout"? As that seems like some rule is being rendered onto those elements that shouldn't. It actually looks like our darkreader--fallback style.

felipecock:
> Hmm looking at the screenshots it rather seems like a bug with dark reader itself that shouldn't be covered up by CSS Patches. Is their by any chance on all elements a rule being applied(background-color and color property) that forces that kind of "layout"? As that seems like some rule is being rendered onto those elements that shouldn't. It actually looks like our darkreader--fallback style.

Thank you, Gusted, for your review and your support. I've not applied any other rule at all but the new patch. I've tested resetting the DYNAMIC THEME EDITOR, but it is still bad styled.

Furthermore, I don't know how to fix it at more global scope without braking some default well-styled sites. I will still be using this workaround patch while anybody fixes it because I work frequently with this site.

Gusted:
It's most likely some variant of #6135 #6075 which is fixed by #6078 so nothing to worry.

felipecock:
> It's most likely some variant of #6135 #6075 which is fixed by #6078 so nothing to worry.

Ok, thank you very much.

PR #6171, opened 2021-07-02
Added KyLeggiero.me, Blog.KyLeggiero.me, and Resume.BenLeggiero.me

KyLeggiero:


src/config/dark-sites.config
```diff
@@ -96,6 +96,7 @@ blog.aractus.com
 blog.counter-strike.net
 blog.dota2.com
 blog.ja-ke.tech
+blog.kyleggiero.me
```
Gusted, line 99: You don't have to add any subdomain. the domain itself is sufficient.

Gusted (changes requested):
That also should resolve the issue with why the tests are failing.

src/config/dark-sites.config
KyLeggiero: Oh interesting! I was just following the lead of others in the file. Will push that, thank you

KyLeggiero: Oh interesting! I was just following the lead of others in the file. Will push that, thank you

Gusted (approved):

src/config/dark-sites.config
KyLeggiero: @Gusted it seems that blog.kyleggiero.me is now being treated as if it's not on this list. What happened?

Issue #6161, opened 2021-06-30
[Performance] Dark Reader adds more than 10 seconds to Gmail load time


MahdiNazemi:
# Performance Issue

## Website
[mail.google.com](mail.google.com)

## How to Reproduce
Open [mail.google.com](mail.google.com)

## Performance Profiler
-
Performance with Dark Reader disabled: -

## System Information

- OS: macOS
- Browser: Firefox 89
- Dark Reader Version: 4.9.31

## Additional Information

- Gmail loads much faster when Dark Reader is disabled.
- The same issue is there in other browsers.
- Hardware acceleration is enabled on all browsers.

I understand that other users have reported this issue in the past. However, because the information shared by other users' profilers has provided insight into speeding up Dark Reader, I thought this new profiling information might be helpful too.

Gusted:
Hi @MahdiNazemi!

Thanks for reporting this issue, the profiler looks pretty interesting and will have a more in-depth look this weekend, especially because some things are out-of-control and might indicate performance issues with Firefox itself. At least that looks to me that way given just a 10 minute look.

Regards,
Gusted

MahdiNazemi:
Thanks for taking the time to look at the profiler's outputs.
I've had the same issue in Brave as well. I'll try to run Brave's profiler and share the results here.

Gusted:
Hi @MahdiNazemi!

I had another look today and some functions that are frequently used in dark reader to be somewhat slow(if you add all the instances of it) and are now being optimized by #6175 and #6174 I found some other issues but that's mainly due to the browser being slow(a lot of built-in functions and properties are somewhat expensive) so that will take some more time to verify and get fixed. But #6175 and #6174 will do it for on gmail. 

Regards,
Gusted

MahdiNazemi:
Thank you for taking the time to improve the performance of Dark Reader!

rakleed:
@Gusted does #6174 and #6175 fix the problem with opening Gmail, using it for a while, then switching to another tab for a few minutes, and after returning to the tab with Gmail, you need to wait 5-15 seconds for the interface to start responding user?

Gusted:
@rakleed These 2 PR should improve the startup performance on Gmail, #6259 should be the best improvement for Gmail and uses an experimental "manager" which I hope to use more in the future which can drastically improve startup performance for websites(by slowly "delaying" certain code).

rakleed:
@Gusted I built a version from the source, the problem I described practically doesn't appear now.

[...]

PR #6174, opened 2021-07-03
Faster iterateCSSRules function

Gusted:
- Use feature detection over the `instanceof` keyword to improve speed.
- See benchmark at: https://jsben.ch/B0eLa
- The order of the property detection is important, because certain types have multiple "detection points", e.g. the CSSImportRule has `media` property, so we ensure we first check for that with the `href` property. 
- Addresses 1 of the performance concerns in #6161

Performance recorded on https://mail.google.com/

139ms(before)
image

63ms(this patch)
image

Which is a 55% improvement. 

Please note that this function is used very frequently without any hesitant so any improvements is overall better for multiple functions that rely on this.


alexanderby (approved):
Nice discovery!

Could you please check, if you can put `if ((rule as CSSStyleRule).selectorText)` at the very top? It is the most popular one (I think 99%), so it would be great to avoid other checks.

Gusted:
> Could you please check, if you can put `if ((rule as CSSStyleRule).selectorText)` at the very top? It is the most popular one (I think 99%), so it would be great to avoid other checks.

I'm not sure what I was thinking, but that's indeed possible and doesn't have any conflicts for other CSSRules types.

PR #6175, opened 2021-07-03
General improvements

Gusted:
- Nothing particular to be note worthy but just found these 2 to be worth fixing.
- Don't create a variable if we don't end up using it multiple times.
- Cache cssText for safari(cssText property is pretty expensive).
- Avoid calculating `!important` when not necessary, a lot of properties coming into `getModifiableDecleration` are worthless to dark reader.

### DIFF
--- a/src/inject/dynamic-theme/css-rules.ts
+++ b/src/inject/dynamic-theme/css-rules.ts
@@ -73,14 +73,14 @@ export function iterateCSSDeclarations(style: CSSStyleDeclaration, iterate: (pro
                     iterate(prop, val);
                 }
             });
-        } else {
-            shorthandVarDependantProperties.forEach((prop) => {
-                const val = style.getPropertyValue(prop);
-                if (val && val.includes('var(')) {
-                    iterate(prop, val);
-                }
-            });
         }
+    } else {
+        shorthandVarDependantProperties.forEach((prop) => {
+            const val = style.getPropertyValue(prop);
+            if (val && val.includes('var(')) {
+                iterate(prop, val);
+            }
+        });
     }
 }
 
--- a/src/inject/dynamic-theme/css-rules.ts
+++ b/src/inject/dynamic-theme/css-rules.ts
@@ -62,9 +62,9 @@ export function iterateCSSDeclarations(style: CSSStyleDeclaration, iterate: (pro
         }
         iterate(property, value);
     });
-    const cssText = style.cssText;
-    if (cssText.includes('var(')) {
-        if (isSafari) {
+    if (isSafari) {
+        const cssText = style.cssText;
+        if (cssText.includes('var(')) {
             // Safari doesn't show shorthand properties' values
             shorthandVarDepPropRegexps.forEach(([prop, regexp]) => {
                 const match = cssText.match(regexp);
--- a/src/inject/dynamic-theme/css-rules.ts
+++ b/src/inject/dynamic-theme/css-rules.ts
@@ -62,15 +62,18 @@ export function iterateCSSDeclarations(style: CSSStyleDeclaration, iterate: (pro
         }
         iterate(property, value);
     });
-    if (isSafari && style.cssText.includes('var(')) {
-        // Safari doesn't show shorthand properties' values
-        shorthandVarDepPropRegexps.forEach(([prop, regexp]) => {
-            const match = style.cssText.match(regexp);
-            if (match && match[1]) {
-                const val = match[1].trim();
-                iterate(prop, val);
-            }
-        });
+    if (isSafari) {
+        const cssText = style.cssText;
+        if (cssText.includes('var(')) {
+            // Safari doesn't show shorthand properties' values
+            shorthandVarDepPropRegexps.forEach(([prop, regexp]) => {
+                const match = cssText.match(regexp);
+                if (match && match[1]) {
+                    const val = match[1].trim();
+                    iterate(prop, val);
+                }
+            });
+        }
     } else {
         shorthandVarDependantProperties.forEach((prop) => {
             const val = style.getPropertyValue(prop);
--- a/src/inject/dynamic-theme/css-rules.ts
+++ b/src/inject/dynamic-theme/css-rules.ts
@@ -62,9 +62,13 @@ export function iterateCSSDeclarations(style: CSSStyleDeclaration, iterate: (pro
         }
         iterate(property, value);
     });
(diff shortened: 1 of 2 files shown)


Gusted:
Because of the https://github.com/darkreader/darkreader/pull/6175/commits/6c7399dfcd505f99b3d0b02b12c301a7a8daa114 commit this PR is now considered another optimization PR.

It will now add the cost of lookinUp cssText property around ~80ms 
image

But avoid the big problem of using going over the shortHandDep properties which is ~140ms

image

So this also improves with 45% and thereby is now 1.75x faster performance.

And once again `iterateCSSDeclarations` is a popular function in dark reader so overall it will help performance and isn't tied to this particular case.

Gusted:
Also addresses the performance issues in #6161

src/inject/dynamic-theme/css-rules.ts
```diff
@@ -58,22 +58,25 @@ export function iterateCSSDeclarations(style: CSSStyleDeclaration, iterate: (pro
...
-        });
-    } else {
-        shorthandVarDependantProperties.forEach((prop) => {
-            const val = style.getPropertyValue(prop);
-            if (val && val.includes('var(')) {
-                iterate(prop, val);
-            }
-        });
+    const cssText = style.cssText;
+    if (cssText.includes('var(')) {
```
alexanderby, line 62: First of all there was previously a bug I forgot to fix. It should be `if (isSafari) { if(style.cssText.includes...) {... } } else { ... }`.
Secondly I'm not sure if `cssText.includes('var(')` should be done for other browsers. It's hard to find out if checking every style for "var(" is faster than running `style.getPropertyValue(prop)` for each style several times. Worth making some benchmark, for Firefox too.

alexanderby (changes requested):

src/inject/dynamic-theme/css-rules.ts
Gusted: Firefox seems to lack behind - Chromium can in some cases have better performance with the `includes`. But not worth it, not sure how the test cases are against actually world code like gmail. But then we are speaking of thousands of such styles. Which all vary

Gusted: https://jsben.ch/3jFp6

src/inject/dynamic-theme/css-rules.ts
```diff
@@ -58,22 +58,25 @@ export function iterateCSSDeclarations(style: CSSStyleDeclaration, iterate: (pro
...
-            }
-        });
-    } else {
-        shorthandVarDependantProperties.forEach((prop) => {
-            const val = style.getPropertyValue(prop);
-            if (val && val.includes('var(')) {
-                iterate(prop, val);
-            }
-        });
+    if (isSafari) {
```
alexanderby, line 61: If not Safari, then do nothing :) Nice optimisation :)

Seriously, it's hard to predict what exact style properties are more common. Here it seems like a bit more performant, but with other properties it will be a bit less performant. https://jsben.ch/CfD6l

Maybe some day we will grab some statistics, e.g. from top 100 sites. Anyway this exact check should not be a bottleneck, I assume it takes some 5%.

Worth comparing if it makes any noticeable difference in Gmail though.

alexanderby (changes requested):

src/inject/dynamic-theme/css-rules.ts
Gusted: > Maybe some day we will grab some statistics, e.g. from top 100 sites. Anyway this exact check should not be a bottleneck, I assume it takes some 5%.

Having some kind of benchmark against real-word data is good in general. As we now have to rely on stripped down test cases which can or cannot have any meaning against real-world sites.

About Gmail, it still seems like a good improvement to have. Especially as once again, it's a popular function so even 1ms shaves of ~30ms in general. so if it shaves of ~5ms you will have shave of  ~150ms etc.

> If not Safari, then do nothing :) Nice optimisation :)

Was this whole hack just targeted to safari? As I thought I did the `else {` check for `if (isSafari) { } else { }`. But doesn't seem like I've correctly coded that. As if it's just to safari this shortHandDep, then I say this PR is good to go.

Gusted:
https://github.com/darkreader/darkreader/pull/6175/commits/cb79d184eae8485e14d15201939bfea574343c91 Should save some more performance

src/inject/dynamic-theme/css-rules.ts
```diff
@@ -62,15 +62,18 @@ export function iterateCSSDeclarations(style: CSSStyleDeclaration, iterate: (pro
...
+            shorthandVarDepPropRegexps.forEach(([prop, regexp]) => {
+                const match = cssText.match(regexp);
+                if (match && match[1]) {
+                    const val = match[1].trim();
+                    iterate(prop, val);
+                }
+            });
+        }
     } else {
         shorthandVarDependantProperties.forEach((prop) => {
```
alexanderby, line 78: Hmm, so, what have you decided, to use `if (cssText.includes('var(')) {` for every browser or not?

src/inject/dynamic-theme/modify-css.ts
```diff
@@ -35,17 +35,15 @@ export function getModifiableCSSDeclaration(
     ignoreImageSelectors: string[],
     isCancelled: () => boolean,
 ): ModifiableCSSDeclaration {
-    const important = Boolean(rule && rule.style && rule.style.getPropertyPriority(property));
-    const sourceValue = value;
     if (property.startsWith('--')) {
         const modifier = getVariableModifier(variablesStore, property, value, rule, ignoreImageSelectors, isCancelled);
         if (modifier) {
-            return {property, value: modifier, important, sourceValue};
+            return {property, value: modifier, important: Boolean(rule && rule.style && rule.style.getPropertyPriority(property)), sourceValue: value};
```
alexanderby, line 41: Great, now I think you should move it to some function, to not repeat the same line.

alexanderby (changes requested):

src/inject/dynamic-theme/css-rules.ts
Gusted: Hmm, not sure yet. Really depends how this code actually shall behave.

Gusted: It's a good performance for gmail so let's include it again.

alexanderby (approved):

PR #6177, opened 2021-07-04
Add translate.metalune.xyz, remove two Nitter instances

Elaborendum:
Those two Nitter instances no longer exist.


src/config/dark-sites.config
```diff
@@ -772,6 +770,7 @@ totalwarwarhammer.gamepedia.com
 tracker.fumik0.com
 tracr.co
 training.azeria-labs.com
+translate.metalune.xyz
```
Gusted, line 773: image
Not dark by default.

Gusted (approved):

PR #6178, opened 2021-07-04
No greeting/unistall pages for debug-watch build

bershanskiy:
This change has no effect in release builds. This just avoids opening install/uninstall pages when developer installs/uninstalls build produced with `npm run debug-watch`.


Gusted (approved):
Seems good.

PR #6196, opened 2021-07-06
Fix recursion of `update` for legit media errors

Gusted:
- Imagine a scenario whereby some media error is legit and cannot be accessed by dark reader(not related to it not being loaded yet).
- With this patch we limit such scenario to only allow once to `update`. This prevents recursion and avoid infinite dead lock within dark reader.

@alexanderby Seems like the patch actually had a little flaw. This PR should be merged before the next release as nobody likes to see "dark reader is slow" because we are recursion on something that shouldn't be happening.

### DIFF
--- a/src/inject/dynamic-theme/style-manager.ts
+++ b/src/inject/dynamic-theme/style-manager.ts
@@ -305,7 +305,7 @@ export function manageStyle(element: StyleElement, {update, loadingStart, loadin
                 isAsyncCancelled: () => cancelAsyncOperations,
             });
             isOverrideEmpty = syncStyle.sheet.cssRules.length === 0;
-            if (sheetModifier.hasNotLoadedImports()) {
+            if (sheetModifier.getNotLoadedLinkStatus() === 'found') {
                 // "update" function schedules rebuilding the style
                 // ideally to wait for link loading, because some sites put links any time,
                 // but it can be complicated, so waiting for document completion can do the trick
--- a/src/inject/dynamic-theme/style-manager.ts
+++ b/src/inject/dynamic-theme/style-manager.ts
@@ -305,7 +305,7 @@ export function manageStyle(element: StyleElement, {update, loadingStart, loadin
                 isAsyncCancelled: () => cancelAsyncOperations,
             });
             isOverrideEmpty = syncStyle.sheet.cssRules.length === 0;
-            if (sheetModifier.getNotLoadedLinkStatus() === 'found') {
+            if (sheetModifier.shouldRebuiltStyle()) {
                 // "update" function schedules rebuilding the style
                 // ideally to wait for link loading, because some sites put links any time,
                 // but it can be complicated, so waiting for document completion can do the trick
--- a/src/inject/dynamic-theme/style-manager.ts
+++ b/src/inject/dynamic-theme/style-manager.ts
@@ -305,7 +305,7 @@ export function manageStyle(element: StyleElement, {update, loadingStart, loadin
                 isAsyncCancelled: () => cancelAsyncOperations,
             });
             isOverrideEmpty = syncStyle.sheet.cssRules.length === 0;
-            if (sheetModifier.shouldRebuiltStyle()) {
+            if (sheetModifier.shouldRebuildStyle()) {
                 // "update" function schedules rebuilding the style
                 // ideally to wait for link loading, because some sites put links any time,
                 // but it can be complicated, so waiting for document completion can do the trick
--- a/src/inject/dynamic-theme/stylesheet-modifier.ts
+++ b/src/inject/dynamic-theme/stylesheet-modifier.ts
@@ -24,20 +24,17 @@ function getThemeKey(theme: Theme) {
 
 const asyncQueue = createAsyncTasksQueue();
 
-// 'not-found' It means dark reader didn't found any non loaded link(s) on this stylesheet.
-// 'found' It means that dark reader did found non loaded link(s) and wasn't rebuilt yet.
-// 'found-and-rebuilt' It means that dark reader previously found non-loaded-links,
-// but has since then rebuilt the stylesheet.
-type NonLoadedLinkStatus = | 'not-found' | 'found' | 'found-and-rebuilt';
-
 export function createStyleSheetModifier() {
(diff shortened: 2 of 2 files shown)


alexanderby (changes requested):
That's great that you've noticed. Could you rename this function or change the way it works?
```typescript
function hasNotLoadedImports() {
    return foundNonLoadedLinks === 1;
}
```
So that the whole thing and naming makes logical sense and tells the story:
1. We calculate errors while loading imports.
2. If there are errored imports found, we rebuild the style when document is ready.
3. If there was an error, the style was rebuilt, document is ready, and it still shows the error, we don't rebuild the style based on this error.

Gusted:
Hmm, numbers are so confusion, so switching to "typed" strings should do the job.

src/inject/dynamic-theme/style-manager.ts
```diff
@@ -305,7 +305,7 @@ export function manageStyle(element: StyleElement, {update, loadingStart, loadin
                 isAsyncCancelled: () => cancelAsyncOperations,
             });
             isOverrideEmpty = syncStyle.sheet.cssRules.length === 0;
-            if (sheetModifier.hasNotLoadedImports()) {
+            if (sheetModifier.getNotLoadedLinkStatus() === 'found') {
```
alexanderby, line 308: It would be a bit better if Style Manager didn't know about statuses, but just knew if it should rebuild the style, or not. But at the same time you can just make sure this update is executed only once. Just add some lock to the Style Manager, e.g. a flag telling it has already reacted on error.

src/inject/dynamic-theme/stylesheet-modifier.ts
```diff
@@ -24,13 +24,19 @@ function getThemeKey(theme: Theme) {
 
 const asyncQueue = createAsyncTasksQueue();
 
+// 'not-found' It means dark reader didn't found any non loaded link(s) on this stylesheet.
+// 'found' It means that dark reader did found non loaded link(s) and wasn't rebuilt yet.
+// 'found-and-rebuilt' It means that dark reader previously found non-loaded-links,
+// but has since then rebuilt the stylesheet.
+type NonLoadedLinkStatus = | 'not-found' | 'found' | 'found-and-rebuilt';
```
alexanderby, line 31: If I don't miss anything, you can switch back to boolean.

src/inject/dynamic-theme/stylesheet-modifier.ts
```diff
@@ -24,13 +24,19 @@ function getThemeKey(theme: Theme) {
 
 const asyncQueue = createAsyncTasksQueue();
 
+// 'not-found' It means dark reader didn't found any non loaded link(s) on this stylesheet.
```
alexanderby, line 27: **D**ark **R**eader didn't **find**

src/inject/dynamic-theme/stylesheet-modifier.ts
```diff
@@ -24,13 +24,19 @@ function getThemeKey(theme: Theme) {
...
+// 'found' It means that dark reader did found non loaded link(s) and wasn't rebuilt yet.
```
alexanderby, line 28: **D**ark **R**eader **found** or **has found** (to the best of my English knowledge).

src/inject/dynamic-theme/stylesheet-modifier.ts
```diff
@@ -41,16 +47,18 @@ export function createStyleSheetModifier() {
...
+        return nonLoadedLinkStatus;
     }
 
     function modifySheet(options: ModifySheetOptions) {
         const rules = options.sourceCSSRules;
         const {theme, ignoreImageAnalysis, force, prepareSheet, isAsyncCancelled} = options;
 
         let rulesChanged = (rulesModCache.size === 0);
-        foundNonLoadedLinks = false;
+        if (nonLoadedLinkStatus === 'found') {
```
alexanderby, line 59: Maybe it should not reset anything here

alexanderby (changes requested):

src/inject/dynamic-theme/stylesheet-modifier.ts
```diff
@@ -30,8 +30,11 @@ export function createStyleSheetModifier() {
     const rulesModCache = new Map<string, ModifiableCSSRule>();
     const varTypeChangeCleaners = new Set<() => void>();
     let prevFilterKey: string = null;
-    let foundNonLoadedLinks = false;
+    // 'false' It means Dark Reader didn't find any non loaded link(s) on this stylesheet.
+    // 'true' It means that Dark Reader has found non loaded link(s).
+    let hasNonLoadedLink = false;
 
+    let hasRebuilt = false;
```
alexanderby, line 37: `didRebuild` or `wasRebuilt`.

Worth adding a new line before the interface declaration.

src/inject/dynamic-theme/stylesheet-modifier.ts
```diff
@@ -41,20 +44,23 @@ export function createStyleSheetModifier() {
         isAsyncCancelled: () => boolean;
     }
 
-    function hasNotLoadedImports() {
-        return foundNonLoadedLinks;
+    function shouldRebuiltStyle() {
```
alexanderby, line 47: `shouldRebuildStyle`

src/inject/dynamic-theme/stylesheet-modifier.ts
```diff
@@ -30,8 +30,11 @@ export function createStyleSheetModifier() {
     const rulesModCache = new Map<string, ModifiableCSSRule>();
     const varTypeChangeCleaners = new Set<() => void>();
     let prevFilterKey: string = null;
-    let foundNonLoadedLinks = false;
+    // 'false' It means Dark Reader didn't find any non loaded link(s) on this stylesheet.
```
alexanderby, line 33: This comment seems like not adding any details to the variable name and type. So you can remove it, or explain why these two booleans are important, or make variables' names more descriptive.

alexanderby (approved):

Gusted:
Having this PR fixed, I will almost call for a update of dark reader as this release will contain probably more than ever a lot of bug fixes code-wise.

PR #6200, opened 2021-07-06
Inverted banner logo on kali.org

ZOASR:


Gusted (commented):
I'm not so sure if this fixes anything.

ZOASR:
> I'm not so sure if this fixes anything.

I did not consider alphapetical order, will fix it!

Gusted:
I actually meant if this fix is necessary as I don't see why this logo should be inverted. As the inverted image has white-ish shadows.

ZOASR:
I just thought the inverted looked alot closer to the original logo, than the default set by the extension.

Gusted:
>  than the default set by the extension

I don't see anything being set by dark reader. I just see the original....

ZOASR:
In Dynamic mode the default looks like this for me, which is not how the logo looks:
image

Gusted:
By any chance on chromium? It seems like I don't have this behavior on Firefox. Let me suggest a fix that won't break on Firefox but will work on Chromium

src/config/dynamic-theme-fixes.config
```diff
@@ -6506,6 +6506,13 @@ IGNORE IMAGE ANALYSIS
 
 ================================
 
+kali.org
+
+INVERT
+#banner-logo
```
Gusted, line 6512:
```suggestion
IGNORE IMAGE ANALYSIS
#banner-logo
```

Gusted (changes requested):

ZOASR:
> By any chance on chromium? 
On chromium yes!
will try it on firefox and see

src/config/dynamic-theme-fixes.config
ZOASR: I tried using `IGNORE IMAGE ANALYSIS` But did not work on chromium

Gusted: You need to give it a reload before it takes effect ;)

ZOASR:
It looks fine on firefox, not sure what is the problem!

Gusted:
> It looks fine on firefox, not sure what is the problem!

Don't worry, Firefox is probably missing up some things with the code. I'll have a look later and see what can be done because this logo should be targeted as "light" and be inverted. Even though I agree with you that the original looks better.

The fix I suggested should disable the analysis altogether for that element.

Gusted (approved):
Thank you!

ZOASR:
> Thank you!

No problem!

PR #6201, opened 2021-07-07
MEGA.nz | Fixed the weird out of place gray border when in List view,…

stankovictab:
… changed the search bar color and removed an ugly fixed color for an element

old

new

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -7638,8 +7638,11 @@ i.transfer-progress-icon
 span.top-icon.language
 
 CSS
-.files-grid-view.fm .column-settings {
-    background-color: ${#181a1b} !important;
+.grid-wrapper {
+    outline: 0px !important;
+}
+.topbar-searcher {
+    background: var(--darkreader-bg--surface-grey-2) !important;
 }
 
 ================================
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -7638,8 +7638,11 @@ i.transfer-progress-icon
 span.top-icon.language
 
 CSS
-.files-grid-view.fm .column-settings {
-    background-color: ${#181a1b} !important;
+.grid-wrapper {
+    outline: 0px !important;
```
ZOASR, line 7642: Maybe changing the border color is bettter than removing it completely!
```suggestion
    outline-color: var(--darkreader-bg--surface-main);
```

ZOASR (changes requested):

src/config/dynamic-theme-fixes.config
Gusted: I agree with this one.

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
stankovictab: Yup, even better!

ZOASR: Great!, Thanks.

Gusted (approved):
Thank you!

PR #6206, opened 2021-07-07
Update `dark-sites.config` with new sites

jakubkaczor:


src/config/dark-sites.config
```diff
@@ -393,6 +393,8 @@ iwanttoeat.app
 ixirc.com
 jackdaniels.com
 jakobneumann.com
+jakubkaczor.com
+jakubkaczor.github.io
```
Gusted, line 397: It seems like a redirect? If so, it shouldn't be needed to be added in the list.

jakubkaczor: That is right. I couldn't see information regarding redirects in `CONTRIBUTING.md`, but now I see it may be unnecessery to add them. I have added two adresses in case they diverge, to not make another pull request. Is it ok to add both in such a scenario? I am the author of the github page and the owner of the domain, so I would make sure both adresses point to dark site. 😉

Gusted: Sure thing for now.

PR #6224, opened 2021-07-08
Update README.md

DJ-Katalina:
I am suggesting some minor grammatical error corrections for the README.

Other than removing a comma in one spot and adding a couple more commas elsewhere, nothing else really major except this sentence:

"Based on its analysis, Dark Reader will generate a dark mode based on this."

I suggest simplifying it to the following sentence: (Dropping the extra "based on this". I believe it sounds better this way without repeating the same thing twice.)

"Based on its analysis, Dark Reader will generate a dark mode." 

--- Also, is this the correct way to submit corrections for typos, grammatical errors, spelling mistakes, etc...?

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 <h2 align="center">Dark Reader</h2>
 <br/>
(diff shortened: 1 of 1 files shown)


Gusted (approved):
LGTM!

PR #6232, opened 2021-07-09
Partial fix for ggmania.com

pexxie:
The background file bg2.gif is being applied across the entire body with DR enabled in Dynamic mode.
This change makes a partial fix to replace the tiled GIF background with the DR neutral background.
Needed for this website as of DR v4.9.34.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5036,6 +5036,15 @@ CSS
 
 ================================
 
+ggmania.com
+
+CSS
+body {
+    background: var(--darkreader-neutral-background) !important;
+}
+
+================================
+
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -5036,6 +5036,15 @@ CSS
 
 ================================
 
+ggmania.com
+
+CSS
+body {
+    background: var(--darkreader-neutral-background) !important;
```
Gusted, line 5043:
```suggestion
    background-image: none !important;
```

Gusted (changes requested):
Just removing the image is good enough.

Gusted (approved):

PR #6240, opened 2021-07-10
Ubereats Map and Logo Fix v2

roberestarkk:
More reliable fix for Ubereats that doesn't require reloading the tracking page to get the map to appear properly, and also now covers the mini-map on the checkout page.
Also removed the unnecessary depth to the selectors as suggested last time.


Gusted (approved):
LGTM

PR #6242, opened 2021-07-10
`rule` can be empty/null

Gusted:
Fix a silent crash/error.

For some weird reason the browsers, won't error/crash when the `rule` is empty(inline-style modifications). And just closes dark reader for no good reason.

Regression introduced in #6175


Gusted:
Merging - As I otherwise cannot debug certain sites that had this weird crash behavior.

PR #6254, opened 2021-07-11
aprs.fi and aprsdirect.com

Luceoria:


src/config/dynamic-theme-fixes.config
```diff
@@ -1070,8 +1070,8 @@ INVERT
 aprs.fi
 
 INVERT
-.gm-style > :first-child
-.gm-style-iw-a
+img[src^="https://maps.googleapis.com/maps/"]
```
Gusted, line 1073: Does this fixes any existing issues?

src/config/dynamic-theme-fixes.config
```diff
@@ -1080,6 +1080,14 @@ CSS
 
 ================================
 
+aprsdirect.com
+
+INVERT
+.leaflet-tile-pane
+img[src^="https://maps.googleapis.com/maps/"]
```
Gusted, line 1087: I couldn't find such element, and would also probably be conflicting with the `leaflet-tile-pane`.

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
Luceoria: It does. Node icons are displayed inverted on the map, but not on the pop-up:
image
After the commit, both icon types are non-inverted.

Gusted: Both should be inverted. The non-inverted version doesn't look that good in dark-mode environment.

Luceoria (commented):
Google Maps tiles are now inverted, so are icons.

Luceoria:
Will submit another PR.

PR #6259, opened 2021-07-12
Add a image manager

### LINKED ISSUE #6237: [Performance] nzherald.co.nz very slow with darkreader enabled
# Performance Issue
Slow page performance on nzherald.co.nz with darkreader enabled
## Website
https://www.nzherald.co.nz/

## How to Reproduce

Go to the homepage or any subpage, it's very slow.
Turn off darkreader for this website, performance is as expected
## Performance Profiler

[Here's both enabled and disabled profiles on a fresh FF 89 profile](https://github.com/darkreader/darkreader/files/6794842/profiles.zip)

## System Information

Firefox Nightly and Stable 89
- OS: 
- Browser: 
- Dark Reader Version: 
Windows 10, Android. DR 4.9.34

Works fine with dark reader disabled. Webrender enabled.

Gusted:
- A long long long overdue PR.
- Add a manager on top of the `analyzeImage` function.
- Site-specific values for how many images can be analyzed per second.
- Make some more use of `async/await` to reduce load on the initial load and spread it out over time.
- Resolves #6237 And all gmail related issues.


src/defaults.ts
```diff
@@ -31,6 +31,7 @@ export const DEFAULT_THEME: Theme = {
     scrollbarColor: isMacOS ? '' : 'auto',
     selectionColor: 'auto',
     styleSystemControls: true,
+    maxImgPerSec: 30,
```
alexanderby, line 34: What if some website uses small images, and another has big?

src/inject/dynamic-theme/image.ts
```diff
@@ -50,93 +71,160 @@ async function urlToImage(url: string) {
...
 }
 
-function analyzeImage(image: HTMLImageElement) {
-    if (!canvas) {
-        createCanvas();
+// AnalyzeImage is a very slow function, so will add a manager on top of it
+// This manager will be responsible for analyzing incoming images of the queue.
+// And will only allow certain amount of images to be analyzed per second.
+// This is to avoid too much memory usage and CPU usage.
+class AnalyzeImageManager {
```
alexanderby, line 90: Nice usage of asynchronous queue here! The problem is one synchronous job takes unpredictable amount of time.

But the solution is already there! [I did it 4 years ago](https://github.com/TargetProcess/tauCharts/blob/master/src/charts/task-runner.ts) :)

It ensures the tasks take ~60FPS (no more 17ms synchronously), and launches other tasks later. Since `requestAnimationFrame` is paused when the tab is hidden, it switches to `requestIdleCallback`. Not sure if it is relevant for us, because we don't want to run the job on hidden tabs and frames.

Take a look on a simpler [example](https://github.com/alexanderby/mova/blob/master/src/content/index.js#L76) I used for another extension.

I think you could improve the code by putting a TaskRunner or AsyncQueue into a separate class or class-like factory function. Maybe it can be useful in other places too.

alexanderby (changes requested):

src/defaults.ts
Gusted: They can change it on a per-website-basis that's why I've listed this option under Theme and not under settings.

src/inject/dynamic-theme/image.ts
Gusted: Done with the latest commit, I'd know a few places which would love to have this kind of `AsyncQueue`, but let's test it out for now with the images and see how it goes. Gmail is noticeably performing better on my special machine and on my PC which I didn't expect given it was already only ~5 seconds overhead, but it's now ~2-3 seconds overhead.

src/inject/dynamic-theme/image.ts
```diff
@@ -15,22 +16,28 @@ export interface ImageDetails {
...
-        width: image.naturalWidth,
-        height: image.naturalHeight,
-        ...info,
-    };
+    return new Promise<ImageDetails>(async (resolve) => {
+        let dataURL: string;
+        if (url.startsWith('data:')) {
+            dataURL = url;
+        } else {
+            dataURL = await getImageDataURL(url);
```
alexanderby, line 27: Worth measuring what actually causes the problem: `getImageDataURL`, `urlToImage` or `analyzeImage`. Maybe they should be put into a queue too.

One problem with a queue is it can cause out of memory issues: all the images, dataURLs etc will be stored in memory until their turn comes and they can be garbage collected. The solution is... no solution, just balance between memory and synchronous intervals lengths. So worth checking if it causes anything for those problematic websites.

src/utils/async-queue.ts
```diff
@@ -0,0 +1,42 @@
+export type QueueEntry = () => void;
+
+// queue manager is an class that helps with managing tasks.
```
alexanderby, line 3: `// AsyncQueue is a class ...`

alexanderby, line 3: `// AsyncQueue is a class ...`

src/utils/async-queue.ts
```diff
@@ -0,0 +1,42 @@
...
+// More specifically, it helps with tasks that are often used.
+// It's fully asyncronous and uses promises and tries to get 60FPS.
```
alexanderby, line 5: `async**h**ronous`

src/utils/async-queue.ts
```diff
@@ -0,0 +1,42 @@
...
+export default class AsyncQueue {
+    private queue: QueueEntry[] = [];
+    private timerID: number = null;
```
alexanderby, line 8: There's a naming convention :) ... timer**Id**

src/utils/async-queue.ts
alexanderby, line 5: async**h**ronous

src/utils/async-queue.ts
```diff
@@ -0,0 +1,42 @@
...
+    stopQueue() {
+        if (this.timerID !== null) {
+            cancelAnimationFrame(this.timerID);
+            this.timerID = null;
+        }
+        this.queue = [];
+    }
+
+    // Ensures 60FPS.
+    private requestQueue() {
```
alexanderby, line 25: Something is wrong with the name `requestQueue`. It requests a queue in a queue :)

But actually it runs the tasks in synchronous chunks, until they start causing a FPS drop. So, try some another short and descriptive name.

Not sure if it is a problem in this case, but the first chunk is not run synchronously. You can leave it as is, but if a flash of images will be noticeable, maybe it could be improved by running the first chunk synchronously later.

alexanderby (changes requested):

src/inject/dynamic-theme/image.ts
Gusted: `analyzeImage` is problematic due to `canvas 2d`, which is using `drawImage`, however upon drawing it will need to decode an base64, png. All while blocking the execution of any other code. And I've seen cases taking it longer than 50ms to do so. I would prefer if we had some API to draw images which doesn't rely on such function that would block the whole execution. But that's only somewhat available for webgl today.

getImageDataURL has a potential bottleneck to due sending message to the background-script which is actually not that performant somehow

urlToImage seems to have no problems so-far. Other than `.href =` being slow sometimes which we cannot do anything about.

src/utils/async-queue.ts
Gusted: Really need to look into the naming-convetions rule from eslint 😄

src/utils/async-queue.ts
Gusted: Hopefully that won't happend the flash of images. As another workaround would be to temporialy have `background-image: none` on those selectors and have asyncRebuild add the correctly modified value.

alexanderby (approved):

Gusted:
Let's be hopeful that all users will notice problematic sites *cough* gmail, run more smoother.

PR #6262, opened 2021-07-13
Move src/inject/utils/log.ts to src/utils/log.ts

bershanskiy:
Move log utils from `src/inject/utils` to `src/utils` because this file is imported by background as well, specifically `src/background/utils/extension-api.ts`.

### DIFF
--- a/src/background/utils/extension-api.ts
+++ b/src/background/utils/extension-api.ts
@@ -1,6 +1,5 @@
 import {isPDF} from '../../utils/url';
 import {isFirefox, isEdge} from '../../utils/platform';
-import {logWarn} from '../../inject/utils/log';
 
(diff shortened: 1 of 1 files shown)


Gusted:
`src/background/utils/extension-api.ts` Seems like this one shouldn't call to `logWarn` it should log it anyway as it's an important storage error. Which we actually want to have on production with such kinds of problems.

bershanskiy:
So should I just replace `logWarn` with `console.error` in `src/background/utils/extension-api.ts`?

Gusted:
Sounds good to me.

bershanskiy:
I just force-pushed a commit doing this.

Gusted (approved):
Thank you!

PR #6263, opened 2021-07-14
Add starship.rs

alexalekseyenko:
This site is already doing their part

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -702,6 +702,7 @@ star-made.org
 stargazing.net
 starlink.com
 starmadedock.net
+starship.rs
 staszic.waw.pl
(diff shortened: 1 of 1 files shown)


Gusted:
image
my eyes

alexalekseyenko:
For future reference, what's our strategy with these sorts of sites? That is, should sites that are responsive to the system theme be inverted? At the moment, this site is doubly inverted on my end (once of its own volition and then again by the extension).

Gusted:
> That is, should sites that are responsive to the system theme be inverted?

User choice. Such sites are very radical in changing the the themes. A lot of this sites seems to be unstable and considered to be on some new list. Which is actually something I want to prevent with #5719

PR #6266, opened 2021-07-14
Fix regression introduced in PR #6113

bershanskiy:
During review of #6113 I [renamed one message object attribute in definition](https://github.com/darkreader/darkreader/pull/6113#discussion_r668071121),
but accidentally left the old name in one instance. This commit updates
that name and adds type annotation (which would have thrown an error at
build time if I annotated the type).

### DIFF
--- a/src/background/messenger.ts
+++ b/src/background/messenger.ts
@@ -131,7 +131,12 @@ export default class Messenger {
 
     reportChanges(data: ExtensionData) {
         if (this.changeListenerCount > 0 || isFirefox) {
-            chrome.runtime.sendMessage({name: 'background', type: 'changes', data});
+            const message: Message = {
+                from: 'background',
+                type: 'changes',
+                data
+            };
+            chrome.runtime.sendMessage(message);
         }
     }
 }


bershanskiy:
Example of a bug fixed by this commit:
1. Open any site and open popup UI.
2. Toggle global extension switch ON/OFF.
3. Notice that switch did not change position, even though site changed color.
4. Try to toggle switch again and notice that it no longer works.

src/background/messenger.ts
```diff
@@ -131,7 +131,12 @@ export default class Messenger {
 
     reportChanges(data: ExtensionData) {
         if (this.changeListenerCount > 0 || isFirefox) {
-            chrome.runtime.sendMessage({name: 'background', type: 'changes', data});
+            const message: Message = {
```
Gusted, line 134: Don't declare variables if you end up using it once or don't modify it in a future reference.

bershanskiy: I declared this variable just to force a type annotation onto it. Is there a better way to do it?

This just "converts" the object into the `Message` type and does not error out the build:
```
            chrome.runtime.sendMessage({
                from: 'background',
                type: 'changes',
                data
            } as Message);
```

Gusted: Not really. As this kind of things just happens once it should be expected to be correct the first time. However in such hotpath the garbage collecter will like to recycle message variable instantly 🤷🏽

Gusted (approved):

src/background/messenger.ts
bershanskiy: I wish TypeScript compiler was aware of connection between `chrome.runtime.sendMessage` and `chrome.runtime.onMessage` and created a union type of all `onMessage` listeners and then forced it onto `sendMessage` arguments.

Gusted: Sounds a bit like Quantum Entanglement ;). Typescript allow for type constructors like `chrome.runtime.sendMessage<Message>()` But it seems like the `@types/chrome` doesn't have this implemented 🤷🏽

bershanskiy: > `chrome.runtime.sendMessage<Message>()`

That was the first thing I tried, but it did't work.

Gusted: The type definition needs explicit support for it. Opening up a issue/discussion https://github.com/DefinitelyTyped/DefinitelyTyped or even a PR for it is the most effective way.

PR #6271, opened 2021-07-15
urbanterror.info is dark

DJ-Katalina:
I am submitting this website to be added to the dark-sites.config file. However, before adding it to the list, I have a question.

The entire site is dark by default, but this one section which is their shop: https://www.urbanterror.info/shop/ is not dark. Does this situation automatically disqualify the site for the dark list? I was not sure in this case and I was just looking for clarity on what to do when this come up.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -791,6 +791,7 @@ unc0ver.dev
 undergroundcellar.com
 underlords.com
 undertale.com
+urbanterror.info
 userdiag.com
 ussr.obys.agency
 uwatchfree.ax


Gusted:
> The entire site is dark by default, but this one section which is their shop: https://www.urbanterror.info/shop/ is not dark. Does this situation automatically disqualify the site for the dark list? I was not sure in this case and I was just looking for clarity on what to do when this come up.

`!urbanterror.info/shop` is possible with #2517

DJ-Katalina:
Can I go ahead and submit this website to the dark-sites.config file or do I need to wait till #2517 is merged?

Gusted:
You will be submitting a site with a non-dark subpage in the current case I would've close the PR.

DJ-Katalina:
Understood. I will close this immediately. Thank you for the clarification.

PR #6284, opened 2021-07-17
Refactor: declare types explicitly

bershanskiy:
This commit adds to a new compiler option `"noImplicitAny": true` to `src/tsconfig.json`.

Test plan: compile before and after this patch and compare the output with `diff -r`, it should match exactly because this commit just changes type annotations which are validated and stripped out during build.

### DIFF
--- a/src/generators/static-theme.ts
+++ b/src/generators/static-theme.ts
@@ -226,7 +226,7 @@ function upperCaseToCamelCase(text: string) {
 export function parseStaticThemes($themes: string) {
     return parseSitesFixesConfig<StaticTheme>($themes, {
         commands: staticThemeCommands,
-        getCommandPropName: upperCaseToCamelCase,
+        getCommandPropName: upperCaseToCamelCase as any,
         parseCommandValue: (command, value) => {
             if (command === 'NO COMMON') {
                 return true;
--- a/src/generators/static-theme.ts
+++ b/src/generators/static-theme.ts
@@ -176,57 +176,45 @@ const ruleGenerators = [
     createRuleGen((t) => t.invert, () => ['filter: invert(100%) hue-rotate(180deg)']),
 ];
 
-const staticThemeCommands = [
-    'NO COMMON',
+const staticThemeCommands = {
+    'NO COMMON': 'noCommon',
 
-    'NEUTRAL BG',
-    'NEUTRAL BG ACTIVE',
-    'NEUTRAL TEXT',
-    'NEUTRAL TEXT ACTIVE',
-    'NEUTRAL BORDER',
+    'NEUTRAL BG': 'neutralBg',
+    'NEUTRAL BG ACTIVE': 'neutralBgActive',
+    'NEUTRAL TEXT': 'neutralText',
+    'NEUTRAL TEXT ACTIVE': 'neutralTextActive',
+    'NEUTRAL BORDER': 'neutralBorder',
 
-    'RED BG',
-    'RED BG ACTIVE',
-    'RED TEXT',
-    'RED TEXT ACTIVE',
-    'RED BORDER',
+    'RED BG': 'redBg',
+    'RED BG ACTIVE': 'redBgActive',
+    'RED TEXT': 'redText',
+    'RED TEXT ACTIVE': 'redTextActive',
+    'RED BORDER': 'redBorder',
 
-    'GREEN BG',
-    'GREEN BG ACTIVE',
-    'GREEN TEXT',
-    'GREEN TEXT ACTIVE',
-    'GREEN BORDER',
+    'GREEN BG': 'greenBg',
+    'GREEN BG ACTIVE': 'greenBgActive',
+    'GREEN TEXT': 'greenText',
+    'GREEN TEXT ACTIVE': 'greenTextActive',
+    'GREEN BORDER': 'greenBorder',
 
-    'BLUE BG',
-    'BLUE BG ACTIVE',
-    'BLUE TEXT',
-    'BLUE TEXT ACTIVE',
-    'BLUE BORDER',
+    'BLUE BG': 'blueBg',
+    'BLUE BG ACTIVE': 'blueBgActive',
+    'BLUE TEXT': 'blueText',
+    'BLUE TEXT ACTIVE': 'blueTextActive',
+    'BLUE BORDER': 'blueBorder',
 
-    'FADE BG',
-    'FADE TEXT',
-    'TRANSPARENT BG',
+    'FADE BG': 'fadeBg',
+    'FADE TEXT': 'fadeText',
+    'TRANSPARENT BG': 'transparentBg',
 
-    'NO IMAGE',
-    'INVERT',
-];
-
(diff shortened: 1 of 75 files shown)


Gusted (commented):
I'm not sure If I'm fan of this typescript compiler function - current IDE's like VSCode and other IDE's that has a language server for typescript can easily detect certain types without having to implicit declare them. The typescript compiler has come very far in terms of control flow analysis and their newest beta has even more support for it https://devblogs.microsoft.com/typescript/announcing-typescript-4-4-beta/#cfa-aliased-conditions.

And in a lot of situations having typed declarations aren't needed and will just drive me personally crazy to see a bunch of `: any` or other obvious types. Because when we use such values it's either already analyzed by typescript correctly or it won't mather because of the nature of javascript(with a possibility of `any` we most likely already have some null-check in place). 

This option seems a bit like insane mode if you're dealing with a lot of uncontrolled inputs, which we 95% aren't.

bershanskiy:
> I'm not sure If I'm fan of this typescript compiler function

If you decide to close this, PR it's fine with me. This is a pretty small nit and does not affect much. I'll just keep it as a patch [here](https://github.com/bershanskiy/darkreader/tree/patches) in case you decide to revisit this idea.

> The typescript compiler has come very far

I like where TypeScript is headed, automatic type and initialization state inference is exactly what people expect from a modern language. Howver, from a quick glance at this, I don't think any of these new features would help in cases touched by this patch. ([Symbol and Template String Pattern Index Signatures](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4-beta/#symbol-template-signatures) is neat, but won't resolve type uncertainty addressed in this patch.

> And in a lot of situations having typed declarations aren't needed and will just drive me personally crazy to see a bunch of : any or other obvious types.

I plan to come back later and remove as much `any` annotations as reasonable.

This patch just makes puts a hard limit on amount of technical debt that can be accumulated over time by making it obvious. For example, one weird thing that this patch highlights is this:
https://github.com/darkreader/darkreader/blob/ea3686317aed286ea003ec7fb331b99e76e39f7b/src/generators/dynamic-theme.ts#L56
On the left of `||` you have `DynamicThemeFix.css` of type `string` and on the right is array. This does not end up making a difference because later there is a join which turns everything into a string anyway.

Gusted:
> If you decide to close this, PR it's fine with me. This is a pretty small nit and does not affect much. I'll just keep it as a patch [here](https://github.com/bershanskiy/darkreader/tree/patches) in case you decide to revisit this idea.

Hmm the added types seems good in my eyes, but having such feature on by-default would drive me insane or anyone during development as you would go type declare a lot, especially if you want to test a simple case or a simplified version or some unsafe code which you know would be x or y.



> On the left of `||` you have `DynamicThemeFix.css` of type `string` and on the right is array. This does not end up making a difference because later there is a join which turns everything into a string anyway.

pretty much hackery around hackery 😄 

If the compiler function could be turned off somehow and only test it on PR's would be great honestly as it would only annoy me once in development.

bershanskiy:
> If the compiler function could be turned off somehow and only test it on PR's would be great

That's possible: we can just create another TypeScript configuration file which extends the existing `src/tsconfig.json` and adds this particuler flag. When specifically should it run? I don't see CI ever building the extension (I see only linting and tests being run).

> pretty much hackery around hackery 😄

So it's intended behavior?

Gusted:
> When specifically should it run? I don't see CI ever building the extension (I see only linting and tests being run).

Hmm, tsc should be run actually but because it's horribly slow(Github CI has 1/2 cores) even with caching, it's asking for huge and long run-times. I would be looking into that, but for it can be disabled. I will hold it in mind when Ill refactor the github workflows to work smarter. 

> So it's intended behavior?

It could be, bot ways would work because we later on do `.filter((s) => s)` which would check if it's undefined or not and than actually use the value.

bershanskiy:
> It could be, both ways would work because we later on do .filter((s) => s) which would check if it's undefined or not and than actually use the value.

OK, I removed the comment.

> Hmm, tsc should be run actually but because it's horribly slow(Github CI has 1/2 cores) even with caching, it's asking for huge and long run-times. I would be looking into that, but for it can be disabled. I will hold it in mind when Ill refactor the github workflows to work smarter.

We can postpone adding `"noImplicitAny"` to the TS compiler flags then because CI does not test it anyway. We can merge the rest of this patch, which just makes type declarations stricter and will reduce the ammount of future work needed to enable the this compiler option on CI?

src/generators/static-theme.ts
```diff
@@ -226,7 +226,7 @@ function upperCaseToCamelCase(text: string) {
 export function parseStaticThemes($themes: string) {
     return parseSitesFixesConfig<StaticTheme>($themes, {
         commands: staticThemeCommands,
-        getCommandPropName: upperCaseToCamelCase,
+        getCommandPropName: upperCaseToCamelCase as any,
```
bershanskiy, line 229: I will avoid this `any` cast by restricting the value to `keyof StaticTheme` in a follow-up patch similarly to how it's done elsewhere in the code. I ommitted this change in this patch to keep LoC reasonable.

bershanskiy: This is done now.

Gusted: Gracious!

alexanderby:
I remember @Gusted used to create a mega PR with many changes to ESLint rules. So maybe you could merge your effort there?

That's great to see explicit types like `const items: string[] = []` instead of `const items = []`. But I find it doubtful to add `: any` everywhere, maybe there's a ESLint (TSLint) rule for this?

Gusted:
> maybe there's a ESLint (TSLint) rule for this?

Not for `implicit any`, however I know a bunch of rules to enforce better types usages in dark reader.

bershanskiy:
> mega PR with many changes to ESLint rules.

This PR was supposed add TSConfig compiler option `"noImplicitAny": true` in `src/tsconfig.json`, but then we decided to pospone this change. I think it's better to keep ESLint and TSConfig changes separate, since they are fundamentally different things.

> But I find it doubtful to add : any everywhere, maybe there's a ESLint (TSLint) rule for this?

For now, I can just remove all useless `any` declarations. This way, we avoid extra `any` noise and still make some progress towards `noImplicitAny` option if we ever decide to use it.

alexanderby:
> For now, I can just remove all useless `any` declarations.

That would be a cool refactoring.

bershanskiy:
> > For now, I can just remove all useless `any` declarations.
> 
> That would be a cool refactoring.

Done! I added a few more [annotations](https://github.com/darkreader/darkreader/pull/6284/commits/f3814c8059e99e4af5b0e86c3204334d15bd99f9) and removed all new `any` annotations.

bershanskiy:
I'm merging this now, sice:
 - this change is safe since the produced output fater this patch is exactly identical to before this patch (tested with `diff -r`)
 - this PR is boring: all existing concerns are adressed and there hasn't been any ongoing discussion for over two weeks

PR #6291, opened 2021-07-18
Refactor: precompute staticThemeCommands

bershanskiy:
This patch merges `staticThemeCommands` array and `upperCaseToCamelCase` method into a simple map literal, which is a sufficiently simple structure for TS to enforce types on. This mathces how other parts of code implement corresponding things ([`src/generators/dynamic-theme.ts`](https://github.com/darkreader/darkreader/blob/master/src/generators/dynamic-theme.ts) and [`src/generators/css-filter.ts`](https://github.com/darkreader/darkreader/blob/master/src/generators/css-filter.ts)).

### DIFF
--- a/src/generators/static-theme.ts
+++ b/src/generators/static-theme.ts
@@ -214,7 +214,7 @@ const staticThemeCommands = {
 export function parseStaticThemes($themes: string) {
     return parseSitesFixesConfig<StaticTheme>($themes, {
         commands: Object.keys(staticThemeCommands),
-        getCommandPropName: (command) => staticThemeCommands[command] || null,
+        getCommandPropName: (command) => staticThemeCommands[command],
         parseCommandValue: (command, value) => {
             if (command === 'NO COMMON') {
                 return true;
--- a/src/generators/static-theme.ts
+++ b/src/generators/static-theme.ts
@@ -176,57 +176,45 @@ const ruleGenerators = [
     createRuleGen((t) => t.invert, () => ['filter: invert(100%) hue-rotate(180deg)']),
 ];
 
-const staticThemeCommands = [
-    'NO COMMON',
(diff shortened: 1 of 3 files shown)


src/generators/static-theme.ts
```diff
@@ -176,57 +176,45 @@ const ruleGenerators = [
...
+    'NO IMAGE': 'noImage',
+    'INVERT': 'invert',
+};
 
 export function parseStaticThemes($themes: string) {
     return parseSitesFixesConfig<StaticTheme>($themes, {
-        commands: staticThemeCommands,
-        getCommandPropName: upperCaseToCamelCase,
+        commands: Object.keys(staticThemeCommands),
+        getCommandPropName: (command) => staticThemeCommands[command] || null,
```
Gusted, line 217:
```suggestion
        getCommandPropName: (command) => staticThemeCommands[command],
```

It will by default return `undefined` or the value, no need to fallback to `null`.

Gusted (changes requested):

src/generators/static-theme.ts
bershanskiy: I agree that `|| null` is unnecessary. Should I also remove it from these two places?
https://github.com/darkreader/darkreader/blob/8a56257b84f8a2d7adef27d1018b5f3ac383943f/src/generators/dynamic-theme.ts#L17
https://github.com/darkreader/darkreader/blob/8a56257b84f8a2d7adef27d1018b5f3ac383943f/src/generators/css-filter.ts#L215

Gusted: Fine by me.

bershanskiy: Done!

Gusted (approved):
Seems to have no problems

PR #6296, opened 2021-07-18
Use Alarms API

bershanskiy:
Use [Alarms API](https://developer.chrome.com/docs/extensions/reference/alarms/) instead of timeouts because those are not persisted after non-persistent background shutdown. Alarms will survive across background shutdowns and re-activations and will actually re-activate unloaded background.

This patch adds one more permission, `alarms`, which is necessary to use this API (it does not generate any extra permission messages during install). This patch migrates two `setInterval` call sites and leaves [few more](https://github.com/darkreader/darkreader/blob/a5c5c9a6e683419befd87cbadde31525193b589d/src/background/extension.ts#L240) because I plan to refactor those in a different way.

### DIFF
--- a/src/background/utils/network.ts
+++ b/src/background/utils/network.ts
@@ -49,7 +49,7 @@ class LimitedCacheStorage {
                 this.removeExpiredRecords();
             }
         });
-        chrome.alarms.create(LimitedCacheStorage.ALARM_NAME, {periodInMinutes: getDurationInMinutes({minutes: 1})});
+        chrome.alarms.create(LimitedCacheStorage.ALARM_NAME, {periodInMinutes: 1});
     }
 
     has(url: string) {
--- a/src/background/utils/network.ts
+++ b/src/background/utils/network.ts
@@ -1,6 +1,6 @@
 import {loadAsDataURL, loadAsText} from '../../utils/network';
 import {getStringSize} from '../../utils/text';
-import {getDuration} from '../../utils/time';
+import {getDuration, getDurationInMinutes} from '../../utils/time';
 
 interface RequestParams {
     url: string;
@@ -38,12 +38,18 @@ interface CacheRecord {
 class LimitedCacheStorage {
     static QUOTA_BYTES = ((navigator as any).deviceMemory || 4) * 16 * 1024 * 1024;
     static TTL = getDuration({minutes: 10});
+    static ALARM_NAME = 'network';
 
     private bytesInUse = 0;
     private records = new Map<string, CacheRecord>();
 
     constructor() {
-        setInterval(() => this.removeExpiredRecords(), getDuration({minutes: 1}));
+        chrome.alarms.onAlarm.addListener(async (alarm) => {
+            if (alarm.name === LimitedCacheStorage.ALARM_NAME) {
+                this.removeExpiredRecords();
+            }
+        });
+        chrome.alarms.create(LimitedCacheStorage.ALARM_NAME, {periodInMinutes: getDurationInMinutes({minutes: 1})});
     }
 
     has(url: string) {
--- a/src/background/utils/network.ts
+++ b/src/background/utils/network.ts
@@ -1,6 +1,6 @@
 import {loadAsDataURL, loadAsText} from '../../utils/network';
 import {getStringSize} from '../../utils/text';
-import {getDuration, getDurationInMinutes} from '../../utils/time';
+import {getDuration} from '../../utils/time';
 
 interface RequestParams {
     url: string;
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -37,6 +37,7 @@
         }
     ],
     "permissions": [
+        "alarms",
         "fontSettings",
         "storage",
         "tabs",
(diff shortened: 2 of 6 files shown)


src/background/utils/network.ts
```diff
@@ -38,12 +38,18 @@ interface CacheRecord {
...
     static TTL = getDuration({minutes: 10});
+    static ALARM_NAME = 'network';
 
     private bytesInUse = 0;
     private records = new Map<string, CacheRecord>();
 
     constructor() {
-        setInterval(() => this.removeExpiredRecords(), getDuration({minutes: 1}));
+        chrome.alarms.onAlarm.addListener(async (alarm) => {
+            if (alarm.name === LimitedCacheStorage.ALARM_NAME) {
```
Gusted, line 48: Hmm, could I ask why we are checking if the the alarm_name is the same? Is this some kind of intentional design of the alarm API?

Gusted (commented):
IMO seems like a good change 👍🏽

src/background/utils/network.ts
bershanskiy: Alarm name allows us to react only to the desired alarm, which is helpful when we have multiple different classes creating their own alarms for different things. Without name check every alarm listener would react to every alarm. E.g., `Newsmaker` would react to `LimitedCacheStorage` alarm and query your server for news every minute when you actually want it to do it every 4 hours.

Gusted: Aha, that makes more change.

Gusted (approved):
You got my 👍🏽, but because it adds an extra permission, I would like @alexanderby to give an approval for this one as well. As I've seen some developers experiences issues when they request new permissions(even non-powerful-ish ones).

src/manifest.json
```diff
@@ -37,6 +37,7 @@
         }
     ],
     "permissions": [
+        "alarms",
```
alexanderby, line 40: One thing I'm afraid it to cause a warning for users after an update.

src/utils/time.ts
```diff
@@ -54,23 +54,27 @@ interface Duration {
...
+        duration += time.hours * 60 * 1000;
     }
     if (time.days) {
-        duration += time.days * 24 * 60 * 60 * 1000;
+        duration += time.days * 24 * 60 * 1000;
     }
     return duration;
 }
 
+export function getDuration(time: Duration) {
```
alexanderby, line 74: Maybe do it vice versa? Leave the `getDuration` function as is.

src/background/utils/network.ts
```diff
@@ -38,12 +38,18 @@ interface CacheRecord {
...
+                this.removeExpiredRecords();
+            }
+        });
+        chrome.alarms.create(LimitedCacheStorage.ALARM_NAME, {periodInMinutes: getDurationInMinutes({minutes: 1})});
```
alexanderby, line 52: Can you just put `1` here without a function?

alexanderby (approved):
Nice usage of proper APIs.

src/manifest.json
bershanskiy: I don't have time to find all the sources about this, so I'll post them tomorrow. Essentially, `alarms` is one of those odd permisions which must be in the manifest but are granted automatically.

src/utils/time.ts
```diff
@@ -54,25 +54,25 @@ interface Duration {
...
     if (time.days) {
-        duration += time.days * 24 * 60 * 1000;
+        duration += time.days * 24 * 60 * 60 * 1000;
     }
     return duration;
 }
 
-export function getDuration(time: Duration) {
-    return getDurationInMinutes(time) * 60;
+export function getDurationInMinutes(time: Duration) {
```
alexanderby, line 74: So `getDurationInMinutes` is no more needed, restore the `getDuration` as it was.

alexanderby (changes requested):
Run the `npm run code-style`, it says some variable is not defined.

src/utils/time.ts
alexanderby: Oops, wrong, it is actually used

src/manifest.json
bershanskiy: I have the following evidence that installing update with the new `alarms` permission does not generate extra warnings:
 - `chrome.management.getPermissionWarningsByManifest` creates the list of user-visible permissions warnings for a given manifest and it returns nothing for a manifest with only `alarms` permission ([Chrome docs](https://developer.chrome.com/docs/extensions/reference/management/#method-getPermissionWarningsByManifest), [MDN](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/management/getPermissionWarningsByManifest))
 - Chrome docs do not list `alarms` among ["Permissions with warnings"](https://developer.chrome.com/docs/extensions/mv2/permission_warnings/#permissions_with_warnings)

Ideally, we would be able to test that Chrome and Firefox do not actually just decide to disable extension anyway upon end-user updating the extension. I'm still researching ways to do this.

bershanskiy:
> Run the `npm run code-style`, it says some variable is not defined.

Fixed it. (It said that a wariable was unused because it was a useless import.)

alexanderby (approved):

PR #6311, opened 2021-07-19
Fix for drive.google.com

Elaborendum:
- Lines that separate folders in the path to a file/folder
- Icons in "add shortcut to Drive" option.
- Dialog box symbol that appears if no file is selected.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3684,9 +3684,9 @@ div[role="menuitem"] svg
 div[role="dialog"][style^="width: 350px;"] div[role="button"][tabindex="0"]
 div[role="dialog"] ~ div[role="menu"] > div[role="menuitem"] > div > div:not([style*="background-image"])
 img[src$="type/audio/mp3"]
-.o-Yc-o > div > div > div > svg
-.pj-Wb > .e-F-v > div > div > div
-.a-q-Xc-Ab-c > svg > .a-s-fa-Ha-pa[fill="#000000"]
+div[role="listitem"] > div > div > div > svg
+div[role="menuitem"] > div > div > div
+div[data-label="nd"] > div > div > svg > path[fill="#000000"]
 
 CSS
 span[data-type="spelling"] {
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3684,6 +3684,9 @@ div[role="menuitem"] svg
 div[role="dialog"][style^="width: 350px;"] div[role="button"][tabindex="0"]
 div[role="dialog"] ~ div[role="menu"] > div[role="menuitem"] > div > div:not([style*="background-image"])
 img[src$="type/audio/mp3"]
+.o-Yc-o > div > div > div > svg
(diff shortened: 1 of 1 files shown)


Gusted (changes requested):
The usage of jiberish CSS is not recommended, especially on this site as I've seen it changed multiple times a day. Everything must be done trough more generic attribute selectors.

Elaborendum:
I hope the changes I made aren't too broad either.

src/config/dynamic-theme-fixes.config
```diff
@@ -3684,6 +3684,9 @@ div[role="menuitem"] svg
 div[role="dialog"][style^="width: 350px;"] div[role="button"][tabindex="0"]
 div[role="dialog"] ~ div[role="menu"] > div[role="menuitem"] > div > div:not([style*="background-image"])
 img[src$="type/audio/mp3"]
+div[role="listitem"] > div > div > div > svg
+div[role="menuitem"] > div > div > div
```
Gusted, line 3688: Only this one concerns me as every "right-click menu" has such class and will definitely have more than 3 div's inside of it. Do you have an example wherefor this is used?

Elaborendum: imagen
I don't exactly understand what you mean, but Line 3688 is for the icons from the "add shortcut to drive" menu that appears once you select the option from the main right-click menu. I tried, but I don't think it is within my ability to narrow this one down further.

On the other hand, would changing Line 3687 to `div[role="listitem"] > div > div > div > svg[fill="#000000"] > path` be of any help in narrowing that one down?

Gusted: The 2 other lines don't concern me for now, but ` svg[fill="#000000"] > path` would be helpful.

Welp, we actually encounter such `invert` problem from the drive.google.com/file viewer and with a slightly modified CSS the same trick will work
```CSS
div[role="menu"] > div > div[role="menuitem"] > div > div > div
div[role="menu"] > div > div[role="menuitem"][class*=" "] > div > :last-child
```

Gusted (approved):

Vedanth-P:
Hello,
Sorry for bringing up this issue again, but I think I'm encountering problems with a line of code that was added in this pull request.

I am seeing that elements that are selected by ```div[role="menu"] > div > div[role="menuitem"][class*=" "] > div > :last-child``` are being wrongly inverted. Perhaps it was different when this pull request was merged, but currently each item on the right-click menus has an image, beside which is text, beside which (sometimes) is a second image. On menu items that do not have this second image, the selector targets the text, and it gets inverted.
Menu Item without Image Menu Item with Picture

So I removed the ```div[role="menu"] > div > div[role="menuitem"][class*=" "] > div > :last-child``` selector from the ```INVERT``` list, and the icons and dividers (which I understand are what this pull request aimed to rectify) seem to look fine. In addition, some of the icons were in fact wrongly inverted (e.g. the arrow on the right of the "Open with" menu item), but they seem fine when I remove the selector.
Wrong Inversion
Corrected Inversion


I just want to clarify that I am not missing anything, before going ahead and submitting this as a fix.

PR #6316, opened 2021-07-20
Enhance site list that able to filter invalid url

ywei390:
Before enhancement, the 'Site list' can accept invalid URL such as '1234412', 'afbdfjb' and etc. After enhancement, the 'Site list' will filter out invalid URL and only accept valid URL such as 'www.google.com'.

### DIFF
--- a/src/ui/popup/components/site-list-settings/index.tsx
+++ b/src/ui/popup/components/site-list-settings/index.tsx
@@ -9,7 +9,13 @@ interface SiteListSettingsProps extends ExtWrapper {
 
 export default function SiteListSettings({data, actions, isFocused}: SiteListSettingsProps) {
     function isSiteUrlValid(value: string) {
-        return /^([^\.\s]+?\.?)+$/.test(value);
+        var pattern = new RegExp('^(https?:\\/\\/)?'+ // protocol
+        '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|'+ // domain name
+        '((\\d{1,3}\\.){3}\\d{1,3}))'+ // OR ip (v4) address
+        '(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*'+ // port and path
+        '(\\?[;&a-z\\d%_.~+=-]*)?'+ // query string
+        '(\\#[-a-z\\d_]*)?$','i'); // fragment locator
+        return !!pattern.test(value);
     }
 
     return (


haowei12:
It is good to handle this, the name 'Site list' should be only accept a valid site.

haowei12 (approved):
Changes does not impact other part. Good to go.

Gusted (commented):
Why do we need such extensive regexp? User is ultimately responsible for it. Also this PR doesn't work with the existing syntax we allow. And it will create a new regexp constructor every time. 

I don't see the value of this PR.

haowei12:
By using the 'User is ultimately responsible for it' as a reason, this shows that the program created will never a good program. Just imagine you say this when you're working in Facebook, Google etc. User behavior is hard to predict and we definitely need to control it.

I would prefer to amend the code to better rather than use 'User is ultimately responsible for it' as a reason for not fix. It's a shame, consider it please. @Gusted

Gusted:
> By using the 'User is ultimately responsible for it' as a reason, this shows that the program created will never a good program.

What? Are you implying that having a configurable list which allows most inputs and then let us deal with the parsing of that input and actually using that dark reader will never be a good program?

>  Just imagine you say this when you're working in Facebook, Google etc. User behavior is hard to predict and we definitely need to control it.

I luckily don't work for them and hopefully never will. 

Yes user behavior is hard to predict, but we are now equalizing extensions to big _shady_ companies with different platforms and different motivations.

But your average use won't use the list, and those who actually will need it, have a good case for it and knows a bit more about URL's than the average user.

> I would prefer to amend the code to better rather than use 'User is ultimately responsible for it' as a reason for not fix.

The code for this would be much much more larger than this regex, because our current parser will allow a good variety of inputs.  Everything from `*` to `^https://google.com/404$` And with the coming up #2517 you can throw this code out, as it will allow an even more variety of inputs.

>  It's a shame, consider it please.

Why it's a shame? I'm not flattered by some code that will restrict the kind of input to the list, It will actually be very unpleasant by having your (maybe) invalid URL being removed from the list without any warning. 
The list was made to 
1. display the the current list and remove entries if you want to.
2. allow to enter edge-cases which aren't handled by simply clicking the site toggle.

I've already considered it and my main points are: It's too restrictive, doesn't allow the current syntax we handle, doesn't correctly handle IPv{4,6}. Bunch of regex screams for performance problems which isn't uncommon for javascript.

If it was such list that the average user had to deal with, I would've think otherwise about it. But it's like adding advanced IDE features to the devtools so the user won't make CSS mistakes and have automatic suggestions, while in reality less than 1% of the users will use it and 99% of those knows CSS anyway.

I don't see the enhancement of it, because we will parse it anyway and if it's invalid, it will do the job that is being asked to do so.

haowei12:
@Gusted I think the name of 'Site list' has causes the confusion to me, I thought by right it should just accept the 'Site' (www, http, https etc.), but seems like now it should and it is allowed to accept anything.

Anyway, it is happy to see #2517, will wait for it to be released. ;) 

Thanks for your explanation.

PR #6327, opened 2021-07-22
Use Alarms API for reconnecting socket on dev build

bershanskiy:
This commit replaces one more instance of `setTimeout` in the background context with an alarm. This is direct continuation of work done in #6296.

Note: This alarm runs in 1s (like the original code) and browsers can log warnings promissing to throttle alarms in release builds to once a minute. This is not a problem, since this code will never run in release builds.

### DIFF
--- a/src/background/index.ts
+++ b/src/background/index.ts
@@ -19,6 +19,7 @@ const WATCH = __WATCH__;
 if (WATCH) {
     const PORT = __PORT__;
     const ALARM_NAME = 'socket-close';
+    const PING_INTERVAL_IN_MINUTES = 1 / 60;
 
     const socketAlarmListener = (alarm) => {
         if (alarm.name === ALARM_NAME) {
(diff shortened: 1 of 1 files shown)


src/background/index.ts
```diff
@@ -48,7 +48,15 @@ if (WATCH) {
                 }
             }
         };
-        socket.onclose = () => setTimeout(listen, 1000);
+        socket.onclose = () => {
+            const ALARM_NAME = 'socket-close';
+            chrome.alarms.onAlarm.addListener((alarm) => {
```
Gusted, line 53: `setTimeout` will only call the code once. But with this, it will call every second and act as an `setInterval`

bershanskiy: This patch specifies only `delayInMinutes`, which means alarm will be triggered only once like `setTimeout`. To get a recurring event I would have to use `periodInMinutes`.

Gusted: Aha, didn't know the `alarms` API allowed that.

Gusted (approved):
Got my approval.

bershanskiy:
I added two refactoring commits which make sure socket alarm listener is registered only when there is no active socket connection. No core logic changes.

src/background/index.ts
```diff
@@ -18,10 +18,21 @@ const WATCH = __WATCH__;
 
 if (WATCH) {
     const PORT = __PORT__;
+    const ALARM_NAME = 'socket-close';
+    const PING_INTERVAL_IN_MINUTES = 1 / 60;
```
alexanderby, line 22: Hmmm, what is a possible outcome of such short intervals? Will it make the page persistent?

alexanderby (approved):

src/background/index.ts
bershanskiy: Yes, short intervals can be wasteful (including staying in memory for prolonged periods of time). Chrome acttually throttles periodic alarms to once a minute and tries to limit non-persistent background lifespan to 5 minutes.

However, in this case performance is not an issue because this code runs only on debug watch build and only when there is no active socket connection.

PR #6328, opened 2021-07-22
Fix for market.yandex.ru

rakleed:
- Invert map.

Example of site page: https://market.yandex.ru/geo?productId=752262035


Gusted (approved):
That's some weird *** selector but LGTM 😄

PR #6333, opened 2021-07-23
Remove UserStorage.constructor argument onRemoteSettingsChange

bershanskiy:
This commit removes one argument of `UserStorage` which is not used anyway.

Comment says:
> TODO: Requires proper handling and more testing
>  to prevent cycling across instances.

Sync inconsistency and races can be fixed in a much simpler way:
 - always use `local` storage area everywhere in core extension logic, background, UI, etc.
 - define dedicated import/export logic for transfering data between `local` and `sync` aread

Basically, every extension install can always have its own coherent state recorded in `local` area. At the same time, the `local` and `sync` areas can be kept in sync by dedicated import/export logic with sanity checks and data merging.

### DIFF
--- a/src/background/extension.ts
+++ b/src/background/extension.ts
@@ -50,7 +50,7 @@ export class Extension {
             },
             onColorSchemeChange: this.onColorSchemeChange,
         });
-        this.user = new UserStorage({onRemoteSettingsChange: () => this.onRemoteSettingsChange()});
+        this.user = new UserStorage();
         this.awaiting = [];
     }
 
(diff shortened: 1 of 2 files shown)


Gusted:
> Sync inconsistency and races can be fixed in a much simpler way:

    always use local storage area everywhere in core extension logic, background, UI, etc.

If said user wants to sync on the browser's sync solution it should be used, as the browser can be closed at any time and we don't have specific code nor can to go and run the code to sync it to `sync`.



>     * define dedicated import/export logic for transfering data between `local` and `sync` aread

Why not use one way or another? It will only increase more code to have certain moments to transfer data between them, which also could have other data from another machine(which is why we use onSettingsChanged). 

> Basically, every extension install can always have its own coherent state recorded in local area. At the same time, the local and sync areas can be kept in sync by dedicated import/export logic with sanity checks and data merging.

It shouldn't be kept in sync with some dedicated logic, it should be done either not done or done while it's saving the other area, as we never have a guarantee it's saved to the other area after x period of time.

bershanskiy:
This patch is just removing an empty function `onRemoteSettingsChange`. If you plan to use `onRemoteSettingsChange` in the future, than it makes sense to keep it.

PR #6342, opened 2021-07-23
Fix for dynamic on wikipedia

sandorex:
It was inverting `.mwe-math-element` which then becomes unreadable as it's black on dark background


Gusted:
Open up a bug report. This bug is on your end.

sandorex:
My bad this was my fault i probably broke something

PR #6347, opened 2021-07-24
Use Alarms API instead of intervals in automation

bershanskiy:
Previously, extension "automation" feature was checking every 10 seconds whether it was "night" or "day", even though typically night-day transition happens only 2 times in a 24 hour period. This commit enables extension to determine when exactly this transition is expected to happen next time and schedule an alarm for exact time.

This commit changes return value of `isInTimeInterval` and `isNightAtLocation` from simple boolen to object which contains the old boolean and a timestamp of when the answer will change. ~~To limit LoC and simplify code review, these functions schedule the new check for a dummy time, which replicates current extension behavior of running the check every 10 seconds.~~

### DIFF
--- a/src/background/extension.ts
+++ b/src/background/extension.ts
@@ -8,18 +8,16 @@ import TabManager from './tab-manager';
 import UserStorage from './user-storage';
 import {setWindowTheme, resetWindowTheme} from './window-theme';
 import {getFontList, getCommands, setShortcut, canInjectScript} from './utils/extension-api';
-import {isInTimeInterval, getDuration, isNightAtLocation} from '../utils/time';
+import {isInTimeInterval, isNightAtLocation} from '../utils/time';
 import {isURLInList, getURLHostOrProtocol, isURLEnabled, isPDF} from '../utils/url';
 import ThemeEngines from '../generators/theme-engines';
 import createCSSFilterStylesheet from '../generators/css-filter';
 import {getDynamicThemeFixesFor} from '../generators/dynamic-theme';
 import createStaticStylesheet from '../generators/static-theme';
 import {createSVGFilterStylesheet, getSVGFilterMatrixValue, getSVGReverseFilterMatrixValue} from '../generators/svg-filter';
-import type {ExtensionData, FilterConfig, News, Shortcuts, UserSettings, TabInfo} from '../definitions';
+import type {ExtensionData, FilterConfig, News, Shortcuts, UserSettings, TabInfo, TimeCheck} from '../definitions';
 import {isSystemDarkModeEnabled} from '../utils/media-query';
 import {isFirefox, isThunderbird} from '../utils/platform';
-
-const AUTO_TIME_CHECK_INTERVAL = getDuration({seconds: 10});
 
 export class Extension {
     ready: boolean;
@@ -33,6 +31,9 @@ export class Extension {
     tabs: TabManager;
     user: UserStorage;
 
+    private isEnabledNow: boolean = null;
+
+    static ALARM_NAME = 'auto-time-alarm';
     constructor() {
         this.ready = false;
 
@@ -54,30 +55,51 @@ export class Extension {
         this.awaiting = [];
     }
 
-    isEnabled() {
-        const {automation} = this.user.settings;
-        if (automation === 'time') {
-            const now = new Date();
-            return isInTimeInterval(now, this.user.settings.time.activation, this.user.settings.time.deactivation);
-        } else if (automation === 'system') {
-            if (isFirefox) {
-                // BUG: Firefox background page always matches initial color scheme.
-                return this.wasLastColorSchemeDark == null
-                    ? isSystemDarkModeEnabled()
-                    : this.wasLastColorSchemeDark;
-            }
-            return isSystemDarkModeEnabled();
-        } else if (automation === 'location') {
-            const latitude = this.user.settings.location.latitude;
-            const longitude = this.user.settings.location.longitude;
+    private alarmListener = (alarm): void => {
+        if (alarm.name === Extension.ALARM_NAME) {
+            this.handleAutoCheck();
+        }
+    };
 
-            if (latitude != null && longitude != null) {
-                const now = new Date();
-                return isNightAtLocation(now, latitude, longitude);
-            }
+    isEnabled(): boolean {
+        if (this.isEnabledNow !== null) {
+            return this.isEnabledNow;
         }
 
-        return this.user.settings.enabled;
+        const {automation} = this.user.settings;
+
+        let goodTime: TimeCheck = null;
+        switch (automation) {
+            case 'time':
+                const now = new Date();
+                goodTime = isInTimeInterval(now, this.user.settings.time.activation, this.user.settings.time.deactivation);
+                break;
+            case 'system':
+                if (isFirefox) {
+                    // BUG: Firefox background page always matches initial color scheme.
+                    return this.wasLastColorSchemeDark == null
+                        ? isSystemDarkModeEnabled()
+                        : this.wasLastColorSchemeDark;
+                }
+                return isSystemDarkModeEnabled();
+            case 'location': {
+                const latitude = this.user.settings.location.latitude;
+                const longitude = this.user.settings.location.longitude;
+
+                if (latitude != null && longitude != null) {
+                    const now = new Date();
+                    goodTime = isNightAtLocation(now, latitude, longitude);
+                }
+                break;
+            }
+            default:
+                return this.user.settings.enabled;
+        }
+        this.isEnabledNow = goodTime.rightNow;
+        if (goodTime.nextCheck) {
+            chrome.alarms.create(Extension.ALARM_NAME, {when: goodTime.nextCheck});
+        }
+        return goodTime.rightNow;
     }
 
     private awaiting: Array<() => void>;
@@ -237,12 +259,8 @@ export class Extension {
     private wasEnabledOnLastCheck: boolean;
 
     private startAutoTimeCheck() {
-        setInterval(() => {
-            if (!this.ready || this.user.settings.automation === '') {
-                return;
-            }
-            this.handleAutoCheck();
-        }, AUTO_TIME_CHECK_INTERVAL);
+        chrome.alarms.onAlarm.addListener(this.alarmListener);
+        this.handleAutoCheck();
     }
 
     private wasLastColorSchemeDark = null;
@@ -259,6 +277,7 @@ export class Extension {
         if (!this.ready) {
             return;
         }
+        this.isEnabledNow = null;
         const isEnabled = this.isEnabled();
         if (this.wasEnabledOnLastCheck !== isEnabled) {
             this.wasEnabledOnLastCheck = isEnabled;
@@ -347,6 +366,7 @@ export class Extension {
     //
 
     private onAppToggle() {
+        this.isEnabledNow = null;
         if (this.isEnabled()) {
             this.icon.setActive();
             if (this.user.settings.changeBrowserTheme) {
--- a/src/background/extension.ts
+++ b/src/background/extension.ts
@@ -86,8 +86,7 @@ export class Extension {
                 const longitude = this.user.settings.location.longitude;
 
                 if (latitude != null && longitude != null) {
-                    const now = new Date();
-                    goodTime = isNightAtLocation(now, latitude, longitude);
+                    goodTime = isNightAtLocation(latitude, longitude);
                 }
                 break;
             }
--- a/src/background/extension.ts
+++ b/src/background/extension.ts
@@ -82,8 +82,7 @@ export class Extension {
                 }
                 return isSystemDarkModeEnabled();
             case 'location': {
-                const latitude = this.user.settings.location.latitude;
-                const longitude = this.user.settings.location.longitude;
+                const {latitude, longitude} = this.user.settings.location;
 
                 if (latitude != null && longitude != null) {
                     timingInformation = isNightAtLocation(latitude, longitude);
--- a/src/background/extension.ts
+++ b/src/background/extension.ts
@@ -71,8 +71,7 @@ export class Extension {
         let goodTime: TimeCheck = null;
         switch (automation) {
             case 'time':
-                const now = new Date();
-                goodTime = isInTimeInterval(now, this.user.settings.time.activation, this.user.settings.time.deactivation);
+                goodTime = isInTimeInterval(this.user.settings.time.activation, this.user.settings.time.deactivation);
                 break;
             case 'system':
                 if (isFirefox) {
--- a/src/background/extension.ts
+++ b/src/background/extension.ts
@@ -68,10 +68,10 @@ export class Extension {
 
         const {automation} = this.user.settings;
 
-        let goodTime: TimeCheck = null;
+        let timingInformation: TimeCheck = null;
         switch (automation) {
             case 'time':
-                goodTime = isInTimeInterval(this.user.settings.time.activation, this.user.settings.time.deactivation);
+                timingInformation = isInTimeInterval(this.user.settings.time.activation, this.user.settings.time.deactivation);
                 break;
             case 'system':
                 if (isFirefox) {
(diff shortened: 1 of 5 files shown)


src/utils/time.ts
```diff
@@ -37,14 +39,24 @@ function compareTime(a: number[], b: number[]) {
     return 1;
 }
 
-export function isInTimeInterval(date: Date, time0: string, time1: string) {
+function dummyTime() {
```
bershanskiy, line 42: I will remove this `dummyTime` function in future commits once `isInTimeInterval` and `isNightAtLocation` can calculate exact time.

Gusted: For the first portion of converting: 
[patches.tar.gz](https://github.com/darkreader/darkreader/files/6874024/patches.tar.gz)

The always night/day time requires some more specialized Math. Otherwise we could just fallback to check it every day.

Gusted: The calculation for it would be more or less brute-forcing which is something I don't like so you could just fallback to check it every day, even though if someone was in this time-range they happen for around ~180 days before going back to the correct cycle. Which is another way to calculate the approximate.

bershanskiy: I excluded the actual calculation changes from this PR to simplify reviewing this PR, however I can add them in if you prefer.

> The calculation for it would be more or less brute-forcing which is something I don't like

As far as I see, calculating sunset/sunrise would be a brute-force for each _day_ of seasonal day/night, which is still better than brute-forcing every 10 seconds. So falling back to checking every day might not be that bad. However, I'd prefer to just fall back to scheduling the next alarm in 24 hours and defering this problem until later.

Gusted: > I excluded the actual calculation changes from this PR to simplify reviewing this PR, however I can add them in if you prefer.

Just add them now. As in both cases I will just approve them and merge them.

>  I'd prefer to just fall back to scheduling the next alarm in 24 hours and defering this problem until later.

👍🏽

bershanskiy: I just added time calculation for isInTimeInterval (with tests). Solar dawn/sunset is next.

src/utils/time.ts
```diff
@@ -196,7 +196,7 @@ function getSunsetSunriseUTCTime(
         return {
             alwaysDay: false,
             alwaysNight: false,
-            time: UT * getDuration({hours: 1}),
+            time: Math.round(UT * getDuration({hours: 1})),
```
bershanskiy, line 199: This rounds number of milliseconds to the nearest integer, which is convenient for debugging and writing tests. This rounding does not actually lead to any perceived precision loss since JS event loop can introduce delays over a millisecond anyway.

src/utils/time.ts
bershanskiy: I added support for calculating solar dawn/sunset and tests. The results make sense (they fall within reasonable ballpark), but do not match exactly any calculators I could find on the web. I suspect that these slight variations (within an hour) can be caused by different definitions of "dawn" and "sunset" in different models. E.g., someone might say dawn is when you see first sunrays creating ambient light, someone might say it is when sun appears on the horizon and someone might say it is when sun is above horizon and provides reasonable illumination.

bershanskiy:
Apparently, tests are failing because two millisecond timing values differ between my PC and GitHub Actions runs. Specifically, 1568093140142 and 1568093042501. That is ~97 seconds of difference, which is a bit much.

Gusted:
Even weirder: I'm getting a totally complete other error on another line...

image

Gusted:
I have a good feeling it might be due to `UTC` being used everywhere which is something that doesn't go well, given that everyone lives in different time-zones and doesn't follow UTC time-zone and will actually receive incorrect timings due to it as they can live +8 hours away from UTC or -5 hours.

bershanskiy:
> I have a good feeling it might be due to UTC being used everywhere

Time zones would have a much larger impact than 97 seconds. I suspect it is due to floating point precission loss somewhere in `getTime`, but I don't know how to find an exact line. As a work around, it's possible to round off a few significant digits in tests.

Gusted:
> I suspect it is due to floating point precission loss somewhere in `getTime`

In that case we actually should return a precise Date object. As we don't want to round off some tests in order to look good.

bershanskiy:
Note that your difference is `97642` and mine is `97641`.

bershanskiy:
> > I suspect it is due to floating point precission loss somewhere in getTime
>
> In that case we actually should return a precise Date object. As we don't want to round off some tests in order to look good.

Could you elaborate? If problem is caused by numeric digit loss, it wouldn't matter whether return value is a plain number or the same number converted to a `Date`.

Gusted:
> Could you elaborate? If problem is caused by numeric digit loss, it wouldn't matter whether return value is a plain number or the same number converted to a `Date`.

Well, I still find it hard to imagine that it's numeric digit loss as the float precision should only happen at around ~53 bits and should be able to handle the precise number up until 9007199254740991 (2^53 -1). Which means that the epoch times `1627419407399` should be perfectly able to be stored without loss of precision, also if such things would happen, it will probably be a known thing and some workaround would be known as it isn't uncommon to store and handle epoch times. 


Date is an internal javascript object, which could work around such issue if that's the case. But like I said, I still think it's due to the usage of UTC, which isn't a good representation of the current date of said user or machine.

Gusted:
Anyway - What I'm trying to say, the current logic resolves around UTC time. However we shouldn't return UTC time back to the user as the user isn't in always in UTC - but if we do we should take into account the timezone difference with UTC to reflect that back into the alarm api. The amount of conversion between numbers could cause for imprecision or whatever the cause is.

bershanskiy:
@Gusted I found the bug and fixed it, now CI passes.

src/utils/time.ts
```diff
@@ -116,9 +116,9 @@ function getSunsetSunriseUTCTime(
     latitude: number,
     longitude: number,
 ) {
-    const dec31 = new Date(date.getUTCFullYear(), 0, 0);
```
bershanskiy, line 83: This is the buggy line. The problem is `new Date(...)` returns an object which internally keeps track of user's time zone. Instead, we can use `Date.UTC(...)` which just returns a timestamp in UTC GMT which does not depend on user's time zone.

Gusted:
> @Gusted I found the bug and fixed it, now CI passes.

That's great, let's forget our past discussion about the theory's 😄.

tests/utils/time.tests.ts
```diff
@@ -36,27 +36,27 @@ test('Duration', () => {
...
-    expect(isNightAtLocation(utcDate(2019, 8, 9, 18, 0), 52, 0)).toBe(false);
-    expect(isNightAtLocation(utcDate(2019, 8, 9, 20, 0), 52, 0)).toBe(true);
-    expect(isNightAtLocation(utcDate(2019, 8, 9, 23, 59), 52, 0)).toBe(true);
+    expect(isNightAtLocation(52, 0, utcDate(2019, 8, 9, 0, 0))).toEqual({rightNow: true, nextCheck: Date.UTC(2019, 8, 9, 5, 24, 2, 501)});
+    expect(isNightAtLocation(52, 0, utcDate(2019, 8, 9, 5, 0))).toEqual({rightNow: true, nextCheck: Date.UTC(2019, 8, 9, 5, 24, 2, 501)});
+    expect(isNightAtLocation(52, 0, utcDate(2019, 8, 9, 7, 0))).toEqual({rightNow: false, nextCheck: 1568053786448});
+    expect(isNightAtLocation(52, 0, utcDate(2019, 8, 9, 12, 0))).toEqual({rightNow: false, nextCheck: 1568053786448});
+    expect(isNightAtLocation(52, 0, utcDate(2019, 8, 9, 18, 0))).toEqual({rightNow: false, nextCheck: 1568053786448});
+    expect(isNightAtLocation(52, 0, utcDate(2019, 8, 9, 20, 0))).toEqual({rightNow: true, nextCheck: 1568093042501});
+    expect(isNightAtLocation(52, 0, utcDate(2019, 8, 9, 23, 59))).toEqual({rightNow: true, nextCheck: 1568093042501});
```
bershanskiy, line 45: I believe that, e.g., `Date.UTC(2019, 8, 9, 5, 24, 2, 501)` is easier to read than timestamp `1568006642501`. Should I convert all other timestamps to this format or should I just use simple numeric timestamps?

Gusted: Yeah go for the read-ability

src/background/extension.ts
```diff
@@ -54,30 +55,49 @@ export class Extension {
...
-            return isSystemDarkModeEnabled();
-        } else if (automation === 'location') {
-            const latitude = this.user.settings.location.latitude;
-            const longitude = this.user.settings.location.longitude;
-
-            if (latitude != null && longitude != null) {
-                const now = new Date();
-                return isNightAtLocation(now, latitude, longitude);
+
+        let goodTime: TimeCheck = null;
```
Gusted, line 71:
```suggestion
        let timingInformation: TimeCheck = null;
```

Or any other variable 😄

src/utils/time.ts
```diff
@@ -27,24 +29,58 @@ function parse24HTime(time: string) {
...
-    if (a[0] < b[0] || (a[0] === b[0] && a[1] < b[1])) {
+    if (time1[0] < time2[0] || (time1[0] === time2[0] && time1[1] < time2[1])) {
         return -1;
     }
     return 1;
 }
 
-export function isInTimeInterval(date: Date, time0: string, time1: string) {
+// a <= b
+function nextIntervalTime(a, b, t, date: Date): number {
```
Gusted, line 43:
```suggestion
function nextIntervalTime(time1, time2, currentTime, date: Date): number {
```

tests/utils/time.tests.ts
bershanskiy: Done!

Gusted (changes requested):
Just some variable naming. The code on it's own and the logic is approved for me.

@alexanderby If you'd like to have a look soon 🤷🏽‍♀️, as this one will be merged soon.

src/background/extension.ts
bershanskiy: Done!

src/utils/time.ts
bershanskiy: Done!

src/utils/time.ts
```diff
@@ -204,8 +247,70 @@ export function isNightAtLocation(
...
     );
 
-    if (sunsetTime > sunriseTime) {
-        return (currentTime > sunsetTime) || (currentTime < sunriseTime);
+    if (sunriseTime < sunsetTime) {
+        // Timeline:
+        // --- sunrise <----> sunset ---
+        let rightNow: boolean;
+        let nextCheck: number;
+        if (currentTime < sunriseTime) {
```
Gusted, line 255: What about moving the if statements a bit around and simplify it a bit for us.

```ts
if ((sunriseTime < currentTime) && (currentTime < sunsetTime)) {
            // Timeline:
            // --- sunrise <----> sunset ---
            //               ^
            //          Current time
            rightNow = false;
            nextCheck = Date.UTC(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate(), 0, 0, 0, sunsetTime);
} else {
            // Timeline:
            // --- sunrise <----> sunset ---
            //   ^                       ^
            //           Current time
            rightNow = true;
            nextCheck = Date.UTC(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate() + (sunsetTime < currentTime ? 1 : 0), 0, 0, 0, sunriseTime);
}
```

src/utils/time.ts
```diff
@@ -204,8 +247,70 @@ export function isNightAtLocation(
...
+        return {
+            rightNow,
+            nextCheck,
+        };
+    }
+    // Timeline:
+    // --- sunset <----> sunrise ---
+    let rightNow: boolean;
+    let nextCheck: number;
+    if (currentTime < sunsetTime) {
```
Gusted, line 288: Same story for these functions.

src/utils/time.ts
```diff
@@ -204,8 +247,70 @@ export function isNightAtLocation(
...
+    if ((sunsetTime < currentTime) && (currentTime < sunriseTime)) {
+        // Timeline:
+        // --- sunset <----> sunrise ---
+        //               ^
+        //          Current time
+        rightNow = true;
+        nextCheck = Date.UTC(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate(), 0, 0, 0, sunriseTime);
     }
-    return (currentTime > sunsetTime) && (currentTime < sunriseTime);
+    if ((sunsetTime < currentTime) && (currentTime < sunriseTime)) {
```
Gusted, line 304: Wrong if condition btw. But should be solved if we go `if () {} else {]`

Gusted (changes requested):
Woops just a couple more 😃

src/background/extension.ts
```diff
@@ -54,30 +55,49 @@ export class Extension {
...
+                }
+                break;
             }
+            default:
+                return this.user.settings.enabled;
         }
-
-        return this.user.settings.enabled;
+        this.isEnabledNow = timingInformation.rightNow;
+        if (timingInformation.nextCheck) {
```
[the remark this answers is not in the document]
bershanskiy: This is an unusual/edge case when time automation is used and start time is equal to end time. As of now, UI allows this to happen, but I feel like it shouldn't.

This behavior is described by the following `isInTimeInterval` tests:
```
expect(isInTimeInterval('10:00', '10:00', new Date(2018, 11, 4, 10))).toEqual({rightNow: false, nextCheck: null});
expect(isInTimeInterval('10:00', '10:00', new Date(2018, 11, 4, 12))).toEqual({rightNow: false, nextCheck: null});
expect(isInTimeInterval('10:00', '10:00', new Date(2018, 11, 4, 8))).toEqual({rightNow: false, nextCheck: null});
```

src/background/extension.ts
```diff
@@ -54,30 +55,49 @@ export class Extension {
...
 
-    isEnabled() {
+    private alarmListener = (alarm): void => {
+        if (alarm.name === Extension.ALARM_NAME) {
+            this.handleAutoCheck();
+        }
+    };
+
+    isEnabled(): boolean {
+        if (this.isEnabledNow !== null) {
```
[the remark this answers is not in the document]
bershanskiy: I'm using `null` on purpose here. Removing it would modify existing logic for case when `isEnabledNow` is `false`.

Gusted (approved):
All seems good for me. If you think it's ready let me know so I can merge it.

bershanskiy:
Yes, I think this is ready now.

Gusted:
Thank you 😃

src/background/extension.ts
```diff
@@ -33,6 +31,9 @@ export class Extension {
     tabs: TabManager;
     user: UserStorage;
 
+    private isEnabledNow: boolean = null;
```
alexanderby, line 34: Maybe you could remove `isEnabledNow`? Is it an optimisation? I think it makes code harder to understand and to keep track of its state. Or at least remove to something like `isEnabledByAutomation`.

src/background/extension.ts
```diff
@@ -33,6 +31,9 @@ export class Extension {
...
+
+    static ALARM_NAME = 'auto-time-alarm';
```
alexanderby, line 36: `ALARM_NAME` seems a generic name.

As an idea, maybe try using 2 alarms? One for the beginning of time interval and another one for the ending?

src/utils/time.ts
```diff
@@ -27,24 +29,57 @@ function parse24HTime(time: string) {
...
+        date.setMinutes(time2[1]);
+        return date.getTime();
+    }
+
+    // a <= b <= t
+    // Schedule for tomorrow at time a
+    return (new Date(date.getFullYear(), date.getMonth(), date.getDate() + 1, time1[0], time1[1])).getTime();
+}
+
+export function isInTimeInterval(time0: string, time1: string, date: Date = new Date()): TimeCheck {
```
alexanderby, line 69: `isInTimeInterval` should only return a boolean value, it should not know anything about next intervals etc.

Try exporting `nextIntervalTime` function, but make its name understandable out of context.

alexanderby (commented):
Sorry for the delay, I was feeling bad these days.

alexanderby:
@bershanskiy By the way, could you please check your email?

src/background/extension.ts
bershanskiy: Yes, `isEnabledNow` is just an optimisation for avoiding to evaluate these interval checks repeatedly. This is still useful for `getTabMessage()`:

https://github.com/darkreader/darkreader/blob/0301db28e7e6c461033d56dd18c8b47d609b8704/src/background/extension.ts#L414-L417

As far as I know, when user toggles extension ON, `getTabMessage` calls `isEnabled()` for every tab, which might be a lot.

src/background/extension.ts
bershanskiy: > `ALARM_NAME` seems a generic name.

I had to create internal alarms for Newsmaker, Network, and Extension classes, so I called them `Newsmaker.ALARM_NAME`, `LimitedCacheStorage.ALARM_NAME`, and `Extension.ALARM_NAME` as a naming convention. If you have a better suggestion, I can rename them and one in #6327. (I just need static string variables with unique names, but their names and even values can be entirely arbitrary.)

> As an idea, maybe try using 2 alarms? One for the beginning of time interval and another one for the ending?

I'm not sure how that would be an improvement, could you elaborate? These alarms can't be recurring alarms anyway because the exact sunset/dawn times vary a bit.

src/utils/time.ts
bershanskiy: Makes sense, will prepare a PR today or tomorrow.

src/background/extension.ts
bershanskiy: I'll bundle this rename together with `isInTimeInterval` PR.

src/utils/time.ts
bershanskiy: PR is at #6405.

PR #6352, opened 2021-07-25
Fix static theme for Youtube

gabrielfin:
This fixes a few issues with the Youtube player. These are mostly for the embedded player, but they also apply to the player in the Youtube website.

Here is the before and after:


Gusted (approved):
LGTM

PR #6354, opened 2021-07-25
Register LimitedCacheStorage alarm only when needed

bershanskiy:
This is direct continuation of #6296. In 6296, we added a new permission `alarms` and replaced `setInterval` with corresponding recurring (periodic) alarm which fires even when cache is empty and therefore there is nothing to do. That alarm is problematic if extension wants to go to sleep for exteneded periods of time but can't because it's resurrected by an alarm every minute. This PR replaces the unconditional periodic alarm with a one-time alarm which is scheduled only when needed (when cache is not empty).

This change will help extension background to go entirely passive (get unloaded) when extension does not do anything. E.g., when extension is set to OFF or when extension is set to ON but user does not load new pages. (After applying this patch and setting `persistent` to `false` in manifest, I observed that extension background consistently goes to sleep as expected.)

### DIFF
--- a/src/background/utils/network.ts
+++ b/src/background/utils/network.ts
@@ -42,7 +42,7 @@ class LimitedCacheStorage {
 
     private bytesInUse = 0;
     private records = new Map<string, CacheRecord>();
-    private hasAlarm: boolean = false;
+    private hasAlarm = false;
 
     constructor() {
         chrome.alarms.onAlarm.addListener(async (alarm) => {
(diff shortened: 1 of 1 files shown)


src/background/utils/network.ts
```diff
@@ -42,14 +42,24 @@ class LimitedCacheStorage {
 
     private bytesInUse = 0;
     private records = new Map<string, CacheRecord>();
+    private hasAlarm = false;
```
Gusted, line 45:
```suggestion
    private alarmIsActive = false;
```

Seems like a more descriptive variable to give.

Gusted (commented):
Must've slipped my email notifications somehow.

src/background/utils/network.ts
bershanskiy: Done!

Gusted (approved):

Issue #6365, opened 2021-07-28
[Feature Request] A list of sites using Light Mode (see description for use case)


AnishDe12020:
# Feature Request

A list of sites specifically using **lightmode** so that they are always converted to dark unless disabled by the user. This would make sense for sites like `github.com` which has dark mode all over the site except some pages like `github.com/marketplace` This way `github.com` can be added to the list of dark mode supported sites and `github.com/marketplace` can be added to the list of sites using light mode (at least until GitHub updates GitHub Marketplace to use a dark UI :smile:) so that Dark Reader applied the required styles only on that specific page.

## Is this feature request related to a problem?

In [`CONTRIBUTING.md`](https://github.com/darkreader/darkreader/blob/master/CONTRIBUTING.md) there is a line:
> All pages are dark by default regardless of the system's preferred color-scheme

If this feature is implemented, say a site has 100 pages and 98 of those use dark mode but 2 use light mode, that site can be added to the list of sites supporting dark mode and the 2 pages can be added to the list of sites using light mode.

## Screenshots

None

## Additional Context

None, already explained in the Feature Request section

Gusted:
> If this feature is implemented, say a site has 100 pages and 98 of those use dark mode but 2 use light mode, that site can be added to the list of sites supporting dark mode and the 2 pages can be added to the list of sites using light mode.

I don't see any real-life example other than github.com being mentioned. And github.com doesn't even qualify for this as they have by default a dark-theme. 



> This would make sense for sites like `github.com` which has dark mode all over the site except some pages like `github.com/marketplace` This way `github.com` can be added to the list of dark mode supported sites and `github.com/marketplace` can be added to the list of sites using light mode (at least until GitHub updates GitHub Marketplace to use a dark UI smile) so that Dark Reader applied the required styles only on that specific page.

See #2517 which would allow to disable specific pages. With `github.com/marketplace$`, other than that it will solve a problem that doesn't exist in my eyes.

PR #6366, opened 2021-07-28
Added GitHub to the list of sites with dark mode

AnishDe12020:
I am also aware of the fact that some GitHub sites like GitHub marketplace (`github.com/marketplace`) still use light mode and hence have propose a solution.
See - #6365

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -329,6 +329,7 @@ gifrun.com
 gifyourgame.com
 gikken.co
 giphy.com
+github.com
 githubuniverse.com
 gitkraken.com
 gitmoji.kaki87.net


Gusted:
And it still doesn't meet the requirements. And we don't add things for "proposed" ideas. Also this isn't a dark-by-default which means regardless of any preference that the user has, the site must render a dark theme, which github.com isn't.

AnishDe12020:
> And it still doesn't meet the requirements. And we don't add things for "proposed" ideas. Also this isn't a dark-by-default which means regardless of any preference that the user has, the site must render a dark theme, which github.com isn't.

GitHub does default to the system default which is fine right? I get the thing about proposed idea though

Gusted:
> GitHub does default to the system default which is fine right?

No, that isn't fine, as stated in the contributing.md "regardless of the system's preferred color-scheme", so if the user wants light theme by the system. It shouldn't render a light theme.

PR #6367, opened 2021-07-28
Fix for mail.google.com (gmail)

rakleed:
- Invert category icons.
- Darken the background of the pop-up side menu.
- Remove obsolete rule.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -7473,7 +7473,6 @@ img.Hk
 img.Ha
 .asor_t0
 .asor_i4
-.ita-icon-0
 .d-Na-Jo.d-Na-N-ax3
 .RK-QJ-Jk
 .RK-Mo.RK-Qq-LF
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -7509,6 +7508,12 @@ CSS
 .buj {
     background-image: url(//ssl.gstatic.com/ui/v1/icons/mail/rfr/density_compact_v1_1x.png) !important;
 }
+div[class*="bym"][role="navigation"] {
+    background-color: rgb(25 27 28) !important;
```
Gusted, line 7512: Could we something that isn't hardcoded?

src/config/dynamic-theme-fixes.config
```diff
@@ -7509,6 +7508,12 @@ CSS
...
+}
+.byl .nZ.aBP {
+    background-color: rgba(255,255,255,0.05) !important;
```
Gusted, line 7515: For this one as well.

src/config/dynamic-theme-fixes.config
rakleed: Fixed.

src/config/dynamic-theme-fixes.config
rakleed: The closest to the current version I found the variable `--lt-color-gray-700: #2F3237;`, but still it turns out not a gray shade, but a brown one, and it looks worse than the current version. Therefore, it is better to leave it as it is.

Gusted (approved):

PR #6385, opened 2021-07-29
Add giggl.app & phineas.io to the dark-sites list

Phineas:
Both giggl.app & phineas.io are dark-themed, regardless of the system theme.


cnrad (approved):
I've visited these sites and I can vouch that they are dark themed by default, darkreader only messes with them further.

Hexiro (approved):
I've visited these sites and I can vouch that they are dark themed, 👍🏼 😊

Gusted (approved):
LGTM

PR #6387, opened 2021-07-30
Use crypto.randomUUID() in Utils UID

bershanskiy:
Chrome 92 added support for `crypto.randomUUID()` function which returns string representation of a Universally Unique Identifiers (UUID) based on RFC4122. This new method is approximatelly 2x faster than the old one based on `crypto.getRandomValues()` and manual hex encoding (when measuring time required for generating 10000 strings).

Note: this method leads to reduced entropy of the resulting string (by 6 bits) because `crypto.randomUUID()` produces UUID with predefined version identifiers. I don't think this entropy reduction is critical given the overall entropy of the string and the way this function is used.

(Of course, `generateUID()` isn't critical for performance, but it would be nice to use the proper API instead of manual byte manipulations.)

## Sources
- [Chrome Platform status entry](https://www.chromestatus.com/feature/5689159362543616)
- [Specification draft](https://wicg.github.io/uuid/)

### DIFF
--- a/src/utils/uid.ts
+++ b/src/utils/uid.ts
@@ -3,9 +3,9 @@ function hexify(number: number) {
 }
 
 export function generateUID() {
-    // TODO: remove any cast once types are updated
+    // TODO: remove any cast once type declarations are updated
     if ((crypto as any).randomUUID) {
-        return (crypto as any).randomUUID().replaceAll('-', ' ');
+        return (crypto as any).randomUUID().replaceAll('-', '');
     }
 
     return Array.from(crypto.getRandomValues(new Uint8Array(16))).map((x) => hexify(x)).join('');
--- a/src/utils/uid.ts
+++ b/src/utils/uid.ts
@@ -3,5 +3,10 @@ function hexify(number: number) {
 }
 
 export function generateUID() {
+    // TODO: remove any cast once types are updated
+    if ((crypto as any).randomUUID) {
+        return (crypto as any).randomUUID().replaceAll('-', ' ');
+    }
+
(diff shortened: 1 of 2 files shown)


src/utils/uid.ts
```diff
@@ -3,5 +3,10 @@ function hexify(number: number) {
 }
 
 export function generateUID() {
+    // TODO: remove any cast once types are updated
+    if ((crypto as any).randomUUID) {
+        return (crypto as any).randomUUID().replaceAll('-', ' ');
```
alexanderby, line 8: Did you mean to replace "-" with "" (remove dashes)?

alexanderby (changes requested):

src/utils/uid.ts
bershanskiy: Yes, I did. I'll add unit tests for UID utils.

tests/utils/uid.tests.ts
```diff
@@ -0,0 +1,32 @@
...
+    globalThis.crypto = {
+        randomUUID: () => {
+            shim2();
+            return 'a19cc926-bf7f-4d5f-bf9c-202bc7a4c7c6';
+        },
+    } as any;
+    expect(generateUID()).toEqual('a19cc926bf7f4d5fbf9c202bc7a4c7c6');
+    expect(shim2).toHaveBeenCalled();
+
+    globalThis.crypto = undefined;
```
alexanderby, line 31: Not sure if it will break anything, but use `delete globalThis.crypto` to delete the property, otherwise `'crypto' in globalThis` will return `true`.

alexanderby (approved):

bershanskiy:
Rebased to make sure patch applies cleanly.

PR #6392, opened 2021-07-30
Disabling Dark Mode for jadenjsj.ddns.net

JadenJSJ:
It would be very good if you can add a meta tag to disable darkreader, instead of needing to add to a list.


Gusted:
There are some open issues/PR's for that matter.

Gusted:
Im not sure what happended but I cannot connect at any point to the server.

PR #6398, opened 2021-07-30
Fix Salesforce docs tabs container

roko-p:
example page to test this: https://developer.salesforce.com/docs/component-library/bundle/lightning-progress-step/documentation


Gusted (approved):
Thank you.

PR #6405, opened 2021-07-31
Follow-up to "Use Alarms API instead of intervals in automation"

bershanskiy:
This PR addresses comments by @alexanderby at #6347.

Specifically:

[Source](https://github.com/darkreader/darkreader/pull/6347#discussion_r679990699)

Fix: Split out `nextIntervalTime` into a separate exported function.

## Comment 2

> Maybe you could remove `isEnabledNow`? Is it an optimisation? I think it makes code harder to understand (messes with `isEnabled`) and harder to keep track of its state. Or at least rename to something like `isEnabledByAutomation`.

[Source](https://github.com/darkreader/darkreader/pull/6347#discussion_r679983368)

Fix: This is an important optimisation, I think, especially for when extension is toggled ON on power-constrained devices. I changed `isEnabledNow` to cover all cases, not just automation. Now the name matches the behavior.

### DIFF
--- a/src/background/extension.ts
+++ b/src/background/extension.ts
@@ -8,7 +8,7 @@ import TabManager from './tab-manager';
 import UserStorage from './user-storage';
 import {setWindowTheme, resetWindowTheme} from './window-theme';
 import {getFontList, getCommands, setShortcut, canInjectScript} from './utils/extension-api';
-import {isInTimeInterval, nextIntervalTime, isNightAtLocation} from '../utils/time';
+import {isInTimeInterval, nextIntervalTime, isNightAtLocation, nextSunriseOrSunset} from '../utils/time';
 import {isURLInList, getURLHostOrProtocol, isURLEnabled, isPDF} from '../utils/url';
 import ThemeEngines from '../generators/theme-engines';
 import createCSSFilterStylesheet from '../generators/css-filter';
@@ -87,9 +87,8 @@ export class Extension {
                 const {latitude, longitude} = this.user.settings.location;
 
                 if (latitude != null && longitude != null) {
-                    const info = isNightAtLocation(latitude, longitude);
-                    this.isEnabledCached = info.rightNow;
-                    nextCheck = info.nextCheck;
+                    this.isEnabledCached = isNightAtLocation(latitude, longitude);
+                    nextCheck = nextSunriseOrSunset(latitude, longitude);
                 }
                 break;
             }
--- a/src/background/extension.ts
+++ b/src/background/extension.ts
@@ -8,7 +8,7 @@ import TabManager from './tab-manager';
 import UserStorage from './user-storage';
 import {setWindowTheme, resetWindowTheme} from './window-theme';
 import {getFontList, getCommands, setShortcut, canInjectScript} from './utils/extension-api';
-import {isInTimeInterval, nextIntervalTime, isNightAtLocation, nextSunriseOrSunset} from '../utils/time';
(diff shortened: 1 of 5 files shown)


Gusted (approved):
LGTM, you can merge it yourself now 😄

bershanskiy:
I'll wait for input from @alexanderby :)

src/utils/time.ts
```diff
@@ -222,7 +222,10 @@ export function isNightAtLocation(
     latitude: number,
     longitude: number,
     date: Date = new Date(),
-): TimeCheck {
+): {
```
alexanderby, line 225: Could you please make `isNightAtLocation` to return boolean only?

Also, maybe if you try adding two alarms, one for the start (sunset) time, and another for end (sunrise time), you will not have to check for `nextCheck` every time.

In future it was planned to add an ability to switch between dark/light color scheme. Who knows, maybe we will consider adding more automation options, time intervals in future.

alexanderby (changes requested):

src/background/extension.ts
```diff
@@ -67,36 +67,40 @@ export class Extension {
...
             case 'time':
-                timingInformation = isInTimeInterval(this.user.settings.time.activation, this.user.settings.time.deactivation);
+                this.isEnabledNow = isInTimeInterval(this.user.settings.time.activation, this.user.settings.time.deactivation);
+                nextCheck = nextIntervalTime(this.user.settings.time.activation, this.user.settings.time.deactivation);
                 break;
             case 'system':
                 if (isFirefox) {
                     // BUG: Firefox background page always matches initial color scheme.
-                    return this.wasLastColorSchemeDark == null
+                    this.isEnabledNow = this.wasLastColorSchemeDark == null
```
alexanderby, line 79: Maybe let's rename `isEnabledNow`? Imagine someone sees `isEnabled` and `isEnabledNow` in code. He will not guess how do they differ.

bershanskiy: What should the new name be? Does `this.isEnabledCached` convey the way it is used?

src/utils/time.ts
bershanskiy: > Could you please make isNightAtLocation to return boolean only?

The logic deciding whether it is no=ight or day right now and time of the next check both need results of `getSunsetSunriseUTCTime()`, so I can move call to `getSunsetSunriseUTCTime()` out of `isNightAtLocation` and then pass result into `isNightAtLocation` and funcion determining next time.

> Also, maybe if you try adding two alarms, one for the start (sunset) time, and another for end (sunrise time), you will not have to check for nextCheck every time.

These would still have to be one-time allarms (as opposed to periodic ones) because sunset/sunrise drift slightly during the year.

alexanderby: The problem is, the answer for the question `is night at location` should be only `yes` or `no` (`true` or `false`). By looking at `rightNow` property name someone can guess its type is Date.

If it's not possible to just return boolean, then rename the function some way, like `getLocationSunInfo` or similar, that will return `{isNight: boolean; nextSunsetOrSunriseTime: Date}`.

src/background/extension.ts
alexanderby: I think yes, `isEnabledCached` is fine.

bershanskiy: Done!

alexanderby (approved):

alexanderby:
Seems like conflicts should be resolved.

PR #6419, opened 2021-08-04
Fix for classroom.google.com

Elaborendum:
1. For the YouTube Icon that appears when watching a YT video within GC, I'm unsure of the result. 
imagen
2. The icons of the options of the 3 dot menu of a file preview. Just used #5930 here, which was meant to fix the same issue, but in Drive.
3. Removed line that caused the icons in the file preview to look misplaced, like this: 
imagen


Gusted (approved):
LGTM, thank you!

PR #6434, opened 2021-08-06
Remove two dead functions in Connector

bershanskiy:
These two methods are not used, instead, UI uses `changeSettings` method:  https://github.com/darkreader/darkreader/blob/master/src/ui/connect/connector.ts#L84

 - These is no receiving end for messages sent by these functions.
 - Nothing happens if I try to emulate these functions by running `chrome.runtime.sendMessage({from: 'ui', type:'enable'})` (or disable)

### DIFF
--- a/src/ui/connect/connector.ts
+++ b/src/ui/connect/connector.ts
@@ -69,14 +69,6 @@ export default class Connector implements ExtensionActions {
         }
     }
 
-    enable() {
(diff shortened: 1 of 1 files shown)


bershanskiy:
@alexanderby Could you look over this PR to make sure I'm not accidentally breaking something?

alexanderby (approved):

PR #6435, opened 2021-08-06
Create MessageType enum

bershanskiy:
This PR changes the content of `type` attribute in messages from an arbitrary string to a member of enum `MessageType`. It also removes the need for `from` attribute.


src/api/chrome.ts
```diff
@@ -22,10 +23,10 @@ async function sendMessage(...args) {
             } else {
                 text = await response.text();
             }
-            messageListeners.forEach((cb) => cb({type: 'fetch-response', data: text, error: null, id}));
+            messageListeners.forEach((cb) => cb({type: MessageType.BACKGROUND_FETCH_RESPONSE, data: text, error: null, id}));
```
alexanderby, line 26: Hmm, `BACKGROUND_FETCH_RESPONSE` is missing in this enum.

alexanderby (changes requested):
Take a look at ESLint warnings, seems like it doesn't like `_` in `[_, {}]`, but `[, {}]` is also odd. Maybe there's a way to configure this array destructuring with ESLint https://eslint.org/docs/rules/prefer-destructuring

src/api/chrome.ts
bershanskiy: Good catch. I'm alarmed that compiler didn't compalin during type checking. I fixed this and looked over all other changes to make sure everything is correct. 

While looking at it, I spotted the following on line 28:

https://github.com/darkreader/darkreader/blob/4ade4bb543b61b97bd8f1b0b68d10226fbc32812/src/api/chrome.ts#L28
Is this correct? I think instead of `err` it should be `error: err` or something like that because `Message` type does not have `err` but has `error`.

bershanskiy: I filed https://github.com/darkreader/darkreader/pull/6467 for that.

src/utils/message.ts
```diff
@@ -0,0 +1,80 @@
+export const MessageType: {[type: string]: number} = {
```
alexanderby, line 1: Maybe take a look at some [alternatives](https://maxheiber.medium.com/alternatives-to-typescript-enums-50e4c16600b1) to enums.
[Record Type](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeystype)

If you just remove the type it will be more type safe:
```typescript
export const MessageType = {
    UI_GET_DATA: 0,
};
```

src/utils/message.ts
```diff
@@ -0,0 +1,80 @@
...
+    CS_FRAME_RESUME: 30,
+    CS_EXPORT_CSS_RESPONSE: 31,
+    UI_SAVE_FILE: 32,
+    UI_REQUEST_EXPORT_CSS: 33,
+    CS_FETCH: 34,
+};
+
+// This map exists only for debugging convenience and is not used in the final build output.
+export const MessageTypeDescriptions = {
+    [MessageType.UI_GET_DATA]: 'ui-get-data',
```
alexanderby, line 43: Why don't you use strings, but numbers in the "enum"? Is that a micro optimisation? Come on :)

alexanderby (approved):
You can leave it as is, just throwing some ideas.

src/utils/message.ts
bershanskiy: Done!

bershanskiy:
Let's get this merged! :)

PR #6441, opened 2021-08-07
Add PromiseBarrier utility

bershanskiy:
~~I'll add tests once we reach agreement on this class API.~~ This is ready for review.

Usage:

```
const barrier = new PromiseBarrier();

async function one(){
  await barrier.entry();
  console.log('One');
}

async function two(){
  await barrier.entry();
  console.log('Two');
}

one();
two();
console.log('This show up first');
barrier.resolve();
```


src/background/utils/promise-barrier.ts
```diff
@@ -0,0 +1,32 @@
+export class PromiseBarrier {
+    private resolves = [];
+    private rejects = [];
```
Gusted, line 3:
```suggestion
    private rejects: Array<Promise<void>> = [];
```

src/background/utils/promise-barrier.ts
```diff
@@ -0,0 +1,32 @@
...
+        if (this.wasResolved) {
+            return Promise.resolve();
+        }
+        return new Promise((resolve, reject) => {
+            this.resolves.push(resolve);
+            this.rejects.push(reject);
+        });
+    }
+
+    resolve(){
```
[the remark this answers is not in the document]
bershanskiy: Done!

src/background/utils/promise-barrier.ts
bershanskiy: Done!

Since this is an array of `reject` arguments from `Promise` constructor (as opposed to array of Promises), it I made it `Array<(value?: any) => void>`.

bershanskiy:
This is ready for review.

src/utils/promise-barrier.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    async resolve(value?: any){
+        if (this.wasRejected || this.wasResolved) {
+            return;
+        }
+        this.wasResolved = true;
+        this.resolution = value;
+        this.resolves.forEach((resolve) => resolve(value));
+        return new Promise<void>((resolve) => setTimeout(() => resolve()));
```
Gusted, line 28:
```suggestion
        return Promise.all(this.resolves);
```

src/utils/promise-barrier.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    async reject(reason?: any){
+        if (this.wasRejected || this.wasResolved) {
+            return;
+        }
+        this.wasRejected = true;
+        this.resolution = reason;
+        this.rejects.forEach((reject) => reject(reason));
+        return new Promise<void>((resolve) => setTimeout(() => resolve()));
```
Gusted, line 38:
```suggestion
        return Promise.all(this.rejects);
```

Gusted (commented):
Promise.all does the job for you ;)

bershanskiy:
`Promise.all` will not do anything usefull if we bass a function to it. `Promise.all` wil just look at a function and say "it's not a Promise, I can just return it".

So like this:
```
const f = () => console.log('Will not print'); // does not log anything because f is never called
(await Promise.all([f]))[0] === f // true
```

Gusted:
Woops that's correct.

PR #6444, opened 2021-08-08
Fix leetcode's tree visualizer node color

Ashesh3:
The background color property no long works as the circles are colored by fill which supersedes the background color, hence nodes appear as white along with the text inside being white. This makes the node unreadable. This commit fixes the node visualizer by making it readable.

Previous: 
image

Now:
image


Gusted (approved):
Thank you!

PR #6448, opened 2021-08-09
Remove outdated CSS for instagram.com

felixonmars:
The mentioned classes are not found on the webpage anymore. These were added back in https://github.com/darkreader/darkreader/pull/2045, I think eventually they regenerated the names.


Gusted (approved):
LGTM

PR #6475, opened 2021-08-11
Network utils: add fallback to fetch()

bershanskiy:


src/background/utils/network.ts
```diff
@@ -9,22 +9,56 @@ interface RequestParams {
...
     return new Promise((resolve, reject) => {
-        const request = new XMLHttpRequest();
-        request.overrideMimeType('text/plain');
-        request.open('GET', params.url, true);
-        request.onload = () => {
-            if (request.status >= 200 && request.status < 300) {
-                resolve(request.responseText);
-            } else {
-                reject(new Error(`${request.status}: ${request.statusText}`));
+        if (XMLHttpRequest) {
```
Gusted, line 12: Can we move these 2 detections to `utils/platform.ts`

In the form of `const is IsXMLHttpRequestSupported = typeof XMLHttpRequest === 'function'` and `typeof isFetchSupported = typeof fetch === 'function'`.

Also should it by default prefer fetch over XMLHttpRequest?

Gusted: Oops, too late 🤷🏽‍♂️

bershanskiy: Oh, I rushed and missed your comments by one minute. I'll create a follow-up for this.

bershanskiy: Here it is: #6486

bershanskiy:
RE to @Gusted:

> Also should it by default prefer fetch over XMLHttpRequest?

I decided to use `XMLHttpRequest` when available because browsers are weird about abort signals in `fetch`. Specifically, some versions of Safari can accept abort signal, but then just ignore when you actually abort it. If we use `fetch` as a fallback, we know that right now it will be used only in service workers in Chrome MV3 build and when other browsers inplement MV3 they will already support `fetch` properly.

Also, I resorted to using an extra boolean `timedOut` because browsers throw different errors when you abort a `fetch` request.

Gusted:
> Specifically, some versions of Safari can accept abort signal, but then just ignore when you actually abort it

Ah, great. Always damn safari.

PR #6491, opened 2021-08-12
Fix for baidu.com

Gusted:
- Resolves #6480

PR #6496, opened 2021-08-12
Fix for ditu.baidu.com

felixonmars:
- A lot of fixes after #6491 fixes the rendering issue.


Gusted (approved):
All looks goods to me

PR #6497, opened 2021-08-12
Add dark sites

jasonhargrove:


src/config/dark-sites.config
```diff
@@ -479,6 +480,7 @@ m4rtyr.github.io
 macbb.org
 magnetoo.io
 magoware.tv
+mainstre.am
```
Gusted, line 483: Doesn't seem that dark mode to me.

image

PR #6506, opened 2021-08-13
Manifest V3 build

bershanskiy:
This PR adds a new build output, `build/*/chrome-mv3/`, which uses Manifest V3 format. This build provides basic functionality, but there are still challenges which need to be resolved:

 - Dark Reader's current script/style injection logic is not considered secure by Chrome's standards and therefore script/style injectionsfail
 - Service workers do not have access to `window` object and therefore Automation CSS media query fails
 - There is a terrible power use spike with this build, my laptop overheats when I install this build

### DIFF
--- a/src/inject/dynamic-theme/index.ts
+++ b/src/inject/dynamic-theme/index.ts
@@ -44,9 +44,11 @@ function createOrUpdateStyle(className: string, root: ParentNode = document.head
 }
 
 function createOrUpdateScript(className: string, root: ParentNode = document.head || document) {
-    let element: HTMLScriptElement = root.querySelector(`.${className}`);
+    let element: HTMLElement = root.querySelector(`.${className}`);
     if (!element) {
-        element = document.createElement('script');
+        element = document.createElement('div');
+        const parent = document.createElement('div');
+        parent.appendChild(element);
         element.classList.add('darkreader');
         element.classList.add(className);
     }
@@ -64,6 +66,15 @@ function setupNodePositionWatcher(node: Node, alias: string) {
 function stopStylePositionWatchers() {
     forEach(nodePositionWatchers.values(), (watcher) => watcher.stop());
     nodePositionWatchers.clear();
+}
+
+/*
+ * We cancell propagation of our event before it reaches any other handlers
+ * by creating an extra 'parent' element and calling stopImmediatePropagation()
+ */
+function injectedProxyEventListener(event) {
+    event.stopImmediatePropagation();
+    injectProxy();
 }
 
 function createStaticStyleOverrides() {
@@ -134,13 +145,13 @@ function createStaticStyleOverrides() {
     document.head.insertBefore(rootVarsStyle, variableStyle.nextSibling);
 
     const proxyScript = createOrUpdateScript('darkreader--proxy');
-    const blob = new Blob([`(${injectProxy})()`], {type: 'text/javascript'});
-    const url = URL.createObjectURL(blob);
-    proxyScript.src = url;
     proxyScript.textContent = '';
-    document.head.insertBefore(proxyScript, rootVarsStyle.nextSibling);
-    URL.revokeObjectURL(url);
-    proxyScript.remove();
+    proxyScript.addEventListener('__darkreader__yeet', injectedProxyEventListener);
+    const event = new Event('__darkreader__yeet');
+    document.head.insertBefore(proxyScript.parentElement, rootVarsStyle.nextSibling);
+    proxyScript.dispatchEvent(event);
+    proxyScript.removeEventListener('__darkreader__yeet', injectedProxyEventListener);
+    proxyScript.parentElement.remove();
 }
 
 const shadowRootsWithOverrides = new Set<ShadowRoot>();
--- a/tasks/bundle-js.js
+++ b/tasks/bundle-js.js
@@ -13,8 +13,10 @@ const {createTask} = require('./task');
 async function copyToBrowsers({cwdPath, debug}) {
     const destPath = `${getDestDir({debug, platform: PLATFORM.CHROME})}/${cwdPath}`;
     const ffDestPath = `${getDestDir({debug, platform: PLATFORM.FIREFOX})}/${cwdPath}`;
+    const mv3DestPath = `${getDestDir({debug, platform: PLATFORM.CHROME_MV3})}/${cwdPath}`;
     const tbDestPath = `${getDestDir({debug, platform: PLATFORM.THUNDERBIRD})}/${cwdPath}`;
     await fs.copy(destPath, ffDestPath);
+    await fs.copy(destPath, mv3DestPath);
     await fs.copy(destPath, tbDestPath);
 }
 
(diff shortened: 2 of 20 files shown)


Gusted:
>     * Dark Reader's current script/style injection logic is not considered secure by Chrome's standards and therefore script/style injectionsfail

What are their recommended ways?

bershanskiy:
Google recommends `chrome.scripting` for injecting code into extension's "clean context". But since we need to inect into page's own context, we probably need to use event handler method.

Gusted:
> Google recommends `chrome.scripting` for injecting code into extension's "clean context".

I see, it's more or less the current chrome.tabs.executeScript but you are even limited more. Great :D The current script we are injecting does really need the page's context due to the prototype overriding. 

The style doesn't really matter, as long it works(and can very easily be edited) and overrides the style's CSS. But the script is a must to have the correct page's context.

alexanderby:
If some API is missing we can ask in a user group or file an issue in Chromium bug tracker.

Gusted:
> If some API is missing we can ask in a user group or file an issue in Chromium bug tracker.

That's possible, I don't have really good experience around their bug tracker nor the user groups 🤷🏽. But currently we just add a script into the page's HTML which is currently reliable(with #4977)

bershanskiy:
> If some API is missing we can ask in a user group or file an issue in Chromium bug tracker.

We probably should. But injecting code into page's own context was never officially supported, it was always an afterthought.

tasks/bundle-js.js
```diff
@@ -28,6 +30,19 @@ function patchFirefoxJS(/** @type {string} */code) {
...
 }
 
+function patchMV3JS(/** @type {string} */code) {
+    // MV3 moves a few APIs around
+    code = replace(code, 'chrome.browserAction.setIcon', 'chrome.action.setIcon');
+    code = replace(code, 'chrome.browserAction.setBadgeBackgroundColor', 'chrome.action.setBadgeBackgroundColor');
+    code = replace(code, 'chrome.browserAction.setBadgeText', 'chrome.action.setBadgeText');
+
+    // Background serivice workers do not have access to "foreground" APIs like chrome.i18n
+    code = replace(code, 'chrome.i18n.getUILanguage()', "navigator.language.split('-')[0]");
```
bershanskiy, line 40: Should I move this to `src/utils/locales.ts`?

alexanderby: Let's do this

Gusted:
> We probably should. But injecting code into page's own context was never officially supported, it was always an afterthought.

https://groups.google.com/a/chromium.org/g/chromium-extensions/c/hQeJzPbG-js

alexanderby:
Try applying this #4977 and see if it helps. I was scared to merge it, but let's see if it helps. If so, than let's merge.

Gusted:
> Try applying this #4977 and see if it helps. I was scared to merge it, but let's see if it helps. If so, then let's merge.

It would enable the script in some places, but Manifest v3 is still another issue :)

tasks/bundle-js.js
```diff
@@ -28,6 +30,19 @@ function patchFirefoxJS(/** @type {string} */code) {
...
+    // TODO: Fix me
+    code = replace(code, 'window.matchMedia(query).matches', 'false');
```
bershanskiy, line 42: There are multiple ways to implement CSS Media queries:
 - use content script context
 - spawn a "sandbox" which has an invisible window. This is the cleanest and the most reliable way, but spawning an entire process just to query a boolean value seems wasteful...

alexanderby: Then try doing it in context script

alexanderby:
> > Try applying this #4977 and see if it helps. I was scared to merge it, but let's see if it helps. If so, then let's merge.
> 
> It would enable the script in some places, but Manifest v3 is still another issue :)

I was afraid it could disable it in other places as well :)

Gusted:
> I was afraid it could disable it in other places as well :) Like Firefox.

Nope seems all good, `blob:` is special like normal it will match the `self` directive in CSP which sites like privacytools.io uses and some other places with strict CSP because in that context a blob made by dark reader is somewhat equal to a blob made on that site. It's a weird CSP but it luckily does the trick.

alexanderby:
> > I was afraid it could disable it in other places as well :) Like Firefox.
> 
> Nope seems all good, `blob:` is special like normal it will match the `self` directive in CSP which sites like privacytools.io uses and some other places with strict CSP because in that context a blob made by dark reader is somewhat equal to a blob made on that site. It's a weird CSP but it luckily does the trick.

Well, then merge it.

bershanskiy:
`blob:` is still considered insecure:

> Insecure CSP value "blob:" in directive 'script-src'.

Gusted:
> `blob:` is still considered insecure:

`self` matches any blob: content, but any added scripts to the page's context relies on their CSP not on the extension's one.

alexanderby:
I don't know, if it's not possible, then the fallbacks should work. We can file some issues. E.g. explain that the extensions need a way to watch for cssRules insertions and deletions. Or maybe allowing such scripts (though it seems insecure).

Gusted:
> I don't know, if it's not possible, then the fallbacks should work. We can file some issues. E.g. explain that the extensions need a way to watch for cssRules insertions and deletions. Or maybe allowing such scripts (though it seems insecure).

The current fallback is something to avoid at all times(rAF polluting 60x times a second).

So far I've seen no chromium dev publicly stating(I'd might missed some conversiations) something about this whole manifest v3 mess and how to migrate for extensions with existing API's that will be gone. You might have luck with Firefox which seems to take a better stake at this whole manifest v3.

alexanderby:
> The current fallback is something to avoid at all times(rAF polluting 60x times a second).

It just compares the count of cssRules. It should be fast, but not accurate.

> You might have luck with Firefox which seems to take a better stake at this whole manifest v3.

As far as I understand we can keep using MV2 there

Gusted:
> It just compares the count of cssRules. It should be fast,

It is, until you will aggressively loop it 60x times It it's fast, but those little cpu time added together will become a problem over time, the main reason it was optimized away to have a passive listener as it seriously impacted performance of webpages and the profiler didn't like the function as well.

alexanderby:
> Hmm, didn't they migrate to MV3 but keep comparability with MV2?

https://blog.mozilla.org/addons/2021/05/27/manifest-v3-update/
> We have not yet set a deprecation date for Manifest v2 but expect it to be supported for at least one year after Manifest v3 becomes stable in the release channel.

Gusted:
> We have not yet set a deprecation date for Manifest v2 but expect it to be supported for at least one year after Manifest v3 becomes stable in the release channel.

That gives us some breathing time with firefox.

src/inject/dynamic-theme/index.ts
```diff
@@ -134,13 +141,13 @@ function createStaticStyleOverrides() {
...
 
     const proxyScript = createOrUpdateScript('darkreader--proxy');
-    const blob = new Blob([`(${injectProxy})()`], {type: 'text/javascript'});
-    const url = URL.createObjectURL(blob);
-    proxyScript.src = url;
     proxyScript.textContent = '';
-    document.head.insertBefore(proxyScript, rootVarsStyle.nextSibling);
-    URL.revokeObjectURL(url);
-    proxyScript.remove();
+    proxyScript.addEventListener('yeet', injectedProxyEventListener);
```
Gusted, line 145: Could `yeet` be changed into `__darkreader__yeet`?

bershanskiy:
This is ready for review, but we should merge #6564 bug fix prior to this.

tasks/bundle-js.js
bershanskiy: Done!

alexanderby (approved):

PR #6512, opened 2021-08-14
Fix for tieba.baidu.com

felixonmars:
- Add many more fixes after #6491 landed

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -12125,20 +12125,30 @@ INVERT
 .add-more-forum
 .all-wraper
 .core_title_btns
+.day_rcmd > .class_title
+.fengchao-wrap-box span
 .focus_btn
+img[src*="sign_err.png"]
 .nav_wrap_add_border
 .save_face_bg
 .share_btn_wrapper
+.sub_nav_wrap
+.threadlist_rep_num
+.u-f-item
 
 CSS
+.card_banner {
+    opacity: 50%;
+}
 .class_title,
 .content-sec,
-.f-d-w,
+.f-d-item,
 .left-sec,
+.page-container,
+.pb_footer,
 .sub_nav_wrap,
-.tb_footer,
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -12125,20 +12125,30 @@ INVERT
...
 .nav_wrap_add_border
 .save_face_bg
 .share_btn_wrapper
+.sub_nav_wrap
+.threadlist_rep_num
+.u-f-item
 
 CSS
+.card_banner {
+    opacity: 50%;
```
Gusted, line 12141: Hmm interesting, what is this change for? We mostly don't need to poke around the opacity.

src/config/dynamic-theme-fixes.config
```diff
@@ -12125,20 +12125,30 @@ INVERT
...
+.f-d-item,
 .left-sec,
+.page-container,
+.pb_footer,
 .sub_nav_wrap,
-.tb_footer,
-.u-f-item {
-    background-color: var(--darkreader-neutral-background) !important;
+.tb_footer {
+    background: none !important;
```
Gusted, line 12151:
```suggestion
    background-image: none !important;
```
Specify it a bit further.

src/config/dynamic-theme-fixes.config
felixonmars: Most banners are too bright for dark theme. Perhaps I should implement it in a different way?

Gusted: Well, we currently uses `filter: brightness(50%) sepia(40%) !important;` to ensure the brightness is down and add  bit of sepia over it to make it more natural.

felixonmars: Looks like turning down some brightness is enough here (and looks the same as opacity). Updated.

Gusted (approved):
All good.

PR #6515, opened 2021-08-14
Fix for mumble.info

pnaf:
Fixes "Downloads" page.

Before:

After:


Gusted (approved):
Seems, good. Thank you!

PR #6520, opened 2021-08-15
Fix native MathML elements on MediaWiki

felixonmars:
`.mwe-math-element` has two children: `.mwe-math-mathml-inline` and `.mwe-math-fallback-image-inline`. The first one is native MathML and should not be inverted, while the later should.

Native MathML could be force enabled with this addon: https://addons.mozilla.org/en-US/firefox/addon/native-mathml/

Tested for wikipedia, wikisource, wikibooks with an example from their corresponding commit, each with MathML on and off.

I updated the one in explainxkcd too but I did not find an example there.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4334,7 +4334,7 @@ canvas
 explainxkcd.com
 
 INVERT
-.mwe-math-element
+.mwe-math-fallback-image-inline
 .mw-ext-score
 .main-footer-menuToggle
 img[src*="Loudspeaker.svg"]
@@ -13212,7 +13212,7 @@ wikibooks.org
 
 INVERT
 .bookend
-.mwe-math-element
+.mwe-math-fallback-image-inline
 
 IGNORE INLINE STYLE
 .infobox td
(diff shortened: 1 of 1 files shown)


Gusted (approved):
Seems all good.

andryandrew:
For me (chrome on Mac), this change disables inversion of display math when on fallback (.mwe-math-fallback-image-display). Maybe we should add it for people not using MathML?

Gusted:
> Maybe we should add it for people not using MathML?

I'm not sure what you're talking about, this changes just makes sure it would only target the image and not the MathML elements.

andryandrew:
inverting `.mwe-math-element` used to target both inline and display math, but this commit only restores inversion for `mwe-math-fallback-image-inline` and not `mwe-math-fallback-image-display`

Gusted:
> inverting `.mwe-math-element` used to target both inline and display math, but this commit only restores inversion for `mwe-math-fallback-image-inline` and not `mwe-math-fallback-image-display`

Uhm, again what? I don't see any `mwe-math-fallback-image-display` elements on sites like wikipedia. But we could add it if people see such classes.

Gusted:
#6958

andryandrew:
Thank you! I see them, for example the display equations here: https://en.wikipedia.org/wiki/Marginal_distribution

PR #6524, opened 2021-08-15
Add more workarounds for thunderbird

### LINKED ISSUE #6521: [Bug] No longer working with Thunderbird 91
# Bug Report

## Bug Description
With the release of Thunderbird 91, Dark Reader is no longer working, despite the add-on being marked as installed.

## Steps To Reproduce

- Install Thunderbird 91
- Setup an account
- Install Dark Reader from the Thunderbird add-ons page

## Expected Behavior

- Dark Reader should load properly as in Thunderbird 78
- 

## Actual Behavior

- Dark Reader shows as installed, but it's stuck in "loading"
- 
- No e-mails are shown in dark mode

## System Information

- Thunderbird version 91
- Dark Reader version 4.9.32
- Windows 10 x64 / Ubuntu 20.04 LTS

Gusted:
- Resolves #6521

@bershanskiy I've disabled the pagehide event for now, as That was erroring(error is currently included in the comment) and dark reader seems to still work when disabled, if you could check https://thunderbird.topicbox.com/groups/addons/T90f2914acf15917b/mx-fails-in-91-promise-rejected-after-context-unloaded to see why this *might* happen would be nice. But like I said in my comment, their are no tabs, so once the tab is hidden, it means it's gone and another email is opened or something otherwise.

@alexanderby If you read this, that probably means on a monday so I won't be able to respond in any quickly time manner, but it will be nice to have a hotfix deployed for thunderbird as mentioned in #6521 it currently doesn't work at all and v91 is currently rolling out(as beta-ish?)

Regards,
Gusted

### DIFF
--- a/src/background/tab-manager.ts
+++ b/src/background/tab-manager.ts
@@ -76,7 +76,9 @@ export default class TabManager {
                         }
                     };
 
-                    const isPanel = sender.tab == null;
+                    // Workaround for thunderbird, not sure how. But sometimes sender.tab is undefined but accessing it.
+                    // Will actually throw a very nice error.
+                    const isPanel = typeof sender === 'undefined' || typeof sender.tab === 'undefined';
                     if (isPanel) {
                         // NOTE: Vivaldi and Opera can show a page in a side panel,
                         // but it is not possible to handle messaging correctly (no tab ID, frame ID).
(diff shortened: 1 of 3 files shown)


bershanskiy:
> I've disabled the pagehide event for now ... But like I said in my comment, their are no tabs, so once the tab is hidden, it means it's gone and another email is opened or something otherwise.

Thanks for letting me know.

> https://thunderbird.topicbox.com/groups/addons/T90f2914acf15917b/mx-fails-in-91-promise-rejected-after-context-unloaded

I'll look at it in greated detail later today. I'm aware that Mozilla codebase has some race conditions and they manifest in curious ways in Firefox and Thunderbird. We already had something similar in #6113 and there we worked around the bug by using a persistent port. More workarounds, I guess.

PR #6528, opened 2021-08-16
Add inventory-profiles-next.github.io as already dark

victorbnl:
Note: the site is dark by default, then checks whether the system theme is dark or light via media queries, so if you go on the site and see it's light, it might be your browser's theme


Gusted:
Hi!

Thank your for contribution, 
unfortunately this website doesn't meet our [standards](https://github.com/darkreader/darkreader/blob/master/CONTRIBUTING.md#adding-a-website-that-is-already-dark) that is required to be added to the list.

Regards,
Gusted

PR #6529, opened 2021-08-16
Fix for todoist.com

rakleed:
- Invert logo.

Example of site page: https://todoist.com/help

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -12202,7 +12202,7 @@ todoist.com
 
 INVERT
 .ist_button_apple > img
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -12202,6 +12202,7 @@ todoist.com
 
 INVERT
 .ist_button_apple > img
+path[d^="M166.805 16.694"]
```
Gusted, line 12205: What about something like:
```suggestion
#td-help-logo_svg__a ~ g:last-of-type
```

Gusted (approved):

PR #6536, opened 2021-08-18
bitwarden.com fixes

ZOASR:


src/config/dynamic-theme-fixes.config
```diff
@@ -1739,6 +1739,19 @@ CSS
 
 ================================
 
+bitwarden.com
+
+INVERT
+img[alt="Github Logo"]
+body > div > section:nth-child(4) > div > div:nth-child(1) > div > blockquote > footer > cite > img
```
Gusted, line 1746: What about to simplify it?
```suggestion
img[src$="forbes.png"]
```

src/config/dynamic-theme-fixes.config
```diff
@@ -1739,6 +1739,19 @@ CSS
...
+
+CSS
+.blockquote-header{
+     background-image: url(../images/quote-marks.png) !important;
+}
```
Gusted, line 1751:
```suggestion
IGNORE IMAGE ANALYSIS
blockquote .blockquote-header
```

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
ZOASR: I tried `IGNORE IMAGE ANALYSIS` didn't work for me so i resorted to just CSS.

Gusted: You have to apply it and then give the website a hard refresh.

ZOASR: > You have to apply it and then give the website a hard refresh.

I did but didn't work either thats' why i got confused

Gusted: image
does seem to work for me.

ZOASR: Ok now i know why:
i was using refrencing only the class in the selector: `.blockquote-header`
but that should work too right?

Gusted: Not really, the current implementation requires the exact selector.

ZOASR: > Not really, the current implementation requires the exact selector.

Ok, got it. 
Thanks!.

Gusted (approved):

PR #6541, opened 2021-08-18
Fix for iqiyi.com

felixonmars:
- Fix background of user info popup

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -6547,6 +6547,13 @@ img[alt="ipla.tv"]
 
 ================================
 
+iqiyi.com
+
+INVERT
+.qy-micro-bg
+
+================================
+
 isbgpsafeyet.com
 
 CSS


Gusted (commented):
"background of user info popup", hmm is this like a `div` acting as a background model? As in such case I don't think a INVERT should be used. Could you point me to where I can find this element?

felixonmars:
> "background of user info popup", hmm is this like a `div` acting as a background model? As in such case I don't think a INVERT should be used. Could you point me to where I can find this element?

It is an <i> with a background-image, acting like a background. If I unset the background-image here, it will become transparent and makes the text unreadable.

It's shown at the top of the popup window when hovering over your avatar after login.

Gusted (approved):
All good then.

felixonmars:
No fix:
DeepinRuudunkaappaus_valitse-alue_20210820213644

Unset background-image:
DeepinRuudunkaappaus_valitse-alue_20210820213659

Invert:
DeepinRuudunkaappaus_valitse-alue_20210820213713

PR #6551, opened 2021-08-20
Added fix for watchasian.in

SaruDesu:
Fixed text color.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -12952,6 +12952,13 @@ a[data-sc-c="headerlogo"] path
 
 ================================
 
+watchasian.in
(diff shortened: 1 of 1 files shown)


Gusted:
Hi @SaruDesu!

Thank you for your time and effort for this contribution. I cannot accept this PR as I don't see anything wrong on this site and as it's using a global selector `*` which is something we discourage to use in the fixes. Please turn it into a issue to explain what is going wrong.

Regards,
Gusted

SaruDesu:
@Gusted 

I created an issue regarding this. Thank you!

SaruDesu:
I just noticed that if I use selectors `div` or `body` it will still fix the issue.

PR #6554, opened 2021-08-20
Refactor: do not store frame URL of and do not send messages to inactive tabs

bershanskiy:
Store URLs only for frames which are either "active" or "passive" in Page Life API terminology. Also, differ sending messages to these frames until they become "active" or "passive". This serves three goals:
 - do less work overall. When user has Dev Tools open and tries various fixes and saves them regularly, extension used to send a message to every frame on every incrimental change. Now it will send messages only to frames which are potentially visible to the user.
 - do less work at one time. When user has extension OFF, and then toggles it ON, Dark Reader sends a message to every frame all at once.
 - store less data about frames which are not visible to the user. Do not store URL of these frames. As mentioned earlier, browsers limit amount of data which can be stored without `unlimitedStorage` permission (5MB). It was theoretically posible to exceed the limit by hoarding lots of tabs with long URLs. Now, such hoarding is even harder.

-------------------------

~~This permission does not display any warnings for user, does not cause Chrome to disable extension upon update, and is not even displayed anywhere in the browser UI. Pinging @alexanderby and @Gusted to sign-off on this.~~

~~I tested this manually. I packaged multiple versions of this extension: (1) one from current `master`; (2) another with `unlimitedStorage` permission, and (3) another with a more powerful permission which is known to throw more warnings (`management`). As expected, upon uprading from 1 to 3 there was a warning and extension was disabled (which validates this testing method). Upon upgrading from 1 to 2 there were no warnings or any interruptions at all.~~

~~Docs (apply to MV2 and MV3 despite mv3 in the URL): https://developer.chrome.com/docs/extensions/mv3/permission_warnings/~~

### DIFF
--- a/src/background/extension.ts
+++ b/src/background/extension.ts
@@ -350,7 +350,7 @@ export class Extension {
         if (this.wasEnabledOnLastCheck === null || this.wasEnabledOnLastCheck !== isEnabled) {
             this.wasEnabledOnLastCheck = isEnabled;
             this.onAppToggle();
-            this.tabs.invalidateTimestamp();
+            this.tabs.sendMessage();
             this.reportChanges();
             this.stateManager.saveState();
         }
@@ -469,7 +469,7 @@ export class Extension {
         }
         await this.stateManager.loadState();
         this.wasEnabledOnLastCheck = this.isEnabled;
-        this.tabs.invalidateTimestamp();
+        this.tabs.sendMessage();
         this.saveUserSettings();
         this.reportChanges();
         this.stateManager.saveState();
--- a/src/background/extension.ts
+++ b/src/background/extension.ts
@@ -61,7 +61,8 @@ export class Extension {
                 }
                 return this.getConnectionMessage(url, frameURL);
             },
-            onColorSchemeChange: this.onColorSchemeChange,
+            getTabMessage: this.getTabMessage,
+            onColorSchemeChange: this.onColorSchemeChange
         });
         this.user = new UserStorage({onRemoteSettingsChange: () => this.onRemoteSettingsChange()});
         this.startBarrier = new PromiseBarrier();
@@ -349,7 +350,7 @@ export class Extension {
         if (this.wasEnabledOnLastCheck === null || this.wasEnabledOnLastCheck !== isEnabled) {
             this.wasEnabledOnLastCheck = isEnabled;
             this.onAppToggle();
-            this.tabs.sendMessage(this.getTabMessage);
+            this.tabs.invalidateTimestamp();
             this.reportChanges();
             this.stateManager.saveState();
         }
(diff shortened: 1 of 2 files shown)


Gusted:
You can add the permission into the manifest.json the Firefox and Thunderbird will override the permissions array anyway.

alexanderby:
What about Firefox, why it should not have this permission?

alexanderby:
Maybe let's move it to Manifest V3? I'm afraid a bit :) I will ask tech support to assure it's OK.

bershanskiy:
> You can add the permission into the manifest.json the Firefox and Thunderbird will override the permissions array anyway.

Good idea.

> What about Firefox, why it should not have this permission?

Mostly because of least priviledge principle: Firefox and Thunderbird don't store much data into storage while Chromium builds theoretically could. Also, there is plenty of documentation about how Chrome decides whether to disable extension on update and I did most testing on Chrome. Firefox and especially Thunderbird have less docs on this specific topic and I didn't reach out to their Matrix groups yet.

alexanderby:
I think you can safely add it to MV3, and I will ask tech support. It's easy to pollute storage with this permission, I feel like it can be revoked some day.

bershanskiy:
> Maybe let's move it to Manifest V3? I'm afraid a bit :) I will ask tech support to assure it's OK.

OK, makes sense.

src/background/extension.ts
```diff
@@ -61,7 +61,8 @@ export class Extension {
                 }
                 return this.getConnectionMessage(url, frameURL);
             },
-            onColorSchemeChange: this.onColorSchemeChange,
+            getTabMessage: this.getTabMessage,
+            onColorSchemeChange: this.onColorSchemeChange
```
alexanderby, line 65: It's OK to leave a comma, it's often used to not highlight unnecessary changes in search results.

src/background/tab-manager.ts
```diff
@@ -190,6 +214,27 @@ export default class TabManager {
         });
     }
 
+    async invalidateTimestamp() {
```
alexanderby, line 217: It looks like the purpose of this function is to send a message, but not just update a timestamp. Could you put it back to see what have changed?

src/background/tab-manager.ts
```diff
@@ -50,22 +53,29 @@ export default class TabManager {
...
                     frames = tabs[tabId];
                 } else {
                     frames = {};
                     tabs[tabId] = frames;
                 }
-                frames[frameId] = {url: senderURL, state: DocumentState.ACTIVE};
+                frames[frameId] = {
+                    url: senderURL,
+                    state: DocumentState.ACTIVE,
+                    timestamp: timestamp,
```
alexanderby, line 77: Can be just `timestamp,`

alexanderby (changes requested):

src/background/tab-manager.ts
bershanskiy: I renamed it back to `sendMessage()` and moved it in old location.

alexanderby (approved):
Maybe I don't get something, but do you still store the frame URL?

bershanskiy:
We store frame URL only for pages which are "active" (have user focus) or "passive" (do not have user focus, but might be visible to the user). When page looses this status, we delete frame URL but keep info about frame hierarchy. When frame comes back to life( becomes "active" or "passive" again), we get a message with frame URL and can insert it back into data structure.

PR #6566, opened 2021-08-21
Indicate when extension does not have proper access

### LINKED ISSUE #6538: [Feature Request] Indication when extension does not have permission to access local URLs
# Feature Request
UI should indicate when extension does not have "local URLs" permission.

People get confused when they open local PDF files and Dark Reader does not work on them despite being enabled. This usually happens because extension does not have proper permission to "open local URLs" (not granted by default).

## Screenshots

image

## Additional Context

To test this:
 - Open Chrome and install extension from store
 - Open extension details at `chrome://extensions/?id=eimadpbcbfnmbkopoojfekhnkhdbieeh` and make sure "local URLs" toggle is off (default)
 - Open a local PDF
 - Notice that Dark Reader disaplays "PDF" with a checkmark, but page is not inverted

bershanskiy:
Fixes #6538.

Relevant section of the popup UI:
image

May be, a more specific message for local files is more appropriate. It could even take user directly to the browser page with the relevant toggle.

@Gusted What do you think of this?

### DIFF
--- a/src/background/extension.ts
+++ b/src/background/extension.ts
@@ -154,7 +154,9 @@ export class Extension {
                     await new Promise<void>((resolve) => this.awaiting.push(resolve));
                 }
                 const url = await this.tabs.getActiveTabURL();
-                return this.getURLInfo(url);
+                const info = this.getURLInfo(url);
+                info.isInjected = await this.tabs.canAccessActiveTab();
+                return info;
             },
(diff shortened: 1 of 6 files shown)


Gusted (approved):
Looks good code-wise. Just make sure logic-wise it's still going good.

PR #6594, opened 2021-08-23
Update browser versions for issue template

paniash:


Gusted:
Thank you for your effort, but I'm not really interested to keep this up-to-date as It doesn't serve that big of a purpose as people should have it replaced anyway.

PR #6597, opened 2021-08-23
Fix for old.reddit.com

c-nagy:
Fix for dark video controls on dark background. Tried to get it working on the new reddit layout as well, but the new reddit layout has a ton of UUIDs everywhere that make it hard to granularly control elements.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -9263,6 +9263,15 @@ INVERT
 
 ================================
 
+old.reddit.com
+
+INVERT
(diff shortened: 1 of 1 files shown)


Gusted:
> but the new reddit layout has a ton of UUIDs everywhere that make it hard to granularly control elements.

It's a hit or miss with the new reddit layout. E.g. the header elements are easy to target, but anything beyond that(and some non-post/subreddit pages) is just a no to have a reliable CSS selector.

Gusted (approved):

PR #6647, opened 2021-08-30
Just-in-time parsing of config files

bershanskiy:
As discussed earlier, current parser implementation is suboptimal. This PR aims to:
 - decrease memory usage by storing only rule offsets for unused rules, and
 - improve performance by creating one more cache for used rules

This optimisation takes advantage of the following observations:
 1. Most users visits only a tiny subset of sites which have configs
 2. Storing two numbers (indecies of rule start and end) is cheaper than storing the whole parsed rule
 
This PR also should improve performance because prior to this `getDynamicThemeFixesFor(...)` calculated `sortedBySpecificity` by iterating over all rules each time. After this commit, it iterates only over url patterns with `*` and looks up exact domain in a map.

This is still a draft because I have a few more questions (below).

Before this patch, heap snapshot was 2.7 MB, after it is 1.9MB.

### DIFF
--- a/src/background/extension.ts
+++ b/src/background/extension.ts
@@ -463,11 +463,11 @@ export class Extension {
                 case ThemeEngines.dynamicTheme: {
                     const filter = {...theme};
                     delete filter.engine;
-                    const fixesNew = getDynamicThemeFixesFor(url, frameURL, this.config.DYNAMIC_THEME_FIXES_RAW, this.config.DYNAMIC_THEME_FIXES_INDEX, this.user.settings.enableForPDF);
+                    const fixes = getDynamicThemeFixesFor(url, frameURL, this.config.DYNAMIC_THEME_FIXES_RAW, this.config.DYNAMIC_THEME_FIXES_INDEX, this.user.settings.enableForPDF);
                     const isIFrame = frameURL != null;
                     return {
                         type: MessageType.BG_ADD_DYNAMIC_THEME,
-                        data: {filter, fixesNew, isIFrame},
+                        data: {filter, fixes, isIFrame},
                     };
                 }
                 default: {
--- a/src/background/extension.ts
+++ b/src/background/extension.ts
@@ -11,7 +11,7 @@ import {getFontList, getCommands, setShortcut, canInjectScript} from './utils/ex
 import {isInTimeIntervalLocal, nextTimeInterval, isNightAtLocation, nextNightAtLocation} from '../utils/time';
 import {isURLInList, getURLHostOrProtocol, isURLEnabled, isPDF} from '../utils/url';
 import ThemeEngines from '../generators/theme-engines';
-import createCSSFilterStylesheet, {parseInversionFixes} from '../generators/css-filter';
+import createCSSFilterStylesheet from '../generators/css-filter';
 import {getDynamicThemeFixesForNew} from '../generators/dynamic-theme';
 import createStaticStylesheet, {parseStaticThemes} from '../generators/static-theme';
 import {createSVGFilterStylesheet, getSVGFilterMatrixValue, getSVGReverseFilterMatrixValue} from '../generators/svg-filter';
@@ -22,7 +22,7 @@ import {MessageType} from '../utils/message';
 import {logInfo, logWarn} from '../utils/log';
 import {PromiseBarrier} from '../utils/promise-barrier';
 
-// TODO: remove parseStaticThemes and parseInversionFixes
+// TODO: remove parseStaticThemes
 
 export class Extension {
     config: ConfigManager;
@@ -421,20 +421,20 @@ export class Extension {
                 case ThemeEngines.cssFilter: {
                     return {
                         type: MessageType.BG_ADD_CSS_FILTER,
-                        data: createCSSFilterStylesheet(theme, url, frameURL, parseInversionFixes(this.config.INVERSION_FIXES_RAW)),
+                        data: createCSSFilterStylesheet(theme, url, frameURL, this.config.INVERSION_FIXES_RAW, this.config.INVERSION_FIXES_INDEX),
                     };
                 }
                 case ThemeEngines.svgFilter: {
                     if (isFirefox) {
                         return {
                             type: MessageType.BG_ADD_CSS_FILTER,
-                            data: createSVGFilterStylesheet(theme, url, frameURL, parseInversionFixes(this.config.INVERSION_FIXES_RAW)),
+                            data: createSVGFilterStylesheet(theme, url, frameURL, this.config.INVERSION_FIXES_RAW, this.config.INVERSION_FIXES_INDEX),
                         };
                     }
                     return {
                         type: MessageType.BG_ADD_SVG_FILTER,
                         data: {
-                            css: createSVGFilterStylesheet(theme, url, frameURL, parseInversionFixes(this.config.INVERSION_FIXES_RAW)),
+                            css: createSVGFilterStylesheet(theme, url, frameURL, this.config.INVERSION_FIXES_RAW, this.config.INVERSION_FIXES_INDEX),
                             svgMatrix: getSVGFilterMatrixValue(theme),
                             svgReverseMatrix: getSVGReverseFilterMatrixValue(),
                         },
--- a/src/background/extension.ts
+++ b/src/background/extension.ts
@@ -12,7 +12,7 @@ import {isInTimeIntervalLocal, nextTimeInterval, isNightAtLocation, nextNightAtL
 import {isURLInList, getURLHostOrProtocol, isURLEnabled, isPDF} from '../utils/url';
 import ThemeEngines from '../generators/theme-engines';
 import createCSSFilterStylesheet from '../generators/css-filter';
-import {getDynamicThemeFixesFor} from '../generators/dynamic-theme';
+import {getDynamicThemeFixesFor, getDynamicThemeFixesForNew} from '../generators/dynamic-theme';
 import createStaticStylesheet from '../generators/static-theme';
 import {createSVGFilterStylesheet, getSVGFilterMatrixValue, getSVGReverseFilterMatrixValue} from '../generators/svg-filter';
 import type {ExtensionData, FilterConfig, News, Shortcuts, UserSettings, TabInfo} from '../definitions';
@@ -21,6 +21,10 @@ import {isFirefox, isThunderbird} from '../utils/platform';
 import {MessageType} from '../utils/message';
 import {logInfo, logWarn} from '../utils/log';
 import {PromiseBarrier} from '../utils/promise-barrier';
+
+// TODO: remove
+import {parseInversionFixes} from '../generators/css-filter';
+import {parseStaticThemes} from '../generators/static-theme';
 
 export class Extension {
     config: ConfigManager;
@@ -419,20 +423,20 @@ export class Extension {
                 case ThemeEngines.cssFilter: {
                     return {
                         type: MessageType.BG_ADD_CSS_FILTER,
-                        data: createCSSFilterStylesheet(theme, url, frameURL, this.config.INVERSION_FIXES),
+                        data: createCSSFilterStylesheet(theme, url, frameURL, parseInversionFixes(this.config.INVERSION_FIXES_RAW)),
                     };
                 }
                 case ThemeEngines.svgFilter: {
                     if (isFirefox) {
                         return {
                             type: MessageType.BG_ADD_CSS_FILTER,
-                            data: createSVGFilterStylesheet(theme, url, frameURL, this.config.INVERSION_FIXES),
+                            data: createSVGFilterStylesheet(theme, url, frameURL, parseInversionFixes(this.config.INVERSION_FIXES_RAW)),
                         };
                     }
                     return {
                         type: MessageType.BG_ADD_SVG_FILTER,
                         data: {
-                            css: createSVGFilterStylesheet(theme, url, frameURL, this.config.INVERSION_FIXES),
+                            css: createSVGFilterStylesheet(theme, url, frameURL, parseInversionFixes(this.config.INVERSION_FIXES_RAW)),
                             svgMatrix: getSVGFilterMatrixValue(theme),
                             svgReverseMatrix: getSVGReverseFilterMatrixValue(),
                         },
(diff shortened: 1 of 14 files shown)


src/generators/utils/parse.ts
```diff
@@ -1,9 +1,22 @@
+import {isURLMatched, isFullyQualifiedDomain} from '../../utils/url';
 import {parseArray} from '../../utils/text';
 
 interface SiteProps {
     url: string[];
 }
 
+interface SitePropsOffsetRecord {
```
bershanskiy, line 8: My algorithm creates an instance of `SitePropsOffsetRecord` for every site prop. That is not very efficient, but very easy to read. I would prefer to sacrifice some readability and store these offsets in serialized form in a flat number array or even a string (hichever is more compact). Thoughts?

src/generators/static-theme.ts
```diff
@@ -249,11 +250,23 @@ export function formatStaticThemes(staticThemes: StaticTheme[]) {
     });
 }
 
-function getCommonTheme(themes: StaticTheme[]) {
-    return themes[0];
+// TODO: do not rely on common theme being the first in the file
+function getCommonTheme(staticThemes: string, staticThemesIndex: SitePropsIndex<StaticTheme>): StaticTheme {
```
bershanskiy, line 254: I'm a bit worried that code relies on the assumption that the first rule is for every domain (has pattern `*`).

src/utils/url.ts
```diff
@@ -199,3 +199,7 @@ export function isURLEnabled(url: string, userSettings: UserSettings, {isProtect
     }
     return (!isInDarkList && !isURLInUserList);
 }
+
+export function isFullyQualifiedDomain(candidate: string) {
+    return /^[a-z0-9.-]+$/.test(candidate);
```
Gusted, line 204: Not sure if it matters but this could be turned into a anti-pattern with simply: `/^(\*|\^|\$)$/` as in terms of regex it would be faster to only loop once trough the text(without tracking the state of the previous character) and check if such character exists.

Gusted (changes requested):
PR is on a good track. We just need to polish the rough edges of this.

src/utils/url.ts
bershanskiy: I created a rudimentary microbenchmark and get slightly better results with `/^[a-z0-9.-]+$/` than with `/^(\*|\^|\$)$/`. Will create a proper benchmark and share result by the end of today.

src/generators/utils/parse.ts
bershanskiy: Yes, this is close to what I was thinking. I want to improve on this design just a bit by storing the length of delimiters (to avoid having to strip them out twice) and encoding the flat number array into a string (because base 10 numbers are not efficient).

src/generators/static-theme.ts
bershanskiy: Then I'll just remove the comment and add a unit test wich ensures published configs have `*` rule.

bershanskiy: Turns out, there is already test for this. I'll just remove the comment then.
https://github.com/darkreader/darkreader/blob/58416b36d7d0ce1dd7700e01b8e56c1bcbdb226a/tests/config/config.tests.ts#L57

src/generators/utils/parse.ts
```diff
@@ -47,3 +60,100 @@ export function parseSitesFixesConfig<T extends SiteProps>(text: string, options
...
+
+        if (commandIndices.length === 0) {
+            return;
+        }
+
+        const urls = parseArray(lines.slice(0, commandIndices[0]).join('\n'));
+        const index = offsets.length;
+        for (const url of urls) {
+            // TODO: make sure this is correct
+            const domain = url.split('/')[0].toLowerCase();
```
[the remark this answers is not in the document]
bershanskiy: Yes, that's exactly the idea. Thanks for letting me know.

src/generators/utils/parse.ts
```diff
@@ -47,3 +60,100 @@ export function parseSitesFixesConfig<T extends SiteProps>(text: string, options
...
+            if (isFullyQualifiedDomain(domain)) {
+                if (typeof domains[domain] === 'undefined') {
```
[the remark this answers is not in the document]
bershanskiy: Done!

src/utils/url.ts
Gusted: Sure things, even though I think it might be more efficient to just do it like: `/[\*\^\$]/` 🤷🏽‍♂️

Gusted: image
Seems like the current one is indeed the fastest, so let it be as is.

src/generators/utils/parse.ts
```diff
@@ -47,3 +60,100 @@ export function parseSitesFixesConfig<T extends SiteProps>(text: string, options
...
+
+export function parseSiteFixConfig<T extends SiteProps>(text: string, options: SitesFixesParserOptions<T>, indexRecord: SitePropsOffsetRecord): T {
+    const block = text.substring(indexRecord.start, indexRecord.end);
+    return parseSitesFixesConfig<T>(block, options)[0];
+}
+
+export function getSitesFixesFor<T extends SiteProps>(url: string, text: string, index: SitePropsIndex<T>, options: SitesFixesParserOptions<T>): T[] {
+    const records: T[] = [];
+    let recordIds: number[] = [];
+    const domain = url.split('/')[0].toLowerCase();
```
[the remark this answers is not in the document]
bershanskiy: Done!

src/generators/utils/parse.ts
```diff
@@ -47,3 +60,100 @@ export function parseSitesFixesConfig<T extends SiteProps>(text: string, options
...
+    const domainPatterns: {[domainPattern: string]: number | number[]} = {};
+    const offsets: SitePropsOffsetRecord[] = [];
+
+    function processBlock(recordStart: number, recordEnd: number) {
+        // TODO: more formal definition of URLs and delimiters
+        const block = text.substring(recordStart, recordEnd);
+        const lines = block.split('\n');
+        const commandIndices: number[] = [];
+        lines.forEach((ln, i) => {
+            if (ln.match(/^\s*[A-Z]+(\s[A-Z]+)*\s*$/)) {
```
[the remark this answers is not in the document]
bershanskiy: I fixed this same issue in two places: my new function and the old function.

src/generators/utils/parse.ts
```diff
@@ -47,3 +60,100 @@ export function parseSitesFixesConfig<T extends SiteProps>(text: string, options
...
+            start: recordStart,
+            end: recordEnd
+        });
+    }
+
+    let recordStart = 0;
+    // Delimiter between two blocks
+    const delimiterRegex = /\r?\s*={2,}\s*\r?/gm;
+    while (true) {
+        const delimiter = delimiterRegex.exec(text);
```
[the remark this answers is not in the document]
bershanskiy: The problem with `while ((delimiter = delimiterRegex.exec(input))) {` is that it skips the very last block because its next delimiter is `null`. I can work around this by explicitly writing another `processBlock()` call.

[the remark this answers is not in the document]
bershanskiy: Done!

src/generators/utils/parse.ts
bershanskiy: I switched to a flat array with even items representing start and odd representing end. This way, if we need offsets of item with index `i`, its start is at `2*i` and end is at `2*i+1`. There is very little memory savings in "shingled" offsets and a bit more code complexity. Most lot of memory usage comes from domain URL strings themselves (keys in index).

Gusted: Sure things.

src/generators/utils/parse.ts
```diff
@@ -47,3 +55,117 @@ export function parseSitesFixesConfig<T extends SiteProps>(text: string, options
 
     return sites;
 }
+
+// URL patterns are guaranteed to not have protocol and leading '/'
+function getDomain(url: string) {
+    return url.split('/')[0].toLowerCase();
+}
+
+function encodeOffsets(offsets: number[]): string {
```
Gusted, line 64: I think we should have at this place a good big comment explaining this magic about storing the offsets into a string.

bershanskiy: Sure thing. I changed the method a bit and added a comment explaining how it works.

src/generators/utils/parse.ts
```diff
@@ -47,3 +60,100 @@ export function parseSitesFixesConfig<T extends SiteProps>(text: string, options
...
     return sites;
 }
+
+export function indexSitesFixesConfig<T extends SiteProps>(text: string): SitePropsIndex<T> {
+    const domains: {[domain: string]: number | number[]} = {};
+    const domainPatterns: {[domainPattern: string]: number | number[]} = {};
+    const offsets: SitePropsOffsetRecord[] = [];
+
+    function processBlock(recordStart: number, recordEnd: number) {
+        // TODO: more formal definition of URLs and delimiters
```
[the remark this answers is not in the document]
bershanskiy: I'll draft a formal document later today.

bershanskiy:
@Gusted I fixed a couple rought edges, I think it is ready for general use. There are still a few things that could be improved (more efficient storing of domain index, saving and restoring index to avoid index reconstruction on every context reload). But there are multiple different ways to do that and I can't pick the right tradeoffs right now.

Gusted:
>  I think it is ready for general use. T

I will play it once I'm finished writing this comment =)

> But there are multiple different ways to do that and I can't pick the right tradeoffs right now.

Put them into a nice issue and add it to the roadmap. Let's try small steps. Big steps will usually mean you will see something over the head and more code = more code review ;)

src/background/extension.ts
```diff
@@ -457,17 +457,17 @@ export class Extension {
...
                             theme.stylesheet :
-                            createStaticStylesheet(theme, url, frameURL, this.config.STATIC_THEMES),
+                            createStaticStylesheet(theme, url, frameURL, this.config.STATIC_THEMES_RAW, this.config.STATIC_THEMES_INDEX),
                     };
                 }
                 case ThemeEngines.dynamicTheme: {
                     const filter = {...theme};
                     delete filter.engine;
-                    const fixes = getDynamicThemeFixesFor(url, frameURL, this.config.DYNAMIC_THEME_FIXES, this.user.settings.enableForPDF);
+                    const fixesNew = getDynamicThemeFixesFor(url, frameURL, this.config.DYNAMIC_THEME_FIXES_RAW, this.config.DYNAMIC_THEME_FIXES_INDEX, this.user.settings.enableForPDF);
```
Gusted, line 466:
```suggestion
                    const fixes = getDynamicThemeFixesFor(url, frameURL, this.config.DYNAMIC_THEME_FIXES_RAW, this.config.DYNAMIC_THEME_FIXES_INDEX, this.user.settings.enableForPDF);
```

src/background/extension.ts
```diff
@@ -457,17 +457,17 @@ export class Extension {
...
                     const isIFrame = frameURL != null;
                     return {
                         type: MessageType.BG_ADD_DYNAMIC_THEME,
-                        data: {filter, fixes, isIFrame},
+                        data: {filter, fixesNew, isIFrame},
```
Gusted, line 470:
```suggestion
                        data: {filter, fixes, isIFrame},
```

Gusted (commented):
Lets hold this variable like this ;) as the content-script still expects `fixes`

src/generators/utils/parse.ts
```diff
@@ -47,3 +55,132 @@ export function parseSitesFixesConfig<T extends SiteProps>(text: string, options
...
+
+export function parseSiteFixConfig<T extends SiteProps>(text: string, options: SitesFixesParserOptions<T>, recordStart: number, recordEnd: number): T {
+    const block = text.substring(recordStart, recordEnd);
+    return parseSitesFixesConfig<T>(block, options)[0];
+}
+
+export function getSitesFixesFor<T extends SiteProps>(url: string, text: string, index: SitePropsIndex<T>, options: SitesFixesParserOptions<T>): T[] {
+    const records: T[] = [];
+    let recordIds: number[] = [];
+    const domain = getDomain(url);
```
Gusted, line 167: Ah stupid stupid me, `url` is the full blown URL and not a sanitized one. I think you can copy paste that sanatizeInput function over and use it here and probably as well for the other one.

bershanskiy: Can we just use the native `URL` constructor? E.g.,`new URL(url).hostname`.

src/generators/utils/parse.ts
```diff
@@ -47,3 +55,132 @@ export function parseSitesFixesConfig<T extends SiteProps>(text: string, options
...
+    // Array of tuples, where first number is an offset of record start and second number is record length.
+    const offsets: Array<[number, number]> = [];
+
+    function processBlock(recordStart: number, recordEnd: number, index: number) {
+        // TODO: more formal definition of URLs and delimiters
+        const block = text.substring(recordStart, recordEnd);
+        const lines = block.split('\n');
+        const commandIndices: number[] = [];
+        lines.forEach((ln, i) => {
+            if (ln.match(/^[A-Z]+(\s[A-Z]+){0,2}$/)) {
```
alexanderby, line 106: Not sure if there will be no commands with 3 and more words in future.

src/generators/utils/parse.ts
```diff
@@ -47,3 +55,132 @@ export function parseSitesFixesConfig<T extends SiteProps>(text: string, options
...
+            } else if (typeof domains[domain] === 'object' && !((domains[domain] as number[]).includes(index))) {
+                (domainPatterns[domain] as number[]).push(index);
+            }
+        }
+        offsets.push([recordStart, recordEnd - recordStart]);
+    }
+
+    let recordStart = 0;
+    // Delimiter between two blocks
+    const delimiterRegex = /\r?\s*={2,}\s*\r?/gm;
```
alexanderby, line 142: Hmmm, maybe we should ban the `\r`?

bershanskiy: I'd gladly remove `\r` from the code, since it is an archaic character not relevant in modern systems. I added this because current code permits source file to have `\r` (it just removes them):
https://github.com/darkreader/darkreader/blob/9c664ff170841c46322424795098e61629919f8a/src/generators/utils/parse.ts#L16

Perhaps, we could just forbid `\r` in configs (and add a unit test and throw a warning if user edits config and introduces this character).

src/generators/utils/parse.ts
bershanskiy: This restriction was suggested by @Gusted, I'll defer to him on this.

Gusted: I live in `new Date()` not in `new Date(new Date().setFullYear(new Date().getFullYear() + 1)) ` ;)

src/generators/utils/parse.ts
Gusted: I can live with that.

src/generators/utils/parse.ts
bershanskiy: Let's leave this restriction as is and then relax it later if ever need to.

src/generators/utils/parse.ts
bershanskiy: Please see #6690.

src/generators/utils/parse.ts
Gusted: I can live with that.

src/generators/utils/parse.ts
bershanskiy: I changed `getDomain()` a bit to accomodate incomplete URLs:
```js
function getDomain(url: string) {
    try {
        return (new URL(url)).hostname.toLowerCase();
    } catch (error) {
        return url.split('/')[0].toLowerCase();
    }
}
```

bershanskiy:
@Gusted I think all code issues should be resolved now, could you take another look?

Gusted (approved):
Implementation-wise you're getting the green light to have it merged. Everything is working and edge cases are being handled, nice work.

bershanskiy:
I added some more tests (now `parse` has 100% coverage) and fixed a comment and a typo.

PR #6648, opened 2021-08-30
Fix for twitter.com video player

c-nagy:
Twitter uses obfuscated class names so I had to get creative to avoid using gibberish CSS selectors.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -12822,8 +12822,7 @@ CSS
 twitter.com
 
 INVERT
-[aria-label="Seek slider"]
-[aria-label="Volume slider"]
+[role="slider"]
 [style^="flex-grow"]
 
 CSS
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -12821,6 +12821,11 @@ CSS
 
 twitter.com
 
+INVERT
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -12821,6 +12821,11 @@ CSS
 
 twitter.com
 
+INVERT
+[aria-label="Seek slider"]
+[aria-label="Volume slider"]
```
Gusted, line 12826: I'm not sure if this stay the same across different languages.

src/config/dynamic-theme-fixes.config
```diff
@@ -12821,6 +12821,11 @@ CSS
...
+[style^="flex-grow"]
```
Gusted, line 12827: Hmm, well it seems good it could also be tricky to have it included. But it should be good for now.

c-nagy: Yeah, it's a trick corner situation for sure. I couldn't find a better way to do it, so I'd be very interested in how else it could be fixed.

Gusted: > I couldn't find a better way to do it, so I'd be very interested in how else it could be fixed.

Nor can I so let's leave it for now and see if it breaks things.

Gusted (approved):

PR #6651, opened 2021-08-31
Fix for psychonautwiki.org

c-nagy:


src/config/dynamic-theme-fixes.config
```diff
@@ -10142,6 +10142,27 @@ INVERT
...
+img[src*="Pencil.svg"]
+#InfoTable > tbody > tr > .Image > div > div > .image
+
+CSS
+.dosechart {
+    background-color: white !important; 
+}
+.thumbimage {
+    background-color: white !important; 
+}
```
Gusted, line 10162:
```suggestion
.dosechart,
.thumbimage {
    background-color: ${black} !important;
}
```

src/config/dynamic-theme-fixes.config
```diff
@@ -10142,6 +10142,27 @@ INVERT
...
 
+psychonautwiki.org
+
+INVERT
+#mw-panel > a > svg
+.mw-headline > a > svg
+a[href="http://www.emcdda.europa.eu/about"] > svg
+img[src*="Tasks.svg"]
+img[src*="Star-o.svg"]
+img[src*="Pencil.svg"]
```
Gusted, line 10153: Are this individual images on wiki pages? Or global ones. I'd like to avoid including individual images as they are pretty much break pages without noticing(we've had this before on wikipedia)

c-nagy: I think those icons are just used on the home page but I'm not sure. I noticed that the alt tags for those icons have the form of `` I can specify that instead of using `src*=` if that would make it safer, as they don't have a shared CSS class.

Gusted (approved):

PR #6659, opened 2021-09-01
doodstream.com dood.la are dark

FadeMind:
https://doodstream.com/

dood.la is subdomain used for share links.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -231,6 +231,8 @@ disneyplus.com
 distrotube.com
 dlive.tv
 dodi-repacks.site
+doodstream.com
+dood.la
 dota2.ru
 dotabuff.com
 dotapicker.com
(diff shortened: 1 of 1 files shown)


src/config/dark-sites.config
```diff
@@ -231,6 +231,8 @@ disneyplus.com
 distrotube.com
 dlive.tv
 dodi-repacks.site
+dood.la
```
Gusted, line 234: We don't allow sites that are used as redirects.

Gusted (changes requested):

src/config/dark-sites.config
Gusted: Wait, let me guess... dood.la is used as "shorten link"?

FadeMind: Sample:
it is video stream file on doodstream.com 
Instead of long `doodstream.com/path/foo/bar.mp4`
You have link dood.la/s/abscdfgr

This is same server provider, using own domain for share link to content and embed player. 

Against DMCA I cannot share full link for valid content.

Gusted: Got it, seems good then.

Gusted (approved):

PR #6668, opened 2021-09-03
Fix for reddit.com

c-nagy:
Fix for seekbar and volume controls in the video player.

There's so much obfuscation on new reddit that I couldn't find a better way to target the volume controls other than `[style^="height"]` which seems to be a unique enough fingerprint at least. It's an assumption though that anything with an inline style tag with the first word being "height" (`<style="height: ...">`) must be a volume slider and needs to be inverted... Seems unique enough to fix the dark controls for now though without accidentally targeting other elements.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -10411,7 +10411,7 @@ reddit.com
 
 INVERT
 [role="slider"]
-[style^="height"]
+video ~ div [style^="height"]
 
 CSS
 [role="slider"] > div:nth-child(4) {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -10409,7 +10409,14 @@ div.fly-but-wrap.left.relative > span
 
 reddit.com
 
+INVERT
+[role="slider"]
+[style^="height"]
```
Gusted, line 10414: Nope, this will need to be a more strict, I know they are applying to posts that are out of the content view these `height` so they appear out of the content view, I don't like to apply any invert over those.

```suggestion
video ~ div [style^="height"]
```

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
c-nagy: Ah nice use of the `~` selector. That's a handy way to make it more targeted, thanks.

Gusted (approved):

PR #6674, opened 2021-09-03
Remove text-shadow on developers.arcgis.com

mikepianka:
Removes the white text shadow that obscures <a> tags on developers.arcgis.com, particularly in the API References. For example, https://developers.arcgis.com/javascript/latest/

Before
arcgis_js_darkreader_default

After
arcgis_js_darkreader_modified


Gusted (approved):
Thank you!

PR #6676, opened 2021-09-03
License 2021

alexanderby:


Gusted:
Should you give a explanation to this change?

alexanderby:
I don't know, nothing changes. But if anything wrong happens, they should blame the company, but not me. I will write ToS too.

PR #6682, opened 2021-09-04
Fix for developer.chrome.com

c-nagy:


src/config/dynamic-theme-fixes.config
```diff
@@ -3389,6 +3389,24 @@ IGNORE IMAGE ANALYSIS
 
 ================================
 
+developer.chrome.com
+
+INVERT
+.logo__text
+.search-box__input
```
Gusted, line 3396: Ehm, I don't think is correct.

Gusted (changes requested):

Gusted (approved):

PR #6683, opened 2021-09-04
Fix for yelp.com

c-nagy:


Gusted (changes requested):
Not sure which are targgeted for the input, but it should be removed. I'm not sure why it should white?

c-nagy:
> Not sure which are targgeted for the input, but it should be removed. I'm not sure why it should white?

It's targeting the search bar on the home page and the results page, just to make it stand out a bit more.

Gusted:
> just to make it stand out a bit more.

Why should it stand more out?. In my eyes that would be a opinionated fix.

c-nagy:
> > just to make it stand out a bit more.
> 
> Why should it stand more out?. In my eyes that would be a opinionated fix.

I guess that's true. Alright, I reverted that part and just left the inversion for the Yelp logo and embedded google map section.

Gusted (approved):

PR #6689, opened 2021-09-04
Add option to manually dispatch release workflow and upload artifacts

bershanskiy:
This is continuation of #6650, but just for release build workflow. Also, it adds option to store the built files as artifacts (via [`upload-artifact`](https://github.com/actions/upload-artifact)).

### DIFF
--- a/.github/workflows/tagged-release.yml
+++ b/.github/workflows/tagged-release.yml
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
     inputs:
       retention_days:
-        description: 'Retantion duration (in days)'
+        description: 'Retention duration (in days)'
         required: true
         default: '1'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/tagged-release.yml
```diff
@@ -38,3 +44,9 @@ jobs:
         with:
           files: |
             build/release/darkreader-*
+
+      - name: Upload built artifact
+        uses: actions/upload-artifact@v2
+        with:
+          path: build/release/darkreader-*
+          retention-days: ${{ inputs.retention_days }}
```
Gusted, line 52: Does the uploading of a artifact work properly? Commits like these are only accessible by linking them or digging into /actions

bershanskiy: Yes, it uploads a single file named `artifact.zip` which contains all Darkreader builds. I tested this by pushing this commit to my own `master` and then dispatching this workflow from "Actions" tab.

Gusted (approved):

PR #6696, opened 2021-09-05
Fix: catch and re-throw error in getImageDetails

bershanskiy:
`urlToImage` can throw error `Unable to load image ${url}` which was not previously caught by `getImageDetails()` and could invalidate the whole context. After this commit, this error is properly caught and re-thrown.


Gusted (changes requested):
Can we limit the `try { }` scope to just the `urlToImage` function?

bershanskiy:
Is this better?

Gusted (approved):

PR #6697, opened 2021-09-05
Fix for Jumia

OmarNamis:
Update to fix some CSS changes

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -6944,9 +6944,8 @@ img[src*="empty-cart.png"]
 div.-fw-w:nth-child(1) > a.-fs0 > img
 article.-df > svg > use
 svg.ic.xprss
-svg.ic[alt*="جوميا"]
-svg.ic[aria-label*="اكسبريس"]
-svg.ic[aria-label*="جلوبال"]
+div.-df.-d-co.-c-bet > h3.-fw > svg.ic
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -6940,13 +6940,13 @@ img[src*="empty-cart.png"]
...
+.col4 > .s-menu > a:not(:nth-child(1)):not(:nth-child(3)):not(:nth-child(7)) > svg
 div.-fw-w:nth-child(1) > a.-fs0 > img
 article.-df > svg > use
 svg.ic.xprss
-svg.ic[alt*="Express"]
-svg.ic[aria-label*="Express"]
-svg.ic[aria-label*="Global"]
+svg.ic[alt*="جوميا"]
+svg.ic[aria-label*="اكسبريس"]
+svg.ic[aria-label*="جلوبال"]
```
Gusted, line 6949: This seems like language-independent? Could these be changed to something that doesn't rely on the aria-label?

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
OmarNamis: Hmm, indeed it's language-dependent.
The css code is obfuscated, so it was a little annoying trying to get a nice and short selector.

Gusted (approved):

PR #6702, opened 2021-09-06
Fix display of star ratings on sephora.com

SithLadyRaven:
Makes star ratings visible when dark mode is enabled.


Gusted (approved):
Sorry for the late review, somehow this email got slipped into my spam.

PR #6729, opened 2021-09-10
Refactor: use enum in getDestDir()

bershanskiy:
This is just a trivial refactor to improve readability.

### DIFF
--- a/tasks/bundle-js.js
+++ b/tasks/bundle-js.js
@@ -5,15 +5,15 @@ const rollupPluginNodeResolve = require('@rollup/plugin-node-resolve').default;
 const rollupPluginReplace = require('@rollup/plugin-replace');
 const rollupPluginTypescript = require('rollup-plugin-typescript2');
 const typescript = require('typescript');
-const {getDestDir} = require('./paths');
+const {getDestDir, PLATFORM} = require('./paths');
 const reload = require('./reload');
 const {PORT} = reload;
 const {createTask} = require('./task');
 
 async function copyToBrowsers({cwdPath, debug}) {
-    const destPath = `${getDestDir({debug})}/${cwdPath}`;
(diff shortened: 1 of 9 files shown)


tasks/bundle-js.js
```diff
@@ -129,7 +129,7 @@ async function bundleJS(/** @type {JSEntry} */entry, {debug, watch}) {
     });
     entry.watchFiles = bundle.watchFiles;
     await bundle.write({
-        file: `${getDestDir({debug})}/${dest}`,
+        file: `${getDestDir({debug, platform: PLATFORM.CHROME})}/${dest}`,
```
Gusted, line 132: Hmm, this will means the watcher function is limited to the chrome folder. We might need to add extra arguments for the watcher then.

Gusted (changes requested):
Hmm, I do like using enums for this repeating string :)

tasks/bundle-js.js
bershanskiy: I think this is intentional, since `bundle.write` creates a generic build (which happens to be Chrome build) and then `entry.postBuild` copies the output to all desired locations simultaniously patching it if needed.

Gusted: Okay.

Gusted (approved):

tasks/bundle-js.js
bershanskiy: I agree that it is harder to read, but if `bundleJS` was called for every target separatelly, compile time would be almost 3 times longer.

bershanskiy:
For some reason, build output differs (checked with `diff -r`). I'll investigate the difference tomorrow, since I don't have time today.

bershanskiy:
Now `diff -r` does not find any differences aside from archives, which have the same content but a bit diffrerent file timestamps.

PR #6731, opened 2021-09-10
added curiositystream.com to dark-sites.config

3tes:


src/config/dark-sites.config
```diff
@@ -120,6 +120,7 @@ bugmenot.com
 buildbot.orphis.net
 bungie.net
 byte.co
+curiositystream.com
```
Gusted, line 123: You have to correctly arrange the site.

Gusted (changes requested):

src/config/dark-sites.config
3tes: sorry it should now be in the right place

Gusted (approved):

PR #6733, opened 2021-09-10
Make selected radio button visible on checkout page of etsy.com

SithLadyRaven:


src/config/dynamic-theme-fixes.config
```diff
@@ -7200,6 +7201,15 @@ img[src$="close.svg"]
 
 ================================
 
+klarna.com
+
+CSS
+input, textarea {
+    -webkit-text-fill-color: var(--darkreader-neutral-text) !important;
```
Gusted, line 7208: Uhm, this is not a standarized property and thus not accepted also we have a default style that does this by default.

Gusted (changes requested):

Gusted (approved):

src/config/dynamic-theme-fixes.config
SithLadyRaven: Sorry! That section of changes was not meant to be included as part of this. It was some separate experimentation. I've now corrected the PR to be limited to the appropriate commit.

PR #6741, opened 2021-09-11
Fix book content inversion on Kindle Cloud Reader

ThatChristianGuy:
read.amazon.com

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -750,6 +750,7 @@ amazon.*.*
 
 INVERT
 #banner-image
+#kindleReader_content
 #ordersContainer .a-box.order-attributes img
 div.a-section.vse-lb-video-metadata
 div.vse-video-content


Gusted (commented):
Hm, why should the content be inverted? Doesn't that include images?

dzaikos:
I just started using DarkReader and while I'm not the author of this PR, I can say that Kindle Cloud Reader now renders pages dynamically, on-demand, as images. This means normal CSS for changing fore- and background text don't affect the actual book content. Inverting images inside `#kindleReader_content` would produce the correct results.

It should be noted that this PR isn't valid anymore because since it was submitted further commits to this file now provide specific instructions for `read.amazon.com`, although the correct domain should be `read.amazon.*` since country-codes are a possibility with Kindle Cloud Reader.

Gusted:
> I just started using DarkReader and while I'm not the author of this PR, I can say that Kindle Cloud Reader now renders pages dynamically, on-demand, as images. This means normal CSS for changing fore- and background text don't affect the actual book content. Inverting images inside `#kindleReader_content` would produce the correct results.
> 
> It should be noted that this PR isn't valid anymore because since it was submitted further commits to this file now provide specific instructions for `read.amazon.com`, although the correct domain should be `read.amazon.*` since country-codes are a possibility with Kindle Cloud Reader.

Thank you for testing out the PR, I will update it a bit and get it merged.

Gusted (approved):

Gusted:
Superseeded by #7523

PR #6743, opened 2021-09-12
Some fixes for odysee.com

ZOASR:


Gusted (changes requested):
Please remove the usages of `[dir=...]`.

ZOASR:
> Please remove the usages of `[dir=...]`.

Ok, i removed it,  but why not use it?

Gusted:
> Ok, i removed it, but why not use it?

As since recently I'm personally trying to avoid such usage, as they add little to no value and can even harm the selector itself. `dir` is commonly used to indicate if the user has a preference for left-to-right rendering or right-to-left rendering.

src/config/dynamic-theme-fixes.config
```diff
@@ -9477,6 +9477,23 @@ IGNORE INLINE STYLE
...
+.header__navigation-logo
+[data-vjs-player] .vjs-control-bar > [title="Autoplay Next Off"]
+[data-vjs-player] .vjs-control-bar > [title="Autoplay Next On"]::after
+
+CSS
+[dir] .content__viewer {
+    border: none;
+}
+[data-vjs-player] .vjs-control-bar .vjs-progress-control.vjs-control {
+    filter: invert(100%) hue-rotate(230deg) contrast(90%) !important;
```
Gusted, line 9492: Hmm, why is their a custom filter? Is the `INVERT` not sufficient enough?

src/config/dynamic-theme-fixes.config
```diff
@@ -9477,6 +9477,23 @@ IGNORE INLINE STYLE
...
 
+odysee.com
+
+INVERT
+.header__navigation-logo
+[data-vjs-player] .vjs-control-bar > [title="Autoplay Next Off"]
+[data-vjs-player] .vjs-control-bar > [title="Autoplay Next On"]::after
+
+CSS
+[dir] .content__viewer {
```
Gusted, line 9488:
```suggestion
.content__viewer {
```

src/config/dynamic-theme-fixes.config
```diff
@@ -9477,6 +9477,23 @@ IGNORE INLINE STYLE
...
+    border: none;
```
Gusted, line 9489:
```suggestion
    border: none !important;
```

src/config/dynamic-theme-fixes.config
ZOASR: No it `INVERT` does the job right,  but it turns this selector a little bit pale,  so i adjusted it a bit.

ZOASR: i only changed the hue it's not that huge of a change

Gusted: >  that huge of a change

Correct, until we're going to add the factor of user settings which might have grayscale turned up all the way or sepia all over the place, in which this fix isn't going to match with the user, I'm trying to avoid such situations.

ZOASR: Well, for now i don't mind removing this filter modification and leaving the default filter place, cause as i said it doesn't contribute much.

Gusted (approved):

PR #6749, opened 2021-09-13
Fix: restore old behavior of implied wildcards

### LINKED ISSUE #6738: [Bug] Domains Index aren't URL Matched
# Bug Report

## Bug Description
The index.domains are currently being checked by 1-on-1 equal check:
https://github.com/darkreader/darkreader/blob/d69522753bff6abea855bcbc7661642ffc950db4/src/generators/utils/parse.ts#L172-L174

While they currently should be parsed by our inefficient and weird isURLMatch implemenation as it will allow `unix.stackexchange.com` to match `stackexchange.com`(Please note that the current config has *.stackexchange.com, but I'm planning on removing all `*.` references for now until the #2517 is merged)

## Website Address
https://unix.stackexchange.com/questions/797/how-can-i-find-the-implementations-of-linux-kernel-system-calls

## Steps To Reproduce
- Go to the site.
- See that the fixes aren't applied.

## Expected Behavior
Have the fixes applied

## Actual Behavior
Fixes aren't found.

## System Information

- OS: Arch Linux
- Browser: Firefox Nightly 93.0a1 (2021-09-04) 
- Dark Reader Version: https://github.com/darkreader/darkreader/commit/d69522753bff6abea855bcbc7661642ffc950db4

## Additional Context
CC @bershanskiy 
Blocker for #6737

bershanskiy:
Fixes #6738.

Commit 2cf54561f7981f243d5d9fafd61e7f3cfd764f80 changed behavior
of getDynamicThemeFixesFor() to require explicit wildcards.
Restore the old bahavior and delegate final URL chack to isURLMatched().

### DIFF
--- a/src/generators/utils/parse.ts
+++ b/src/generators/utils/parse.ts
@@ -178,8 +178,13 @@ export function getSitesFixesFor<T extends SiteProps>(url: string, text: string,
             recordIds = recordIds.concat(index.domainPatterns[pattern]);
         }
     }
-    if (index.domains[domain]) {
-        recordIds = recordIds.concat(index.domains[domain]);
+
+    const labels = domain.split('.');
+    for (let i = 0; i < labels.length; i++) {
+        const substring = labels.slice(i).join('.');
+        if (index.domains[substring] && isURLMatched(url, substring)) {
+            recordIds = recordIds.concat(index.domains[substring]);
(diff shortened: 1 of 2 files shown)


Gusted (approved):
Seems good.

PR #6761, opened 2021-09-14
Add context menus

### LINKED ISSUE #6717: [Feature Request] Add context menus?
# Feature Request
Hey, I would like to suggest that in the feature that already exists of enabling or disabling the extension you could not disable the whole extension, but set default mode for ON/OFF. So, I could set it to be ON/OFF by default, and change its behavior if needed for the active tab. **Right now, you can't set it OFF by default and enable it only for the active tab, if its off, the tab option is off too.**

Also, it might be interesting to add this option of turning on/off when you right click in the page, like other extensions do. 

Thanks a lot!

No. 

## Additional Context
--

bershanskiy:
Fixes #6717.

Preview:

image

image

image

### DIFF
--- a/src/_locales/en.config
+++ b/src/_locales/en.config
@@ -150,6 +150,18 @@ Default
 Change the browser theme
 
 
+#==== Context menu ====
+
+@context_menu_setting
+Use context menus
+
+@context_menu_setting_disabled
+Context menu integration is disabled
+
+@context_menu_setting_enabled
+Context menu integration is enabled
+
+
 #==== Footer ====
 
 @privacy
--- a/src/_locales/en.config
+++ b/src/_locales/en.config
@@ -150,18 +150,6 @@ Default
 Change the browser theme
 
 
-#==== Context menu ====
-
-@context_menu_setting
-Use context menus
-
-@context_menu_setting_disabled
-Context menu integration is disabled
-
-@context_menu_setting_enabled
-Context menu integration is enabled
-
-
 #==== Footer ====
 
 @privacy
--- a/src/background/extension.ts
+++ b/src/background/extension.ts
@@ -51,12 +51,6 @@ export class Extension {
     static ALARM_NAME = 'auto-time-alarm';
     static LOCAL_STORAGE_KEY = 'Extension-state';
     constructor() {
-        // We need to stash a reference to extension object to use it within
-        // context menu activation handler.
-        // Storing this extra reference does not have any adverse effects on
-        // GC, since Extension object is long-living and does not need to be collected.
-        globalThis.extension = this;
-
         this.config = new ConfigManager();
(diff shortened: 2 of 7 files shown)


bershanskiy:
I think it's ready to be merged. I'm pretty happy with the way code turned out, but I have no idea how to add unit tests for this.

src/background/extension.ts
```diff
@@ -49,6 +51,14 @@ export class Extension {
     static ALARM_NAME = 'auto-time-alarm';
     static LOCAL_STORAGE_KEY = 'Extension-state';
     constructor() {
+        /*
```
alexanderby, line 54: What do you think about using `//` for inline comments and `/**/` for JSDoc comments? I think it's a common pattern.

src/ui/popup/settings-page/context-menus.tsx
```diff
@@ -0,0 +1,36 @@
...
+        }
+    }
+
+    return !isMobile && (
+        <CheckButton
+            checked={props.data.settings.enableContextMenus}
+            label="Use context menus"
+            description={props.data.settings.enableContextMenus ?
+                'Make use of context menus when appropriate' :
+                'Do not use context menus'}
```
alexanderby, line 32: I think this description should represent the current state, e.g. "Context menu actions disabled" etc.

src/background/extension.ts
```diff
@@ -181,37 +212,82 @@ export class Extension {
     }
 
     async onCommand(command: string, url: string) {
-        if (this.startBarrier.isPending()) {
-            await this.startBarrier.entry();
+        /*
+         * We need the following line to re-use this function for
+         * context menu handlers.
+         */
+        const extension = this || globalThis.extension;
```
alexanderby, line 219: What happens to "this" here? Is it "null"? Maybe try function assignment?
```typescript
onCommand = async (command: string, url: string) => {
    // ...
};
```
This way it will be bound to Extension instance when called as a function untied from an instance.

alexanderby (changes requested):

bershanskiy (commented):
New UI messages:
image

src/_locales/en.config
```diff
@@ -150,6 +150,18 @@ Default
 Change the browser theme
 
 
+#==== Context menu ====
```
alexanderby, line 153: Lets currently remove it, because it requires much work with all the locales.

The design will change and many messages will totally disappear, and some will appear. So we will start another campaign to ask people translate.

src/background/extension.ts
```diff
@@ -212,6 +233,46 @@ export class Extension {
         }
     }
```
alexanderby, line 234: ESLint should have warned about `;`

alexanderby (approved):
Let's remove the locales here, it will be all changed later.

alexanderby (approved):

PR #6764, opened 2021-09-14
Refactor: move getFontList() to UI

bershanskiy:
This is a part of series of PRs which will eventually allow popup UI to retreive all necessary info from storage bypassing background context.


Gusted (approved):
Good refactor.

PR #6771, opened 2021-09-15
update: patch the pairs fix from darkreader#2129

ashtonx:
fixes issue
https://github.com/darkreader/darkreader/pull/2129#issuecomment-919761982

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4129,7 +4129,7 @@ INVERT
 ._24NNT
 
 CSS
-._2OVuy{
+.LhRk3{
   opacity: 0.3;
 }
 
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4129,8 +4129,8 @@ INVERT
 ._24NNT
 
 CSS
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -4129,7 +4129,7 @@ INVERT
 ._24NNT
 
 CSS
-._2OVuy{
+.LhRk3{
```
Gusted, line 4132: Is it possible to change this selector that is less obscure and don't rely on this random "string"?

ashtonx: Will try to look for a way, that said webdev and css is black magic for me so if i fail i won't be able to tell for certain.

src/config/dynamic-theme-fixes.config
```diff
@@ -4129,8 +4129,8 @@ INVERT
 ._24NNT
 
 CSS
-._2OVuy{
-  opacity: 0.3;
+button[aria-disabled="true"] > span {
```
Gusted, line 4132: Phew, okay this seems okay.

Gusted (approved):

Gusted:
Thank you!

Issue #559, opened 2018-06-12
Translations


alexanderby:
### How to improve a translation
- Open [locales folder](https://github.com/darkreader/darkreader/tree/master/src/_locales).
- Find a language code you know.
- Edit messages directly on GitHub (or on local machine if you are familiar with Git) under `@message_id`. Please, preserve the lines count.
- Submit a pull request.

### How to add a new translation
- Leave a comment that you are going to add a new translation, check that somebody else has not started it yet.
- See the list of [supported Chrome locales](https://developer.chrome.com/webstore/i18n#localeTable).
- Download [English locale](https://raw.githubusercontent.com/darkreader/darkreader/master/src/_locales/en.config).
- Rename the file into corresponding language code.
- Edit messages under `@message_id`. Please, preserve the lines count.
- Open [locales folder](https://github.com/darkreader/darkreader/tree/master/src/_locales).
- Click **Upload files**.
- Submit a pull request.

### Help page translations
- Help page translations are listed [here](https://github.com/darkreader/darkreader.org/blob/master/src/help/topics).
- Follow the same steps described above.

alexriabtsev:
Hi, team! uk (Ukrainian) translation going to be done.

alexanderby:
@alexriabtsev Great, here is [Belarusian](https://github.com/darkreader/darkreader/blob/master/src/_locales/be.config) and [Russian](https://github.com/darkreader/darkreader/blob/master/src/_locales/ru.config) for your reference.

alexanderby:
Thank you @svemal, great work! It will be included in next release.

Underday:
hello turkish translation is ready soon

PsiPhiTheta:
currently working on french translation for help page

UchihaSR:
can i do it for indian languages?

alexanderby:
Hi @UchihaSR, here is my Hindi translation using Google Translate, feel free to suggest a fix https://github.com/darkreader/darkreader/blob/master/src/_locales/hi.config

UchihaSR:
Greeting @alexanderby  your translation seems to be pretty good, need to do some correction with store listing section for grammatical error and small changes in a few places. shall i make changes and then send a PR.

reymon359:
Hi! I just finished the spanish translation for the help on [darkreader.org](https://github.com/darkreader/darkreader.org/tree/master/src/help/topics)

alexanderby:
[...]

PR #6777, opened 2021-09-16
Fix german translation

ElegyD:
Also mentioned in this comment: https://github.com/darkreader/darkreader/issues/559#issuecomment-763387007


Gusted (approved):
Thanks!

Issue #6789, opened 2021-09-17
[Bug] Proxy doesn't work


Gusted:
# Bug Report

## Bug Description

The proxy doesn't work, due to #6506 changes to the proxy function.

## Website Address

[a.tar.gz](https://github.com/darkreader/darkreader/files/7188052/a.tar.gz)

## System Information

N.V.T. Just the latest build of dark reader is needed

## Additional Context
Sigh, I just only found about this recently as I don't visit many sites anymore since school started, especially sites that has usages of CCSOM... Sorry for this being a last moment release blocker #6737
CC @bershanskiy

bershanskiy:
What is the desired behavior? Is there a known-good revision to test against? It's odd, but for me the current released version does not work (it does absolutely nothing), while current `master` properly inverts background but does not react to the text color change.

Gusted:
> What is the desired behavior?

The desired behavior of the proxy is to be executed within the page's context. When that happens the correct prototype's can be overridden within the context and thus when certain things happens it will execute our proxy function instead of the native function. Currently the proxy is just being executed in our extension's context, which is not correct.

> Is there a known-good revision to test against?

https://github.com/darkreader/darkreader/commit/b104227d7fd88026eab38260e09ff7fd6b22eb88

> while current `master` properly inverts background but does not react to the text color change.

The proxy function is able to hook into the functions that can change style's and thus let dark reader know when their are changes.

Gusted:
BTW ^ because of the cross-context behavior it's very hard to actually create a test for it. Which is probably why it wasn't catched earlier on.

bershanskiy:
I'm not sure I understand the bug, since repro test case looks almost identical:
![image](https://user-images.githubusercontent.com/45960703/133881358-100fc704-4579-46c5-924a-164cb0b96a83.png)

Gusted:
> I'm not sure I understand the bug, since repro test case looks almost identical:

Easiest way to check if the bug is affected if the text color is getting modified(devtools -> Select `<p>` -> See if their's a override from the default `pink` color).

PR #6791, opened 2021-09-18
Fix: restore old method of script injection

bershanskiy:
Discussed in #6789. I marked this as a "draft" until I validate this fixes the bug.


Gusted (approved):
Approved to be merged.

image
The easiest way to check if it works, when one of the proxied function's can be seen when using the `.toString()` function.

PR #6795, opened 2021-09-19
Fix for codeberg.org

Elaborendum:
When submitting a bug/pull request on Codeberg, the blinking typing cursor on the title of said bug/pull request was dark, much like the background. This didn't happen on the body of the bug/pull request (it's typing cursor was white), so I'm not sure if it's okay this way.

imagen

I think this would also apply to other Gitea instances?


Gusted:
I'm going to leave this open as it seems like rather a bug within dark reader, as we should just modify the caret-color property. But in this case DR isn't doing it.

Gusted:
Thanks for the unusual bug report 😄

PR #6796, opened 2021-09-19
Add router.asus.com to dark sites config

bershanskiy:
This is router admin panel of ASUS routers. It is already dark by default:
image

When Dark Reader is on it looks slightly broken:
image


Gusted (approved):
Does every router have the same UI? Looking from [the images](https://duckduckgo.com/?q=asus+router+dashboard&t=ffab&iax=images&ia=images) it seems like they are all dark-by-default.

Approved :)

PR #6806, opened 2021-09-19
Add `codeiguess.github.io`

CodeIGuess:


src/config/dark-sites.config
```diff
@@ -143,6 +143,8 @@ clash3d.com
 clashofstats.com
 clt.gg
 codepen.io
+codeiguess.github.io
```
Gusted, line 146: image

Seems to be not dark.

CodeIGuess: Sorry mate, site's not up yet.

Gusted: Then why should I merge this PR?

CodeIGuess: codeiguess.github.io/Notepad is dark mode and standalone. It would be nice to add it. codeiguess.github.io itself isn't up yet, however.

Gusted: Then let's merge /notepad for now

Gusted: Please remove the other entry.

PR #6811, opened 2021-09-20
Tests: devtools E2E test

bershanskiy:
This adds a test for Devtools UI which loads a dummy page (with some dummy styles), changes config (and checks that page styles changed), then resets styles (and checks that page styles went back to the original ones).

This test probably would have caught broken Devtools if we had it (I can check this if needed).

### DIFF
--- a/tests/browser/environment.js
+++ b/tests/browser/environment.js
@@ -26,6 +26,7 @@ class PuppeteerEnvironment extends JestNodeEnvironment {
         this.global.browser = this.browser;
 
         this.extensionPopup = await this.openPopupPage();
+        this.extensionDevtools = await this.openDevtoolsPage();
 
         this.page = await this.createTestPage();
         this.global.page = this.page;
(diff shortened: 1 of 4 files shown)


tests/browser/globals.d.ts
```diff
@@ -10,6 +10,9 @@ declare global {
     const popupUtils: {
         click: (selector: string) => Promise<void>;
         exists: (selector: string) => Promise<boolean>;
-        getBoundingRect: (selector: string) => Promise<{left: number; top: number; width: number; height: number}>;
```
bershanskiy, line 13: I removed `popupUtils.getBoundingRect` because it is not used.

tests/browser/environment.js
```diff
@@ -174,7 +185,7 @@ class PuppeteerEnvironment extends JestNodeEnvironment {
                 ws.on('close', () => sockets.delete(ws));
             });
 
-            function sendToPopup(message) {
```
bershanskiy, line 177: `sendToPopup()` currently broadcasts every message to every socket for simplicity. I don't think it's worth making it smarter yet.

tests/browser/globals.d.ts
Gusted: Seems good.

tests/browser/environment.js
Gusted: If it works, don't touch it ;)

Gusted (approved):
First glance seems to be fine for me. No weird code shenanigans. Nice to have a E2E for Dev tools.

PR #6816, opened 2021-09-20
perf: faster color parsing

### LINKED ISSUE #6814: [Performance Issue] Poor performance on bytescout
# Performance Issue Report

## Website Address
https://bytescout.com/products/developer/documentparsersdk/index.html

## Performance Profile
https://share.firefox.dev/2XAIQ3l

## System Information

- OS: (Arch) Linux
- Browser: Firefox Nightly
- Dark Reader Version: https://github.com/darkreader/darkreader/commit/47826398ccdada8520fb6681f6e50f9be55c9fcd

## Additional Context
- Does this happen when Dark Reader is disabled? Nope
- Does this happen in other browsers? Yup
- Do you have hardware acceleration enabled? I honestly don't know, Linux is weird.
 
#6794 Was the originally issue but not in the right format
CC @Gitoffthelawn

Gusted:
- Partially resolves #6814
- This should remove the old slow regex splitter and replace it by a simple parser.
- Tests are passing and seems to be all good and working.
- Perf results on Firefox Nightly | Arch Linux | (Split functionallity)
Old: 873,027 ops/sec ±1.54% (95 runs sampled) 
New: 1,598,297 ops/sec ±0.63% (97 runs sampled)


Gusted:
I'm probably missing some kind of edge case, which isn't covered by the tests. But only testing this change out on master will tell us the edge-case =)

PR #6861, opened 2021-09-22
Fix: context menus on Firefox

bershanskiy:
Firefox does not support `contextMenus` as an optional permission. It must be among required `permissions` in manifest.

[MDN docs](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/optional_permissions).

### DIFF
--- a/src/ui/popup/settings-page/context-menus.tsx
+++ b/src/ui/popup/settings-page/context-menus.tsx
@@ -1,11 +1,15 @@
 import {m} from 'malevic';
-import {isMobile} from '../../../utils/platform';
+import {isFirefox, isMobile} from '../../../utils/platform';
 import CheckButton from '../check-button';
 import type {ViewProps} from '../types';
 
 export default function ContextMenusGroup(props: ViewProps) {
     function onContextMenusChange(checked: boolean) {
         if (checked) {
+            if (isFirefox) {
+                props.actions.changeSettings({enableContextMenus: true});
+                return;
+            }
             // We have to request permissions in foreground context in response to user action.
             // 'contextMenus' permission is granted automatically without any permission prompts.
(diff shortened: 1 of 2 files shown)


alexanderby:
Will the menu show up by default?

bershanskiy:
No, menus will be disabled by default. We technically could make menus enabled by default, but (1) I don't like when extensions do that and clutter the UI and (2) this would not be consistent with Chrome, potentially causing user confusion.

Gusted (changes requested):
We should add a note in the ui just for firefox users that the permission won't be requested as it can't be a optional permissions as some will definitely open issues about it :)

bershanskiy:
Chrome (no changes)

image

Firefox (extra message only for "OFF" state)
image

src/ui/popup/settings-page/context-menus.tsx
```diff
@@ -29,7 +33,7 @@ export default function ContextMenusGroup(props: ViewProps) {
             label="Use context menus"
             description={props.data.settings.enableContextMenus ?
                 'Context menu integration is enabled' :
-                'Context menu integration is disabled'}
+                `Context menu integration is disabled${isFirefox ? '\nNote: conext menus permission is not used.' : ''}`}
```
Gusted, line 36:
```suggestion
                `Context menu integration is disabled${isFirefox ? '\nNote: context menus permission is not used.' : ''}`}
```

Gusted (changes requested):

src/ui/popup/settings-page/context-menus.tsx
bershanskiy: Good catch!

Gusted (approved):

PR #6869, opened 2021-09-22
Fix: make browser tests run entirely automatically

bershanskiy:
This makes browser tests run entirely automatically. This change closes `about:blank` page created by Puppeteer which was preventing browser shutdown. Now, at the end of every set of tests all tabs are closed automatically, allowing Puppeteer to close the whole browser and start the next set of tests.

### DIFF
--- a/tests/browser/environment.js
+++ b/tests/browser/environment.js
@@ -34,7 +34,12 @@ class PuppeteerEnvironment extends JestNodeEnvironment {
         this.assignTestGlobals();
 
         // Close sentinel about:blank page created by Puppeteer but not used in tests.
-        (await this.browser.pages())[0].close();
+        this.browser.pages().then((pages) => {
+            const sentinel = pages[0];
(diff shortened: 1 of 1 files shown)


bershanskiy:
Note that we probably could re-use this page in `createTestPage()`, but that would make code slightly harder to read.

tests/browser/environment.js
```diff
@@ -32,6 +32,9 @@ class PuppeteerEnvironment extends JestNodeEnvironment {
         this.global.page = this.page;
 
         this.assignTestGlobals();
+
+        // Close sentinel about:blank page created by Puppeteer but not used in tests.
+        (await this.browser.pages())[0].close();
```
Gusted, line 37: Good as is, but might be add a guard that the first page is `about:blank`

bershanskiy: I added this check, does it look better now?

Gusted (approved):

PR #6879, opened 2021-09-23
Virustotal text mode is dark

nathanielcwm:


Gusted (commented):
The link gets redirected to a 404

nathanielcwm:
@Gusted weird looks like it only works if you keep the trailing slash

Gusted (approved):

PR #6888, opened 2021-09-24
Fix for bankofamerica.com

andryandrew:
The first fix (.fsdnav-sub-nav-left) fixes the hover submenus (Accounts, Bill Pay...) which are otherwise unreadable over a transparent background.
The second section is cosmetic, as the background images of some elements get badly inverted, but that's IMHO. Feel free to just push the first modification if you don't agree.

If needed, I can provide before/after photos from the user dashboard (which is accessible only with an account)

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1514,6 +1514,18 @@ img[alt="Bankier.TV"]
 
 ================================
 
+bankofamerica.com
+
+CSS
+.fsdnav-sub-nav-left {
+    background-color: var(--darkreader-neutral-background) !important;
+}
+.ad-acct-layout,.thrfwl-body,.ad-summary-container,.olb-summary-widget-container {
+    background-image: none !important;
+}
+
+================================
+
 basecamp.com
 
 INVERT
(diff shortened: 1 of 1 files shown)


Gusted (approved):
Thank you!

src/config/dynamic-theme-fixes.config
```diff
@@ -1514,6 +1514,20 @@ img[alt="Bankier.TV"]
...
 
+bankofamerica.com
+
+CSS
+.fsdnav-sub-nav-left {
+    background-color: var(--darkreader-neutral-background) !important;
+}
+.ad-acct-layout,.ad-acct-detail-card-layout,
+.thrfwl-body,
+.ad-summary-container,.olb-summary-widget-container {
```
Gusted, line 1525: Woops, missed this. `ad` are those advertisements? While I do support and recommend using adblock we shouldn't be the one removing it.

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
andryandrew: I was worried too, but even with no adblocker active they do not seem to contain any ad.
For example, these are `ad-summary-container` and `.olb-summary-widget-container`:

**No Darkreader**
No Darkreader

**Darkreader**
Darkreader

**Darkreader fixed**
Darkreader fixed

Gusted (approved):

PR #6897, opened 2021-09-24
Fix: Make chromium happy

### LINKED ISSUE #6892: [Bug] Chrome Extension: Unchecked runtime.lastError: The message port closed before a response was received.
# Bug Report

## Bug Description
Viewing any website/page causes an error in the console: 
```
Unchecked runtime.lastError: The message port closed before a response was received.
```

## Website Address
Any website.

## Steps To Reproduce
- open any website
- open the dev tools > console
- an error is written to the console  (see above)
- refresh the page; error is repeated

## Expected Behavior
No errors are thrown or written to the console.

## Actual Behavior
Error is written out to the console.

## Screenshots
n/a

## System Information
- OS: Windows 10 x64 
- Windows Version: 20H2
- Windows Build: 19042.1237
- Windows Feature Experience Pack 120.2212.3530.0
- Browser: Chrome Version 94.0.4606.54 (Official Build) (64-bit)
- Dark Reader Version: 4.9.37

## Additional Context
Toggled all my extensions off until I found the extension throwing the above error. Disabling Dark Reader and the error is no longer thrown. Enabling Dark Reader and the error is shown again.

## Possible Resolution?
- Stackoverflow: https://stackoverflow.com/a/56483156/9664
- Be sure to read the comments of the linked answer as well; accepted answer is crap.

Gusted:
- Fixes the `chrome.runtime.lastError` message.
- Ensure Chromium does like our current message system.
- Resolves #6892

### DIFF
--- a/src/background/tab-manager.ts
+++ b/src/background/tab-manager.ts
@@ -102,7 +102,7 @@ export default class TabManager {
                         } else {
                             sendResponse('unsupportedSender');
                         }
-                        return;
+                        return true;
                     }
 
                     const tabId = sender.tab.id;
@@ -211,6 +211,8 @@ export default class TabManager {
                     break;
                 }
             }
+            sendResponse('Nothing-to-be-seen-here');
+            return true;
         });
     }
 
--- a/src/background/tab-manager.ts
+++ b/src/background/tab-manager.ts
@@ -211,7 +211,7 @@ export default class TabManager {
                     break;
                 }
             }
-            sendResponse('Nothing-to-be-seen-here');
+            sendResponse({type: '¯\\_(ツ)_/¯'});
             return true;
         });
     }


Gusted:
@bershanskiy or @alexanderby Could one of you have a look and merge it?

src/background/tab-manager.ts
```diff
@@ -211,6 +211,8 @@ export default class TabManager {
                     break;
                 }
             }
+            sendResponse('Nothing-to-be-seen-here');
```
alexanderby, line 214: Maybe worth sending something like {type: 'anything'}? Less chance to break anything, like object destructuring.

alexanderby (approved):

src/background/tab-manager.ts
Gusted: A lemmy face should do it :)

src/background/tab-manager.ts
```diff
@@ -211,6 +211,8 @@ export default class TabManager {
                     break;
                 }
             }
+            sendResponse({type: '¯\\_(ツ)_/¯'});
```
alexanderby, line 214: The most hilarious response 🤣

alexanderby (approved):

bershanskiy:
I'm late, but technically we could write `return false;` since all calls to `sendResponse()` happen within the function. But that doesn't change anything AFAIK.

I'm curious why I didn't notice the error during development.

PR #6915, opened 2021-09-25
Fix etsy.com stars when rating item

SithLadyRaven:
Fixes review stars when rating an item. Previously all stars would be marked regardless of what was actually selected.

First fix:
Before:
Screen Shot 2021-09-25 at 4 21 15 PM

After:
Screen Shot 2021-09-25 at 4 22 01 PM

Second fix:
Before:
Screen Shot 2021-09-25 at 4 19 36 PM

After:
Screen Shot 2021-09-25 at 4 20 20 PM


Gusted (approved):
Thank you!

PR #6916, opened 2021-09-25
Fix: use null for missing empty node

### LINKED ISSUE #6912: [Bug] Automation settings page is blank on Firefox Android 92
# Bug Report

## Bug Description
The automation page is blank and cannot change automation settings. 

## Steps To Reproduce

Go to darkreader settings and the settings and automation

## Expected Behavior

Display time schedule etc

## Actual Behavior

Blank page
Screenshot_20210925-125022_Fennec

## System Information

- OS: Android 11
- Browser: Firefox Android 92.1.1
- Dark Reader Version: 4.9.37.1

bershanskiy:
Fixes #6912.

My earlier commit, #6761 incorrectly returned `false` to denote empty node. Now it correctly returns `null`.

### DIFF
--- a/src/ui/popup/settings-page/context-menus.tsx
+++ b/src/ui/popup/settings-page/context-menus.tsx
@@ -27,7 +27,7 @@ export default function ContextMenusGroup(props: ViewProps) {
         }
     }
 
-    return !isMobile && (
+    return isMobile ? null : (
         <CheckButton
             checked={props.data.settings.enableContextMenus}
             label="Use context menus"


bershanskiy:
I think we should add E2E tests to avoid such bugs in the future. However, I'm not sure how we would run Firefox for Android builds. Perhaps, a VM?

Gusted:
> I think we should add E2E tests to avoid such bugs in the future. However, I'm not sure how we would run Firefox for Android builds. Perhaps, a VM?

Its not really possible to test that.

Gusted (approved):

PR #6923, opened 2021-09-26
Fix Vivaldi detector; update other browsers detectors

bershanskiy:
Fixes `isVivaldi` utility. While at it, also updates `isOpera` and `isMobile` to use the new `userAgentData` API.


src/utils/platform.ts
```diff
@@ -4,9 +4,15 @@ const platform = typeof navigator === 'undefined' ? 'some platform' : navigator.
 export const isChromium = userAgent.includes('chrome') || userAgent.includes('chromium');
 export const isThunderbird = userAgent.includes('thunderbird');
 export const isFirefox = userAgent.includes('firefox') || isThunderbird;
-export const isVivaldi = (function(){if (typeof navigator === 'object' && typeof (navigator as any).userAgentData === 'object') return false; const match = userAgent.match(/chrome\/[0-9]*/); return match && match[0] && Number(match[0].substring('chrome/'.length)) > 92})() || userAgent.includes('vivaldi');
+export const isVivaldi = (function () {
```
Gusted, line 7: A comment maybe?

bershanskiy: I added a few comments.

src/utils/platform.ts
```diff
@@ -4,14 +4,33 @@ const platform = typeof navigator === 'undefined' ? 'some platform' : navigator.
 export const isChromium = userAgent.includes('chrome') || userAgent.includes('chromium');
 export const isThunderbird = userAgent.includes('thunderbird');
 export const isFirefox = userAgent.includes('firefox') || isThunderbird;
-export const isVivaldi = userAgent.includes('vivaldi');
+// Vivaldi attempts to hide itself from User Agent sniffing so we use the following facts:
+//  1. Vivaldi does not support navigator.userAgentData at all (it is undefined)
```
Gusted, line 8: image
I assume this is only true for non-devtools context?

bershanskiy: I looked into it and in extension contexts `userAgentData` is supported, but does not contain Vivaldi name.

bershanskiy:
Current Vivaldi detector is still defective: it looks for `Vivaldi` in User Agent (which works for very old versions) and looks for `navigator.userAgentData.brands.length` (which is supported by modern Vivaldi). This still leaves a gap for Vivaldi which does not have either attribute. I think detector works incorrectly on Vivaldi [2.10](https://vivaldi.com/press/releases/vivaldi-2-10-no-strings-attached/) to 3.7 (because it [updates to Chromium 90](https://vivaldi.com/blog/vivaldi-crumbles-cookie-dialogs-raises-privacy/), which is [the first version to support `userAgentData`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/userAgentData#browser_compatibility)).

PR #6924, opened 2021-09-26
Add Safari to the README

amadejpapez:
Dark Reader is now available on Safari for Mac, iPhone and iPad. Having it in the README offers more people using these devices to discover and use it.

This PR adds Safari button with the same style to the README. When clicked it goes to the https://darkreader.org/safari/ page as that is what the Safari button does on the website. If you want, I can change it to link directly to the App Store. :)


Gusted (approved):
Looks all right to me.

PR #6939, opened 2021-09-28
fix: ensure proxy is run immeditally

### LINKED ISSUE #5508: [Broken Website]  ilsole24ore.com
Articles pages (not the main page) in this website return the message: "Something went wrong."
A workaround is to disable darkreader for the website (Alt+Shift+A), reload the page (F5), and enable darkreader again (Alt+Shift+A).

Try for example:
https://www.ilsole24ore.com/art/vacanze-mare-italia-e-pochi-viaggi-all-estero-ecco-dove-andremo-quest-estate-ADNxswVB
but it's the same for most of the articles.

Gusted:
- Fixes a issue whereby the proxy script would be run as last entry due to the usage of `src` it would be seen as a external link and put on the back of the "stack" before getting it loaded.
- Removing the blob and going back to the old behavior of textContent should be fine for now.
- Resolves #5508

### DIFF
--- a/src/inject/dynamic-theme/index.ts
+++ b/src/inject/dynamic-theme/index.ts
@@ -134,12 +134,8 @@ function createStaticStyleOverrides() {
     document.head.insertBefore(rootVarsStyle, variableStyle.nextSibling);
 
     const proxyScript = createOrUpdateScript('darkreader--proxy');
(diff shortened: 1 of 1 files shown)


Gusted:
@alexanderby just FYI that the blob change is now reverted ;) You were right it was a dangerous PR.

alexanderby:
Glad you found out, welldone.

PR #6942, opened 2021-09-28
perf: only do 1 passtrough a expansive task

Gusted:
- The `iterateCSSRules` and `iterateCSSDeclarations` function can be expensive on a lot of rules.
- With this commit we're making sure that the `rulesQueue` is only passed trough this expensive functions once.
- This removes around 200ms+ from sites with a lot of variable rules(e.g. github.com).


Gusted:
Tests are passing, no comprises were made. Merging.

PR #6950, opened 2021-09-29
Add Need for Speed fandom wiki to dark sites exception list

nobelharvards:


src/config/dark-sites.config
```diff
@@ -636,6 +637,7 @@ pr0gramm.com
 pre.fyp.nl
 premid.app
 primevideo.com
+privacytools.io
```
Gusted, line 640: privacytools.io doesn't meet [the requirements](https://github.com/darkreader/darkreader/blob/master/CONTRIBUTING.md#adding-a-website-that-is-already-dark).

Gusted (approved):
Woops, sorry for the late response seems like this Email got caught up in spam.

PR #6951, opened 2021-09-29
fix: add mutex to writing

### LINKED ISSUE #6848: [Bug] Dark reader breaks upon attempting to modify any settings
# Bug Report

## Bug Description

Dark reader breaks upon attempting to modify any settings.

## Website Address

Any

## Steps To Reproduce

1. Install Dark reader on firefox 92
2. Attempt to change the settings on any website

## Expected Behavior

Options work as intended

## Actual Behavior

Extension just stops working

## Video

https://ttm.sh/eDd.mp4

## System Information

- OS: Arch Linux
- Browser: Firefox 92, also tested on Firefox Developer Edition 93
image
image

- Dark Reader Version: 
Latest

## Additional Context

This occurred after I updated the firefox package on my system (https://archlinux.org/packages/extra/x86_64/firefox/)

### LINKED ISSUE #6873: [Bug] Settings won't persist anymore
# Bug Report

## Bug Description
Since last update (4.9.36), whenever I change any setting, close the Dark Reader dialog, and reopen it, the setting is reverted to the previous value.
This did not happen in previous versions (at least in 4.9.34)

## Steps To Reproduce
- Update to latest version (4.9.36)
- Change any setting (reduce brightness for example)
- You'll see that the setting is changed correctly (the brightness is indeed reduced)
- Close the settings dialog
- Reopen the settings dialog

## Expected Behavior
The brightness value should have persisted

## Actual Behavior
The brightness value is shown with the original value

## Screenshots
https://i.imgur.com/ckjBkn0.mp4

## System Information

- OS: Ubuntu MATE 20.04
- Browser: Firefox 92.0
- Dark Reader Version: 4.9.36

## Additional Context
I can reproduce the issue on a clean profile.

I can see the following error in the browser console:
```
Promise rejected after context unloaded: Actor 'Conduits' destroyed before query 'RuntimeMessage' was resolved
index.js:1312
    disconnect moz-extension://07fd650a-7e0d-4a0a-aa37-791aa18fae1c/ui/popup/index.js:1312
```

EDIT: I have just realized that the settings *do* persist, if I immediately close Firefox after changing them (i.e. before opening the settings dialog for a second time)

Gusted:
- This patch fixes 2 issues.
- 1) For some reason it was possible for a data race to occur(I still haven't a clue how this happend) when writing data. So it's now behind a mutex lock which solves this, and could be removed later when I find why the f*ck a data race can occur.
- 2) The mutex provides a method to check if it's locked. And the lock functionallity is async I've added a setTimeout to 500ms(this should be within error or of marigin) so that the call from `subscribeToOuterSettingsChange` can be executed and seen that the mutex is locked and shouldn't overwrite the data with the old data which it previously had.
- Resolves a bunch of issues.
- Resolves #6848
- Resolves #6873

CC @alexanderby And there we go. Only took more than 7 days to find the stupid root cause and fix it properly.

### DIFF
--- a/src/background/utils/extension-api.ts
+++ b/src/background/utils/extension-api.ts
@@ -72,6 +72,7 @@ export async function writeSyncStorage<T extends {[key: string]: any}>(values: T
         chrome.storage.sync.set(values, () => {
             if (chrome.runtime.lastError) {
                 reject(chrome.runtime.lastError);
+                mutexStorageWriting.unlock();
                 return;
             }
             resolve();
--- a/src/background/utils/extension-api.ts
+++ b/src/background/utils/extension-api.ts
@@ -1,5 +1,6 @@
 import {isPDF} from '../../utils/url';
 import {isFirefox, isEdge} from '../../utils/platform';
+import {Mutex} from '../../utils/mutex';
 
 declare const browser: {
     commands: {
@@ -37,7 +38,7 @@ export function canInjectScript(url: string) {
     );
 }
 
-let isWriting = false;
+const mutexStorageWriting = new Mutex();
 
 export async function readSyncStorage<T extends {[key: string]: any}>(defaults: T): Promise<T> {
     return new Promise<T>((resolve) => {
@@ -66,32 +67,32 @@ export async function readLocalStorage<T extends {[key: string]: any}>(defaults:
 }
 
 export async function writeSyncStorage<T extends {[key: string]: any}>(values: T): Promise<void> {
-    return new Promise<void>((resolve, reject) => {
-        isWriting = true;
+    return new Promise<void>(async (resolve, reject) => {
+        await mutexStorageWriting.lock();
         chrome.storage.sync.set(values, () => {
             if (chrome.runtime.lastError) {
                 reject(chrome.runtime.lastError);
                 return;
             }
             resolve();
-            setTimeout(() => isWriting = false);
+            setTimeout(() => mutexStorageWriting.unlock(), 500);
         });
     });
 }
 
 export async function writeLocalStorage<T extends {[key: string]: any}>(values: T): Promise<void> {
-    return new Promise<void>((resolve) => {
-        isWriting = true;
+    return new Promise<void>(async (resolve) => {
+        await mutexStorageWriting.lock();
         chrome.storage.local.set(values, () => {
             resolve();
-            setTimeout(() => isWriting = false);
+            setTimeout(() => mutexStorageWriting.unlock(), 500);
         });
     });
 }
 
 export const subscribeToOuterSettingsChange = (callback: () => void) => {
     chrome.storage.onChanged.addListener(() => {
-        if (!isWriting) {
+        if (!mutexStorageWriting.isLocked()) {
             callback();
         }
     });
--- a/src/utils/mutex.ts
+++ b/src/utils/mutex.ts
@@ -1,6 +1,4 @@
-
 import {logWarn} from './log';
-
 
 /**
   * This is the Mutex(Not to confuse with RW-Mutex).
--- a/src/utils/mutex.ts
+++ b/src/utils/mutex.ts
@@ -1,0 +1,52 @@
+
+// This is the Mutex(Not to confuse with RW-Mutex). This is to ensure that certain access are only done
+// by 1 callback/operation(a current fallpit of PromiseBarrier). This ensures that their is a global mutex
+// Which can be called by a finite amount of callbacks. We should ensure that the mutex can only be locked
+// by the first callback which has requrested it and only give the next callback the mutex lock once it's
+
+import {logWarn} from './log';
+
+// unlocked by the callback.
+export class Mutex {
+    private awaitingResolves = [] as Array<() => void>;
+    private locked = false;
+
+    public isLocked() {
+        return this.locked;
+    }
+
+    // Request a lock from the mutex.
+    public async lock() {
+        // Easiest path, the mutex is not locked.
+        // We lock the mutex and
+        if (!this.locked) {
+            this.locked = true;
+            return;
+        }
+        return new Promise<void>((resolve) => {
+            // Harder path, we will push the resolve
+            // into the array and let it be for now.
+            this.awaitingResolves.push(resolve);
+        });
+    }
+
+    // Unlock the mutex.
+    public unlock() {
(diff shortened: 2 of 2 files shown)


Gusted:
To add more context of why it only happening now, the ` chrome.storage.onChanged.addListener` callback got behind the `setTimeout(() => isWriting = false);` callback due to our new usage of promiseBarriers in the `saveSettingsIntoStorage` function. So it wouldn't see that `isWriting` was true. Which is now properly fixed with this patch ;)

bershanskiy:
There is a [Web Locks API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API) which provides this functionality natively. Have you considered using it? (Available in Chromium but not Firefox and Safari.)

src/background/utils/extension-api.ts
```diff
@@ -66,32 +67,32 @@ export async function readLocalStorage<T extends {[key: string]: any}>(defaults:
 }
 
 export async function writeSyncStorage<T extends {[key: string]: any}>(values: T): Promise<void> {
-    return new Promise<void>((resolve, reject) => {
-        isWriting = true;
+    return new Promise<void>(async (resolve, reject) => {
+        await mutexStorageWriting.lock();
         chrome.storage.sync.set(values, () => {
             if (chrome.runtime.lastError) {
```
bershanskiy, line 73: In case of an error this code seems to never release the lock leading to a deadlock. Error might be temporary, but there is no mechanism for resolving the deadlock.

src/background/utils/extension-api.ts
```diff
@@ -66,32 +67,32 @@ export async function readLocalStorage<T extends {[key: string]: any}>(defaults:
...
                 reject(chrome.runtime.lastError);
                 return;
             }
             resolve();
-            setTimeout(() => isWriting = false);
+            setTimeout(() => mutexStorageWriting.unlock(), 500);
```
bershanskiy, line 78: Why 500? Is it a magic number?

bershanskiy:
Why do sync storage and local storage share the same mutex? These appear to be two different resources in browser's view so locking both of them on write to either one might be unnecessary.

bershanskiy:
I don't like `setTimeout` with some fixed timeout (`500` in this case) because the real timeout will most likely depend on current system and current resource load. I believe, `StateManager` solves the same problem but in a bit more elegant manner: it just records when we are in overcommitted state (it calls it "Saving Override") and waits for current obsolete write operation to end and starts the next write operation right after the obsolete one ends.

Gusted:
> There is a [Web Locks API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API) which provides this functionality natively. Have you considered using it? (Available in Chromium but not Firefox and Safari.)

I did consider it, but we will end up with more code to this. As the whole syntax of the WebLock is different it cannot really be used in par with the JS solution.

src/background/utils/extension-api.ts
Gusted: Woops, yup correct. Good catch!

src/background/utils/extension-api.ts
Gusted: It is a magic number, but it will ensure that the promises of `saveSettingsIntoStorage` are run etc. So that the call from `chrome.storage.onChanged.addListener` can be runned. Also it should prevent that this callback is given a priority due to the none-usage of timeout Number. `500` is a pretty big number so we wouldn't hit any timing races etc.

Gusted:
> Why do sync storage and local storage share the same mutex? These appear to be two different resources in browser's view so locking both of them on write to either one might be unnecessary.

Correctly, that will be fixed ;)

Gusted:
> I don't like `setTimeout` with some fixed timeout

Neither do I, luckily we're on the same page. But we could add the unlock into the onChanged callback, as that is the main culprit currently.



>  I believe, `StateManager` solves the same problem but in a bit more elegant manner: it just records when we are in overcommitted state (it calls it "Saving Override") and waits for current obsolete write operation to end and starts the next write operation right after the obsolete one ends.

Hmm, that is possible. I'm not sure if it fit's the use-case. As we would just ensure that the write(either sync or local) shouldn't be happening within each other within a short time-frame and ensure that's 1 global mutex/lock state, so that `onChanged` correctly can see if we are writing it or another copy of the DR(on another machine) and should save those settings.

src/utils/mutex.ts
```diff
@@ -0,0 +1,52 @@
+
+// This is the Mutex(Not to confuse with RW-Mutex). This is to ensure that certain access are only done
```
alexanderby, line 2: I suggest to add it to the /** ... */ JSDoc comment before a class declaration. Please make the comment lines not so long

src/utils/mutex.ts
```diff
@@ -0,0 +1,54 @@
+
```
alexanderby, line 1: Come on, too much whitespaces :)

alexanderby (approved):
Looks great.

alexanderby:
Let's merge and test it all together. Or did you want to add something?

bershanskiy:
There is still one more thing from above discussion:
> >Why do sync storage and local storage share the same mutex? These appear to be two different resources in browser's view so locking both of them on write to either one might be unnecessary.
>
>Correctly, that will be fixed ;)

src/utils/mutex.ts
```diff
@@ -0,0 +1,52 @@
...
+    private locked = false;
+
+    public isLocked() {
+        return this.locked;
+    }
+
+    // Request a lock from the mutex.
+    public async lock() {
+        // Easiest path, the mutex is not locked.
+        // We lock the mutex and
```
bershanskiy, line 21: > We lock the mutex and

Such a cliffhanger. Let's finish the sentence. :)

Gusted:
> There is still one more thing from above discussion:

I've played with it a little bit, and it will just add more complexity to the current system. As their is no real multi-threading the using 1 mutex shouldn't be a big problem. The `onChanged` seems to like it as it is.

src/utils/mutex.ts
Gusted: Ah sigh, not sure what VSCode did with my comments today. Maybe time to search for alternatives.

Gusted:
Welp, let's hope it doesn't break other code :)

PR #6990, opened 2021-10-02
Refactor: use unique message ID for WS test messages

bershanskiy:
Previously, resolvers and rejecters were keyed on message type, which is not unique. After this patch, they are keyed on unique ID (which is just a simple counter). This is part of work which will allow us to run more tests simultaneously.

Also, previously error handling was messed up. It was:
```
if (message.type === 'error') {
    const reject = rejectors.get(message.type);
    reject(message.data);
```
But there was not any value in `rejectors` map with key `'error'`. This bug was never triggered since there are no tests which are expected to error out.


Gusted (approved):
Logic seems good to me.

PR #6992, opened 2021-10-02
Add livesplit.org to global dark list

CryZe:
Especially the web version of LiveSplit is harmed by Dark Reader, because you can set your own colors there which end up looking different than what you choose if Dark Reader is active.


Gusted (approved):
Thanks!

PR #7001, opened 2021-10-03
Fix for nytimes.com

c-nagy:
Invert WhatsApp share icon. In this edge case, the SVG icon is set as the CSS background-image, so I don't see a way to more elegantly fix this without needing to override the border and background colors as well to match the other share icons.


Gusted (approved):
Thanks!

PR #7086, opened 2021-10-13
fix: correctly remove moves out additions

Gusted:
- Fix a special case with the `getElementsTreeOperations` function whereby the 
element would first be removed(and still be connected while not) and than 
another element would be back. But the moves.forEach would remove it from the 
additions set because it doesn't take the order into account.
- Moving the .delete call to within the existing loop fixes this issue(as it 
will now take into account the order) and also removes another loop.
- Fixes a github navigation bug.


Gusted:
More details: This fixes a issue with the github header which has a inline element which would be otherwise not have get modified again due to removal and addition of the element when navigation to another page(Github avoids whole new page loads).

PR #7093, opened 2021-10-14
Fix MURAL tabs in dynamic theme

alejandro5042:
Icons were almost impossible to see.

Note: The sidebar is not always open. Just when you are looking for icons or other content. So that it's not fully dark is probably OK, since its not always in your view. Besides, MURALs are generally not dark.

Before:

image

After:

image

Open for color suggestions!

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1060,6 +1060,15 @@ img:not([src*="png"]):not([src*="svg"]) {
 
 ================================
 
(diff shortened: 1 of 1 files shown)


Gusted:
Hi @alejandro5042!

What do you think about changing this fix?
I think it will be nicer to invert the icons(so they will be white) and leave the background color as-is. 

The fix will look something like:

```CSS
app.mural.co

INVERT
.generic-selector-for-icons
```

PR #7110, opened 2021-10-16
Add pre-defined color-schemes

Gusted:
- Add a _first iteration_ contributing guide.
- Add a human readable config syntax and a parser with a lot of guards. Performance is around (150K Op/sec).
- Add _popular_ pre-defined color schemes to Dark Reader and be able to select them. If any of the colors are changed by the user the color-scheme option is reset as we _currently_ do not have this ability and should be done for a futher PR.
- Resolves

### DIFF
--- a/src/background/config-manager.ts
+++ b/src/background/config-manager.ts
@@ -100,7 +100,6 @@ export default class ConfigManager {
         this.handleColorSchemes();
     }
 
-
     private async loadDarkSites({local}: Config) {
         const sites = await this.loadConfig({
             name: 'Dark Sites',


Gusted:
> The syntax here looks a lot like YAML, so perhaps instead of re-implementing a special parser we could re-use a YAML library?

That would be possible - even though that will make it more out-of-sync with our current syntax-like configs.

Another thing is that we will have to "loop" twice trough it - once to parse the YAML and then to check for specifications errors within the defined YAML object, I'm not sure if that's able to be done within < 250 LOC as this parser without it's comments is already large in my opinion.

bershanskiy:
Yes, this looks ready to merge (aside from a minor nit). Good job!

metafates:
sorry if it's a silly question, but how can I use theses themes? I couldn't find an option to do, I'm really confused...

Gusted:
Click Dark Reader icon.
    Click Dev tools (in the bottom-right corner).
    Click Preview new design.
    under Theme click See all options, Colors.

PR #7132, opened 2021-10-19
Refactor: stricter type declarations

bershanskiy:
This commit sets `@typescript-eslint/no-explicit-any` to `error` and resolves all the errors which arise from it. If resolving the root cause of `any` declaration is very cumbersome, this commit just adds `// eslint-disable-next-line @typescript-eslint/no-explicit-any` to ignore a particular occurrence of `any`.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -654,6 +654,7 @@ protondb.com
 proxx.app
 proxy.vulpes.one
 psi.cynicaloptimist.me
+psyoptions.io
 ptt.cc
 pxseu.com
 pylon.bot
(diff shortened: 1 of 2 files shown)


Gusted:
>  this commit just adds `// eslint-disable-next-line @typescript-eslint/no-explicit-any` to ignore a particular occurrence of `any`.

Never really been a fan of disabling the linter in the code-base, in fact we never had such occurrence.
I'd prefer to fix-up the types than to have such code living a long time in the code-base. 

If the tests are passing just-fine, give it a merge. But a follow-up PR will be made to remove those ignore lines.

Just a FYI 🤷🏽. Seems like a good a chance tho.

bershanskiy:
I think this is ready, but I want to do some manual testing before merging. It's getting late, so I'll test and merge this tomorrow.

PR #7152, opened 2021-10-21
fix: allow API to be compiled

Gusted:


bershanskiy:
Interesting, what was failing before this PR?

PR #7233, opened 2021-11-02
Add Geometry Dash Fandom Wiki to dark sites exceptions list

nobelharvards:


Gusted:
I personally don't think this site can be qualified of having a dark theme.

PR #7245, opened 2021-11-03
fix: Remove unnecessary  `|| null`

Gusted:
- Removing unneeded `|| null` as it otherwise already return null of some sort.

### DIFF
--- a/src/background/devtools.ts
+++ b/src/background/devtools.ts
@@ -237,7 +237,7 @@ export default class DevTools {
     }
 
     private async getSavedDynamicThemeFixes() {
-        return this.store.get(DevTools.KEY_DYNAMIC) || null;
+        return this.store.get(DevTools.KEY_DYNAMIC);
     }
 
     private saveDynamicThemeFixes(text: string) {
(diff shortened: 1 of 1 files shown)


bershanskiy:
Does this change do anything? As far as I know, this would just unwrap a promise and then re-wrap it again.

For example, the following two functions `passthroughNoAwait()` and `passthroughAwait()` behave identically:
```
async function f(){
    return new Promise((resolve) => setTimeout(() => resolve(true), 1000));
}

async function passthroughAwait() {
    return await f();
}

async function passthroughNoAwait() {
    return f();
}
```

Gusted:
> Does this change do anything? As far as I know, this would just unwrap a promise and then re-wrap it again.

Correctly, but that isn't necessary the case I'm trying to fix. Because in the `getXXX` function (For some reason I have included the `hasXXX`, which I will revert shortly).

Because we currently have a conditional `||`(or) in the code which as you know will return the right-hand value if the left-hand value is undefined/null.
Currently because we don't `await` the returning result from the storage.get (which is always a Promise) the left-hand will never be null. Looking at the code a bit better, we're better off by removing those `|| null` as it will only return the right-hand when left-hand is already null of some sort.

bershanskiy:
The updated commit makes sense to me, let's merge this!

PR #7256, opened 2021-11-04
fix: only add Microsoft fix to Microsoft

### LINKED ISSUE #7255: [Bug] Firefox white flashing with dynamic theme since 4.9.40.
# Bug Report

## Bug Description

Since the latest update, pages will flash white upon loading. Both foreground and background pages will flash white for a brief moment before the dynamic theme gets applied. This was not an issue before 4.9.40. Static filter continues to work fine.

## Website Address

Any website with dynamic mode enabled.

## Steps To Reproduce
Open website in foreground:
 - Page loads
 - White flash occurs
 - Dynamic theme is applied

or
Open website in background:
 - Wait for page to fully load in background
 - Upon switch to the loaded tab, white flash occurs
 - Dynamic theme is applied

## Expected Behavior

Not sure how it worked before the update, but background pages already had the dynamic theme applied when switching to tab.

Pages loaded in the foreground would honor the `nglayout.initialpaint.delay` configured in Firefox, and the dynamic theme would be applied before the page was shown, thus no white flashing.

## System Information

- OS: Windows 10 21H1
- Browser: Firefox 94
- Dark Reader Version: 4.9.40

Gusted:
- Only add the hack to microsoft.com domains and not to other websites, this fixes a regression with white flashes.
- Resolves #7255

### DIFF
--- a/src/inject/dynamic-theme/modify-css.ts
+++ b/src/inject/dynamic-theme/modify-css.ts
@@ -206,7 +206,8 @@ function getModifiedScrollbarStyle(theme: Theme) {
 export function getModifiedFallbackStyle(filter: FilterConfig, {strict}: {strict: boolean}) {
     const lines: string[] = [];
     // https://github.com/darkreader/darkreader/issues/3618#issuecomment-895477598
-    lines.push(`html, body, ${strict ? 'body :not(iframe):not(div[style^="position:absolute;top:0;left:-"]' : 'body > :not(iframe)'} {`);
+    const isMicrosoft = location.hostname.endsWith('microsoft.com');
+    lines.push(`html, body, ${strict ? `body :not(iframe)${isMicrosoft ? ':not(div[style^="position:absolute;top:0;left:-"]' : ''}` : 'body > :not(iframe)'} {`);
     lines.push(`    background-color: ${modifyBackgroundColor({r: 255, g: 255, b: 255}, filter)} !important;`);
     lines.push(`    border-color: ${modifyBorderColor({r: 64, g: 64, b: 64}, filter)} !important;`);
     lines.push(`    color: ${modifyForegroundColor({r: 0, g: 0, b: 0}, filter)} !important;`);


bershanskiy:
Even with this fix I still get white flashes.

bershanskiy:
I think this fix has a typo, 
```
lines.push(`html, body, ${strict ? `body :not(iframe)${isMicrosoft ? ':not(div[style^="position:absolute;top:0;left:-"]' : ''}'` : 'body > :not(iframe)'} {`);
```
should be
```
lines.push(`html, body, ${strict ? `body :not(iframe)${isMicrosoft ? ':not(div[style^="position:absolute;top:0;left:-"]' : ''}` : 'body > :not(iframe)'} {`);
```

bershanskiy:
@Gusted I can confirm that after applying this patch on top of this PR I no longer get white flashes:
```diff
diff --git a/src/inject/dynamic-theme/modify-css.ts b/src/inject/dynamic-theme/modify-css.ts
index e29d3cfb..b2052c34 100644
--- a/src/inject/dynamic-theme/modify-css.ts
+++ b/src/inject/dynamic-theme/modify-css.ts
@@ -207,7 +207,7 @@ export function getModifiedFallbackStyle(filter: FilterConfig, {strict}: {strict
     const lines: string[] = [];
     // https://github.com/darkreader/darkreader/issues/3618#issuecomment-895477598
     const isMicrosoft = location.hostname.endsWith('microsoft.com');
-    lines.push(`html, body, ${strict ? `body :not(iframe)${isMicrosoft ? ':not(div[style^="position:absolute;top:0;left:-"]' : ''}'` : 'body > :not(iframe)'} {`);
+    lines.push(`html, body, ${strict ? `body :not(iframe)${isMicrosoft ? ':not(div[style^="position:absolute;top:0;left:-"]' : ''}` : 'body > :not(iframe)'} {`);
     lines.push(`    background-color: ${modifyBackgroundColor({r: 255, g: 255, b: 255}, filter)} !important;`);
     lines.push(`    border-color: ${modifyBorderColor({r: 64, g: 64, b: 64}, filter)} !important;`);
     lines.push(`    color: ${modifyForegroundColor({r: 0, g: 0, b: 0}, filter)} !important;`);
```

GitAcc53:
I have this problem https://github.com/darkreader/darkreader/issues/12944

What do I do exactly to fix the white flash bug?

PR #7264, opened 2021-11-05
Fix for udemy.com

felipebarbosaferreira:
The practice questions radio buttons were the same color even when selected. Now when selected it has highlighted color ;)

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -13725,6 +13725,15 @@ g#ubuntu-logo > path
 
 udemy.com
 
+CSS
+.radio:not(.radio-button) input[type="radio"]:checked + .radio-label::before, .radio-inline:not(.radio-button) input[type="radio"]:checked + .radio-label::before {
+    background-color: rgb(93, 122, 136);
+    box-shadow: rgb(27, 29, 29) 0px 0px 0px 2px inset;
+}
+div[class^='curriculum-item-view--scaled-height-limiter'], div[class*=' curriculum-item-view--scaled-height-limiter'] {
+    max-height: 100% !important;
+}
+
 IGNORE INLINE STYLE
 [data-purpose="star-rating-mask"] > rect
 


src/config/dynamic-theme-fixes.config
```diff
@@ -13725,6 +13725,15 @@ g#ubuntu-logo > path
 
 udemy.com
 
+CSS
+.radio:not(.radio-button) input[type="radio"]:checked + .radio-label::before, .radio-inline:not(.radio-button) input[type="radio"]:checked + .radio-label::before {
+    background-color: rgb(93, 122, 136);
+    box-shadow: rgb(27, 29, 29) 0px 0px 0px 2px inset;
+}
+div[class^='curriculum-item-view--scaled-height-limiter'], div[class*=' curriculum-item-view--scaled-height-limiter'] {
```
Gusted, line 13733: Ehm, we should dark reader change the max-height?

src/config/dynamic-theme-fixes.config
```diff
@@ -13725,6 +13725,15 @@ g#ubuntu-logo > path
 
 udemy.com
 
+CSS
+.radio:not(.radio-button) input[type="radio"]:checked + .radio-label::before, .radio-inline:not(.radio-button) input[type="radio"]:checked + .radio-label::before {
+    background-color: rgb(93, 122, 136);
```
Gusted, line 13730:
```suggestion
    background-color: ${#6f92a3} !important;
```
What about changing it a little bit.

src/config/dynamic-theme-fixes.config
```diff
@@ -13725,6 +13725,15 @@ g#ubuntu-logo > path
...
+    box-shadow: rgb(27, 29, 29) 0px 0px 0px 2px inset;
```
Gusted, line 13731: Is the box-shadow necessary to be modified?

src/config/dynamic-theme-fixes.config
```diff
@@ -13725,6 +13725,15 @@ g#ubuntu-logo > path
 
 udemy.com
 
+CSS
+.radio:not(.radio-button) input[type="radio"]:checked + .radio-label::before, .radio-inline:not(.radio-button) input[type="radio"]:checked + .radio-label::before {
```
Gusted, line 13729:
```suggestion
.radio:not(.radio-button) input[type="radio"]:checked + .radio-label::before,
.radio-inline:not(.radio-button) input[type="radio"]:checked + .radio-label::before {
```

src/config/dynamic-theme-fixes.config
felipebarbosaferreira: my 2560x1080 (ultra wide) screen

### without the change

Screenshot_20211105_092014


### with the change, the space of the content is for the questions is proportional to the screen 

Screenshot_20211105_091926

Gusted: Is that a bug that Dark Reader caused? Otherwise it isn't a fix, just a cosmetic change.

felipebarbosaferreira: ```we should dark reader change the max-height?```


for me the change makes sense, but I think the site that should fit my screen

dark reader is more for theme, not layout. but it has helped me a lot at this point

the decision is yours ;)

if not added by you, I'll have to keep putting it in the dynamic theme every time (it's not saving the settings)

src/config/dynamic-theme-fixes.config
felipebarbosaferreira: I agree \o/

src/config/dynamic-theme-fixes.config
felipebarbosaferreira: it's not necessary, I was trying to make it smoother

src/config/dynamic-theme-fixes.config
felipebarbosaferreira: agree, clean look

src/config/dynamic-theme-fixes.config
felipebarbosaferreira: > Is that a bug that Dark Reader caused? Otherwise it isn't a fix, just a cosmetic change.

not a bug

Gusted: > for me the change makes sense, but I think the site that should fit my screen

I get that, but this config is exclusively for fixes that is caused by Dark Reader.

src/config/dynamic-theme-fixes.config
Gusted: I consider it as a cosmetic change and not necessary a fix.

PR #7276, opened 2021-11-06
Fix Cloudflare Community forums

WalshyDev:
On the community forums, you cannot see visited posts because the background colour does not change. This adds a lighter dark to distinguish between visited and non-visited. 

Without fix:
image

With fix:
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2871,8 +2871,8 @@ CSS
 community.cloudflare.com
 
 CSS
-.visited {
-  background-color: #262626;
+:root {
+    --darkreader-text--primary-medium: ${gray} !important;
 }
 
 ================================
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -2868,6 +2868,15 @@ CSS
 
 ================================
 
+community.cloudflare.com
+
+CSS
+.visited {
+  background-color: #262626;
+}
```
Gusted, line 2876: What about having a fix like:
```suggestion
:root {
    --darkreader-text--primary-medium: ${gray} !important;
}
```

WalshyDev: Yeah, could work. My eyes aren't the best so I find mine much easier to distinguish between the two. If you wish to merge your fix in though, I am happy with it. I can still keep my fix for just me

image

Gusted: > If you wish to merge your fix in though

Yeah, the fixes config should stay as the original page(which also just modifies the text's color).

Gusted (approved):

PR #7281, opened 2021-11-06
Add beepbox.co

Sobsz:
https://www.beepbox.co/
Technically the [synth example](https://www.beepbox.co/synth_example.html) subpage isn't dark theme, but barely anyone visits it (let alone more than once), so I don't think it matters too much?


Gusted:
> Technically the [synth example](https://www.beepbox.co/synth_example.html) subpage isn't dark theme, but barely anyone visits it (let alone more than once), so I don't think it matters too much?

Let it be a exception for this time :)

Gusted (approved):

PR #7287, opened 2021-11-07
Add kemono.party to dark-sites.config

oSumAtrIX:


Gusted:
On each request I'm getting a DDOS Guard page which is not dark.

PR #7324, opened 2021-11-10
refactor: prevent infinite loops

Gusted:
- Prevent infinite loops with `getRulesSync` and `getRulesAsync`.
- Prevent unresponsive tabs and prevent tabs from crashing, instead just let skip the CSSRules, which is beter than sucking up the CPU.

### DIFF
--- a/src/inject/dynamic-theme/style-manager.ts
+++ b/src/inject/dynamic-theme/style-manager.ts
@@ -23,8 +23,9 @@ declare global {
 
 export type StyleElement = HTMLLinkElement | HTMLStyleElement;
 
+export type detailsArgument = {secondRound: boolean};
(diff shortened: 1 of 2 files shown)


src/inject/dynamic-theme/style-manager.ts
```diff
@@ -23,8 +23,9 @@ declare global {
 
 export type StyleElement = HTMLLinkElement | HTMLStyleElement;
 
+export type detailsArgument = {secondRound: boolean};
```
alexanderby, line 26: Wonderful! One note: Uppercase names for types.

Gusted: Will be remembered for the next time!

PR #7327, opened 2021-11-10
Touchup some wealthfront.com items

mordavidj:
Icon wouldn't invert and button background was not showing against general background.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -14248,6 +14248,18 @@ INVERT
 
 ================================
 
+wealthfront.com
+
+INVERT
+.dashboard-path-goal-icon-component
+
+CSS
+.default__9vSXGQvh8MlK1JxJBq-k4 {
+    background-color: rgba(51, 42, 173, 0.2) !important;
+}
+
+================================
+
 weather.com
 
 INVERT


src/config/dynamic-theme-fixes.config
```diff
@@ -14248,6 +14248,18 @@ INVERT
 
 ================================
 
+wealthfront.com
+
+INVERT
+.dashboard-path-goal-icon-component
+
+CSS
+.default__9vSXGQvh8MlK1JxJBq-k4 {
```
Gusted, line 14257: Can this CSS selector be changed? As such CSS Selectors that include randomness will most likely change every few so.

src/config/dynamic-theme-fixes.config
```diff
@@ -14248,6 +14248,18 @@ INVERT
...
+    background-color: rgba(51, 42, 173, 0.2) !important;
```
Gusted, line 14258: `rgba(51, 42, 173, 0.2)`, could this be in the form of `${original color}`?

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
mordavidj: I'm not sure what you mean by `${original color}`. I just found the right color against the background and with a different transparency.

Gusted: I meant if you replace `original color` with the color that the element has(e.g. `#fef` so it becomes `${#fef}`) dark reader will modify that color according to the user's settings. We try to avoid hardcoded colors as much as possible.

src/config/dynamic-theme-fixes.config
Gusted: > I think you're right that it will change, but it is still the same after almost 7 months.

I prefer to not set such code-style, as other new contributes will point at this and say "but it's allowed here". So I'd prefer a consistent way of not using such random CSS selector.

PR #7333, opened 2021-11-11
Add rain.thecomicseries.com to dynamic-theme-fixes.config

abitenum:
This fixes a few issues with the webcomic's webpage using this. The page expects a few background colors to be specific values, so this fixes some CSS attributes to make it work correctly.

* By default, `body` tags have a small margin. On the website, the `html` tag doesn't have a background color, so this margin just uses the body's background. Dark Reader adds a background color attribute to the `html`, which causes a border to appear on the website. This removes the background color value so the body's background is used instead.
* The `body`'s background works by using a gradient PNG, followed by a fixed background color to continue the end of the gradient. Dynamic mode doesn't affect the background gradient, but does change the background color, which breaks the continuous color. This sets the background color to what it was originally.
* The comic pages on occasion have transparency (usually on multi-page posts) which expect the background color behind it to be white. This extension changes the background color, which messes it up a bit. This sets the background color for the `img` tag to white so that the image transparency works as intended.

The only other thing that I noticed was that the border color around the main `#container` is the same as its background color both on the original site and in light mode, but is different in dark mode. Attempting to fix this by setting the border color to transparent looked worse due to the background gradient, so it seems that having the border color helps. If anything, the border color should maybe be forced to be brighter than the gradient, but I couldn't figure out a good way of doing that, so I left that out.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -11040,11 +11040,10 @@ INVERT
 rain.thecomicseries.com
 
 CSS
-html {
-    background-color: revert !important;
-}
 body {
     background-color: #333333 !important;
+    margin: 0px !important;
+    padding: 8px !important;
 }
 #comicimage {
     background-color: #ffffff !important;
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -11037,6 +11037,21 @@ INVERT
 
 ================================
 
+rain.thecomicseries.com
+
+CSS
+html {
+    background-color: revert !important;
+}
+body {
+    background-color: #333333 !important;
+}
+#comicimage {
+    background-color: #ffffff !important;
+}
+
+================================
+
 rapidtables.com
 
 IGNORE INLINE STYLE
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -11041,7 +11041,8 @@ rain.thecomicseries.com
 
 CSS
 body {
-    background-color: #333333 !important;
+    background-color: var(--darkreader-neutral-background) !important;
+    background-image: none !important;
     margin: 0px !important;
     padding: 8px !important;
 }


src/config/dynamic-theme-fixes.config
```diff
@@ -11037,6 +11037,21 @@ INVERT
 
 ================================
 
+rain.thecomicseries.com
+
+CSS
+html {
+    background-color: revert !important;
+}
```
Gusted, line 11045: We shouldn't remove the background-color of `html` - especially not with `revert`. The background-color of html is used for initial rendering, removing the background-color will make the initial coloring white, which causes problems like white flashes/

src/config/dynamic-theme-fixes.config
```diff
@@ -11037,6 +11037,21 @@ INVERT
...
+body {
+    background-color: #333333 !important;
```
Gusted, line 11047:
```suggestion
    background-color: var(--darkreader-neutral-background) !important;
```

src/config/dynamic-theme-fixes.config
```diff
@@ -11037,6 +11037,21 @@ INVERT
...
+}
+#comicimage {
+    background-color: #ffffff !important;
```
Gusted, line 11050:
```suggestion
    background-color: var(--darkreader-neutral-text) !important;
```

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
abitenum: Ok, I can see how this does make sense.

The alternative fix would be to explicitly set the margin for the `body` tag to `0px`, which would remove the outline area. To keep the content layout looking the same, it should add padding equal to the margin being removed. I'll update the PR with this.

src/config/dynamic-theme-fixes.config
abitenum: This actually goes in the wrong direction. As I explained, the issue is that the background color should match the background image, and the only real way to do this is to fix the color to an exact color value. Below is a composite screenshot of what is happening. Left is what Dark Reader does by default, middle is my proposed change that is identical to the original site (which uses a light mode `#container` with a dark background behind it), and left is your proposed change. The screenshots were taken using the dark mode setting, and the issue is reversed when using the light mode setting in Dark Reader.

darkrain

src/config/dynamic-theme-fixes.config
abitenum: Like with my last comment, the fixed color is the point. For context, here's an example of what's going on before applying any fixes:

gaprain

You can see there's a black line in that transparent gap that's supposed to act like a line break. On the original page, the white background blended into the white page borders, making the gap invisible other than the black line. I'm working on the assumption that this is how it's intended to look, hence why I set the background color for the `img` tag to white. Changing it to the neutral text color just looks off. In dark mode, the black line is more visible, but it still doesn't blend with the page borders, and in light mode it just looks horrible.

src/config/dynamic-theme-fixes.config
Gusted: > This actually goes in the wrong direction. As I explained, the issue is that the background color should match the background image, and the only real way to do this is to fix the color to an exact color value. Below is a composite screenshot of what is happening. Left is what Dark Reader does by default, middle is my proposed change that is identical to the original site (which uses a light mode `#container` with a dark background behind it), and left is your proposed change. The screenshots were taken using the dark mode setting, and the issue is reversed when using the light mode setting in Dark Reader.

Just remove the background-image?

src/config/dynamic-theme-fixes.config
```diff
@@ -11037,6 +11037,21 @@ INVERT
 
 ================================
 
+rain.thecomicseries.com
+
+CSS
+body {
```
abitenum, line 11043: Just to add a note, I considered using this instead:

```CSS
body {
    background-color: inherit !important;
}
```

This would use the `html` tag background color that Dark Reader adds. However, this doesn't look terribly good. Notably, this results in the loss of the intended center column in light mode.

Gusted: You could simply add `background-color: var(--darkreader-neutral-background) !important;` to `html` tag, such that you would keep consistency between body and html

src/config/dynamic-theme-fixes.config
```diff
@@ -11037,6 +11037,21 @@ INVERT
...
+    background-color: var(--darkreader-neutral-background) !important;
+    background-image: none !important;
+    margin: 0px !important;
+    padding: 8px !important;
```
Gusted, line 11047: I'm still not a big fan of these, the config is for 99.99% made out of color changing, we shouldn't touch stuff like these as a extension.

abitenum: I do agree with you on this point generally. I find it kind of weird that it doesn't seem to be a more common issue considering that the 8px margin is the default if one isn't specified for the body tag. If you think setting the HTML's background color is the right move as suggested in your other comment, then I can update the PR with that, but it feels like just as much of a hack as this to me.

Gusted: I think this will hit the 00.01% case - so let it be.

Gusted (approved):

PR #7334, opened 2021-11-11
Add comicfury.com to dynamic-theme-fixes.config

abitenum:
Because of the way Flexbox is used on the site, the `body` tag only has a height of the browser window. The background color extends to the whole page normally because the `html` tag doesn't have a background color, but Dark Reader adds a background color to it. This simply removes the background color from the `html` tag so that the background color works as the website intends.

Honestly, this website is kind of weird, seemingly using a very bright dark mode, so there's probably more that need to be done to get the light mode setting to work right on the site, but this at least fixes the issue of the inconsistent background color.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2836,6 +2836,15 @@ button
 
 ================================
 
+comicfury.com
+
+CSS
+html {
+    background-color: revert !important;
+}
+
+================================
+
 comixology.com
 
 INVERT
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2839,8 +2839,12 @@ button
 comicfury.com
 
 CSS
-html {
-    background-color: revert !important;
+#wrapper {
+    background-color: inherit !important;
+    padding-bottom: 10px !important;
+}
+#footer {
+    margin-top: 0px !important;
 }
 
 ================================


src/config/dynamic-theme-fixes.config
```diff
@@ -2836,6 +2836,15 @@ button
 
 ================================
 
+comicfury.com
+
+CSS
+html {
+    background-color: revert !important;
+}
```
Gusted, line 2844: Same thing as in your previous PR.

abitenum: Ok, I did some experimenting to figure out another solution. It looks like setting the background color of `#wrapper` to `inherit` also solves the core issue. The only issue is the top margin on the `#footer`, but replacing it with a bottom padding on `#wrapper` fixes that.

src/config/dynamic-theme-fixes.config
```diff
@@ -2836,6 +2836,19 @@ button
...
 
+comicfury.com
+
+CSS
+#wrapper {
+    background-color: inherit !important;
+    padding-bottom: 10px !important;
+}
+#footer {
+    margin-top: 0px !important;
```
Gusted, line 2847: ??

src/config/dynamic-theme-fixes.config
```diff
@@ -2836,6 +2836,19 @@ button
 
 ================================
 
+comicfury.com
+
+CSS
+#wrapper {
+    background-color: inherit !important;
+    padding-bottom: 10px !important;
```
Gusted, line 2844: ??

src/config/dynamic-theme-fixes.config
```diff
@@ -2836,6 +2836,19 @@ button
 
 ================================
 
+comicfury.com
+
+CSS
+#wrapper {
+    background-color: inherit !important;
```
Gusted, line 2843: Seems good.

src/config/dynamic-theme-fixes.config
abitenum: This "??" is related to the other one.

The issue is that without these two lines, a small bit of the `html` tag background still comes through when it shouldn't.

comicfury

The issue is is happening between `#wrapper` and `#footer`. The `#footer` has a `10px` top margin to add spacing between the two. The best way to fix this that I found was to replace the `#footer` margin with a `#wrapper` padding, as the padding uses the background color.

Gusted: Sure thing, let it be. As it's simply changing which element is adding the margin/padding.

Gusted (approved):

PR #7345, opened 2021-11-11
Mikufan.com

Stephhawks:
Canceled the header image inversion

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -8949,10 +8949,8 @@ body,
 
 mikufan.com
 
-CSS
-#header-container {
-    background-image: url(https://www.mikufan.com/wp-content/uploads/2018/04/header_main3.jpg) !important;
-}
+IGNORE IMAGE ANALYSIS
+#header-container
 
 ================================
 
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -8947,6 +8947,15 @@ body,
 
 ================================
 
+mikufan.com
+
+CSS
+#header-container {
+    background-image: url(https://www.mikufan.com/wp-content/uploads/2018/04/header_main3.jpg) !important;
+}
```
Gusted, line 8955:
```suggestion
IGNORE IMAGE ANALYSIS
#header-container
```

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
Stephhawks: It didn't work when i tried

Gusted: You've to reload the page before it takes effect.

Stephhawks: I see, thanks

Gusted (approved):
Thanks!

PR #7353, opened 2021-11-13
codingame visible variable fields

fifididi:
From this: https://i.imgur.com/5G47fwp.png
To this: https://i.imgur.com/btPpt8F.png

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2813,11 +2813,12 @@ INVERT
 
 ================================
 
-codingame.com/ide/*
+codingame.com/ide
 
 CSS
-.cg-statement .statement-body span.var, .cg-statement .statement-body var {
-    background-color: rgb(0, 10, 164);
+.cg-statement .statement-body span.var,
+.cg-statement .statement-body var {
+   color: var(--darkreader-neutral-background) !important;
 }
 
 ================================
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2813,6 +2813,15 @@ INVERT
 
 ================================
 
+codingame.com/ide/*
+
+CSS
+.cg-statement .statement-body span.var, .cg-statement .statement-body var {
+    background-color: rgb(0, 10, 164);
+}
+
+================================
+
 coinbase.com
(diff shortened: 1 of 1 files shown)


Gusted:
Hi @fifididi!

Can you give a example link where this happens?

Regards,
Gusted

fifididi:
Hello, 
[Here](https://www.codingame.com/ide/puzzle/don't-panic-episode-1)'s an example where it could come in handy

You can modify the colour to your liking and integrate it, the main problem I want fixed is to be able to read the fields with ease. 

It turns out there are other classes that use the same colour for buttons on the site such as `.navigation-optin`, `.cg-cookies-banner_button_accept`, `#content-details .content-details_main-header .banner-overlay .header .solve-button`, `.cg-profile-settings input[type="radio"]:checked + label` that are not included in this PR but could still use the same background-color change. 

Thank you!

src/config/dynamic-theme-fixes.config
```diff
@@ -2813,6 +2813,15 @@ INVERT
 
 ================================
 
+codingame.com/ide/*
```
Gusted, line 2816:
```suggestion
codingame.com/ide
```

src/config/dynamic-theme-fixes.config
```diff
@@ -2813,6 +2813,15 @@ INVERT
...
+
+CSS
+.cg-statement .statement-body span.var, .cg-statement .statement-body var {
```
Gusted, line 2819:
```suggestion
.cg-statement .statement-body span.var,
.cg-statement .statement-body var {
```

src/config/dynamic-theme-fixes.config
```diff
@@ -2813,6 +2813,15 @@ INVERT
...
+    background-color: rgb(0, 10, 164);
```
Gusted, line 2820:
```suggestion
   color: var(--darkreader-neutral-background) !important;
```

Gusted (approved):

src/config/dynamic-theme-fixes.config
fifididi: I thought that changing the text colour would remove from the logic of the page, since the original text color doesn't change. But that works too.

PR #7354, opened 2021-11-13
Prevent darkreader on Braze previews

torma616:
Braze.com has a preview window to see how messages will look on iOS and Android once published. Parts of these previews use static images, and so once darkreader does its thing it's an incomplete conversion.

This change will prevent darkreader from modifying the Braze iOS/Android preview view.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -45,6 +45,57 @@ div[class^='scrollableList'] {
 
 ================================
 
+*.braze.com
+
+CSS
+.preview-background,
+.iphone-rich-notification-preview,
+.iphone-rich-notification-divider {
+    background-color: #dddde2 !important;
+}
+.iphone-push-preview-title, 
+.iphone-push-preview-message-container,
+.iphone-push-preview-preview-message {
+    color: #ffffff !important;
+}
+.iphone-push-preview-now{
+    color: #788186 !important;
+}
+.tab-pull {
+    background: #aaaaaa !important;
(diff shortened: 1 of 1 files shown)


Gusted:
I am not so sure if we should have such huge CSS fix, if someone doesn't want dark reader modifying something, they should simply disable dark reader. Otherwise the broken part should be fixed.

torma616:
That's fair enough. 

I find myself manually adding this to the dynamic rules whenever I have to use the site (I don't want to turn darkreader off site-wide, and the non-"dynamic" theme generation modes don't look nearly as good), so I figured I'd share my changes in case you wanted to merge them. If not, I'll just continue loading them in manually locally.

As a side note, I can't really understand why there isn't an option to point darkreader to an additional, personal "inversion fixes" file, whether it be a local one on your computer or a gist,etc.

Gusted:
> As a side note, I can't really understand why there isn't an option to point darkreader to an additional, personal "inversion fixes" file, whether it be a local one on your computer or a gist,etc.

Because the dev-tools is meant for temporary fixes, long-term ones should be applied by a extension like [stylus](https://github.com/openstyles/stylus)



> I find myself manually adding this to the dynamic rules whenever I have to use the site (I don't want to turn darkreader off site-wide, and the non-"dynamic" theme generation modes don't look nearly as good), so I figured I'd share my changes in case you wanted to merge them. If not, I'll just continue loading them in manually locally.

The fix is fine, but it doesn't really match the way how fixes in the config should be.

PR #7288, opened 2021-11-07
Comments on selectors; Added GitHub and GitLab style

mattalexx:

PR #7377, opened 2021-11-14
Revert "Comments on selectors; Added GitHub and GitLab style"

mattalexx:
Reverts darkreader/darkreader#7288

@Gusted i think this needs to be reverted. I noticed what I think is a regression, with some icons that used to load but don't now. 

I am not able to fix until the end of the week. If you revert this, i will submit a better report, fix, test, and resubmit then.

### DIFF
--- a/src/generators/text-style.ts
+++ b/src/generators/text-style.ts
@@ -1,27 +1,10 @@
 import type {FilterConfig} from '../definitions';
 
 export function createTextStyle(config: FilterConfig): string {
+    const lines: string[] = [];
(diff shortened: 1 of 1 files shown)


mattalexx:
If I were a betting man, 

```ts
'.material',
```
Should be changed to
```ts
'.material-icon',
```

But when I'm in front of an actual computer I'll test.

Gusted (approved):

Gusted:
Let's revert it for now.

mattalexx:
Thanks

PR #7380, opened 2021-11-15
Improved conversions on colors.dopely.top

CrispyMiner:
I made some changes for this site where the shades of colors are displayed correctly, as well made the tool icon symbols more visible.

NOTE: There is a bug that if you load a page on site with darkreader already on, things are displayed incorrectly for some reason. However, turning off and on darkreader instantly fixes it.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2866,35 +2866,34 @@ button
 colors.dopely.top
 
 INVERT
-div[class*="navbar__container"]
-div[class="navbar__buttons"]
-div[class="posts-container"]
-div[class*="WhatsNew__Bloks"]
-div[class*="-slider"]
-nav.navbar
 .footer
-.toner-filter-button
-.toner-number-selector-container
 
 CSS
 img[src*=".svg"]:not([alt*="color wheel icon"]):not([class*="dropdown"]):not([src="/static/allInOne.fd51fb9d.svg"]):not([bottom-toolbar-left-item]):not(.ToolsLanding__image),
 div[class*="assetsbar-mobile-block"],
-.contrast-checker-like-icon {
+.contrast-checker-like-icon,
+.accountDopely__dropdownIcon,
+div[class="title"] {
     filter: invert(50%) saturate(100%) !important;
 }
-div[class$="toner-container"],
-.bottom-toolbar-left-item {
-    background-color: var(--darkreader-neutral-background) !important;
+input[class*="number-slider"] {
+    filter: invert(10%) saturate(100%) !important;
 }
-div[class*="navigation"],
-div[class*="userNavigation"],
-div[class="navbar__buttons"],
-div[class*="controller-container"] {
-    color: var(--darkreader-neutral-background) !important;
+img[class*="dropdown"],
+img[src*="allInOne"],
+.profile-header {
+    filter: saturate(0%) brightness(50%) !important;
+}
+div[class*="profile-pic"] {
+    filter: brightness(50%) !important;
+.palette-generator-container *,
+.toner-colors * {
+    --darkreader-inline-bgcolor: initial !important;
 }
 
 IGNORE INLINE STYLE
-span(colors)
+.ExpandPalette *
+.toner-colors-container *
 .color-picker
 
 ================================
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -8740,14 +8740,6 @@ section[class$="container"][style^="background-image"] {
 
 ================================
 
-mdigi.tools
-
-IGNORE IMAGE ANALYSIS
-.color-preview *
-.color-shade *
-
-================================
-
 mediawiki.org
 
 INVERT
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2890,6 +2890,7 @@ a[class*="header"],
 .toner-colors *,
 .toner-colors-container * {
     --darkreader-inline-bgcolor: initial !important;
+}
 
 IGNORE INLINE STYLE
 .ExpandPalette *
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2863,6 +2863,42 @@ button
 
 ================================
 
+colors.dopely.top
+
+INVERT
+div[class*="navbar__container"]
+div[class="navbar__buttons"]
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -2863,6 +2863,42 @@ button
...
+}
+div[class*="navigation"],
+div[class*="userNavigation"],
+div[class="navbar__buttons"],
+div[class*="controller-container"] {
+    color: var(--darkreader-neutral-background) !important;
+}
+
+IGNORE INLINE STYLE
+span(colors)
```
Gusted, line 2897: This is a invalid CSS selector.

CrispyMiner: How would I word it in CSS in that case? (For example on the [palette part of the site](https://colors.dopely.top/palette-generator/)) and I'm struggling to prevent the colors from getting inverted and I can't find the anything for the color tiles outside their "ExpandPalette" class name and colors are their background-color.  Because I am stumped

(First image is the inverted result I **don't** want, second image is the result I **do** want)
color_result
G)

Gusted: What about using the `.PaletteGenerator__item--insideItem *` selector? It should cover the case.

CrispyMiner: > .PaletteGenerator__item--insideItem *

I tried that, did nothing in INVERT, it changed something when I put it in IGNORE INLINE and CSS (as:
> .PaletteGenerator__item--insideItem * {
+    filter: invert(0%) !important;
+ }

But all that did was invert the color for the text labels of the color containers

CrispyMiner: Ok so,
> CSS
> .palette-generator-container * {
>     --darkreader-inline-bgcolor: initial !important;
> }
made all of the colors black (and only the colors), but I just need to be able to reword the statement so it reverts back to the un-inverted color it originally had

Gusted: Hmm, what about using `.ExpandPalette *` in the IGNORE INLINE STYLE?

CrispyMiner: didn't anything either

Gusted: Did you refresh the site after applying?

CrispyMiner: > Did you refresh the site after applying?

Ahh yeah it worked, I also was implying .ExpandPalette would still be in CSS, but changed it to the .palettegen
lemme fix up the other generators with the same technique and clean up the code. Thanks! (just i reply the fix once done?)

Gusted: > (just i reply the fix once done?)

I automatically get a notification if a edit is made.

src/config/dynamic-theme-fixes.config
```diff
@@ -2863,6 +2863,41 @@ button
...
+    filter: invert(10%) saturate(100%) !important;
+}
+img[class*="dropdown"],
+img[src*="allInOne"],
+.profile-header {
+    filter: saturate(0%) brightness(50%) !important;
+}
+div[class*="profile-pic"] {
+    filter: brightness(50%) !important;
+.palette-generator-container *,
```
Gusted, line 2889:
```suggestion
}
.palette-generator-container *,
```

Gusted (changes requested):
Things like `div[class*="assetsbar-mobile-block"]` can be converted to `div.assetsbar-mobile-block`

src/config/dynamic-theme-fixes.config
```diff
@@ -2863,6 +2863,41 @@ button
...
+}
+img[src*="svg"], .title {
+    filter: invert(50%) !important;
+}
+.accountDopely__letter,
+.editProfile__changeProfilePicture {
+    filter: saturate(0%) brightness(50%) !important;
+}
+div[class*="text"],
+a[class*="header"],
```
Gusted, line 2884:
```suggestion
div.text,
a.header,
```

src/config/dynamic-theme-fixes.config
```diff
@@ -2863,6 +2863,41 @@ button
...
+.navbar-dropdown-item__logo,
+.navbar__listItem__button {
+    filter: saturate(0%) !important;
+}
+.palette-generator-container *,
+.toner-colors *,
+.toner-colors-container * {
+    --darkreader-inline-bgcolor: initial !important;
+
```
Gusted, line 2893:
```suggestion
}

```

src/config/dynamic-theme-fixes.config
CrispyMiner: This is done so that it includes all div with "text" or all a with "header" in their names across the whole site.
Doing it the way you suggested excludes some out. This site is very unorganized with the code and how things are referred as

src/config/dynamic-theme-fixes.config
```diff
@@ -2863,6 +2863,42 @@ button
...
+}
+div[class*="text"],
+a[class*="header"],
+.navbar-dropdown-item__logo,
+.navbar__listItem__button {
+    filter: saturate(0%) !important;
+}
+.palette-generator-container *,
+.toner-colors *,
+.toner-colors-container * {
```
Gusted, line 2891: Their shouldn't be any inline styles happening on these. as it's already specified to ignore inline styles.

CrispyMiner (commented):
Additionally for clarity sake, anything in the IGNORE INLINE STYLE, disables the inversion for elements specified under it, correct?

Gusted:
> anything in the IGNORE INLINE STYLE, disables the inversion for elements specified under it, correct?

Well the current IGNORE INLINE STYLE has the `*` selector at the end which means it will match all element under that element. So it will disable the styling of inline styles for that element and all elements under it.

Gusted (approved):

CrispyMiner:
Crap, accidentally sent a review change for mdigi.tools 3 times

src/config/dynamic-theme-fixes.config
```diff
@@ -8709,6 +8740,14 @@ section[class$="container"][style^="background-image"] {
...
 ================================
 
+mdigi.tools
+
+IGNORE IMAGE ANALYSIS
+.color-preview *
+.color-shade *
+
+================================
+
```
Gusted, line 8750:
```suggestion
```

CrispyMiner:
For mdigi.tools I can't get the light colors within color-shades to not invert, any help?

PR #7385, opened 2021-11-16
Improve Grammarly Theme

BellCubeDev:
It's probably messy CSS (this is my first time editing CSS outside of basic text color variables), it might need a little bit of review, but it is at least better-looking than it was (and you can see the text in question when you select an issue)

Also apparently I typoed the commit name. My bad.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1032,7 +1032,7 @@ use.valueCircle_f6otssy {
 }
 [class*="-navigation-counterWrapper"] [class*="-navigation-counterContent"] {
     fill: rgb(0, 0, 0, .9);
-
+}
 
 IGNORE INLINE STYLE
 use[class*="valueCircle_"]
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1030,6 +1030,9 @@ use.valueCircle_f6otssy {
 [class*="-paidview-counter"] [class*="-paidview-counterContent"] {
     fill: rgb(0, 0, 0, .9);
 }
+[class*="-navigation-counterWrapper"] [class*="-navigation-counterContent"] {
+    fill: rgb(0, 0, 0, .9);
+}
 @keyframes f15bts9f {
     to {
         background-color: rgb(255, 219, 227, 0.25)
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1028,7 +1028,7 @@ use.valueCircle_f6otssy {
     color: rgb(14, 16, 26) !important;
 }
 [class*="-paidview-counter"] [class*="-paidview-counterContent"] {
-    fill: rgb(0, 0, 0, 105);
+    fill: rgb(0, 0, 0, .9);
 }
 @keyframes f15bts9f {
     to {
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1027,10 +1027,8 @@ use.valueCircle_f6otssy {
 [class*="-alerts-markSelectedHigh"], span[class*="markSelectedFocused"] {
     color: rgb(14, 16, 26) !important;
 }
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -1021,9 +1021,53 @@ IGNORE IMAGE ANALYSIS
 app.grammarly.com
 
 CSS
+use.valueCircle_f6otssy {
+    stroke: url(#pb-gradient-0);
+}
 [class*="-alerts-markSelectedHigh"], span[class*="markSelectedFocused"] {
     color: rgb(14, 16, 26) !important;
 }
+._c0560d9f-paidview-counter ._a7150ba2-paidview-counterContent {
```
Gusted, line 1030:
```suggestion
[class*="-paidview-counter"] [class*="-paidview-counterContent"] {
```

src/config/dynamic-theme-fixes.config
```diff
@@ -1021,9 +1021,53 @@ IGNORE IMAGE ANALYSIS
...
+    fill: rgb(0, 0, 0, 105);
```
Gusted, line 1031: This doesn't seem like a valid CSS Color.

src/config/dynamic-theme-fixes.config
```diff
@@ -1021,9 +1021,53 @@ IGNORE IMAGE ANALYSIS
...
+@keyframes f1gz3vdc {
+    to {
+        background-color: rgb(219, 235, 255, 0.25)
+    }
+}
+@keyframes fjzahul {
+    to {
+        background-color: rgb(215, 247, 240, 0.25)
+    }
+}
```
Gusted, line 1067: These all use obscure naming, not sure if it should be added.

src/config/dynamic-theme-fixes.config
```diff
@@ -1021,9 +1021,53 @@ IGNORE IMAGE ANALYSIS
...
+
+IGNORE INLINE STYLE
+use.valueCircle_f6otssy
```
Gusted, line 1070:
```suggestion
use[class*="valueCircle_"]
```

src/config/dynamic-theme-fixes.config
BellCubeDev: That's all on the part of Grammarly there. Those are the names for the different animated highlights. The only change I made was turning the opacity down to 25% because, at 100%, the text was near-impossible to see at pure white.

Gusted: The problem with such naming is that they are automatically generated and when grammarly pushes out another update(small or big) those naming will be gone.

src/config/dynamic-theme-fixes.config
BellCubeDev: My bad. First time doing CSS, you see.
I'm used to the typical 0–255 opacity range. You can see later that I did change it to the 0–1 range.

src/config/dynamic-theme-fixes.config
BellCubeDev: The question is then, how do I modify those colors?

Gusted: Not really, keyframes are a edge-case within this config.

src/config/dynamic-theme-fixes.config
```diff
@@ -1021,9 +1021,21 @@ IGNORE IMAGE ANALYSIS
...
+use.valueCircle_f6otssy {
+    stroke: url(#pb-gradient-0);
+}
 [class*="-alerts-markSelectedHigh"], span[class*="markSelectedFocused"] {
     color: rgb(14, 16, 26) !important;
 }
+[class*="-paidview-counter"] [class*="-paidview-counterContent"] {
+    fill: rgb(0, 0, 0, .9);
+}
+[class*="-navigation-counterWrapper"] [class*="-navigation-counterContent"] {
```
Gusted, line 1033:
```suggestion
[class*="-navigation-counterWrapper"] [class*="-navigation-counterContent"],
[class*="-paidview-counter"] [class*="-paidview-counterContent"] {
```

Gusted (approved):

PR #7390, opened 2021-11-16
Dynamic theme for *.service-now.com

n21lv:
Better colour scheme for CodeMirror script editor used in SN
Fixes for colour scheme used in SN's diff manager

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -54,6 +54,61 @@ CSS
 
 ================================
 
+*.service-now.com
+
+CSS
+.CodeMirror-merge-r-inserted, .CodeMirror-merge-l-inserted {
+    background-image: url("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAMAAAACCAIAAAASFvFNAAAAFUlEQVQImWOQLOaSLOZiYGBgYoABABOcAS8huNzvAAAAAElFTkSuQmCC") !important;
+}
+.CodeMirror-merge-r-deleted, .CodeMirror-merge-l-deleted {
+    background-image: url("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAMAAAACCAYAAACddGYaAAAAFUlEQVQYGWP4Hxn5H4QZGBj+MyBzALBLDL9PePLnAAAAAElFTkSuQmCC") !important;
+}
+.CodeMirror-merge-copy {
+    color: #7dd0d8 !important;
+}
+.CodeMirror-merge-l-connect {
+    fill: #082b30 !important;
+    stroke: #53535d !important;
+}
+.CodeMirror-merge-l-chunk{
+    background-color: #082b30 !important;
+}
+.CodeMirror-merge-l-chunk-start {
+    border-top: 1px solid #53535d !important;
+}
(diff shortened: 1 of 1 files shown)


Gusted:
Does this actually fix anything? Or is this just to improve the current coloring? The config should be exclusively used to fix things that Dark Reader is doing wrong.

n21lv:
> Does this actually fix anything? Or is this just to improve the current coloring? The config should be exclusively used to fix things that Dark Reader is doing wrong.

The colour scheme Dark Reader applies doesn't have enough contrast. All the colour fixes I used pass at least WCAG AA (a contrast ratio of at least 4.5:1) and most pass WCAG AAA (a contrast ratio of at least 3:1). The first image shows how default colour scheme looks like
image
This is how the same script looks with contrast fixes
image

Gusted:
> The colour scheme Dark Reader applies doesn't have enough contrast. All the colour fixes I used pass at least WCAG AA (a contrast ratio of at least 4.5:1) and most pass WCAG AAA (a contrast ratio of at least 3:1)

While it's nice to pass the WCAG contrast ration minimum - it's not a requirement that should be fixed. The applied fix seems to not be a variant of the original color, which means it's just giving another color and not beholding the original color.

n21lv:
Okay, but then the default scheme isn't properly inverting some of the classes there, which _is_ what I am trying to fix. If the point of having a dark theme is to make your eyes less strained, why issues such as dark text on dark background shouldn't be fixed? 
I would happily let DarkReader do its job without any corrections, but if I have to squint my eyes to read dark text on dark background or find a dark rectangle that is so very slightly brighter than the dark background it's on, that looks like a bug to me. In either case, if you don't feel that this submission is a valuable addition, just reject it. I don't really see a point in continuing this discussion if we can't agree.

Gusted:
I'm not saying or trying to imply that the fix should be disregarded as all - their definitely should be a fix - the problem is the current fix, it should be changed to match the current style of fixes by e.g. using `${...}` and the dynamic variables. I'm sorry if I came off as that this fix is very incorrect.

n21lv:
> I'm not saying or trying to imply that the fix should be disregarded as all - their definitely should be a fix - the problem is the current fix, it should be changed to match the current style of fixes by e.g. using `${...}` and the dynamic variables. I'm sorry if I came off as that this fix is very incorrect.

Apologies for misunderstanding. I am not really sure about using `${...}`, though -- I tried that before and couldn't get it working with hex values, as that's how original colours are defined. What is the correct way of doing that?

Gusted:
> -- I tried that before and couldn't get it working with hex values, as that's how original colours are defined. What is the correct way of doing that?

It will "invert" the given color between `${...}` e.g. `${red}` will transform to the invert of red. It should work with hex colors such as `${#f0f0f0}`

n21lv:
Yeah, that actually seems to work. I was dumbfounded at first because SN's UI uses iframes and they don't automatically apply new changes, so you have to reload the page.
So, basically your main criticism is that I am applying arbitrary colour values instead of inverting the original ones? If so, I will try to come up with a proper fix.

Gusted:
> So, basically your main criticism is that I am applying arbitrary colour values instead of inverting the original ones?

Yap

Gusted:
> Do you have any plans for allowing custom overrides that don't get removed with updates?

Nope.

PR #7399, opened 2021-11-16
Fix for coolors.co

CrispyMiner:
Saturation of the color picker on coolors.co was inverted so  needed to be inverted.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3077,7 +3077,7 @@ header, div#sticky-cad_img, .page-title {
     filter: saturate(0) brightness(65%) !important;
 }
 .color-picker_saturation {
-    filter: invert(100%) brightness(120%) !important;
+    background-image: -webkit-linear-gradient(right, rgba(255, 255, 255, 0), #fff) !important;
 }
 
 IGNORE INLINE STYLE
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3070,6 +3070,9 @@ CSS
 .generator_color.is-light {
     color: #000 !important;
 }
+.color-picker_saturation {
+    filter: invert(100%) brightness(110%) !important;
+}
 
 IGNORE INLINE STYLE
 #generator*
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -3070,6 +3070,9 @@ CSS
 .generator_color.is-light {
     color: #000 !important;
 }
+.color-picker_saturation {
```
Gusted, line 3073: I'm not able to find this element.

CrispyMiner: It's in the palette generator part of the site, click generator and it should be in there.

To view the actual color picker, do the following:
Click the hex color number
Click the drop down arrow in that menu
Select Picker
image

src/config/dynamic-theme-fixes.config
```diff
@@ -3070,6 +3070,9 @@ CSS
...
+    filter: invert(100%) brightness(110%) !important;
```
Gusted, line 3074:
```suggestion
    background-image: linear-gradient(right,rgba(255,255,255,0),#fff) !important;
```

CrispyMiner: That didn't work for me, I think that while it's not the way to do it, it still works at least I guess?

src/config/dynamic-theme-fixes.config
```diff
@@ -3070,6 +3070,15 @@ CSS
...
     color: #000 !important;
 }
+header, div#sticky-cad_img, .page-title {
+    filter: saturate(0) brightness(50%) !important;
+}
+.sub-header {
+    filter: saturate(0) brightness(65%) !important;
+}
+.color-picker_saturation {
+    filter: invert(100%) brightness(120%) !important;
```
Gusted, line 3080:
```suggestion
    background-image: -webkit-linear-gradient(right, rgba(255, 255, 255, 0), #fff) !important;
```

src/config/dynamic-theme-fixes.config
```diff
@@ -3070,6 +3070,15 @@ CSS
 .generator_color.is-light {
     color: #000 !important;
 }
+header, div#sticky-cad_img, .page-title {
+    filter: saturate(0) brightness(50%) !important;
+}
+.sub-header {
+    filter: saturate(0) brightness(65%) !important;
+}
```
Gusted, line 3078: The elements already looks fine to me?

PR #7409, opened 2021-11-18
update change mcloud.entsvcs.net color 

metalbreeze:


Gusted:
Include these changes into #7408

PR #7428, opened 2021-11-20
Fix for khronos.org

pnaf:
Replaces top logo with inverted version.

Before:

After:

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -7618,10 +7618,8 @@ CSS
 
 khronos.org
 
-CSS
-.navbar-brand {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -7616,6 +7616,15 @@ CSS
 
 ================================
 
+khronos.org
+
+CSS
+.navbar-brand {
```
Gusted, line 7622: What about using:
```
INVERT
.navbar-brand
```

pnaf: That works!

Gusted (approved):

PR #7429, opened 2021-11-20
New interface usage documentation

kruegead:
Added documentation for the new interface design


Gusted:
Why should there be a documentation about the UI? I think that the UI should be intuitive enough that it doesn't need a documentation.

kruegead:
There's already some documentation for the older interface viewable through the help link in the app [https://darkreader.org/help/en/](https://darkreader.org/help/en/).  This documentation is for the newer interface.

Gusted:
Hmm yeah that's a good point, eh, I personally think we can "discontinue" the documentation about the UI(as it really should be made that self-explanatory).

Gusted:
@alexanderby

alexanderby:
Hi @kruegead, thank you for your contribution. It is unclear what will the final design look like yet. It's totally different and experimental on iOS. The final desktop design will be something in the middle.

https://user-images.githubusercontent.com/9638588/143497079-58381dce-a86a-46aa-8892-99d42f4cca3c.mov

ViRo3:
On iOS, especially devices of smaller height it feels quite cramped. On first glance, in the sharesheet it would be better to be limited to showing the on/off/auto toggle alongwith the site toggle and first row of themes while removing the settings button. This would make it feel less cramped and more refined.
x
Something like this where it gets rid of the additional screen.
For the theming screen, if possible it would be recommended to use the OS colorpicker (available via safari too) as it has additional options to pick color from the current screen.

alexanderby:
Hi @ViRo3! What's your device?

alexanderby:
@ViRo3, did you draw this screenshot yourself?

alexanderby:
Oh I got it, this is your proposal. I nearly got a heart attack thinking the layout is broken on some devices. You are right, until there are not very much menus to display, it's possible to show all the controls on the full screen. Thanks for the hint with `input[type="color"]`, I didn't know it exists.

ViRo3:
My device is 12 Pro but the text beneath the sliders makes it seem cramped. And yes the screenshot is stitches of what already exists. Color picker was added recently on safari (present on others) so would make it feel more native on platforms. (Sorry for not clarifying that its a proposal haha)

PR #7449, opened 2021-11-23
Fixes for gog.com

n21lv:
Fixed things like icons not being inverted, applied overlays for unprocessed background images that were too bright

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -6206,14 +6206,58 @@ gog.com
 
 INVERT
 i.icn.icn--close
+svg.icon-wrapper-icon
+svg.big-spot__add-to-cart-icon
+svg.ic-svg:not(.button__icon):not(.productcard-slider__nav-icon)
+.productcard-thumbnails-slider-pagination
+.carousel-pagination__page
+.big-spot__carousel-pages-container
+.big-spots__arrow-icon
+.discover-games-more__icon
 
 CSS
 .product-tile__info:hover {
     background-color: var(--darkreader-neutral-background) !important;
 }
+a.product-tile__content {
+    background-color: var(--darkreader-neutral-background) !important;
+}
+.explore-galaxy {
+    border-top: 0 !important;
+}
+.galaxy-tooltip__layer {
+    background: ${#d9d9d9} !important;
+    color: var(--darkreader-neutral-text) !important;
+}
+.big-spot__content, .big-spot__text, .big-spot__action {
+    backdrop-filter: brightness(50%) !important;
+}
+.big-spot__title, .big-spot__super-title, .big-spot__action {
+    filter: brightness(130%) !important;
+}
+.wishlist-button, .review__read-more, .review-new__add-button, .review-new__guideline-button {
+    background: none !important;
+}
 
 IGNORE IMAGE ANALYSIS
 .menu-anonymous__shelf
+
+================================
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -6206,17 +6206,61 @@ gog.com
...
 CSS
 .product-tile__info:hover {
     background-color: var(--darkreader-neutral-background) !important;
 }
+a.product-tile__content {
+    background-color: var(--darkreader-neutral-background) !important;
+}
+.explore-galaxy {
+    border-top: 0 !important;
+}
```
Gusted, line 6227: Hmm interesting that dark reader is modifying the border-top, I would like to fix this in the code. You can remove it for now.

src/config/dynamic-theme-fixes.config
```diff
@@ -6206,17 +6206,61 @@ gog.com
...
+gog.com/promo
+
+CSS
+.wrapper.cf > .content.cf::before {
+    position: fixed;
+    left: 0;
+    top: 0;
+    width: 100%;
+    height: 100%;
+    background-color: rgba(0,0,0,0.5);
```
Gusted, line 6256: A lot of scary code, could you link me a example page where this is used?

src/config/dynamic-theme-fixes.config
```diff
@@ -6206,17 +6206,61 @@ gog.com
...
+    color: var(--darkreader-neutral-text) !important;
+}
+.big-spot__content, .big-spot__text, .big-spot__action {
+    backdrop-filter: brightness(50%) !important;
+}
+.big-spot__title, .big-spot__super-title, .big-spot__action {
+    filter: brightness(130%) !important;
+}
+.wishlist-button, .review__read-more, .review-new__add-button, .review-new__guideline-button {
+    background: none !important;
```
Gusted, line 6239:
```suggestion
    background-image: none !important;
```
Let's specify it a bit further.

src/config/dynamic-theme-fixes.config
```diff
@@ -6206,17 +6206,61 @@ gog.com
...
 INVERT
 i.icn.icn--close
+svg.icon-wrapper-icon
+svg.big-spot__add-to-cart-icon
+svg.ic-svg:not(.button__icon):not(.productcard-slider__nav-icon)
+.productcard-thumbnails-slider-pagination
+.carousel-pagination__page
+.big-spot__carousel-pages-container
+.big-spots__arrow-icon
+.discover-games-more__icon
```
Gusted, line 6216: I will also have a look at these elements as these are hitting a case that dark reader _rightfully_ doesn't handle yet, but in this specific case it can be handled correctly.

Gusted (commented):
You've found some interesting cases for me to look at :D

src/config/dynamic-theme-fixes.config
n21lv: image

Sure, just click any of the three promos here to see how they look without that overlay. I tried multiple approaches, but the problem is that they use really bright background-images for container elements and after inversions you get bright text on bright background.

Gusted (approved):

n21lv:
> You've found some interesting cases for me to look at :D

Oh yeah, I can see why. I was up all night two days ago trying to get those fixes work properly, because CSS is just so awesome :D
You're welcome!

PR #7450, opened 2021-11-23
Fix for hdlbits.01xz.net

linsui:
E.g. https://hdlbits.01xz.net/wiki/Xnorgate

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -6533,6 +6533,13 @@ INVERT
 
 ================================
 
+hdlbits.01xz.net
+
+INVERT
+.image
+
+================================
+
 heise.de
 
 INVERT
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -6536,7 +6536,7 @@ INVERT
 hdlbits.01xz.net
 
 CSS
-.image {
+img.image, .image > img {
     background-color: white !important;
     filter: brightness(50%) sepia(40%) !important;
 }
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -6533,6 +6533,13 @@ INVERT
 
 ================================
 
+hdlbits.01xz.net
+
+INVERT
+.image
```
Gusted, line 6539:
```suggestion
CSS
.image {
    background-color: white !important;
    filter: brightness(50%) sepia(40%) !important;
}
```

linsui:
@Gusted Anything else can I do?

src/config/dynamic-theme-fixes.config
linsui: This doesn't work well.

https://hdlbits.01xz.net/wiki/Fsm1

image

Gusted: Hmm weird, seems like this has a `.image` wrapper and thus the class isn't on the `img` element(which it has to be in order to work), I think you can change it to:

`img.image, .image > img`

linsui: It works well now, thanks!

src/config/dynamic-theme-fixes.config
```diff
@@ -6533,6 +6533,16 @@ INVERT
 
 ================================
 
+hdlbits.01xz.net
+
+CSS
+.image {
```
linsui, line 6539:
```suggestion
img.image, .image > img {
```

Gusted (approved):

PR #7474, opened 2021-11-25
Update `rollup-plugin-typescript2`

Gusted:
- Fixes compiling the master on node v17.0+
- See https://github.com/ezolenko/rollup-plugin-typescript2/issues/286

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,7 +41,7 @@
         "puppeteer-core": "10.4.0",
         "rollup": "2.59.0",
         "rollup-plugin-istanbul2": "2.0.2",
-        "rollup-plugin-typescript2": "0.30.0",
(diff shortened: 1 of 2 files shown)


alexanderby:
There was also `@rollup/plugin-typescript`, maybe a bit more supported. I used it for some projects, don't know if it is better in any way.

Just in case I've started migrating those build scripts to ES Modules. A bit painful, but the end is near.

Gusted:
> There was also `@rollup/plugin-typescript`, maybe a bit more supported. I used it for some projects, don't know if it is better in any way.

They should give you the same end-result, ezolenko's one is just faster, which is especially when I was coding on my special machine very appreciated. But I should get some time around for #4492

PR #7484, opened 2021-11-27
Removed ezgif.com logo inversion

Mad182:
ezgif.com now supports dark mode, so there is no more need for logo inversion.


Gusted (approved):

Gusted:
Woops, wrong button.

- No, let's not remove it as not everyone would be using their dark mode. If they want to use their dark mode they will have dark reader(as that's recommended)

PR #7485, opened 2021-11-27
Straigten Dark Reader ASCII art

WebMechanic:
The ASCII art seemed a bit shakey


Gusted (commented):
Thanks for the PR,

but could you explain why this needs to be changed? The current result looks fine to me.
It doesn't seem to be that shaky.
image

WebMechanic:
Then there must be two copies of this precious work of art, 'cos the one I saw (and changed) has "doubled" 3d-ish lines and looks a bit different from your screenshot -- which also looks fine to me :)

Have a look at the diff and you shall see :-)

Gusted:
Are you referring to how the ASCII looks in the code? 
That's normal, I only care about the output, which is on the console. We have to use escape character, but in the output those escape characters are removed.

WebMechanic:
oh, gosh ... right! should've thought of that 🤦‍♂️ 

my bad, sorry for bothering :)
Now, how do I withdraw a PR? 🤔 do I just close this and delete my branch?

Gusted:
> oh, gosh ... right! should've thought of that man_facepalming
> 
> my bad, sorry for bothering :)

No worries 😄, everyone could make that same thought.

PR #7516, opened 2021-11-30
Removed sverigesradio.se

Regnander:
Today I discovered that _sverigesradio.se_ - Sweden's national publicly funded radio broadcaster - has a native dark theme when your device has dark mode activated, so I've now removed all code that relates to this site.

sr


Gusted:
Not everyone has their device enabled on dark mode, so let's keep the changes and not add it to the dark-sites config list.

PR #7542, opened 2021-12-03
Fixes black text on black Google Translate bubble

Andrew15-5:
With this line text in Google Translate bubble is black (Google Chrome +  chrome://flags#enable-force-dark "Enabled with selective inversion of non-image elements").

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1,7 +1,6 @@
 *
 
 INVERT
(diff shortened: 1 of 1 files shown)


Gusted:
Thanks for the PR.

Dark Reader shouldn't be run along-side with things like `chrome://flags#enable-force-dark`, removing this fix will impact users who doesn't have this forced-dark mode of chromium. Closing.

PR #7560, opened 2021-12-06
dell.com icons svg

FadeMind:
Simple as SVG 😁 

SVG files are tricky coded there in many places.


src/config/dynamic-theme-fixes.config
```diff
@@ -3603,6 +3603,13 @@ CSS
 
 ================================
 
+dell.com
+
+INVERT
+svg
```
Gusted, line 3609: Ehhh... All SVG's? That's pretty optimistic.

FadeMind: See yourself. 😁

Gusted: It also inverts the icon on the top-right corner, which is are already correct... Ehm what about `.hpg-cat-item-wrap svg`

Gusted (approved):

PR #7564, opened 2021-12-07
Add fix for fullstackopen code comment highlights

jeansibelius:
Define a higher contrast background color for the code comments in fullstackopen.com when using the dynamic mode of Dark reader.

Example can be viewed for example here: https://fullstackopen.com/en/part8/fragments_and_subscriptions#subscriptions-on-the-client


src/config/dynamic-theme-fixes.config
```diff
@@ -5724,6 +5724,14 @@ img[alt="[NLUUG]"]
 
 ================================
 
+fullstackopen.com
+
+CSS
+.gatsby-highlight-code-line {
+    background-color: rgba(77, 80, 83, 0.3);
```
Gusted, line 5731: So what about:
```suggestion
    background-color: ${hsla(0,0%,48.2%,.3)} !important;
```

jeansibelius: Hi! Thanks a lot for the review and suggestion. If it's ok to ask a couple of questions, I'm curious why hsla over rgba? And the other one: why is the expression wrapped in `${ ... }`? Is that for a preprocessor used in this project?

Gusted: > , I'm curious why hsla over rgba?

It's what the CSS(from fullstackopen) used, I only modified 88.2% -> 48.2% to make it more highlighting

> why is the expression wrapped in `${ ... }`? Is that for a preprocessor used in this project?

https://github.com/darkreader/darkreader/blob/master/CONTRIBUTING.md#how-to-use-the-dev-tools -> Look at the Rules table

Gusted (approved):

src/config/dynamic-theme-fixes.config
```diff
@@ -5724,6 +5724,14 @@ img[alt="[NLUUG]"]
...
 ================================
 
+fullstackopen.com
+
+CSS
+.gatsby-highlight-code-line {
+    background-color: ${hsla(0,0%,48.2%,.3)} !important;
+}
+
+================================
```
Gusted, line 5734:
```suggestion
================================

```

PR #7565, opened 2021-12-07
Add support for wanikani.com

TheColtron:
WaniKani is a popular flashcard based Japanese learning website. Typically English meanings are presented in a different style than Japanese readings, but Dark Readers default reskin causes both to have the same style.

This PR fixes that by updating the style of the English meanings as shown below.

image

image


TheColtron (commented):
Sorry, I'm still figuring out git. This file should not be deleted, but rather should be left unchanged. Trying to correct this now

src/config/dynamic-theme-fixes.config
```diff
@@ -14811,6 +14811,16 @@ IGNORE INLINE STYLE
 
 ================================
 
+wanikani.com
+
+CSS
+#question-type.meaning {
+    color: rgb(214, 209, 201) !important;
```
Gusted, line 14818: What about changing this to the dynamic variable?
```suggestion
    color: var(--darkreader-neutral-color) !important;
```

TheColtron: Good idea. Just tested and this looks solid to me.

Gusted (approved):

PR #7583, opened 2021-12-08
Adding dark background to AWS SSO sign in

JamesTrain:


src/config/dynamic-theme-fixes.config
```diff
@@ -12383,6 +12383,16 @@ INVERT
 
 ================================
 
+*.signin.aws
```
Gusted, line 12386: The `*.` part is not needed.

PR #7589, opened 2021-12-09
Prevent Inversion of Title Text on Webtoons.com

KornFlaks:
New / Proposed:
image

Old / Current:
image

Dark Reader off:
image

(NOT DONE) Inversion of title cards:
image

Choice of pure black `rgb(0,0,0)` was selected due to the original website CSS selecting black as title card text. This will prevent Dark Reader from acting on the text located only on the title card. `IGNORE INLINE STYLE` was not used due to the declaration of this black text being inherited from a global font color. Ignoring dark reader's change to this font color would cause the rest of the website's text to break or require custom overrides for all other text using the global font color.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -15013,6 +15013,15 @@ CSS
 
 ================================
 
+webtoons.com
+
+CSS
+.info {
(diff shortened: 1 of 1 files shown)


Gusted (approved):
LGTM! Thanks

PR #7628, opened 2021-12-13
Change Hacker News BG to --darkreader-neutral-bg

EducatedAlmost:
The inverted background colour of Hacker News (news.ycombinator.com) was a sickly green. I've changed it to `--darkreader-neutral-background` instead, and it looks much better.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -10026,6 +10026,11 @@ news.ycombinator.com
 INVERT
 .votearrow
 
+CSS
+#hnmain {
+    background-color: var(--darkreader-neutral-background) !important;
+}
+
 ================================
 
 newyorker.com


Gusted:
I'm personally not sure about this one - the original website has a yellow background, the modified background color is dark yellow. Which is how it should behave, the dynamic theme is to preserve the original's background-color.

So I'm not sure if I should merge this one.

EducatedAlmost:
The original background colour is #F6F6EF or CMYB(0%, 0%, 3%, 4%). Ivory (#FFFFF0) is twice as yellow, that's white in my book.

EducatedAlmost:
The original background colour is #F6F6EF, CMYB(0,0,3,4). Ivory (#FFFFF0) is twice as yellow, in my book they're both white.

Gusted:
> The original background colour is #F6F6EF, CMYB(0,0,3,4). Ivory (#FFFFF0) is twice as yellow, in my book they're both white.

Well IMO they are both yellow(they are both dominant yellow in CMYB and are in the yellow-range in HSL). The code of dark reader agree with this, it's has a tint, it's not perfect yellow. So I'm still not sure about merging this, I think using this is a better idea:
```css
#hnmain {
    background-color: ${#f6f6ef} !important;
}
```

EducatedAlmost:
> Well IMO they are both yellow(they are both dominant yellow in CMYB and are in the yellow-range in HSL).

By this logic CMYB(0, 0, 99, 99) is also yellow. But to the eye it is black and white.

> I think using this is a better idea:

That works well.

PR #7633, opened 2021-12-13
Invert TN seal

Squaduck:
It shows up as black on a dark background normally.

Before:
before

After:
after

### DIFF
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -2243,7 +2243,7 @@ NO INVERT
 tn.gov
 
 INVERT
-img[src="/etc/clientlibs/tncore/main/img/tn-seal.png"]
+img[src$="tn-seal.png"]
 
 ================================
 
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -2240,6 +2240,13 @@ NO INVERT
 
 ================================
 
+tn.gov
+
+INVERT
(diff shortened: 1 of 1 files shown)


Squaduck:
Would it be more correct to refer to it as `img[src="/etc/clientlibs/tncore/main/img/tn-seal.png"]` or `img[alt="Tennessee State Seal"]`? This is my first time actually doing anything like this.

Gusted:
> Would it be more correct to refer to it as `img[src="/etc/clientlibs/tncore/main/img/tn-seal.png"]` or `img[alt="Tennessee State Seal"]`? This is my first time actually doing anything like this.

Well, it's both correct. In this case `img[src$="tn-seal.png"]` can be used ;)

Squaduck:
Thanks for letting me know! I will add a new commit to use img[src$="tn-seal.png"].

src/config/inversion-fixes.config
```diff
@@ -2240,6 +2240,13 @@ NO INVERT
 
 ================================
 
+tn.gov
```
Gusted, line 2243: It seems like it's at the wrong position:

```
   -   "tn.gov",
        "toggl.com",
        "tpondemand.com",
        "tproger.ru",
        "translate.yandex.ru",
        "travis-ci.org",
        "trip101.com",
    +   "tn.gov",
    ```
    
    It should be moved above tn.gov.

PR #7640, opened 2021-12-14
Makes TVLand.com Dark

torma616:
TVLand.com isn't dark when DarkReader is active. This change fixes that.

(NOTE: TVLand.com may be inaccessible outside of the US)

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -115,16 +115,6 @@ IGNORE INLINE STYLE
 
 ================================
 
-tvland.com
-
-CSS
-.module-container,
-.header-container {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -115,6 +115,16 @@ IGNORE INLINE STYLE
 
 ================================
 
+*.tvland.com
```
Gusted, line 118:
```suggestion
tvland.com
```

Also re-order as appriopate.

Gusted (changes requested):

Gusted (approved):

torma616:
Made the changes for this, and also for vh1.com (an older PR of mine that still had the wildcard left in) and reordered both.

PR #7641, opened 2021-12-14
Makes paramountnetwork.com dark

torma616:
[ParamountNetwork.com](https://www.paramountnetwork.com/) is not dark when DarkReader is active. This change fixes that.

(NOTE: ParamountNetwork.com may not be accessible outside of the US).

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -45,6 +45,15 @@ div[class^='scrollableList'] {
 
 ================================
 
+*.paramountnetwork.com
+
+CSS
+.module-container {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -45,6 +45,15 @@ div[class^='scrollableList'] {
 
 ================================
 
+*.paramountnetwork.com
```
Gusted, line 48:
```suggestion
paramountnetwork.com
```
Re-order as appropriate.

Gusted (changes requested):

Gusted (approved):

PR #7648, opened 2021-12-15
Add fixes for CISCE

Vedanth-P:
This fixes certain issues with the [CISCE website's](cisce.org) header.

The navbar in the header is originally dark with light text, but on inversion becomes light with light text.
The rest of header is originally light with dark text, but on inversion becomes dark with dark text.
Attached are pictures of the original page, the inverted page before the fix, and the inverted page after the fix, in that order
Screenshot from 2021-12-15 09-48-35
Screenshot from 2021-12-15 10-01-22
Screenshot from 2021-12-15 09-59-56
.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2699,6 +2699,14 @@ INVERT
 
 ================================
 
+cisce.org
+
+INVERT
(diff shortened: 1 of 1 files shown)


Gusted (approved):
LGTM

PR #7657, opened 2021-12-16
Added Dicetower.com to dynamic styles

Telemain:
Removed background image from Dicetower.com which was very bright.


Gusted (approved):
LGTM

PR #7669, opened 2021-12-17
Prevent rule and variable names from breaking

Gitoffthelawn:
After a silly amount of time spent, finally figured out a way to prevent rule and variables names from line-breaking within tables displayed in a reasonably large browser viewport.  Actually, figured out 2 different ways because each type had unique solution better for it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,11 +86,11 @@ IGNORE IMAGE ANALYSIS
 ```
 
 | Rule | Description | Notes / Examples |
-|-|-|-|
-| `INVERT` | Inverts specified elements. | **Dynamic Mode**: INVERT only for dark images that are invisible on dark backgrounds. |
-| `CSS` | Adds custom CSS to a web page. | `!important` keyword should be specified for each CSS property to prevent overrides by other stylesheets.<br>**Dynamic mode** supports `${COLOR}` template, where `COLOR` is a color value before the inversion. <br>*Example*: `${white}` will become `${black}` in dark mode. |
-| `IGNORE INLINE STYLE` | Prevents inline style analysis of matched elements. | *Example*: `<p style="color: red">` element's style attribute will not be changed. |
-| `IGNORE IMAGE ANALYSIS` | Prevents background images from being analyzed for matched selectors. |  |
+|---|---|---|
+| **INVERT** | Inverts specified elements. | **Dynamic Mode**: INVERT only for dark images that are invisible on dark backgrounds. |
+| **CSS** | Adds custom CSS to a web page. | `!important` keyword should be specified for each CSS property to prevent overrides by other stylesheets.<br>**Dynamic mode** supports `${COLOR}` template, where `COLOR` is a color value before the inversion. <br>*Example*: `${white}` will become `${black}` in dark mode. |
+| **IGNORE&nbsp;INLINE&nbsp;STYLE** | Prevents inline style analysis of matched elements. | *Example*: `<p style="color: red">` element's style attribute will not be changed. |
+| **IGNORE&nbsp;IMAGE&nbsp;ANALYSIS** | Prevents background images from being analyzed for matched selectors. |  |
 
 ## Adding a new color scheme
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -86,11 +86,11 @@ IGNORE IMAGE ANALYSIS
 ```
 
 | Rule | Description | Notes / Examples |
-|-|-|-|
-| `INVERT` | Inverts specified elements. | **Dynamic Mode**: INVERT only for dark images that are invisible on dark backgrounds. |
-| `CSS` | Adds custom CSS to a web page. | `!important` keyword should be specified for each CSS property to prevent overrides by other stylesheets.<br>**Dynamic mode** supports `${COLOR}` template, where `COLOR` is a color value before the inversion. <br>*Example*: `${white}` will become `${black}` in dark mode. |
-| `IGNORE INLINE STYLE` | Prevents inline style analysis of matched elements. | *Example*: `<p style="color: red">` element's style attribute will not be changed. |
-| `IGNORE IMAGE ANALYSIS` | Prevents background images from being analyzed for matched selectors. |  |
+|---|---|---|
+| **INVERT** | Inverts specified elements. | **Dynamic Mode**: INVERT only for dark images that are invisible on dark backgrounds. |
```
Gusted, line 90: Why making it bold?

CONTRIBUTING.md
```diff
@@ -132,13 +132,12 @@ CSS
...
 Here is a full table of available CSS variables:
-
 | Variable | Description | Use |
-|-|-|-|
-| `--darkreader-neutral-background` | Neutral background color that corresponds to the user's settings. | Mostly used for elements that have a wrong background color |
-| `--darkreader-neutral-text` | Neutral text color that corresponds to the user's settings. | Used for elements with a wrong text color |
-| `--darkreader-selection-background` | The text color setting defined by the user. | The user's Text Color setting |
-| `--darkreader-selection-text` | The background color setting defined by the user. | The user's Background Color setting |
+|---|---|---|
+| **`--darkreader-neutral-background`** | Neutral background color that <br>corresponds to the user's settings. | Mostly used for elements that have <br>a wrong background color |
```
Gusted, line 137: Same as above.

CONTRIBUTING.md
```diff
@@ -86,11 +86,11 @@ IGNORE IMAGE ANALYSIS
...
-| `IGNORE INLINE STYLE` | Prevents inline style analysis of matched elements. | *Example*: `<p style="color: red">` element's style attribute will not be changed. |
-| `IGNORE IMAGE ANALYSIS` | Prevents background images from being analyzed for matched selectors. |  |
+|---|---|---|
+| **INVERT** | Inverts specified elements. | **Dynamic Mode**: INVERT only for dark images that are invisible on dark backgrounds. |
+| **CSS** | Adds custom CSS to a web page. | `!important` keyword should be specified for each CSS property to prevent overrides by other stylesheets.<br>**Dynamic mode** supports `${COLOR}` template, where `COLOR` is a color value before the inversion. <br>*Example*: `${white}` will become `${black}` in dark mode. |
+| **IGNORE&nbsp;INLINE&nbsp;STYLE** | Prevents inline style analysis of matched elements. | *Example*: `<p style="color: red">` element's style attribute will not be changed. |
+| **IGNORE&nbsp;IMAGE&nbsp;ANALYSIS** | Prevents background images from being analyzed for matched selectors. |  |
```
Gusted, line 93: `&nbsp;` I'd find this a very very hacky way.

CONTRIBUTING.md
Gitoffthelawn: Just to make it stand out a little more.  Because of the `&nbsp;` characters, these can't use the backtick (code) style, leaving bold, italics, and font-size as the only real choices for emphasis.  I felt that bold was the best choice.

Not sure where you are in your career, but you might benefit from being mindful not to wander too far into [this realm](https://www.thefreedictionary.com/micromanage). :)

CONTRIBUTING.md
Gitoffthelawn: Consistency with above, although was able to also use the backtick (code) style with these.

CONTRIBUTING.md
Gitoffthelawn: Me too.  Do you know a better technique?  I tried about 10 different techniques and spent over 30 minutes reading documentation and other resources before I settled on `&nbsp;`.  Ideally, one could just add a little CSS, but every resource I could find says that GitHub prohibits it.

Gitoffthelawn: BTW, I copied the result into a text editor and was pleased that the resulting spaces are true ASCII 0x20 spaces.

I couldn't find an equivalent non-breaking character for dashes, as everything that looks like a dash (but isn't really one) will not be ASCII  0x2D.

I wanted copy-paste to work for people and not mess up the destination file with characters that looked correct visually but weren't actually the correct ASCII value.

Gusted: Well, let it be then, it indeed seems to be the only way.

CONTRIBUTING.md
Gusted: 👍🏽

Gusted (approved):
Thank you!

Gitoffthelawn:
> Thank you!

You're welcome.  Happy to help.

PR #7675, opened 2021-12-17
Update dark-sites.config

marekmaskarinec:
Removed `marekmaskarinec.github.io` as it will become offline soon and added `mrms.cz`.


Gusted (approved):
LGTM, thanks!

PR #7677, opened 2021-12-18
Update dark-sites.config

TrevorBrown-dev:
added a site


Gusted (changes requested):
Please move the site into alphabetic order.

PR #7704, opened 2021-12-21
Added dark site

sharkiller:
Added dark site with minigames that darkreader mess with some background colours needed for some minigames.

Eg: https://sharkiller.ddns.net/nopixel_minigame/lowervault/


Gusted (approved):
LGTM

PR #7705, opened 2021-12-21
Makes jira subtask outer border visible

samuelgfeller:


src/config/dynamic-theme-fixes.config
```diff
@@ -1347,6 +1347,10 @@ span.loader-wrapper a,
 code:first-of-type {
     background-image: linear-gradient(to right, rgba(255,255,255,0.03),rgba(255,255,255,0.03)calc(1ch + 16px),transparent calc(1ch + 16px),transparent) !important;
 }
+.sc-1jjaulc-0 {
+    box-shadow: 0 0 4px rgba(240,246,252,0.15) !important;
+    background-color: rgba(240,246,252,0.02) !important;
```
Gusted, line 1352: Could these 2 colors be in the form of `${rgba(original color)}`?

src/config/dynamic-theme-fixes.config
```diff
@@ -1347,6 +1347,10 @@ span.loader-wrapper a,
 code:first-of-type {
     background-image: linear-gradient(to right, rgba(255,255,255,0.03),rgba(255,255,255,0.03)calc(1ch + 16px),transparent calc(1ch + 16px),transparent) !important;
 }
+.sc-1jjaulc-0 {
```
Gusted, line 1350: Hmm, this is using randomized selectors(which tends to change every few days/weeks on sites using them), can this be changed to not rely on the randomized selectors?

src/config/dynamic-theme-fixes.config
samuelgfeller: I already tried but they dont look good at all. I could however use `${other color that is good inverted}` if you prefer.

src/config/dynamic-theme-fixes.config
samuelgfeller: We talked about this in #5953. Jira isn't cooperative with this unfortunately.

Gusted: > We talked about this in #5953. Jira isn't cooperative with this unfortunately.

Woops. That's correct.

src/config/dynamic-theme-fixes.config
Gusted: Well you can use a little trick, you can use the value(to be used in the fixes's `${...}` which dark reader generates from `${rgba(240,246,252,0.15)}` as dark reader creates "perfect" inverts.

samuelgfeller: That was my bad, you [already told me](https://github.com/darkreader/darkreader/pull/5953#discussion_r646749678) this.  
> If you take the the correct color and put that in the ${}, get the result of that color from the dev-tools and use that as the value for the fix. It should be the exact same color after you put it in ${}

Its that what you mean, right?
```css
.sc-1jjaulc-0 {
    box-shadow: 0 0 4px ${rgba(3,9,15,0.15)} !important;
    background-color: ${rgba(3, 9, 15, 0.02)} !important;
}
```
To obtain the colors I wanted 
```css
box-shadow: 0 0 4px rgba(240,246,252,0.15) !important;
background-color: rgba(240,246,252,0.02) !important;
```

I correct it.

Gusted (approved):

PR #7725, opened 2021-12-23
Fix for `zombo.com`

Myzel394:


src/config/dynamic-theme-fixes.config
```diff
@@ -17070,6 +17070,18 @@ INVERT
 
 ================================
 
+zombo.com
+
+CSS
+body {
+    background: none !important;
```
Gusted, line 17077: Why should the background be removed?

src/config/dynamic-theme-fixes.config
```diff
@@ -17070,6 +17070,18 @@ INVERT
...
+}
+img.thefade {
```
Gusted, line 17079: Why should the opacity be lowered?

src/config/dynamic-theme-fixes.config
Myzel394: I think it fits better when there's no gradient

src/config/dynamic-theme-fixes.config
Myzel394: The colors are very light, that's why I thought they should only be partially visible. To accomplish this in dark mode, I lowered the opacity a bit

src/config/dynamic-theme-fixes.config
Gusted: So, the color of the gradient is fine and modified correctly, the original site has it as well, so it shouldn't be removed.

src/config/dynamic-theme-fixes.config
Gusted: Well, the colors of the logo are more light than the image, but that could also be because of my monitor settings.

Myzel394:
@q1800 yes you're

PR #7727, opened 2021-12-24
Fix for Google Drive

Vedanth-P:
Each menu item on the right-click context menu consists of an image, a text label, and (sometimes) another image, arranged horizontally. The selector ```div[role="menu"] > div > div[role="menuitem"][class*=" "] > div > :last-child``` selects and inverts the last of these elements on every menu item. However, when the last element is a text label, the text often seems to be incorrectly inverted.
image image
In some cases, images also get wrong inversions when they are returned by the selector (notice the little black arrow on the right end of the menu item).
image
(The images are from the right-click menu for a file in Google Drive)

However, when the selector is removed, things seem to look fine.

For reference, this selector was added in #6311.


Gusted (approved):
LGTM, seems to be indeed a faulty selector.

PR #7739, opened 2021-12-25
Remove Youtube's notification border

NguyenASang:


src/config/dynamic-theme-fixes.config
```diff
@@ -16690,6 +16690,9 @@ CSS
 html[hide-scrollbar] ::-webkit-scrollbar {
     display: none !important;
 }
+#notification-count.ytd-notification-topbar-button-renderer {
```
Gusted, line 16693: The notification border is on the original site, so we shouldn't be removing it.

NguyenASang:
Change the color of it like the background ? (default youtube like that)

PR #7777, opened 2021-12-28
linuxcompatible.org is dark

FadeMind:
https://linuxcompatible.org


Gusted:
image

Doesn't seem to be dark-by-default to me.

PR #7790, opened 2021-12-29
Update for developers.google.com

hjpoe:


src/config/dynamic-theme-fixes.config
```diff
@@ -3915,6 +3915,8 @@ developers.google.com
 
 INVERT
 .devsite-google-wordmark
+a.devsite-nav-title.devsite-nav-active
+div.devsite-article-meta.nocontent
```
Gusted, line 3919: I couldn't find these 2 elements, could you point me to a example site?

hjpoe: Here is the site I was on. 

https://developers.google.com/photos/library/reference/rest/

Gusted: Both seems to be impacted by a edge case of the variables which is fixed by #6107 (Which needs some improvements, before it can be merged)

Issue #6955, opened 2021-09-29
[Bug] Getting endless "Loading, please wait" in LibreWolf


CuriousApe2020:
I run "LibreWolf Version 92.0.1 (64-bit)" and "Dark Reader Version 4.9.37.1"
And I still only see "Loading please wait" when clicking the Extension icon similar to: issue: https://github.com/darkreader/darkreader/issues/6819
Version 4.9.34 Works fine all newer are broken.

bershanskiy:
That bug was fixed by #6927. I don't remember exactly which commit went into 4.9.37.1, but I think you are running a known-faulty version. Could you please update to current `master`? I can give you detailed instructions how to do that and build it for you.

CuriousApe2020:
You are right, I reinstalled LibreWolf and everything works now sorry for the "false" report.

NyaomiDEV:
I am getting the same error with Librewolf 94.0.2 and Dark Reader 4.9.43.

Gusted:
> I am getting the same error with Librewolf 94.0.2 and Dark Reader 4.9.43.

Yeah, that's to be expected... Librewolf has decided to change their user-agent to a unique one. I haven't come around to add the new user-agent yet as it's really weird to see that it has their own user-agent.

Gusted:
#7632

PR #7802, opened 2021-12-30
Using `gecko` to define a firefox-based browser (fixing #6955)

Syudagye:
Fixing #6955 for all firefox-based browsers, by targeting the Gecko engine and not the browser name.
Tested and working for WaterFox, LibreWolf, Icecat, Dot and Firefox.
I couldn't install the extension on FireDragon, but i didn't spent more than 2 minutes on it to be honest. It should be working just fine.

### DIFF
--- a/src/utils/platform.ts
+++ b/src/utils/platform.ts
@@ -3,7 +3,7 @@ const platform = typeof navigator === 'undefined' ? 'some platform' : navigator.
 
 export const isChromium = userAgent.includes('chrome') || userAgent.includes('chromium');
 export const isThunderbird = userAgent.includes('thunderbird');
-export const isFirefox = userAgent.includes('firefox') || userAgent.includes('librewolf') || isThunderbird;
+export const isFirefox = userAgent.includes('gecko') || isThunderbird;
 export const isVivaldi = userAgent.includes('vivaldi');
 export const isYaBrowser = userAgent.includes('yabrowser');
 export const isOpera = userAgent.includes('opr') || userAgent.includes('opera');


Syudagye:
Forgot to mention it, but of course this improves #7632.

src/utils/platform.ts
```diff
@@ -3,7 +3,7 @@ const platform = typeof navigator === 'undefined' ? 'some platform' : navigator.
 
 export const isChromium = userAgent.includes('chrome') || userAgent.includes('chromium');
 export const isThunderbird = userAgent.includes('thunderbird');
-export const isFirefox = userAgent.includes('firefox') || userAgent.includes('librewolf') || isThunderbird;
+export const isFirefox = userAgent.includes('gecko') || isThunderbird;
```
Gusted, line 6: I think it's really bad and not correct go target render engines. And that's all because of how messy user-agents have become these days. Because this PR will break functionality chromium, safari, kiwi browser, vivaldi etc.

Because they all include a simple sentence `AppleWebKit/537.36 (KHTML, like Gecko)`, thus this will evaluate to `true` to a lot of browsers, which shouldn't. We want to know which browsers differ in certain implementation, so we can apply workaround. We shouldn't target a render engine(which also just so happens to be on every user-agent on all major browsers).

I cannot accept such patch in this form, and I personally don't like any workarounds either to accomplish the same deal. It's not ideal in the current form, nor in a workaround patched form that targets a render engine.

Gusted (changes requested):

src/utils/platform.ts
Syudagye: Oh i didn't checked chromium, that's unfortunate. I'll fix this a bit later, thanks for the review !

Syudagye:
Looking back at this, I think it is fixed now. [`src/utils/platform.ts`](https://github.com/darkreader/darkreader/blob/main/src/utils/platform.ts) seems to contains a lot more browsers.

PR #7809, opened 2021-12-30
Added landchad and based.cooking to dark site list

paniash:


src/config/dark-sites.config
```diff
@@ -72,6 +72,7 @@ avengedsevenfold.com
 azeria-labs.com
 azm.azerialabs.com
 badboybill.com
+based.cooking
```
Gusted, line 75: image

This doesn't seem to be dark-by-default.

Gusted (approved):

PR #7825, opened 2022-01-01
fix stackexchange/overflow closed info box, rpg logo

isaec:
This makes the info box readable, and fixes the white outline on the rpg logo.
image
I tested the change to the info box on some subdomains, here is a sampling of how they look after the change:
rpg
image
meta
image
codegolf
image
image
worldbuilding
image
and of course, stackoverflow:
image

before shot of stackoverflow (they all looked like this, approximately):
image

Given how many sites this change applies to, I am sure there is a subdomain with particularly crazy styling - please let me know if you know of one I missed.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -13486,8 +13486,11 @@ INVERT
 .h-auto[alt="Theoretical Computer Science"]
 .h-auto[alt="Unix & Linux"]
 .h-auto[alt="Web Applications"]
+.h-auto[alt="Role-playing Games"]
 a.js-gps-track::before
 img[alt="The Stack Exchange Network"]
+aside.s-notice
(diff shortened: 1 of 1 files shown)


Gusted (approved):

crissthiandi:
Hi ✋, when I visit stackoverflow the class `.s-notice` doesn't need to will "invert" because it was dark, then I need to reinvert the class using `!  .s-notice` in the dynamic theme editor. 
My questing is, how users with "dark mode activate on some pages as StackOverflow" could use commit's changes?

there is a method to detect if users have a dark theme in pages before inverting CSS?

Gusted:
> there is a method to detect if users have a dark theme in pages before inverting CSS?

Nope, we generally discourage to use dark reader on already dark pages, because the results are mediocre at best.

PR #7826, opened 2022-01-02
Fix for wacom.com

hjpoe:
This fixes the logos at the top and bottom of the page. This site does a lot of the pictures with words on top of it thing. Dark Reader flips the text to white, so they are white text on top of white backgrounded images. I'm not sure what to do about this, and I don't want to just make them dark text on white background, so I've done nothing about that part, even though as is, the text is not readable mostly everywhere.


hjpoe:
This CSS makes it better than it is, but I don't think it's a real solution.

CSS
h3, p {
    color: #525252 !important;
}

Gusted:
> This fixes the logos at the top and bottom of the page. This site does a lot of the pictures with words on top of it thing. Dark Reader flips the text to white, so they are white text on top of white backgrounded images. I'm not sure what to do about this, and I don't want to just make them dark text on white background, so I've done nothing about that part, even though as is, the text is not readable mostly everywhere.

So first of, let's apply a niece filter to these pictures, so we don't blind our eyes 😄 
```css
selector_here {
}
```

That should make the text more readable already.

hjpoe:
IDK how I managed to do the same thing 3 times without noticing it actually went... Sorry about that. I guess I'm still learning how GitHub works. 

Also, @Gusted, your CSS is magical, and my eyes no longer hurt. 😄

Gusted (approved):

Gusted:
Ready to be merged?

hjpoe:
Yes, ready to be merged.

PR #7857, opened 2022-01-04
Add dark sites

nikkehtine:
note: aceship.github.io has two images that take up half of the screen, both link to pages that are dark

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -160,7 +160,6 @@ cinema-city.pl
 cl00e9ment.gitlab.io/kingdom-blazon-generator/
 clash3d.com
 clashofstats.com
-clowncore.computer
 clt.gg
 codepen.io
 codesandbox.io
@@ -690,7 +689,6 @@ polsatboxgo.pl
 polsatgo.pl
 pony.tube
 pool.pm
-pornhub.com
 powercord.dev
 ppluss.de
 pr0gramm.com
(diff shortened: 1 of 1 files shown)


src/config/dark-sites.config
```diff
@@ -686,6 +690,7 @@ polsatboxgo.pl
 polsatgo.pl
 pony.tube
 pool.pm
+pornhub.com
```
Gusted, line 693: NSFW Sites aren't permitted.

src/config/dark-sites.config
```diff
@@ -159,6 +160,7 @@ cinema-city.pl
 cl00e9ment.gitlab.io/kingdom-blazon-generator/
 clash3d.com
 clashofstats.com
+clowncore.computer
```
Gusted, line 163: Questionable, as it's not really dark themed either.

src/config/dark-sites.config
nikkehtine: Is that a rule? I've checked [CONTRIBUTING.md](https://github.com/darkreader/darkreader/blob/7daacbeddd59e7742c21bca8a5fb371cd72962e7/CONTRIBUTING.md) and searched through the project's files and it doesn't seem to be listed anywhere. Can you add that to CONTRIBUTING.md?

Gusted: It's more or less a unwritten rule at the moment.

Gusted (approved):

PR #7860, opened 2022-01-04
Update dark-sites.config

danilohere:
Adding my site to Dark sites config


src/config/dark-sites.config
```diff
@@ -197,6 +197,7 @@ daily-fire.com
 daksh.eu.org
 danidev.net
 dankmemer.gg
+danstudioapps.com
```
Gusted, line 200: image
This isn't that dark to me.

PR #7863, opened 2022-01-05
Fix planetminecraft.com

ModProg:
Especially for project pages such as https://www.planetminecraft.com/texture-pack/mc-ctm-overhaul/

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -11543,8 +11543,9 @@ CSS
 #resource-log,
 .tag,
 .view_more:not(:hover),
-#popup {
-    background: var(--darkreader-bg--content-bg) !important;
+#popup,
+code {
+    background: var(--darkreader-neutral-background) !important;
 }
 
 ================================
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -11532,6 +11532,23 @@ p.site-title
...
+#container,
+.forum_replies_container,
+.forum_reply,
+.contents,
+.spoiler_header,
+#resource-log,
+.tag,
+.view_more:not(:hover),
+#popup {
+    background: var(--darkreader-bg--content-bg) !important;
```
Gusted, line 11547: I feel that a bit too much selectors are using this specific variables. E.g. `#container` should be using `--darkreader-neutral-bg`.

Gusted (approved):

src/config/dynamic-theme-fixes.config
ModProg: makes sense, I don't think there is a reason for any of those to be a specific color, so I used neutral-background for all of them now.

PR #7873, opened 2022-01-06
Removing SVG inversions from Finn.no

martinetruls:
The dark mode switch is creating some very unlucky color changes to our svg illustrations, hence removing all SVGs from inverting.

Skjermbilde 2022-01-06 kl  10 40 33


src/config/inversion-fixes.config
```diff
@@ -2722,3 +2722,10 @@ yscec.yonsei.ac.kr
 INVERT
 iframe
 #page-sidebar
+
+================================
+
+finn.no
```
Gusted, line 2728: Fix is not at the correct order, and I think you're using `dynamic` theme, so this is the wrong config.

martinetruls: @Gusted Thanks for the quick reply! I have update the PR now, but was unsuccessful to view any changes when using darkreader devtools. Please let me know if I'm not using the correct syntax for making the darkreader ignore inverting my SVGs.

Gusted (approved):
Thanks!

martinetruls:
@Gusted It looks like my PR also needs a workflow approval. Is that something you are able to do?

Gusted:
> @Gusted It looks like my PR also needs a workflow approval. Is that something you are able to do?

Woops, I forgot to merge this 😄 Thanks for the reminder.

martinetruls:
@Gusted I tried to test and see if my svgs look ok now at finn.no, but they still don't. I looked into the devtools of dark reader extension in chrome, but I can't find my fixes for finn.no there? Do I have to wait for a new release of the extension maybe?

Gusted:
> Do I have to wait for a new release of the extension maybe?

See https://github.com/darkreader/darkreader#site-fixes

PR #7881, opened 2022-01-06
Add dark sites

nikkehtine:


src/config/dark-sites.config
```diff
@@ -973,6 +974,7 @@ wowaudit.com
 wowhead.com
 wowinterface.com
 wtfast.com
+www.canalplus.com
```
nikkehtine, line 977: I don't know about other languages, in Polish the main page seems to be dark, but help and other pages are light

Gusted (approved):

src/config/dark-sites.config
FadeMind: @nikkehtine  see

https://kup.pl.canalplus.com/oferta/wybierz-pakiet

Rule should be more general (without `www.` prefix)

CC @Gusted

PR #7887, opened 2022-01-07
Add Dark sites

New-dev0:


src/config/dark-sites.config
```diff
@@ -36,6 +36,7 @@ anon.sx
 aonprd.com
 app.destinyitemmanager.com
 app.keeweb.info
+app.koyeb.com
```
Gusted, line 39: image
Not dark on the login page.

Gusted (changes requested):

src/config/dark-sites.config
New-dev0: Removed. side element didn't appeared on mobile view 😅

Gusted (approved):

PR #7966, opened 2022-01-16
Fix for docs.soliditylang.org

slopra:
- Fix overflow-y background-color mismatch.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4513,6 +4513,15 @@ CSS
 
 ================================
 
+docs.soliditylang.org
+
+CSS
+html {
+    background-color: unset !important;
+}
+
+================================
+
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -4513,6 +4513,15 @@ CSS
 
 ================================
 
+docs.soliditylang.org
+
+CSS
+html {
+    background-color: unset !important;
```
Gusted, line 4520: We shouldn't be touching the background-color of `html` in general. It's a important property for the browsers and used for initial rendering. I don't think that this is fixable.

slopra: I can appreciate that rule of thumb. Pushing an alternate idea.

Gusted (approved):

PR #7987, opened 2022-01-19
Fix Slack.com dynamic theme

InfernoZeus:
There are two issues with Slack.com's current dynamic fixes:
  - The invert logo on the homepage no longer works (the logo is now an SVG, and has a newly named class)
  - Buttons in the actual "app" aren't darkened properly

The first issue is resolved by coloring the first path of the logo's SVG element. This is the only way to select the text part of the logo as they have not named the paths, and the colorful "pips" of the logo are part of the same SVG element.

The second issue is resolved by letting the buttons inherit their text color and background. With that change, there is no indication when hovering, so I added a basic underline decoration for that. I also took care not to enable it for primary and secondary buttons which seem to be handled fine by the dynamic mode. Unfortunately, it does have a side effect - some buttons which were previously "colored" (see example below) are now greyer, but it was the only solution I could find that works in both light and dark schemes.

Fixing only button background works fine in the dark scheme
Screenshot_2022-01-19_00-58-39
but less so in the light scheme:
Screenshot_2022-01-19_00-58-09

The result when fixing both looks like this - first in the dark scheme:
image
and in the light scheme:
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -13408,8 +13408,17 @@ body {
 
 slack.com
 
-INVERT
-.slack_logo > img
+CSS
+.c-slacklogo svg path:first-child {
+    fill: var(--darkreader-selection-text) !important;
+}
+.c-button:not(.v--primary,.v--secondary) {
+    background: inherit !important;
+    color: inherit !important;
+}
+.c-button:not(.v--primary,.v--secondary):hover {
+    text-decoration: underline !important;
+}
 
 ================================
 
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -13412,11 +13412,11 @@ CSS
 .c-slacklogo svg path:first-child {
     fill: var(--darkreader-selection-text) !important;
 }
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -13408,8 +13408,17 @@ body {
 
 slack.com
 
-INVERT
-.slack_logo > img
+CSS
+.c-slacklogo svg path:first-child {
+    fill: var(--darkreader-selection-text) !important;
+}
+.c-button:not(.v--primary,.v--secondary) {
```
Gusted, line 13415:
```suggestion
.c-button:not(.v--primary, .v--secondary) {
```

src/config/dynamic-theme-fixes.config
```diff
@@ -13408,8 +13408,17 @@ body {
...
+    background: inherit !important;
+    color: inherit !important;
+}
+.c-button:not(.v--primary,.v--secondary):hover {
```
Gusted, line 13419:
```suggestion
.c-button:not(.v--primary, .v--secondary):hover {
```

Gusted (approved):
LGTM

PR #7993, opened 2022-01-19
Fix for weather.com

### LINKED ISSUE #7986: [Feature Request (or bug?)] Option to exclude images from inverting (or a tweak setting)
# Feature Request

Potentially, it could be considered intended behavior or at the same time a bug.

## Screenshots
Use weather.com with the zip code 90210:
https://weather.com/weather/tenday/l/79a8aa24a1dbb73f308d77b3de429eef4da1cd0d83198c44757b11e3fea6dbf8

## Additional Context
For the most part, DarkReader is amazing and incredibly accurate. For 99% of the time, Dynamic mode is what I use because it's most compatible and pleasing to the eye. The amount of customization is excellent too to further tweak things (I'm on version 5 preview).

However there are certain sites where Dynamic gets a little too 'aggressive' with images. The weather.com example above shows it. With Dynamic mode on, it will make any icon that has a sun in it black. If you use a near black background like I do, it turns essentially invisible. The other themes don't necessarily have that problem on that site, but they also make the rest of the website look very poor, so really not an option.

It would be nice to have a feature/tweak/setting that could allow some customization that accomplishes this. That said, I do see there's a dev section which seems to allow you to do exclusions or addition via HTML, but as someone who's not intimately familiar with that sort of thing (and I feel completely out of joe user's knowledge), it would be nice to have that option somehow.

If this has already been implemented and I'm just not seeing it, I greatly apologize. Also if something like that would be a coding nightmare or impossibility I also totally understand that as well.

Thank you so much for this great and free product. I recommend it to everyone I know.

Gusted:
- Resolves #7986

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -16145,6 +16145,8 @@ text[class*="DonutChart"]
 
 IGNORE INLINE STYLE
 [id^="svg-symbol"] *
+svg[class*="alerts--"] *
+#svg-symbol-sun
 
 ================================
 


godzfire:
Thank you for doing Sunny, but it seems that there are more that suffer from the same issue. Is there a way to disable/whitelist it for all of the weather conditions?

Screen Shot 2022-01-20 at 6 15 18 PM
Screen Shot 2022-01-20 at 6 15 26 PM

PR #7995, opened 2022-01-19
Detect website's dark theme

alexanderby:
- Disable Dark Reader styles during the check.
- Check Root and Body background colors.
- Works for Dynamic, Filter and Static modes.


alexanderby:
Hi @Gusted! You are right about the reflow. I've upgraded it to make a check when there's some style and not so much content. Haven't seen it took more that 2ms, even in Gmail.

Gusted (approved):
The code looks fine, interesting approach definetly. My PR for this can be closed. However I think this can be merged AFTER a new release is made, as this will most likely be able to cause regressions(when people enable this feature) and the current release will already contain 2 months of commits.

alexanderby:
Yes, like always, anything can break. Also the UI should be well thought. It should tell somewhere that dark theme was detected, but also let toggle the website back.

rakleed:
@alexanderby what about misbehaving? For example, https://meduza.io/, https://mac.getutm.app/, https://addons.mozilla.org/. After all, now you can only completely disable the function of recognizing the dark theme of the site, which is inconvenient. Perhaps you should add a list of sites where DarkReader will always work?

alexanderby:
@rakleed Did you mean on iOS? The implementation is a bit different there, the check launches at a wrong moment. I will fix it.

rakleed:
@alexanderby no, I'm talking about the desktop version as well.

And maybe I didn’t express myself very clearly, but I would like to have a list of exceptions to which you can manually add a site (for example, meduza.io) so that Dark Reader still works on it, even if the automatic dark theme recognition function does not work correctly.

GitExploraa:
> @alexanderby no, I'm talking about the desktop version as well.
> 
> And maybe I didn’t express myself very clearly, but I would like to have a list of exceptions to which you can manually add a site (for example, meduza.io) so that Dark Reader still works on it, even if the automatic dark theme recognition function does not work correctly.

did this get any updates please?

PR #8029, opened 2022-01-22
Update dark-sites.config

MKSSO5:
Added mksso.dev to the list.


MKSSO5:
(Example: [https://mksso.dev/newlanding/](url) )

Gusted:
We currently don't accept WIP sites([See 3.](https://github.com/darkreader/darkreader/blob/master/CONTRIBUTING.md#adding-a-website-that-is-already-dark)), I will let this PR open for now, you can let me know when the page is in production.

PR #8036, opened 2022-01-23
Don't invert wikipedia images.

msuozzo:
This also fixes the lightbox element's white background.

Try as I might, I was unsuccessful getting any of the "invert", "noinvert", or "removebg" properties to work for either of these image tags so the custom "rules" was the best I could do.


Gusted (changes requested):
We don't use the fix_inversion.json anymore.

PR #8059, opened 2022-01-28
Use longer timeout for mutex

### LINKED ISSUE #8056: [Bug] Sync in Firefox or Chrome causes Dark Reader settings to reset to an earlier version
# Bug Report

## Website Address
N/A

## Bug Description
When using Dark Reader on multiple computers and Firefox Sync or Chrome Sync is enabled, Dark Reader settings will frequently reset to a previous version of the settings. This seems to occur even if I only have the browser open on one computer. This is annoying because I lose several of the sites I've added to the whitelist/blacklist and have to add them multiple times.

## Steps To Reproduce
- Have Dark Reader on two browsers which are synced to each other.
- Add a blacklist item or change from "Ignore Listed" to "Not Ignore Listed"
- Wait a few minutes
- The setting you changed will be reset to what it was before

## Expected Behavior
- When syncing, sync should always prefer the most recent version of settings

## Actual Behavior
- Sync seems to prefer the version of settings in the cloud, regardless of if it is the most recent, and will override local settings with it frequently

## Screenshots
N/A

## System Information
I've been experiencing this issue for a few months, across several Firefox and Chrome versions, on Windows, macOS, and Linux machines. I have been using the latest Dark Reader version from either Chrome Web Store or Firefox Extensions, respectively, through the entire time. This is always with the latest stable version of the browser at the time.

## Additional Context
N/A

Gusted:
- Use a longer mutex in order to take into account for `sync` storage to get notified for a "update".
- Resolves #8056

### DIFF
--- a/src/background/utils/extension-api.ts
+++ b/src/background/utils/extension-api.ts
@@ -131,7 +131,7 @@ export async function writeSyncStorage<T extends {[key: string]: any}>(values: T
                 return;
             }
             resolve();
-            setTimeout(() => mutexStorageWriting.unlock(), 500);
+            setTimeout(() => mutexStorageWriting.unlock(), 5000);
         });
     });
 }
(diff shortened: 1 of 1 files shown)


Gusted:
Firefox & Chromium binaries of this PR:
[binaries.zip](https://github.com/darkreader/darkreader/files/7957957/binaries.zip)

Gusted:
Superseeding by #8064

PR #8103, opened 2022-02-01
Fixed the images not visible in chem.libretexts.org

Sahil2004:
Before this, the images weren't visible. This change fixes it there and now at least images are visible but they have a white background.

Before:
image

After:
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2703,11 +2703,8 @@ header .logo
 
 chem.libretexts.org
 
-CSS
-.internal {
-    background-image: none !important;
-    background-color: ${black} !important;
-}
+INVERT
+.internal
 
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -2701,6 +2701,16 @@ header .logo
 
 ================================
 
+chem.libretexts.org
+
+CSS
+.internal {
+    background-image: none !important;
+    background-color: ${black} !important;
+}
```
Gusted, line 2710: What about:

```

INVERT
.internal
```
?

Sahil2004: It is working better. It gives a dark background with a proper image. I will change that in the pr.

Gusted (approved):

PR #7427, opened 2021-11-20
[WIP] Work on text styles and added testing

mattalexx:

PR #8122, opened 2022-02-03
Fix "DeprecationWarning: Pattern %d{DATE} is deprecated" warnings whe…

mattalexx:
…n running tests

Here's a screencast that shows before and after:

https://asciinema.org/a/LwVOMURYYCfbhB7F4cFYkiZC9

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,7 +28,7 @@
         "globby": "11.0.4",
         "jasmine-core": "4.0.0",
         "jest": "27.4.7",
-        "karma": "6.3.11",
+        "karma": "6.3.12",
         "karma-chrome-launcher": "3.1.0",
         "karma-coverage": "2.1.0",
         "karma-firefox-launcher": "2.1.2",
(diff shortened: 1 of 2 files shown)


Gusted (approved):

mattalexx:
@Gusted By the way, why are you not running the Karma tests in the CI pipeline?

Gusted:
> @Gusted By the way, why are you not running the Karma tests in the CI pipeline?

Because 99% of the PR's don't need to run Karma tests, there are more config PR's than actual code updates. Also karma would require a screen(as we didn't configure with headless, which for firefox would be a problem, as that doesn't have a headless configuration at the moment).

mattalexx:
Ah, that makes sense. 

Github Actions allows you to run steps conditionally. So we could set up a Karma run to test changes in certain directories. 

I would like to set that up for my upcoming changes for PR #7427, would that be alright?

Gusted:
> Github Actions allows you to run steps conditionally. So we could set up a Karma run to test changes in certain directories. https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths

Well that must be a new feature of the github actions - last time I tried to set it up I'd had to use hacky code to achieve that.



> I would like to set that up for my upcoming changes for PR #7427, would that be alright? I could just use Chrome Headless. But you might consider eventually using Sauce. They're free for open source projects, I believe.

Sure thing, for the CI only running chrome-headless is fine, as long we don't forget to run the firefox tests before releases as well 😄. 

I'm not really fancy on using third-party services.

mattalexx:
> as long we don't forget to run the firefox tests before releases as well smile.

Just to get something and working, I'll concentrate on getting Chrome working first. Depending on when you're releasing and how long I can afford to bang on FF, you may need to keep running FF the old way. Trying to make sure I don't bite off more than I can chew is all. Getting Chrome headless automated is almost sure to be easier.

Have you seen [playwright](https://playwright.dev/)? It's like the next step for Puppeteer. In fact, the two projects share several team members. But it supports FF, Webkit, etc. Seems to require fewer lines of code than Puppeteer too. Seems like the next step.

Anyway, you might consider switching over to that from Puppeteer to ease automation. The interface is similar.

mattalexx:
What would be really slick is if it could be a replacement for Karma. What a pain Karma is. It seems like everything that anyone contributed to that project is at least five years old and broken.

Gusted:
> What would be really slick is if it could be a replacement for Karma. What a pain Karma/Jasmine is. It seems like everything that anyone contributed to that project is at least five years old and broken.

Yeah, I'm not sure about that. Karma more or less tests the code not as a extension but as it's own code(the code for the extension is loaded into the page, not as a extension) just to test certain edge cases etc. Puppeteer is really for the E2E testing.

I've heard about playwright, never tried it out as migrating the amounts of tests is _painful_ as both projects use different API's and have other code infrastructure to get used to.

PR #8124, opened 2022-02-03
Fix auto-watch when running `npm run test:inject:debug`

mattalexx:
When running `npm run test:inject:debug`, watching/reloading is done by the karma-rollup-preprocessor plugin. That plugin needed to be upgraded to its prerelease version at master before this was fixed. 

A screencast that shows the problem and the fix:

- [asciicast](https://asciinema.org/a/96heBpMmTJK2VUkZYROfSMc7q)

Side note, [I've left a note](https://github.com/jlmakes/karma-rollup-preprocessor/issues/72#issuecomment-1029080510) over at jlmakes/karma-rollup-preprocessor asking when the new version will be coming out, at which time you would be able to remove the URL from package.json.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,8 +23,11 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v2
         with:
-          node-version: 14.x
-      
+          node-version: 15.x
+
+      - name: Verify NPM version
+        run: npm --version
+
       - name: Set up npm cache
         uses: actions/cache@v2
         with:
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 # Build files
 #-----------------------------------
-build/
+build*/
 darkreader.js
 debug.log
 
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "karma-coverage": "2.1.0",
     "karma-firefox-launcher": "2.1.2",
     "karma-jasmine": "4.0.1",
-    "karma-rollup-preprocessor": "7.0.7",
+    "karma-rollup-preprocessor": "github:jlmakes/karma-rollup-preprocessor#master",
     "karma-safari-launcher": "1.0.0",
     "less": "4.1.2",
     "malevic": "0.18.6",
--- a/tests/inject/karma.conf.js
+++ b/tests/inject/karma.conf.js
@@ -5,6 +5,7 @@ const rollupPluginNodeResolve = require('@rollup/plugin-node-resolve').default;
 const rollupPluginReplace = require('@rollup/plugin-replace');
 const rollupPluginTypescript = require('@rollup/plugin-typescript');
 const typescript = require('typescript');
+const {getTestDestDir} = require("../../tasks/paths");
 
 module.exports = (config) => {
     config.set({
--- a/tests/inject/karma.conf.js
+++ b/tests/inject/karma.conf.js
@@ -5,7 +5,7 @@ const rollupPluginNodeResolve = require('@rollup/plugin-node-resolve').default;
 const rollupPluginReplace = require('@rollup/plugin-replace');
 const rollupPluginTypescript = require('@rollup/plugin-typescript');
 const typescript = require('typescript');
-const {getTestDestDir} = require("../../tasks/paths");
+const {getTestDestDir} = require('../../tasks/paths');
 
 module.exports = (config) => {
     config.set({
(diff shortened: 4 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -23,8 +23,11 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v2
         with:
-          node-version: 14.x
-      
+          node-version: 15.x
```
mattalexx, line 26: Ran into a bug cloning the repo, even though it's public. NPM 7 does away with that restriction.

Read [here](https://github.com/actions/setup-node/issues/214#issuecomment-1014695382) that NPM 7 fixes the bug, then read [here](https://github.com/actions/setup-node/issues/213#issuecomment-739685886) that upping node to 15 will get us that.

.gitignore
```diff
@@ -2,6 +2,9 @@
 #-----------------------------------
 build/
 darkreader.js
+tests/inject/**/*.tests.js
+tests/inject/customize.js
+tests/inject/polyfills.js
```
Gusted, line 7: I don't think they should be gitignored, they are not generated and are required for the tests.

.github/workflows/test.yml
Gusted: Oh yeah, you can upgrade it to node 16, the node version shouldn't really matter for the CI, as long it does it's job it's fine. We don't depend on specific node versions as a browser extension.

package.json
```diff
@@ -88,7 +88,7 @@
     "karma-coverage": "2.1.0",
     "karma-firefox-launcher": "2.1.2",
     "karma-jasmine": "4.0.1",
-    "karma-rollup-preprocessor": "7.0.7",
+    "karma-rollup-preprocessor": "github:jlmakes/karma-rollup-preprocessor#master",
```
Gusted, line 91: Let's wait what they say, maybe we can avoid using this and use the next update.

.gitignore
Gusted: I don't seem to be having that behavior, or is this caused by the upgraded package?

```
-> % node --version
v16.13.2
-> % npm --version
8.4.0
```

mattalexx: To be clear, you're referring to having JS files next to the TS? Where are yours placed when they are generated? Do they just remain in memory, then passed to Karma? That's what it seemed to be doing before.

.github/workflows/test.yml
mattalexx: Ok, if the rollup plugin upgrade request is ignored, I'll circle back.

mattalexx: On the other hand, 16 is the stable version. But I'm it would no longer be "in scope" (aka annoying me) :-).

.gitignore
mattalexx: I can't think how the node/npm versions would be affecting it, but who knows? 

```
% node --version; npm --version
v14.18.1
8.3.0
```

mattalexx: > I don't seem to be having that behavior,

Wait, are you talking about the inability to update when autowatch is active?

package.json
mattalexx: :+1:

.gitignore
Gusted: > Wait, are you talking about the inability to update when autowatch is active?

No, the generated files aren't being stored on disk... autowatch still seems to be working with this PR. I just don't have the behavior that the generated files are being stored on the disk.

tests/inject/karma.conf.js
```diff
@@ -41,6 +42,7 @@ module.exports = (config) => {
                 }),
             ],
             output: {
+                dir: `${getTestDestDir()}`,
```
mattalexx, line 45: This fixes the JS file problem for me. so there's:

```
- ./build/debug
- ./build-test/...
```

Gusted (approved):
LGTM, agreed to merge it?

mattalexx:
Everything's resolved except for this:

https://github.com/darkreader/darkreader/pull/8124#discussion_r798801355

You did mention you wanted to wait for jlmakes/karma-rollup-preprocessor to be updated so we wouldn't need to deal with the URL in the package.json.

mattalexx:
The push from last night was just to rebase onto master

Gusted:
> You did mention you wanted to wait for jlmakes/karma-rollup-preprocessor to be updated so we wouldn't need to deal with the URL in the package.json.

Well seems like they didn't update so it's fine then.

Gusted:
Thank you!

PR #8149, opened 2022-02-05
Revert "Affect img elements with filters (#6044)"

### LINKED ISSUE #8145: Firefox extension inverts Image in Dynamic mode since last update.
Hello, I am using DarkReader since a long time with no issue in firefox (windows) last update. But since the last DarkReader update, it now inverts image content of website too while it was not before. I cannot use it anymore on most trading sites because of it.

I tried uninstalling it and reinstalling, but issue still is there, and as it was with the last update, little chances are this bug is on my side.

Thank you !

Gusted:
This reverts commit 29a9a12b6f10260040b0a666cdd088ae90b36575.

This is causing by far more issues than I'd anticipated, removing for now and can be merged back in once I'd have a better plan to correctly use those.

- Resolves #8145

### DIFF
--- a/src/inject/dynamic-theme/index.ts
+++ b/src/inject/dynamic-theme/index.ts
@@ -9,7 +9,7 @@ import {removeNode, watchForNodePosition, iterateShadowHosts, isDOMReady, remove
 import {logInfo, logWarn} from '../../utils/log';
 import {throttle} from '../../utils/throttle';
 import {clamp} from '../../utils/math';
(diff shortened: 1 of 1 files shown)


SixelAlexiS90:
so how does it work, the extension will be updated with the fix automatically? Thanks :)

Gusted:
> so how does it work, the extension will be updated with the fix automatically? Thanks :)

Yep.

PR #8196, opened 2022-02-13
Fix for piazza.com

hyerrakalva:
Updated toolbars on the top and bottom of the page to use a darker gradient.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -11966,9 +11966,8 @@ img.math-display
 piazza.com
 
 CSS
-[data-darkreader-scheme="dark"] .dashboard_toolbar,
-[data-darkreader-scheme="dark"] .page_feed_bucket_header {
-    background: linear-gradient(to bottom, #33353b 0%, #1d1f20 100%);
+.dashboard_toolbar, .page_feed_bucket_header {
+    background: linear-gradient(to bottom, ${#c5c7cd} 0%, ${#dfe1e2} 100%) !important;
 }
 
 ================================
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -11963,6 +11963,16 @@ img.math-display
 
 ================================
 
+piazza.com
+
+CSS
+[data-darkreader-scheme="dark"] .dashboard_toolbar,
+[data-darkreader-scheme="dark"] .page_feed_bucket_header {
+    background: linear-gradient(to bottom, #33353b 0%, #1d1f20 100%);
+}
+
+================================
+
 picrew.me
 
 IGNORE INLINE STYLE


src/config/dynamic-theme-fixes.config
```diff
@@ -11963,6 +11963,16 @@ img.math-display
 
 ================================
 
+piazza.com
+
+CSS
+[data-darkreader-scheme="dark"] .dashboard_toolbar,
+[data-darkreader-scheme="dark"] .page_feed_bucket_header {
+    background: linear-gradient(to bottom, #33353b 0%, #1d1f20 100%);
```
Gusted, line 11971: Is it possible to use it in a form like `linear-gradient(to bottom, ${color here} 0%, ${color here} 100%);`? So it will play nice regardless of the dark reader scheme.

hyerrakalva: Is there supposed to be a difference between your suggestion vs using the `[data-darkreader-scheme="dark"]` selector? I use the attribute selector so that the website toolbars revert to their original gradient if the user selects light mode from the extension menu. When I do `${linear-gradient(to bottom, #33353b 0%, #1d1f20 100%)};` instead of using the attribute selector, the toolbar gradient remains black even when the light scheme is selected.

Gusted: So when you're using my suggestion, it will modify the colors on-the-fly with the factor of the user's custom settings. And it only works for the dark mode, so light mode will not be bothered by it.

> When I do `${linear-gradient(to bottom, #33353b 0%, #1d1f20 100%)};` instead of using the attribute selector, the toolbar gradient remains black even when the light scheme is selected.

The input of `${...}` only expects a color, and not the whole property, so something like: `linear-gradient(to bottom, ${#33353b} 0%, ${#1d1f20} 100%);`.

hyerrakalva: When I set the `background` to `linear-gradient(to bottom, ${#33353b} 0%, ${#1d1f20} 100%);`, it seems to change the toolbars to a different shade of white than the original website color. When I inspect the toolbar using Edge's dev tools, it shows me that `.dashboard_toolbar, .page_feed_bucket_header` elements have a `background` of `linear-gradient(to bottom, #c5c7cd 0%, #dfe1e2 100%)`, which is much different than what I had set.

However, when I switch to light mode from the extension menu, the toolbars turn dark and the Edge dev tools show the original gradient I had set in the Dark Reader dev tools.

Is the `${...}` syntax an official CSS standard or is it something specific to this extension?

hyerrakalva: Just pushed a commit which switches to the `${COLOR}` template.. I was able to get my desired colors by just inverting the hex values I had originally. I also added an `!important` keyword as requested by `CONTRIBUTING.md`.

Gusted (approved):
LGTM!

PR #8197, opened 2022-02-13
Fix for mitx

AnhadSawhney:
MITx has images with black diagrams on a transparent background. When the colors get inverted, it becomes black diagrams on a dark background, so they are not visible. This sets the background of the images to white.


src/config/dynamic-theme-fixes.config
```diff
@@ -10109,6 +10109,16 @@ INVERT
 
 ================================
 
+*.mitx.*
```
Gusted, line 10112:
```suggestion
mitx.*
```

src/config/dynamic-theme-fixes.config
```diff
@@ -10109,6 +10109,16 @@ INVERT
...
+
+CSS
+img {
+	border-color: currentcolor;
```
Gusted, line 10116: Is this needed?

src/config/dynamic-theme-fixes.config
```diff
@@ -10109,6 +10109,16 @@ INVERT
...
+	background: rgb(255, 255, 255) !important;
```
Gusted, line 10117:
```suggestion
	background-color: white !important;
```

src/config/dynamic-theme-fixes.config
AnhadSawhney: The site is https://lms.mitx.mit.edu

src/config/dynamic-theme-fixes.config
AnhadSawhney: I think this is the selector that works to find the images that are problematic

src/config/dynamic-theme-fixes.config
Gusted: Dark reader already by default match all subdomains.

src/config/dynamic-theme-fixes.config
Gusted: But why changing the border color?

AnhadSawhney: Ok seems it works without that. Sorry, I don't know anything about CSS, I just tried a couple of things until one worked.

AnhadSawhney:
Should be good to merge now? @Gusted

Gusted (approved):

PR #8236, opened 2022-02-19
www.google.* material bar

albertopasqualetto:
Not sure if this makes sense because it may be too bright for someone but that black bar was too unaesthetic for me.
Review it please and change whatever you want to make it darker.

Before:
image

After:
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -17529,9 +17529,6 @@ img[src^="https://www.google.com/maps"]
 
 CSS
 .RNNXgb,
-.aajZCb {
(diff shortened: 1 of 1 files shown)


albertopasqualetto:
Bad  commit

q1800:
I'm sorry, but I have to say this is a _**HARD NO**_ for me.

The less blinding white color, the better. I have enough trouble with my sight now that I believe this is an unnecessary change. Please leave it the way it is.

albertopasqualetto:
> I'm sorry, but I have to say this is a _**HARD NO**_ for me.
> 
> The less blinding white color, the better. I have enough trouble with my sight now that I believe this is an unnecessary change. Please leave it the way it is.

Ok, I understand, I am closing this. If you find a more aesthetic pattern please add it.

PR #8241, opened 2022-02-20
Replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.


src/inject/dynamic-theme/variables.ts
```diff
@@ -644,7 +644,7 @@ function parseRawValue(color: string) {
         splitted.forEach((number) => {
             resultInRGB += `${number.trim()}, `;
         });
-        resultInRGB = resultInRGB.substr(0, resultInRGB.length - 2);
+        resultInRGB = resultInRGB.slice(0, resultInRGB.length - 2);
```
alexanderby, line 647: Personally I prefer using `substring` method for Strings and `slice` for Arrays.

CommanderRoot: I like `slice()` more as it's [stricter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring#differences_between_substring_and_slice) but your repo, your rules. I will change it to `substring()`

PR #8242, opened 2022-02-21
Added platzi.com/ into dark-sites.config

AlbFR:


Gusted:
It seems like that many other of the pages aren't dark.

Issue #8246, opened 2022-02-21
[Broken Website] pipiwiki.com


paniash:
# Broken Website Report

## Website Address

https://pipiwiki.com/wiki/Purcell_effect

## Problematic Observation

The math equations are not visible at all.

## Screenshots

![pic-selected-220221-1820-33](https://user-images.githubusercontent.com/59497618/154958770-d7b886ca-187f-41e6-8f6f-72242a293fdf.png)

## System Information

- OS:  Linux
- Browser:  Firefox 97
- Dark Reader Version:  4.9.45

c-nagy:
I was able to reproduce the issue and submitted a fix here: https://github.com/darkreader/darkreader/pull/8249/files

PR #8249, opened 2022-02-22
Fix for pipiwiki.com

c-nagy:
- Invert equations
- Invert logo

Fix for issue #8246

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -12078,6 +12078,14 @@ body, img, video, iframe {
 
 ================================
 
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -12078,6 +12078,14 @@ body, img, video, iframe {
 
 ================================
 
+pipiwiki.com
+
+INVERT
+.mwe-math-fallback-image-inline
```
Gusted, line 12084: You might want to do:

```css
```

This is also how we fix these equations on wikipedia.org

Issue #8252, opened 2022-02-22
[Broken Website] fandom.com


Elaborendum:
# Broken Website Report

## Website Address

Various wikis under https://fandom.com

## Steps To Reproduce
Main Background and Logo issues:
- Go to, for example, marvel.fandom.com/dc.fandom.com/disney.fandom.com/monsterhunter.fandom.com
- See the background and logo are missing.
- See that 

Footer and Fan Feed issues:
- Go to any article in, for example, gijoe.fandom.com/godzilla.fandom.com/marvel.fandom.com/dc.fandom.com
- Scroll down to the bottom and see the footer and other articles section (for marvel.fandom.com and dc.fandom.com, the problem happens as you scroll down)

Certain wikis, like uncharted.fandom.com and zelda.fandom.com, are unaffected by both of the above issues

Unique case
- Go to spongebob.fandom.com
- See content

## Problematic Observation

- Main Background and Logo issues: The background image gets cut off and the logo with name of the Wiki disappears
  - In pokemon.fandom.com, it appears faded out
  - In gijoe.fandom.com, it disappears despite being simply text
  - In pokemon.fandom.com/es/, the navigation menu at the top loses contrast with the background
- Footer and Fan Feed issues: The main footer with other Fandom.com links is missing, except for the mobile app icons (and "Get the App" badges, which are whited out and unreadable). The Fan Feed that shows various other articles and discussions has no background, making the discussion titles difficult to read and erasing the "top articles" section. (for marvel.fandom.com/dc.fandom.com, the cutoff background covers this section when scrolling down)
- Unique case: Something I discovered when looking for examples of the above, the text of the main content is inverted, but not the background, resulting in the text being difficult to read.

## Screenshots

- Main Background and Logo issues:
With DR:
![imagen](https://user-images.githubusercontent.com/71190696/155045003-e9c2a5b2-59ce-4abc-aba0-23459389a72c.png)
Without DR:
![imagen](https://user-images.githubusercontent.com/71190696/155045190-eba10a8d-c1b4-4a82-be80-845839a873ed.png)

- Footer and Fan Feed issues:
With DR:
![imagen](https://user-images.githubusercontent.com/71190696/155045091-a76f6112-0e41-46ce-9be7-b156cf1cd1c1.png)
![imagen](https://user-images.githubusercontent.com/71190696/155045119-0c574324-4e1b-4fd3-abb9-fb9553e4e2bb.png)
Without DR:
[...]

PR #8257, opened 2022-02-22
Fix for fandom.com

c-nagy:
Resolved the first issue in #8252 regarding the hidden logo and navigation bars due to weird z-indexing.


c-nagy:
This fix breaks the pokemon fandom site for some reason so closing this for now.

PR #8259, opened 2022-02-22
Invert GitHub icon for bugs.python.org

xxyzz:
Invert GitHub icon inside the home page and issue page(login icon).

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2362,6 +2362,14 @@ th {
 
 ================================
 
+bugs.python.org
+
+INVERT
+img[title="has PR"]
+img[title="GitHub"]
+
+================================
+
 build-electronic-circuits.com
 
 INVERT


src/config/dynamic-theme-fixes.config
```diff
@@ -2362,6 +2362,14 @@ th {
 
 ================================
 
+bugs.python.org
+
+INVERT
+img[title="has PR"]
+img[title="GitHub"]
```
Gusted, line 2369: Should this rather be title="has patch"?

Gusted: Yes... that should also be inverted(barely visible). This title="Github" doesn't seem to match any icons.

xxyzz: title="Github" is at the left side of each [issue page](https://bugs.python.org/issue40255):
Screen Shot 2022-02-22 at 16 04 56

Invert the clip icon doesn't have any effect.

Gusted (approved):
LGTM

PR #8262, opened 2022-02-22
Added img inversion to dynamic theme

nitishg:


Gusted:
Hi @nitishg!

Thank you for this PR, unfortunately I cannot accept this PR as it will break every website that includes images, inverted images is not a desired behavior of dark reader's modifications.

Regards,
Gusted

nitishg:
Ah okay got it! Thanks. :)

PR #8263, opened 2022-02-23
Fix for distrowatch.com

c-nagy:
- Invert logo
- Darken bright table backgrounds

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4379,6 +4379,19 @@ INVERT
 
 ================================
 
+distrowatch.com
+
+INVERT
+img[src="images/cpxtu/dwbanner.png"]
+
+CSS
+.NavMenu, 
+td:not(.NewsDate, .NewsHeadline) {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -4379,6 +4379,19 @@ INVERT
...
 ================================
 
+distrowatch.com
+
+INVERT
+img[src="images/cpxtu/dwbanner.png"]
+
+CSS
+.NavMenu, 
+td:not(.NewsDate, .NewsHeadline) {
```
Gusted, line 4389: As per MDN:

```
Using two selectors at the same time is not yet supported in all browsers. Example: :not(.foo, .bar). For wider support you could use, :not(.foo):not(.bar)
```
So let's not hop into fancy features, and revert to a safe default:

```suggestion
td:not(.NewsDate):not(.NewsHeadline) {
```

c-nagy: Ah I didn’t know that, thanks for the heads up.

Gusted (approved):

PR #8268, opened 2022-02-23
add some dark by default sites

sauce420:


dtbj:
@sauce420 Please add 9to5google.com also

sauce420:
@dtbj looks like it uses prefers-color-scheme which means we can't add it to the list: https://github.com/darkreader/darkreader/blob/master/CONTRIBUTING.md

dtbj:
@sauce420 ohk thanks for the info.

Gusted (approved):

PR #8280, opened 2022-02-25
dark-site: add guix.gnu.org

linsui:


Gusted:
Only the homepage is dark-by-default, pages like blog, download, help and about aren't.

PR #8281, opened 2022-02-25
Fix for cnn.com

c-nagy:
- Invert politics section icon as CNN has switched it to a dark style.


c-nagy:
Getting inconsistent results, closing until further testing.

PR #8304, opened 2022-03-01

Elaborendum:
- Crowdin: Logo at the top left when viewing projects
- Fandom: Didn't know if it was right to add these. They are ad sections, but their current absence could also confuse a user who isn't using an ad blocker.
- TVT: Backgrounds of Darth and WMG wikis + subpages icons + loading icon when searching
- Medium: Separator
- Weblate: Icon that indicates a string has a suggestion
- Tumblr: Logos and other elements in the main page, made white because the randomly selected background sometimes has low contrast with those unless they are white + Login with Google/Apple buttons + Video Progress Bar

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -18045,7 +18045,7 @@ a[aria-label="Tumblr"] > svg {
 nav.FhRnI > ul > li {
     filter: invert(100%) hue-rotate(180deg) contrast(200%) !important;
 }
-footer[role="contentinfo"] {
+footer[role="contentinfo"] > ul > li > a {
     filter: invert(0%) hue-rotate(180deg) contrast(200%) !important;
 }
 ul.AohpR > li > a {
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -18029,23 +18029,27 @@ INVERT
 div[role="progressbar"]
 
 CSS
-a[data-testid="google-login-button"] {
-    background-color: #0d0d0d !important;
-}
+a[data-testid="google-login-button"],
 a[data-testid="apple-login-button"] {
-    background-color: #0d0d0d !important;
+    background-color: var(--darkreader-neutral-background) !important;
 }
 a[href="/"] > div > svg {
-    filter: invert(100%) hue-rotate(180deg) contrast(200%) !important;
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -17995,7 +18024,29 @@ a[title="Tinkoff"]
...
+.QnWzN > div
+div[role="progressbar"]
+
 CSS
+a[data-testid="google-login-button"] {
+    background-color: #0d0d0d !important;
+}
+a[data-testid="apple-login-button"] {
+    background-color: #0d0d0d !important;
+}
```
Gusted, line 18037:
```suggestion
a[data-testid="google-login-button"],
a[data-testid="apple-login-button"] {
    background-color: var(--darkreader-neutral-background) !important;
}
```

src/config/dynamic-theme-fixes.config
```diff
@@ -17995,7 +18024,29 @@ a[title="Tinkoff"]
...
+a[href="/explore"] > svg {
+    filter: invert(0%) hue-rotate(180deg) contrast(200%) !important;
+}
+a[aria-label="Tumblr"] > svg {
+    filter: invert(100%) hue-rotate(180deg) contrast(200%) !important;
+}
+nav.FhRnI > ul > li {
+    filter: invert(100%) hue-rotate(180deg) contrast(200%) !important;
+}
```
Gusted, line 18049: Is it possible to `INVERT` these elements?

Elaborendum: The problem is that `INVERT` turns them light grey, but the randomly selected background is chosen/slightly altered with pure white in mind, so on occasion the objects may get lost in the background unless they are white.

For example:
imagen

They are still there, they are just (almost) the same color as the background.

imagen

Although some of them simply work with `fill`, so I've changed those in the new commit. Plus two others I noticed while doing this.

Gusted (approved):
Thanks!

src/config/dynamic-theme-fixes.config
```diff
@@ -16965,6 +16987,13 @@ CSS
 
 ================================
 
+weblate.org
+
+INVERT
+h5.list-group-item-heading > svg
```
Gusted, line 16993:
```suggestion
h5.list-group-item-heading > svg
```

PR #8316, opened 2022-03-02
Add developer.mozilla.org to dark sites list

mattalexx:
New design for MDN: https://hacks.mozilla.org/2022/03/a-new-year-a-new-mdn/

https://developer.mozilla.org/

Also, I removed a stale link that I happened to check out.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -225,7 +225,6 @@ deadbydaylight.com
 deadbydaylight.fandom.com
 deepweblinks.net
 defcon.org
-del.dog
 deltarune.com
 demonforums.net
 denims.tv
@@ -235,6 +234,7 @@ destiny.gg
 destinytracker.com
 devanbuggay.com
 devart.withgoogle.com
+developer.mozilla.org
 developerinsider.co
 developpement-systeme-exploitation.github.io/documentation
 di.fm


src/config/dark-sites.config
```diff
@@ -235,6 +234,7 @@ destiny.gg
 destinytracker.com
 devanbuggay.com
 devart.withgoogle.com
+developer.mozilla.org
```
Gusted, line 237: MDN cannot be added they don't offer their dark mode by-default:
image

mattalexx: Ah. Makes sense.

Gitoffthelawn: Hmmm... I get the dark mode by default if I have the user agent set to use a dark theme.  The little switch in the upper-right corner is flipped the other way for me.

Gusted: Then it detects the system's color scheme, but isnt qualified to be listed here.

Gitoffthelawn: Ah, so do we only list sites that are *always* dark and not ones that auto-detect?

It will be nice to not have DR overriding dark themed sites that are dark because they auto-detected a dark themed user agent.

Gusted: I'm against adding more lists, those lists have become a nightmare and PR's like: https://github.com/darkreader/darkreader/pull/7995 should be focused on instead.

Gitoffthelawn: I agree completely.  Love that idea.

PR #8321, opened 2022-03-02
Fix for azuresynapse.net cursor color

DaveRuijter:


src/config/dynamic-theme-fixes.config
```diff
@@ -1535,6 +1535,13 @@ CSS
 
 ================================
 
+azuresynapse.net
```
Gusted, line 1538: Is this a internal website?

DaveRuijter: No, it is the portal of the suite for the Microsoft Azure Synapse Analytics cloud offering.

Gusted: Is it only reachable by a subdomain?

DaveRuijter: I think it is yes. Full url is: https://web.azuresynapse.net/

Gusted (approved):

PR #8322, opened 2022-03-02
Add a Immediate modify option

### LINKED ISSUE #7682: [Bug] Styles not applied when tab loaded in the background
# Bug Report

## Bug Description
Sites are not styled correctly when loaded in the background

## Website Address
Various sites. Example: https://en.wikipedia.org/wiki/Darkness

## Steps To Reproduce
Load page in the background.

## Expected Behavior
image

## Actual Behavior
I loaded the page and took the above screen shot.
I ctrl+tabbed to this tab to paste the screen shot
I right-click the above tab and select reload, while I am still in this tab
ctrl+tab back to the page and this is what I see:
image

## Screenshots
See above

## System Information

- OS: Win10
- Browser: Firefox Nightly 97
- Dark Reader Version:  4.9.43

## Additional Context
This started happening recently,

Gusted:
- Add a per-site option to enable "Immediate modify", within this mode(only useable for dynamic theme), Dark Reader will modify the page without waiting for it to be shown to the user. With this mode you can open a new tab in the background and switch to it and notice it's already been modified by Dark Reader.
- Because users(myself included to a certain extend) might find this behavior unexpected and prefer to let Dark Reader wait until the tab is shown, this option is disabled by default. And because it has chances it will break certain sites, it's implemented to be a per-site toggle(Under `font & more`).
- Resolves #7682.

### DIFF
--- a/src/defaults.ts
+++ b/src/defaults.ts
@@ -34,6 +34,7 @@ export const DEFAULT_THEME: Theme = {
     styleSystemControls: !isCSSColorSchemePropSupported,
     lightColorScheme: 'Default',
     darkColorScheme: 'Default',
+    immediateModify: false,
 };
 
 export const DEFAULT_COLORSCHEME: ParsedColorSchemeConfig = {
--- a/src/definitions.d.ts
+++ b/src/definitions.d.ts
@@ -64,6 +64,7 @@ export interface Theme {
     styleSystemControls: boolean;
     lightColorScheme: string;
     darkColorScheme: string;
+    immediateModify: boolean;
 }
 
 export type FilterConfig = Theme;
--- a/src/inject/dynamic-theme/index.ts
+++ b/src/inject/dynamic-theme/index.ts
@@ -5,7 +5,7 @@ import type {StyleElement, StyleManager} from './style-manager';
 import {manageStyle, getManageableStyles, cleanLoadingLinks} from './style-manager';
 import {watchForStyleChanges, stopWatchingForStyleChanges} from './watch';
 import {forEach, push, toArray} from '../../utils/array';
-import {removeNode, watchForNodePosition, iterateShadowHosts, isDOMReady, removeDOMReadyListener, cleanReadyStateCompleteListeners, addDOMReadyListener} from '../utils/dom';
+import {removeNode, watchForNodePosition, iterateShadowHosts, isDOMReady, removeDOMReadyListener, cleanReadyStateCompleteListeners, addDOMReadyListener, setIsDOMReady} from '../utils/dom';
 import {logInfo, logWarn} from '../../utils/log';
 import {throttle} from '../../utils/throttle';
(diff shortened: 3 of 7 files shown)


xcasxcursex:
Thanks that's the idea!

I was able to force a failure mode though, where DR would continue before the DOM was ready. It would update immediately afterwards, but I could get it to create a 'white flash' when loading a tab in the foreground (I think maybe it applied the fallback style too soon?). I had to really spam ctrl+r on a very heavy page to make it happen, but I did that because I was concerned it might be possible. I fear this may be worse on slower machines or specific sites. 

In my experiments, I bypassed the check for `!document.hidden` (just made ` if (true) {... ` ) and then fired the visibilitychange event from a userscript, but I left the DOMReady stuff untouched. This seemed to work out alright but I am very far from an expert on this.

I tried to replicate that approach using this PR, [I just commented out the lines that redefine the function (429 - 435) ](https://github.com/darkreader/darkreader/pull/8322/files#diff-3bff23553e95457e59dc4879e23780a6ef7192429d3a63d2873ea3b106c5b6d0) and it seems to be working well. Obviously the change to line 8, and dom.ts, are then redundant. But do you think this would be enough?

Either way, it is a really slick experience. And there's that privacy bonus, so that's nice :) Thanks for being so helpful with it all!

Gusted:
> I was able to force a failure mode though, where DR would continue before the DOM was ready. It would update immediately afterwards, but I could get it to create a 'white flash' when loading a tab in the foreground (I think maybe it applied the fallback style too soon?). I had to really spam ctrl+r on a very heavy page to make it happen, but I did that because I was concerned it might be possible. I fear this may be worse on slower machines or specific sites.

It could be that the browser was "defering" the extension in order to not waste a lot of CPU/Memory at the page startup.



> In my experiments, I bypassed the check for `!document.hidden` (just made `if (true) {...` ) and then fired the visibilitychange event from a userscript, but I left the DOMReady stuff untouched. This seemed to work out alright but I am very far from an expert on this.

Hmm `document.hidden` shouldn't care anymore when `filter.immediateModify = true`, Even if you forcibly have it the whole time set to true or false, Dark Reader should still work when the option is enabled.


> I tried to replicate that approach using this PR, [I just commented out the lines that redefine the function (429 - 435) ](https://github.com/darkreader/darkreader/pull/8322/files#diff-3bff23553e95457e59dc4879e23780a6ef7192429d3a63d2873ea3b106c5b6d0) and it seems to be working well. Obviously the change to line 8, and dom.ts, are then redundant. But do you think this would be enough?

I don't think so `isDOMReady` is uses on some other places and could be a blocker when it checks for the `readyState`.

xcasxcursex:
OK that all makes sense. I am using this patch full-time in my main profile now and I have had no problems at all.

BTW, your build instructions are very simple and the build worked perfectly. Really nice.

PR #8328, opened 2022-03-03
Fix distrowatch.com (dynamic dark mode)

Gitoffthelawn:


src/config/dynamic-theme-fixes.config
```diff
@@ -4393,11 +4393,11 @@ distrowatch.com
...
 INVERT
 img[src="images/cpxtu/dwbanner.png"]
-.NewsLogo > a:first-of-type
-.TablesTitle > img
 
 CSS
-.NavMenu, 
+input,
+.NavMenu,
+select,
```
Gusted, line 4400: Don't think the input/select's are needed. The modification done by dark reader seems good already.

Gitoffthelawn: I added them because, at least for my eyes, the contrast between the foreground text and background was minimal enough that it makes it hard to discern the text.  Also, it's was hard to notice that the dropdowns were even present as they blend with the surroundings.

Gusted (approved):

PR #8329, opened 2022-03-03
Fix eff.org (dynamic dark mode)

Gitoffthelawn:
Tested in both dynamic dark and dynamic light modes.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5169,6 +5169,9 @@ select > optgroup,
 select > optgroup > option {
     background-color: var(--darkreader-neutral-background) !important;
 }
+#masthead .branding a {
+    filter: invert(0.35) saturate(2.6) !important;
+}
 #masthead #main-menu-nav .pane-main-nav-menu > ul > li a::before {
     background-color: var(--darkreader-neutral-text) !important;
 }


src/config/dynamic-theme-fixes.config
```diff
@@ -5169,6 +5169,9 @@ select > optgroup,
 select > optgroup > option {
     background-color: var(--darkreader-neutral-background) !important;
 }
+#masthead .branding a {
```
Gusted, line 5172: Why not `INVERT` the gray text doesn't really make it any better IMO.

Gitoffthelawn: I started by trying to use `INVERT`.  When I  `INVERT`ed it, the "Electronic Frontier Foundation" words between the large "EFF" letters was not visible in dynamic dark mode.

Gusted: image

Gitoffthelawn: Let me take another look...

Gitoffthelawn: Ah, now I'm remembering.  I chose the other implementation for 2 reasons:  
1) The white text becomes blinding white.  It gets rendered as `#f2f2f2`, which is *way* too bright for a dark theme.  (I think I previously recommended changing `INVERT` to lower the max brightness (such as what `filter: invert(0.85)` will accomplish).
2) The red "EFF" letters become a considerably different colour as compared to their well-recognized logo colours.  As such, I figured you would deny the pull request if I used `INVERT`.

Gusted: > 2\. The red "EFF" letters become a considerably different colour as compared to their well-recognized logo colours.  As such, I figured you would deny the pull request if I used `INVERT`.

That's a common side-effect which is accepted.



> 1. The white text becomes blinding white.  It gets rendered as `#f2f2f2`, which is _way_ too bright for a dark theme.  (I think I previously recommended changing `INVERT` to lower the max brightness (such as what `filter: invert(0.85)` will accomplish).

We already apply contrast(90%) by default.

Gitoffthelawn: > That's a common side-effect which is accepted.

It's not a big deal, but submitting PRs for Dark Reader is turning into a bit of an unfulfilling time-waster for me.

I implement the changes locally using other tools, so I only submit PRs as an act of kindness to benefit the project and our community.  I receive no benefit myself.  It's rare that anyone even says "thank you".

I think I have 1 pull request that hasn't been merged because you said you "weren't amused" because it changed a section of the site's colours (a sidebar, IIRC) to make the text legible.  IIRC, you mentioned that it deviated from the author's intent (or something akin to that).   Based on those comments, I naturally thought only PRs in which no colours changed from the site author's intent would be accepted.

Since changing the colour of the EFF logo obviously doesn't match the EFF's intent, I worked hard to create an invert filter that would come close to matching their official colours and still would work in both dark and light dynamic modes (not an easy task).  I spent about 15 minutes of my free time on it to get it looking good.

> > 1. The white text becomes blinding white.  It gets rendered as `#f2f2f2`, which is _way_ too bright for a dark theme.  (I think I previously recommended changing `INVERT` to lower the max brightness (such as what `filter: invert(0.85)` will accomplish).
> 
> We already apply contrast(90%) by default.

The result is `#f2f2f2`, so it's much too bright for a dark mode (for humans, `f2f2f2` is almost indistinguishable from full bright white `#fff`.).

Except for highly emphasized items, I try to keep primary elements around `#bebebe` with peaks not exceeding `#dcdcdc`.  Based on the neurophysiology of the human optic system combined with the neuropsychology of human perception, I adjust peaks based on hue, as humans perceive different levels of brightness for different colour hues.

When one looks at the mathematical formulae used for `filter`, you'll see that `contrast(0.9)` isn't usually what works best.  Often, an effective solution is to `invert` at less than `1`.  Lowering the contrast isn't all bad, but it compresses the colour spectrum.

Another technique I find useful is to increase brightness and then lower it.  This acts like a low-pass filter and reduces the high-end of the amplitude of the visible light frequencies reaching the receptors within the retina.  For example, I commonly include something like `filter: brightness(1.25) brightness(0.7)` to make images look good (I have dozens more fixes for Dark Reader for which I haven't created pull requests because of the time commitment to do so).  Generally, I'll increase saturation between the each brightness adjustment to maintain perceived colour appearance.

Gusted: > It's not a big deal, but submitting PRs for Dark Reader is turning into a bit of an unfulfilling time-waster for me.
> 
> I implement the changes locally using other tools, so I only submit PRs as an act of kindness to benefit the project and our community. I receive no benefit myself. It's rare that anyone even says "thank you".

If you want me to say on every PR "thank you" I can do that, but I don't feel that there is any meaning behind that. If I do that unconditionally on every Incoming PR it wouldn't have any kind of truth.




> I think I have 1 pull request that hasn't been merged because you said you "weren't amused" because it changed a section of the site's colours (a sidebar, IIRC) to make the text legible. IIRC, you mentioned that it deviated from the author's intent (or something akin to that). Based on those comments, I naturally thought only PRs in which no colours changed from the site author's intent would be accepted.

Yes, while these fixes changes the modification of dark reader as dark reader can't cover any use-case, the dynamic mode was built to hold true to the original's color scheme. If a certain page has a overall red color-theme, dark reader will generate darker variant of the red colors, if the fixes suddenly will make it black-white-ish color scheme then it's just better to force a static color-scheme over it. This probably should be mentioned in contributing.md



> Since changing the colour of the EFF logo obviously doesn't match the EFF's intent, I worked hard to create a custom invert filter that would come close to matching their official colours and still would work in both dark and light dynamic modes (not an easy task). I spent about 15 minutes of my free time on it to get it looking good.

^^ There are exceptions, logo's and icons are free from that because they use `filter` and how best you can try to make the filter it will never match anything like the original colors, it has now been generally accepted to that, this also should be mentioned.




> The result is `#f2f2f2`, so it's much too bright for a dark mode (for humans, `f2f2f2` is almost indistinguishable from full bright white `#fff`.).
> 
> Except for highly emphasized items, I try to keep primary elements around `#bebebe` with peaks not exceeding `#dcdcdc`. Based on the neurophysiology of the human optic system combined with the neuropsychology of human perception, I adjust peaks based on hue, as humans perceive different levels of brightness for different colour hues.
> 
> When one looks at the mathematical formulae used for `filter`, you'll see that `contrast(0.9)` isn't usually what works best. Often, an effective solution is to `invert` at less than `1`. Lowering the contrast isn't all bad, but it compresses the colour spectrum.
> 
> Another technique I find useful is to increase brightness and then lower it. This acts like a low-pass filter and reduces the high-end of the amplitude of the visible light frequencies reaching the receptors within the retina. For example, I commonly include something like `filter: brightness(1.25) brightness(0.7)` to make images look good (I have dozens more fixes for Dark Reader for which I haven't created pull requests because of the time commitment to do so). Generally, I'll increase saturation between each brightness adjustment to maintain perceived colour appearance.

This should be a separate issue - but there are many cases like this(see fixes at store.google.com) whereby `INVERT` just lacks. The problem is that once something is accepted that anyone can go running around and also use such filter. We also have user-defined filters as we don't want to make the logo, icon to stand out from the overall color-scheme. It's sometimes a slippery slope.

Thank you for the many contributions.

Gitoffthelawn: > ... a bunch of important stuff ...
> Thank you for the many contributions.

Thank you for the thanks.  I don't have sufficient time to write back in any depth right now, but I just wanted to take a quick moment to thank you for your response and let you know I received it.

PR #8340, opened 2022-03-04
Fix for Wikipedia map

xxyzz:
Example page: https://en.wikipedia.org/wiki/Zaporizhzhia_Nuclear_Power_Plant
Screen Shot 2022-03-04 at 13 56 44

`oo-ui-icon-close` was added from #8284, that element also has `oo-ui-iconElement-icon` class, so I deleted it.

Map zoom button in full screen mode:
Screen Shot 2022-03-04 at 14 00 38

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -17355,9 +17355,8 @@ body > .oo-ui-windowManager .vega .marks
 .minerva-footer-logo img
 .music-symbol
 .tool.tool-button[src$="background-image:"]
-.wikiEditor-ui .oo-ui-iconElement-icon
+.oo-ui-iconElement-icon
 #p-personal .mw-echo-notifications-badge
-.oo-ui-icon-close
 .mw-ui-icon::before
 
 CSS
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -17355,10 +17355,11 @@ body > .oo-ui-windowManager .vega .marks
 .minerva-footer-logo img
 .music-symbol
 .tool.tool-button[src$="background-image:"]
-.wikiEditor-ui .oo-ui-iconElement-icon
+.oo-ui-iconElement-icon
```
Gusted, line 17358: Does this cover the general `.oo-ui-icon-close` element?

xxyzz: The only case I've seen so far is the close video button in MinervaNeue skin and it has `oo-ui-iconElement-icon` class.

Gusted: Lgtm then

Gusted (approved):

PR #8341, opened 2022-03-04
fix: make `cssImportRegex` case insensitive

felipewil:
- Seems like some links return `@IMPORT ...` with capital import. The original regex `cssImportRegex` does not match them;
- It can be observed in the following page: https://m.post.naver.com/viewer/postView.naver?volumeNo=33377956&memberNo=31740578

Before:
Screen Shot 2022-03-04 at 09 58 29

After:
Screen Shot 2022-03-04 at 10 02 36


Gusted (approved):

Gusted:
Thank you for your contribution.

PR #8210, opened 2022-02-14
Adding the eslint-plugin-compat plugin revealed some violations with …

mattalexx:
…the browser support list as listed in `package.json`.

I was reviewing the TS config, the browser support list, etc to see what's allowed and what's not. And I can across these errors:

```
% npm run code-style

> darkreader@4.9.45 code-style
> eslint --ignore-pattern "!.eslintplugin.js" --cache --fix "src/**/*.ts" "src/**/*.tsx" "tasks/**/*.js" "tests/[!coverage]**/*.js" "tests/**/*.ts" ".eslintrc.js" "index.d.ts" ".eslintplugin.js"


/home/matt/Projects/darkreader/src/api/fetch.ts
2:12  error  Promise.reject() is not supported in bb 7  compat/compat

/home/matt/Projects/darkreader/src/background/config-manager.ts
148:15  error  Promise.all() is not supported in bb 7  compat/compat

/home/matt/Projects/darkreader/src/background/devtools.ts
45:16  error  Promise is not supported in bb 7  compat/compat
93:16  error  Promise is not supported in bb 7  compat/compat

/home/matt/Projects/darkreader/src/background/extension.ts
398:16  error  Promise is not supported in bb 7  compat/compat

/home/matt/Projects/darkreader/src/background/newsmaker.ts
74:36  error  fetch is not supported in Firefox 33, bb 7  compat/compat

/home/matt/Projects/darkreader/src/background/tab-manager.ts
11:12  error  Promise is not supported in bb 7  compat/compat
212:30  error  URL is not supported in bb 7      compat/compat

/home/matt/Projects/darkreader/src/background/utils/extension-api.ts
41:12  error  Promise is not supported in bb 7  compat/compat
85:12  error  Promise is not supported in bb 7  compat/compat
121:12  error  Promise is not supported in bb 7  compat/compat
134:12  error  Promise is not supported in bb 7  compat/compat
142:12  error  Promise is not supported in bb 7  compat/compat

/home/matt/Projects/darkreader/src/background/utils/network.ts
12:12  error  Promise is not supported in bb 7  compat/compat

/home/matt/Projects/darkreader/src/background/utils/state-manager.ts
144:24  error  Promise is not supported in bb 7  compat/compat

/home/matt/Projects/darkreader/src/generators/utils/parse.ts
62:17  error  URL is not supported in bb 7  compat/compat

/home/matt/Projects/darkreader/src/inject/dynamic-theme/css-collection.ts
[...]

PR #8348, opened 2022-03-04
Fix tasks/build.js ignoring Ctrl+C

mattalexx:


package.json
```diff
@@ -83,6 +84,7 @@
     "eslint": "8.10.0",
     "eslint-plugin-import": "2.25.4",
     "eslint-plugin-local": "1.0.0",
+    "get-stream": "^6.0.1",
```
Gusted, line 87: Exact version

mattalexx: Darn. I keep missing those, I need to write myself a git commit hook or something.

tasks/build.js
```diff
@@ -55,9 +56,28 @@ async function api() {
...
+async function executeChildProcess(args) {
+    if (process.env.BUILD_CHILD) {
+        throw new Error('Infinite loop');
+    }
+    const env = {...process.env, BUILD_CHILD: '1'};
+    const child = fork(__filename, args, {env});
+    // Send SIGINTs as SIGKILLs, which are not ignored
+    process.on('SIGINT', () => {
+        child.kill('SIGKILL');
+        process.exit(130);
```
Gusted, line 68: Why 130?

Gusted, line 68: We might need to clean it up, to not leave a partial-mess behind?

tests/project/build.tests.ts
```diff
@@ -1,27 +1,52 @@
...
+});
+
+describe('tasks/build.js', () => {
+    // Slow test (skipped by default)
+    it.skip('should build successfully from an unexpected working directory', async () => {
+        const tmpDir: string = await mkdtemp(join(tmpdir()));
 
         // Fire
-        const {code, stdout} = await forka(buildModule, ['--debug', '--release', '--api'], {cwd});
+        const child = fork(buildModule, ['--debug', '--release', '--api'], {cwd: tmpDir, silent: true});
```
Gusted, line 46: Wait... a debug release api?!?

tests/project/tsconf.tests.ts
```diff
@@ -26,27 +26,25 @@ describe('TypeScript project config', () => {
...
     it.skip.each(tsProjects)('should compile without errors: $tsconfig', async ({tsconfig}) => {
         // Compile config from temp dir instead of root dir
         const cwd = await mkdtemp(join(tmpdir()));
         const project = join(rootDir, tsconfig, 'tsconfig.json');
-        const {code, stdout} = await tsc(['--project', project, '--noEmit'], {cwd});
-        expect(stdout).toBeUndefined();
-        expect(code).toBe(ExitStatus.Success);
+
+        // Fire
+        const child = fork(tsc, ['--project', project, '--noEmit'], {silent: true, cwd});
```
Gusted, line 46: `--noEmit` should already be included in tsconfig.json now.

tasks/build.js
mattalexx: Well, for Linux, a fatal error is 128, and they add the signal code to that, which in this case is 2. At least I think that's how it goes.

In this case, it just needs to be non-zero. I can do that if you want

tests/project/build.tests.ts
mattalexx: Originally I thought it would conflict, especially the initial clean step perhaps targeting the same directory. But they only attend to the directories they write. Very nice :-)

tests/project/tsconf.tests.ts
mattalexx: I was thinking about this.. There are a few of those "compiler options" that should really be under a heading more along the lines of "output options". They follow different rules. Options like `noEmit` and `declarationOnly` and `outDir`, they may change depending on how you want to compile the code (say, rollup vs dev REPL, etc). 

So I thought it would be alright to "pin" this option. If for some reason you decided to ditch rollup and use tsc to handle the job, you would need to set noEmit to false, but I would still want it set this way here.

tasks/build.js
mattalexx: I don't have a test for that, but I did try that and made sure that no processes were left alive. In fact, that used to be a problem, but I fixed it. the main process died and didn't kill the child.

package.json
```diff
@@ -83,6 +84,7 @@
     "eslint": "8.10.0",
     "eslint-plugin-import": "2.25.4",
     "eslint-plugin-local": "1.0.0",
+    "get-stream": "6.0.1",
```
mattalexx, line 87: FYI, this was already in node_modules, a dependency of web-ext and ts-jest, and a few others.

package.json
```diff
@@ -24,7 +24,8 @@
     "test:inject:debug": "node tests/inject/run.js --debug",
     "test:project": "jest --config=tests/project/jest.config.js",
     "test:unit": "jest --config=tests/unit/jest.config.js",
-    "test:unit:debug": "node --inspect-brk ./node_modules/jest/bin/jest --config=tests/unit/jest.config.js --runInBand --no-cache --watch"
+    "test:unit:debug": "node --inspect-brk ./node_modules/jest/bin/jest --config=tests/unit/jest.config.js --runInBand --no-cache --watch",
+    "test:update-snapshots": "npm run test -- --updateSnapshot && npm run test:project -- --updateSnapshot"
```
mattalexx, line 28: Here's a script you can use to update the snapshots.

tests/project/utils.ts
```diff
@@ -1,25 +1,54 @@
...
-        proc.on('close', (code, signal) => {
-            resolve({code, signal, stdout: stdout || undefined, stderr: stderr || undefined});
+/**
+ * Returns a promise that resolves when a certain condition has been met, based on the content of a stream.
+ *
+ * @example
+ * await watchStream(child.stdout).forMatch(/Config loaded/)
+ * thisRunsAfterConfigLoads();
+ */
+export const watchStream = (readable: Readable, options?: { encoding: BufferEncoding }) => {
```
Gusted, line 39: Not sure why TS isn't erroring about this:
```suggestion
export const watchStream = (readable: Readable, options?: {encoding: BufferEncoding}) => {
```

tests/project/utils.ts
```diff
@@ -1,25 +1,54 @@
...
+    if (!readable.readableEncoding) {
+        readable.setEncoding(options?.encoding || 'utf-8');
+    }
+    const forCondition = async (cb: { (value: string): boolean }) => await new Promise<void>((resolve, reject) => {
```
Gusted, line 43:
```suggestion
    const forCondition = async (cb: {(value: string): boolean}) => await new Promise<void>((resolve, reject) => {
```

Also `async` `await` and twice `=>` on the same line doesn't really look good.

tests/project/utils.ts
```diff
@@ -1,25 +1,54 @@
...
+        readable.on('close', reject).on('data', (chunk: string) => {
+            if (chunk && cb(chunk)) {
+                resolve();
+            }
         });
     });
-}
+
+    const forMatch = async (regExp: RegExp) => await forCondition((chunk: string) => chunk.match(regExp) != null);
```
Gusted, line 51: Ditto

tests/project/utils.ts
mattalexx: I'll check eslint.

mattalexx: Looks like two things are happening: 

1. The rule (https://eslint.org/docs/rules/object-curly-spacing) doesn't apply to this type of bracket. I put this into a file under `src/`:

Screenshot_2022-03-08_05-32

2. The typescript test files aren't being included in the ESlint scan unless their `tsconfig.json` is explicitly referenced in `eslintrc.js` under `*.parserOptions.project`. I'm really not sure about how to put them all in there (like if there's a way to do a wildcard or something). For now, I'll add `tests/project/tsconfig.json` here. I'll add the rest of the projects to https://github.com/darkreader/darkreader/pull/8210.

mattalexx: #2 was a false alarm.. Only #1 is an issue, from what I'm discovering.

Gusted (approved):

PR #8365, opened 2022-03-05
Fix for patreon.com

marblenix:
- invert google docs and forms preview
- no image analysis

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -12150,6 +12150,17 @@ pre {
 
 ================================
 
+patreon.com
+
+INVERT
+a[href^="https://docs.google.com"] img
+a[href^="https://forms.gle"] img
+
+IGNORE IMAGE ANALYSIS
+*
+
+================================
+
 paulgraham.com
 
 CSS


src/config/dynamic-theme-fixes.config
```diff
@@ -12150,6 +12150,22 @@ pre {
...
+a[href^="https://docs.google.com"] img
+a[href^="https://forms.gle"] img
+
+CSS
+div[data-tag="media-container"] > img {
+    filter: brightness(75%);
+}
+
+IGNORE IMAGE ANALYSIS
+*
```
Gusted, line 12165: Why should this be disabled?

src/config/dynamic-theme-fixes.config
```diff
@@ -12150,6 +12150,22 @@ pre {
...
+
+INVERT
+a[href^="https://docs.google.com"] img
+a[href^="https://forms.gle"] img
+
+CSS
+div[data-tag="media-container"] > img {
+    filter: brightness(75%);
+}
+
```
Gusted, line 12163: Does this target all content(e.g. user's uploaded post)?

src/config/dynamic-theme-fixes.config
marblenix: Mostly because the avatars are inverted, which causes unsightly, bright white spots on the page.

src/config/dynamic-theme-fixes.config
marblenix: It does, and that's intentional. A lot of YouTube creators will uploaded incredibly bright thumbnails for their videos (since bright == eye catching).

src/config/dynamic-theme-fixes.config
marblenix: I should mention I tried a lot of different css selector ways to target just the avatar images, but patreon randomizes the css classes and doesn't use any id tags.

src/config/dynamic-theme-fixes.config
Gusted: Yeah, I cannot allow such fix, we have user settings to deal with that. This is the same of lowering the brightness of youtube videos.

src/config/dynamic-theme-fixes.config
Gusted: If nothing otherwise breaks then LGTM for this case.

src/config/dynamic-theme-fixes.config
marblenix: Okay, I'll remove these two lines.

Gusted (approved):

PR #8372, opened 2022-03-05
Toggle detected dark theme

alexanderby:


src/utils/url.ts
```diff
@@ -221,14 +221,16 @@ export function isURLEnabled(url: string, userSettings: UserSettings, {isProtect
...
     const isURLInEnabledList = isURLInList(url, userSettings.siteListEnabled);
 
-    if (userSettings.applyToListedOnly && !isURLInEnabledList) {
-        return isURLInUserList;
+    if (userSettings.applyToListedOnly) {
+        return isURLInEnabledList || isURLInUserList;
     }
-
-    if (isURLInEnabledList && isInDarkList) {
+    if (isURLInEnabledList) {
```
Gusted, line 227: It seems like `isURLInEnabledList` will now always return true, is it still possible for people to still enable dark reader on those sites?

alexanderby: For DR v5 there will be 2 lists, one for enabled and one for disabled websites. But currently there is a single `siteList` for both modes. The `siteListEnabled` is currently used to override websites from Dark List. So yes, websites from this list will always be active. And the dark theme detection will be disabled for them (otherwise it will automatically disable them).

rakleed:
@alexanderby could you please add this feature in iOS extension?

PR #8359, opened 2022-03-04
Update dynamic-theme-fixes.config

Golem642:
Added fix for two major french school websites (Toutatice and Pronote)

PR #8373, opened 2022-03-05
Dynamic dark theme fix for educational french websites

Golem642:
Added fix for two major french school websites (Toutatice and Pronote) (Updated and fixed from #8359 )


Golem642:
Closed old PL because of my mistakes of not applying the suggested changes that i wanted to apply, and couldn't revert back to do it

Gusted (approved):

PR #8375, opened 2022-03-05
Add palettte.app and colorbox.io to dark sites

Jorgen1040:
Add [palettte.app](https://palettte.app/) and [colorbox.io](https://colorbox.io/) to dark-sites.config.

Kind of important considering they're palette generation sites and darkreader messes with the colors

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -178,6 +178,7 @@ codesandbox.io
 codestackr.com
 codingame.com
 coinpot.co
+colorbox.io
 comicextra.com
 comicpunch.net
 conficturaindustries.com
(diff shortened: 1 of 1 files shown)


Gusted:
> Kind of important considering they're palette generation sites and darkreader messes with the colors

That isn't a factor of why a site should be included in the dark sites config. Closing.

Jorgen1040:
> That isn't a factor of why a site should be included in the dark sites config. Closing.

I guess I forgot to mention that both sites have a dark theme by default as well

Gusted:
> I guess I forgot to mention that both sites have a dark theme by default as well

Hmm seems that only 1 of the sites are by-default dark.

PR #8377, opened 2022-03-06
Fix for chessprogramming.org

0xhtml:
- Correctly invert logo
- Invert embedded images

Example page: [https://www.chessprogramming.org/Texel's_Tuning_Method](https://www.chessprogramming.org/Texel%27s_Tuning_Method)

Before:
20220306_12h14m56s_grim

After:
20220306_12h09m23s_grim

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2819,6 +2819,17 @@ INVERT
 
 ================================
 
+chessprogramming.org
+
+INVERT
+img:not(.thumbimage):not(.thumbborder)
+.mw-wiki-logo
+
+IGNORE IMAGE ANALYSIS
+.mw-wiki-logo
+
+================================
+
 chilkatsoft.com
 
 CSS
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -2819,6 +2819,17 @@ INVERT
 
 ================================
 
+chessprogramming.org
+
+INVERT
+img:not(.thumbimage):not(.thumbborder)
```
Gusted, line 2825: This will invert all kind of images, and not just the ones with latex expressions on it. This is too broad to be included.

Gusted (approved):

src/config/dynamic-theme-fixes.config
0xhtml: I tried to narrow the selector down as far as possible in a new commit. As far as I looked through the pages all images that aren't equations or diagrams integrated into the text have the `thumbborder` or `thumbimage` class set.

PR #8386, opened 2022-03-07
Banned dark mode for a .io game

Golem642:
Added shellshock.io (with its alt geometry.best) in the darkreader blacklist because glitches occur when dark mode is on.


Gusted (approved):
Seems like dark reader cannot do much either to the site.

PR #8371, opened 2022-03-05
Fix for wikipedia.org

rakleed:
- Invert editor toolbar icons.
- Invert notification icons.
- Invert audio speaker icon.

Example of site page: https://en.wikipedia.org/wiki/Wikipedia?action=edit

Resolves #8384

PR #8401, opened 2022-03-09
Fix for Wikimedia websites

neonr4in:
Related to #8371 , for the rest of Wikimedia Foundation websites

Example: https://en.wiktionary.org/w/index.php?title=look_after&action=edit

notification_bar
toolbar

commons.wikimedia.org, mediawiki.org, wikibooks.org, wikidata.org, wikimedia.org, wikinews.org, wikiquote.org, wikisource.org, wiktionary.org, wikiversity.org, wikivoyage.org, wikitech.wikimedia.org

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3364,6 +3364,14 @@ INVERT
 .mw-mmv-filepage-buttons .mw-mmv-view-expanded .mw-ui-icon::before
 .mw-mmv-filepage-buttons .mw-mmv-view-config .mw-ui-icon::before
 .licensetpl td[style^="width"]
+#p-personal .mw-echo-notifications-badge
+.mw-ui-icon::before
+.mw-kartographer-map
+.mw-kartographer-mapDialog-map
+.oo-ui-iconElement-icon
+.oo-ui-indicatorElement-indicator
+.mw-echo-notifications-badge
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -3364,6 +3364,14 @@ INVERT
...
 .mw-mmv-filepage-buttons .mw-mmv-view-config .mw-ui-icon::before
 .licensetpl td[style^="width"]
+#p-personal .mw-echo-notifications-badge
+.mw-ui-icon::before
+.mw-kartographer-map
+.mw-kartographer-mapDialog-map
+.oo-ui-iconElement-icon
+.oo-ui-indicatorElement-indicator
+.mw-echo-notifications-badge
+img[alt="audio speaker icon"]
```
Gusted, line 3374: I'm not sure if all these elements will actually apply something onto those websites. Certain things like `.mw-kartographer-map` would need to be checked because I don't think every website will have maps to be displayed.

neonr4in: After recheck, I think Kartographer isn't needed outside Wikipedia (as originally applied by #8371)

I removed Kartographer lines and left the remaining (they are used by notification badge and editing toolbar per screenshots)

Gusted (approved):

PR #8433, opened 2022-03-13
Apply fix for header on yettel.rs

kizule:
yettel.rs is a site of mobile provider in Serbia, previously known as Telenor Serbia.

For start, this pull request makes content in header readable. This isn't perfect, but does the job.


kizule:
Before:
image

After:
image

Gusted (approved):
Yeah good start indeed.

PR #8468, opened 2022-03-18
Add dark themed websites

alfawal:
and normalize some lines.


src/ui/popup/index.tsx
```diff
@@ -72,16 +72,16 @@ if (DEBUG) {
         try {
```
Gusted, line 72: Please revert these lines. Code-style should take of styling issue. Also not related to adding dark themes PR. Don't clutter 1 PR with multiple things.

src/config/dark-sites.config
```diff
@@ -726,6 +726,7 @@ polsatboxgo.pl
 polsatgo.pl
 pony.tube
 pool.pm
+popper.js.org
```
Gusted, line 729: I don't find the start page that dark, quite flashy when you open the site.

src/config/dark-sites.config
```diff
@@ -769,6 +770,7 @@ redditinvestigator.com
 redeclipse.net
 reelgood.com
 refind.com
+regex101.com
```
Gusted, line 773: Not dark-by-default.

Gusted (changes requested):

src/config/dark-sites.config
alfawal: This is weird, can you provide some efficient ways to check? incognito mode is not enough I guess.

src/config/dark-sites.config
alfawal: same here

src/ui/popup/index.tsx
alfawal: Okay, is there a label or something called "Code-style"? should I do these lines there?

Gusted (approved):

PR #8478, opened 2022-03-19
Several smash.gg fixes

FluentCoding:
- Removed white border below "Close"
- Removed the white gradient background from the bracket header
- Fixed "path" svg that connects two sets inside of a bracket
- Adjusted shadow of tournament toaster to fit into the dark design


src/config/dynamic-theme-fixes.config
```diff
@@ -14524,6 +14524,27 @@ CSS
 
 ================================
 
+smash.gg
+
+CSS
+.ModalContext__close {
+    border-bottom: none !important;
+}
+.scroller-sgg2hsxy.hasHinting-sggj0FSG.horizontal-sgg1mjhU .controller-sggb2Rt_ {
```
Gusted, line 14533: All these `sgg2hsxy` should not be used as they are randomly generated and can change any time. Something along the lines of `div[class*="scroller"]` should be used. Also, a [good resource](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors) about the attribute selectors.

FluentCoding: done! thanks for pointing it out.

FluentCoding:
pretty much done now btw

src/config/dynamic-theme-fixes.config
```diff
@@ -14524,6 +14524,30 @@ CSS
...
+    --darkreader-bg--shadow--card: 0 2px 4px rgba(0,0,0,0.4) !important;
+}
+.ModalContext__close {
+    border-bottom: none !important;
+}
+div[class^="controller"] {
+    background: none !important;
+}
+div[class^="BracketContainer"] > svg > path {
+    stroke: #373c3e !important;
```
Gusted, line 14540: Can this be converted into the `#373c3e` -> `${original color here}`?

src/config/dynamic-theme-fixes.config
```diff
@@ -14524,6 +14524,30 @@ CSS
...
+}
+div[class^="ToasterSlider"] {
+    box-shadow: rgb(177 157 157 / 10%) 1px -2px 10px !important;
```
Gusted, line 14543: Ditto

src/config/dynamic-theme-fixes.config
```diff
@@ -14524,6 +14524,30 @@ CSS
 
 ================================
 
+smash.gg
+
+CSS
+:root {
+    --darkreader-bg--shadow--card: 0 2px 4px rgba(0,0,0,0.4) !important;
```
Gusted, line 14531: Ditto

src/config/dynamic-theme-fixes.config
FluentCoding: do you mean that i should just put the hex into the ${...} or should i find in the source code of smash.gg if there is a css color variable for it.

(btw there is no css color name for this exact hex)

Gusted: You should put the original hex from smash.gg's theme into `${HERE}`, such that darkreader will modify it correctly according the user's settings.

FluentCoding: "the original hex" => so basically the inverted hex of what is in here? what about custom colors that i chose here, e.g.:

rgb(177 157 157 / 10%)

FluentCoding: also, i have another problem: smash.gg updates its page automatically (spa) and when said elements are being created, the darkreader styles don't apply.

Gusted: > so basically the inverted hex of what is in here?

Yup, but they can also be in RGB(A) format if needed, dark reader can automatically convert them.



> what about custom colors that i chose here, e.g.:
> 
> rgb(177 157 157 / 10%)

Avoid them at all time, you can insert `${#B19D9D}` and see what dark reader's result it and put that in this PR(as modification function that dark reader uses, is as well the inverse function of itself).

Gusted: > also, i have another problem: smash.gg updates its page automatically (spa) and when said elements are being created, the darkreader styles don't apply.

I don't think that's a problem, many sites are these days and Dark Reader watches for updates just fine.

FluentCoding: > > also, i have another problem: smash.gg updates its page automatically (spa) and when said elements are being created, the darkreader styles don't apply.
> 
> I don't think that's a problem, many sites are these days and Dark Reader watches for updates just fine.

I can confirm that it is an issue because reapplying the styles when the page is loaded already makes it work as it should.

Gusted: Please open up a new issue about it then, it's _quite_ unrelated to this PR.

FluentCoding: alright. i will do the color fixes asap

PR #8497, opened 2022-03-23
Update NYT CSS matchers for new puzzle DOM

mark8128:
The New York Times is rolling out a new UI for the crossword puzzle which has different class names for the cells breaking the CSS matchers.

This PR adds matches for the new puzzle layout while maintaining the old format as both versions of the board appear to be live in production.

Some readability issues with penciled text being gray on gray have also been addressed.

image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -11765,6 +11765,21 @@ a[data-testid] > svg {
 #xwd-board rect[class^="Cell-related--"] {
     fill: ${goldenrod} !important;
 }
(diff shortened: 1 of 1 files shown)


Gusted (commented):
Are the old selectors still needed?

mark8128:
> Are the old selectors still needed?

NYT appears to be rolling out the changes gradually (not all my browsers had the change) so I kept them in for compatibility.

Let me know the preferred strategy for gradual release changes in the repo. If it's preferred to only have the new matchers I'll strip the old ones out. I can't imagine they have a very long release window.

Gusted:
Nah, then it's all fine to still include them.

PR #8500, opened 2022-03-23
Update dev dependencies

alexanderby:
Fixed vulnerability warnings.


Gusted:
@alexanderby Could you merge the latest master branch into this? https://github.com/darkreader/darkreader/pull/8504 should fix the CI errors.

alexanderby:
Maybe we should seek for some more stable way, e.g. fire some custom DOM event when Dark Reader finishes processing. Though it can be tough as far as styles can be added or modified at any time.

Issue #8537, opened 2022-03-28
[Broken Website] discuss.pixls.us


q1800:
# Broken Website Report

## Website Address
https://discuss.pixls.us/

## Steps To Reproduce
Visit the website with Dark Reader enabled.
Notice the logo in the screenshot needs fixing.
Click on the Sign Up button near the top right on the webpage.
Notice the white area in the screenshot needs fixing.

## Problematic Observation
Logo
White Area

## Screenshots
![Screenshot 2022-03-28 at 16-17-39 discuss pixls us - Free Software Photography](https://user-images.githubusercontent.com/95879668/160490019-0c40a3cd-7c3b-42ed-9c47-d38d13447299.png)

![Screenshot 2022-03-28 at 16-17-59 discuss pixls us - Free Software Photography](https://user-images.githubusercontent.com/95879668/160489924-c23a7b33-f13a-455b-8d4d-aec3be3d3d5d.png)

## System Information

- OS: Windows 10
- Browser: Firefox Nightly 100
- Dark Reader Version: 4.9.47

Gusted:
Re-opening due white area.

Elaborendum:
Just in case, this happens in all pages using Disqus when signing up, like https://community.signalusers.org/ or https://community.jitsi.org/.

PR #8545, opened 2022-03-29
Fix for discuss.pixls.us

q1800:
Fix for #8537

Hey @Gusted,

This is my first real attempt at fixing website issues for Dark Reader. Please let me know how I did or if anything is wrong.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4548,11 +4548,6 @@ discuss.pixls.us
 INVERT
 #site-logo.logo-big
 
-CSS
-:root {
-    --secondary-rgb: var(--darkreader-bg) !important;
-}
-
 ================================
 
 disk.yandex.*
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -4543,6 +4543,18 @@ body
...
 
+discuss.pixls.us
+
+INVERT
+#site-logo.logo-big
+
+CSS
+:root {
+    --secondary-rgb: var(--darkreader-bg) !important;
+}
```
Gusted, line 4554: So while this is correct... I would like to fix this within the code, because this would currently happen for any discourse instance out their. So if I fix this within the code, every discourse instance would be fixed. If you could leave this out for this PR that would be nice.

q1800: Sure thing! Please check my work.

Gusted (approved):

PR #8549, opened 2022-03-29
Fix for mesa3d.org

q1800:
Additional fix for Dropdown Caret when section is expanded.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -10472,6 +10472,7 @@ mesa3d.org
 
 INVERT
 .collapsed a svg
+a svg
 
 ================================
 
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -10472,6 +10472,7 @@ mesa3d.org
 
 INVERT
 .collapsed a svg
+a svg
```
Gusted, line 10475: Could this be _specified_ a bit further? E.g. adding a specific class-name? `a svg` is too _generic_ and could unwanted target other elements.

q1800: Would it be the most specific if I put the following:

mesa3d.org

INVERT
[data-bs-toggle="collapse"].collapsed > a > svg
[data-bs-toggle="collapse"] > a > svg

This way both lines in the fix for mesa3d.org are corrected to be as specific as possible?

Gusted: That would indeed be more specific! But `[data-bs-toggle="collapse"].collapsed > a > svg` would be redundant then. so you can replace it with a one-liner :D

Gusted (approved):

PR #8551, opened 2022-03-29
Update README.md

q1800:
Corrected grammatical errors, fixed passive voice use, reworded several sentences and paragraphs, fixed format issues and removed unnecessary extra spaces and formatting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Open the terminal in the root folder and run:
 - `npm install`
 - `npm run build`
 
-A `build/release/darkreader-chrome.zip` file will generate for use in a Chromium-based browser, and also a `build/release/darkreader-firefox.xpi` file will generate for use in Firefox.
+This will create a `build/release/darkreader-chrome.zip` file for use in a Chromium-based browser and a `build/release/darkreader-firefox.xpi` file for use in Firefox.
 
 ## Using Dark Reader for a website
 
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ const CSS = await collectCSS();
 const isEnabled = isDarkReaderEnabled();
 ```
 
-Be aware that darkreader will add the chrome object onto the window object. These are to stub certain functions that the code will use. They originate from the webextension-api.
+Be aware that darkreader will add the `chrome` object onto the `window` object. These are to stub certain functions that the code will use. They originate from the webextension-api.
 
 ## Site fixes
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,12 +22,12 @@ Read more about contributing to Dark Reader in [CONTRIBUTING.md](https://github.
...
-Open terminal in the root folder and run:  
+Open the terminal in the root folder and run:
 
-- `npm install`  
-- `npm run build`  
+- `npm install`
+- `npm run build`
 
-This will generate a `build/release/darkreader-chrome.zip` file that is useable in a Chromium-based browser and also a `build/release/darkreader-firefox.xpi` file that is useable in Firefox.
+A `build/release/darkreader-chrome.zip` file will generate for use in a Chromium-based browser, and also a `build/release/darkreader-firefox.xpi` file will generate for use in Firefox.
```
Gusted, line 30:
```suggestion
A `build/release/darkreader-chrome.zip` file will generated for use in a Chromium-based browser, and also a `build/release/darkreader-firefox.xpi` file will generated for use in Firefox.
```

README.md
```diff
@@ -90,14 +90,13 @@ const CSS = await collectCSS();
 const isEnabled = isDarkReaderEnabled();
 ```
 
-Be aware, that darkreader will add the `chrome` object onto the `window` object, these are to stub certain functions that
-the code will use. They originate from the webextension-api.
+Be aware that darkreader will add the chrome object onto the window object. These are to stub certain functions that the code will use. They originate from the webextension-api.
```
Gusted, line 93: I think we should keep `

q1800: You are 100% correct. I stupidly removed them without even noticing. Completely my mistake.

README.md
q1800: The word "generate" is the correct tense form of the word to use here.

The word "generated" is the past tense form of the word "generate".

If we were to use "generated" here, then the correct wording would be "be generated". However, doing that invokes passive voice construction. It is always best to use active voice construction to provide a clear, concise, and straight to the point description of what is happening.

q1800: That was my explanation and thoughts on the wording. What are your thoughts on the wording after my explanation?

How do you wish to proceed? I can change it if needed.

Gusted: I think we should change the whole sentence structure, because the current form be generated isn't that good and just `generate` feels odd to me(it seems to be technically correct, but it doesn't feel like it).

q1800: How about now? I took the original sentence and reworded it/reduced the wordiness. I believe this is the best of both worlds and perhaps something we can agree on? 🤣

Gusted (approved):
LGTM! Thanks!

PR #8561, opened 2022-03-31
Update dynamic theme with cb2.com

jacobsgomez:
Updating to include cb2.com. The proposed change inverts the menu buttons to be white on the dark background (black previously)


src/config/dynamic-theme-fixes.config
```diff
@@ -19398,3 +19398,10 @@ CSS
 .zyimage {
     background-color: var(--darkreader-neutral-text) !important;
 }
+
+================================
+
+cb2.com
```
Gusted, line 19404: Please correctly order the site alphabetic order.

PR #8579, opened 2022-04-03
Remove image bg on dnd.su

BANOnotIT:
Remove background images on content blocks

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4691,15 +4691,12 @@ INVERT
 dnd.su
 
 CSS
-
 input,
 select,
 textarea,
 .input-cast_time,
 .input-style,
-.input-select, 
-.input-select-multiple,
-.input-select-multiple-options,
+.input-select,
 .button,
 button:not(.lg-icon),
 .workshop__checkbox .workshop__label::before,
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4688,6 +4688,38 @@ INVERT
 
 ================================
 
+dnd.su
+
+CSS
+.card-type,
+.paper-1,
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -4688,6 +4688,38 @@ INVERT
 
 ================================
 
+dnd.su
```
Gusted, line 4691: Is this the correct domain? I'm not able to visit it.

BANOnotIT: image
Yeah. It pretty is.

src/config/dynamic-theme-fixes.config
```diff
@@ -4688,6 +4688,38 @@ INVERT
...
+.input-cast_time,
+.input-style,
+.workshop__checkbox .workshop__label::before,
+#ya-site-form0 .ya-site-form__submit,
+.button,
+button:not(.lg-icon),
+input[type="submit"],
+input[type="button"] {
+   background-image: none !important;
+   background-color: var(--darkreader-neutral-background) !important;
```
Gusted, line 4718: Is this one necessary for **all** elements listed?

BANOnotIT: I've removed input types but yeah, most of the elements are special containers with background-image

Gusted: I meant, if all these elements needs to have the same background-color forced. It seems like only some of them needs a new background-color.

BANOnotIT: All these elements originally has no background-color, just background-image. The extension doesn't have any instruments to make background-image look ok. So I just replace whitish background-image with neutral background-color. If we don't set background-color all these elements become transparent.
изображение

Gusted (approved):

PR #8638, opened 2022-04-14
Revert timer timing

alexanderby:
- Revert #8634 and #8600.


Gusted (changes requested):
This will break the code again. We should use setUTCxxx instead then.

PR #8644, opened 2022-04-15
Add Neeva.com to global dark list

dwaxe:
Neeva has its own dark mode that one can set on the top right of its interface.

Neeva Dark Mode


Gusted:
Thank you for your PR, but this site doesn't meet the first requirement: https://github.com/darkreader/darkreader/blob/master/CONTRIBUTING.md#adding-a-website-that-is-already-dark

PR #8474, opened 2022-03-19
Non-zero exit code for failing Karma; Fixed browser execute boolean f…

mattalexx:
…lags from CI

These issues were brought up here:

> @Gusted This needs tweaking, it looks like.
> 
> * This is the ESR run, and Chrome should not be used. [darkreader/darkreader/runs/5591884298?check_suite_focus=true#step:10:259](https://github.com/darkreader/darkreader/runs/5591884298?check_suite_focus=true#step:10:259)
> * Firefox is failing but the test run is succeeding: [darkreader/darkreader/runs/5591884298?check_suite_focus=true#step:10:472](https://github.com/darkreader/darkreader/runs/5591884298?check_suite_focus=true#step:10:472)
> 
> I'll figure it out and post a PR.

https://github.com/darkreader/darkreader/pull/8393#issuecomment-1072097576

As a change of pace, I am removing more lines than I'm adding in this PR :-)

PR #8648, opened 2022-04-15
Stable async tests

alexanderby:
- Listen to async events in Dynamic mode tests.
- Restore CORS echo server (launched in a hacky way).


alexanderby:
Hi @mattalexx! It is important that our echo test server to run on a separate port to make the requests cross-origin (CORS style sheets are handled differently). Since #8474 it runs on the same port, thus the links became same-origin. Maybe you could suggest how to run the CORS server?

PR #8670, opened 2022-04-17
Fix comments on banki.ru

jtraub:
Comments section on banki.ru has white light background for even comments

Example URL: https://www.banki.ru/services/responses/bank/response/10649002/

How it looks without the fix

image

How it looks with the fix

image


Gusted (approved):

Gusted:
Thanks!

PR #8690, opened 2022-04-19
Scribd.com fixes 1

Mirm0n:
-Logo fix
-White highlight fix
-Contact form fix
-other fixes

scribd fixes
scribd fixes2
scribd fixes3
scribd fixes4
s
scribd fixes5

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -14356,7 +14356,33 @@ CSS
 scribd.com
 
 INVERT
-.logo
+.wrapper__megamenu_top_bar .logo_button
+.scribd_blue_long_logo, .scribd_white_logo
+.auto__upload_index .icon_row .reason_cell .cell_icon
+.auto__static_about .about_content .fact, .auto__static_about .about_content .fun_facts
+.e584-3.x-image
+.auto__shared_fancy_header_header .logo_image
+._1gtMJh.f6Cwhy::before
+
+CSS
+.flat_btn.outline_btn.negate {
+    border-color: white !important;
+    color: rgb(232, 230, 227) !important; 
+}
+.jfk-bubble.gtx-bubble, .captcheck_answer_label > input + img, span#closed_text > img[src^="https://www.gstatic.com/images/branding/googlelogo"], span[data-href^="https://www.hcaptcha.com/"] > #icon, #bit-notification-bar-iframe, ::-webkit-calendar-picker-indicator, .wrapper__megamenu_top_bar .logo_button, .scribd_blue_long_logo, .scribd_white_logo, .auto__upload_index .icon_row .reason_cell .cell_icon, .auto__static_about .about_content .fact, .auto__static_about .about_content .fun_facts, .auto__static_about .about_content .fact, .auto__static_about .about_content .fun_facts
+{
+filter: invert(100%) hue-rotate(180deg) contrast(0%) brightness(170%)  !important;
+}
+.e584-3.x-image{
+filter: invert(100%) hue-rotate(180deg) contrast(0%) brightness(170%)  !important;
+}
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -14356,7 +14356,33 @@ CSS
 scribd.com
 
 INVERT
-.logo
+.wrapper__megamenu_top_bar .logo_button
+.scribd_blue_long_logo, .scribd_white_logo
+.auto__upload_index .icon_row .reason_cell .cell_icon
+.auto__static_about .about_content .fact, .auto__static_about .about_content .fun_facts
+.e584-3.x-image
```
Gusted, line 14363: Try to avoid these kind of selectors, they are automatically generated and have a low guarantee that they won't be changed.

src/config/dynamic-theme-fixes.config
```diff
@@ -14356,7 +14356,33 @@ CSS
...
+.e584-3.x-image{
+filter: invert(100%) hue-rotate(180deg) contrast(0%) brightness(170%)  !important;
+}
+.auto__shared_fancy_header_header .logo_image
+{
+filter: invert(100%) hue-rotate(180deg) contrast(0%) brightness(170%)  !important;
+}
+.icon-ic_menuclosed:before{
+filter: invert(100%) hue-rotate(180deg) contrast(0%) brightness(170%)  !important;
+}
```
Gusted, line 14385: Why isn't regular `INVERT` used for this?

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
Mirm0n: This is more about the brightness increase and not the invert, it makes the whole logo white instead of a weird pale blue that doesn't look that great on the black background. (basically removes the saturation)

Mirm0n: If I remove the invert(100%) part it still works the same way and turns it to an all white logo , which looks very sleek on the dark background.

src/config/dynamic-theme-fixes.config
Mirm0n: What do you mean by "This type of selector" ? What kind should I be looking to use?

Gusted: The types that relies on random generated selectors. Try to see if this selector can be selected via a parent element that doesn't use a random selector.

src/config/dynamic-theme-fixes.config
```diff
@@ -14355,8 +14355,19 @@ CSS
 
 scribd.com
 
+scribd.com
+
 INVERT
```
Gusted, line 14360:
```suggestion
scribd.com

INVERT
```

src/config/dynamic-theme-fixes.config
```diff
@@ -14355,8 +14355,19 @@ CSS
...
-.logo
+.header_logo
+.logo_container
+._1gtMJh.f6Cwhy::before
+
+CSS
+.jfk-bubble.gtx-bubble, .captcheck_answer_label > input + img, span#closed_text > img[src^="https://www.gstatic.com/images/branding/googlelogo"], span[data-href^="https://www.hcaptcha.com/"] > #icon, #bit-notification-bar-iframe, ::-webkit-calendar-picker-indicator, .wrapper__megamenu_top_bar .logo_button, .scribd_blue_long_logo, .scribd_white_logo, .auto__upload_index .icon_row .reason_cell .cell_icon, .auto__static_about .about_content .fact, .auto__static_about .about_content .fun_facts, .auto__static_about .about_content .fact, .auto__static_about .about_content .fun_facts
```
Gusted, line 14366: You're now using too much CSS selectors here...

PR #8738, opened 2022-04-22
Remove YT Studio from dynamic-theme-fixes

zayadur:
YT Studio has native dark mode. This extension does help darken elements properly. However these custom INVERTs make it more difficult to see the radio buttons and logo. The default extension behavior is better.


zayadur:
> Does removing the fixes still works on both with and without the native dark mode enabled?
I overlooked this.
With **native dark mode enabled** and **DarkReader enabled**, logo and radio buttons look normal.
With **native dark mode disabled** and **DarkReader enabled**, logo is dark on dark and needs inverting, and "#onRadio tp-yt-paper-radio-button" is still dark on dark with inverting.

Issue #8742, opened 2022-04-22
[Broken Website] Replace with title


DeFaLT-dj:
# Broken Website Report
So, I just visited Bing today and the dark mode was broken. I first thought it was something to do with extension, but it is Bing testing some visual effect that basically adds accent color to the background. There is also a new icon under Bing's icon something similar to Google Doodles. I'll attach the screenshot to explain it more.
## Website Address

https://www.bing.com/search?q=world+book+day+observance&filters=HpDate:%2220220423_0700%22&setlang=en-us&cc=US&form=SHHPIH
## Steps To Reproduce
 
1- go to the https://www.bing.com/search?q=world+book+day+observance&filters=HpDate:%2220220423_0700%22&setlang=en-us&cc=US&form=SHHPIH

## Problematic Observation

It seems like Bing is trying a new visual style which breaks the dark reader. 
## Screenshots

I mentioned something similar to Google doodles added to Bing. I circled it with red. I think the main website issue is also visible. Bing puts some accent color to the top left corner depending on content
![image](https://user-images.githubusercontent.com/64617514/164814359-5444e061-8300-4118-ae5b-186751156cbd.png)

## System Information

- OS:  Windows 11
- Browser:  Edge 100
- Dark Reader Version:  4.9.47

## Additional Context

This new design is still in testing so you may not see it on your first try. If it doesn't show up, try opening it on incognito or on different browsers

Gusted:
Not reproduce-able *yet* for me on multiple browsers/machines.

DeFaLT-dj:
> Not reproduce-able _yet_ for me on multiple browsers/machines.

I think they removed it now since I didn't see it in a while but now the icons look a little glitchy. I don't know what they are trying to do. 
![image](https://user-images.githubusercontent.com/64617514/166212813-5c440fab-57ff-4821-89a4-ada1074ab8db.png)

TotallyYourGrandpa:
I can reproduce it on my machine with Arch Linux, Edge Version 101.0.1210.39, and Dark Reader 4.9.48.

PR #8781, opened 2022-04-28
Bing Results with summary side panel

ChriZ0529:
When you search for something that has an extended summary panel, it adds a gradient to the background that was being inverted strangely. This change removes the gradient.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2047,7 +2047,7 @@ CSS
     z-index: 2 !important;
 }
 #b_content {
-    background: none !important
+    background-image: none !important
 }
 
 IGNORE INLINE STYLE
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2046,6 +2046,9 @@ CSS
 .l_ecrd_imcolheader.gradient {
     z-index: 2 !important;
 }
+#b_content {
+    background: none !important
+}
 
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -2046,6 +2046,9 @@ CSS
 .l_ecrd_imcolheader.gradient {
     z-index: 2 !important;
 }
+#b_content {
+    background: none !important
```
Gusted, line 2050:
```suggestion
    background-image: none !important
```

Gusted:
Does this solve https://github.com/darkreader/darkreader/issues/8742?

ChriZ0529:
> Does this solve #8742?

It won't do anything about the circled "doodle" mentioned there but fixes the gradient.
I actually can't recreate this anymore, so I don't see the html in question. I do remember it being a background, rather than background-image though.

Gusted:
> I actually can't recreate this anymore, so I don't see the html in question. I do remember it being a background, rather than background-image though.

`background` is a shorthand for all the `background-*` properties. But in this case it's a gradient which are implemented under the `background-image` property, so just specifying that would be better.

Gusted:
> I have no idea if this interferes or not, but I would like to say that bing.com has been removed from the global dark site blacklist.

_Shouldn't_ interfere. For context: It was there for historical reason, just a matter of time before it was removed(as it shouldn't be there)

Gusted (approved):
Thanks!

PR #8785, opened 2022-04-29
Two minor changes to the PT-BR translation

Tiagoquix:
Hello. I'm sorry to be creating another pull request in such a short time. But it's because the change in `@setup_hotkey_toggle_extension` I would have liked to do, but I wasn't sure if it would fit in the extension (there are texts that are cut off and you can't really get a sense of the limit). I did the "calculations" here and it looks like it will fit.

About the change in `@more`, I'm changing it because, at least, in my language, the translation looks weird. I would suggest you change the English to "Extras" too, just like I did. I didn't include this change from "More" to "Extras" in the English version in this pull request because I don't know if you would like it.

This will be my last pull request regarding PT-BR translations. I'll only do another one if there are new translations.

Again, I apologize for bothering you.

Thanks in advance.

### DIFF
--- a/src/_locales/pt-BR.config
+++ b/src/_locales/pt-BR.config
@@ -24,8 +24,8 @@ alternância (site atual)
 Ativar/des. extensão
 
 @setup_hotkey_toggle_extension
-Atalho de alternância
-da extensão
+Configurar atalho de
+alternância da extensão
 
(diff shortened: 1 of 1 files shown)


Gusted (approved):
Thank you!

PR #8857, opened 2022-05-10
Fix for bloomberg.com

c-nagy:
- Fix top menu bar border color.
- Invert dark magnifying glass icon, Live Now icon, and current tab indicator.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2328,7 +2328,7 @@ INVERT
 
 CSS
 .navi {
-    border-color: rgb(140, 130, 115) !important;
+    border-color: var(--darkreader-border--color-black) !important;
 }
 
 ================================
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -2319,6 +2319,20 @@ CSS
...
+
+INVERT
+.navi-bar__button-icon
+.navi-sections__item--LiveNow::before
+.navi-sections__link--current::after
+
+CSS
+.navi {
+    border-color: rgb(140, 130, 115) !important;
+}
```
Gusted, line 2332: Could this use the `${original color}` for the color part?

c-nagy: Good idea. It was a solid black border turned off-white, but using a var() it looks good in alternative color schemes.

Gusted (approved):

PR #8866, opened 2022-05-11
Fix Stud.IP portal of the University of Passau

ThexXTURBOXx:
This fixes the following artifacts:
grafik

After applying the patch, it looks like this:
grafik

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -16002,7 +16002,7 @@ studip.uni-passau.de
 
 CSS
 #layout_container > div {
-  background: 0;
+  background-image: none;
 }
 
 ================================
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -15998,6 +15998,15 @@ paper-radio-button
 
 ================================
 
+studip.uni-passau.de
+
+CSS
+#layout_container > div {
+  background: 0;
```
Gusted, line 16005: This isn't valid CSS. Maybe try:
```suggestion
  background-image: none;
```

ThexXTURBOXx: It actually is valid for Stud.IP itself as can be seen here (this is the standard configuration which gets overriden by Dark Reader):
grafik

Gusted: It's preferred to avoid such syntax.

Gusted (approved):

ThexXTURBOXx:
@Gusted In that case, I have applied your suggested change. Thanks for the information!

PR #8886, opened 2022-05-14
Fix for icrc.org

pnaf:
Fixes logo on most pages.

Before:

After:

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -8359,7 +8359,8 @@ CSS
 icrc.org
 
 CSS
-.logo, .logo img {
+.logo,
+.logo img {
     background-color: #fff !important;
 }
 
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -8356,6 +8356,15 @@ CSS
 
 ================================
 
+icrc.org
+
+CSS
+.logo, .logo img {
```
Gusted, line 8362: I don't think `.logo img` does need this fix as well, just applying it to `.logo` should be sufficient.

pnaf: `.logo img` is still needed on most pages (for example, https://www.icrc.org/en/what-we-do/covid-19-pandemic).

Gusted:
```suggestion
.logo,
.logo img {
```

Gusted (approved):

PR #8887, opened 2022-05-14
Fix for understandingwar.org

c-nagy:
- Ignore image analysis throughout site to fix images.
- Set text in header search bar and top menu to be the correct color. Not using ${color} to avoid messing with the site's color scheme.
- Remove bright background images from posts and various other areas of the site.
- Force white background behind transparent ISW logos on home page and publications sections.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -17665,6 +17665,28 @@ a.logo[title="United Nations"]
 
 ================================
 
+understandingwar.org
+
+CSS
+#edit-search-block-form--2,
+#nice-menu-1 > li > a {
+    color: #665c39 !important;
+    text-shadow: none !important;
+}
+#mainwrap,
+.secondarybox,
+.content,
+.view-filters {
+    background-image: none !important;
+}
+img[src*="ISW\%20LOGO"] {
+    background-color: white !important;
+}
+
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -17665,6 +17665,30 @@ a.logo[title="United Nations"]
 
 ================================
 
+understandingwar.org
+
+CSS
+#edit-search-block-form--2,
+#nice-menu-1 > li > a {
+    color: #665c39 !important;
```
Gusted, line 17673: Could this be in the form of `color: ${#original hex color} !important`?

c-nagy: I couldn't get that to work well because the original color is what is desired in this case. Overall a tricky site to work with and I think this is as clean as the patch can be unfortunately.

c-nagy: I switched it out with `var(--darkreader-neutral-background)` which allows it to be readable and not have a hardcoded color. @Gusted

Gusted: > I couldn't get that to work because the original color is what is desired in this case.

You can take the output of `${original hex color}` and use that in the fix(it's a "perfect invert")

PR #8889, opened 2022-05-14
Remove valh.io from dark sites

va1entin:
My personal site is light by default now and relies on Dark Reader for a dark theme which is why this PR removes it from dark-sites.config.


Gusted (approved):
Thanks for the PR!

PR #8891, opened 2022-05-15
fix doc.qt.io

linsui:
https://doc.qt.io/qt-6/qtlabsplatform-index.html

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4961,8 +4961,8 @@ header a[href="/"] svg {
 doc.qt.io
 
 CSS
-body {
-  background-color: var(--darkreader-neutral-background) !important;
+:root {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -4958,6 +4958,15 @@ header a[href="/"] svg {
 
 ================================
 
+doc.qt.io
+
+CSS
+body {
+  background-color: var(--darkreader-neutral-background) !important;
+}
```
Gusted, line 4966:
```suggestion
:root {
    --content-bg-color: var(--darkreader-neutral-background) !important;
}
```

I think it's better to fix the css variable that's used.

linsui: Thanks!

Gusted (approved):

PR #8914, opened 2022-05-19
Fix unthemed DIVs in JIRA Epic issues list (atlassian.net)

terminalmage:
NOTE: The other changes in this file are due to a vim autocmd I have to remove trailing whitespace on buffer write.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1558,7 +1558,7 @@ code:first-of-type {
     background-color: ${rgba(3, 9, 15, 0.02)} !important;
 }
 .sc-186t4q4-1 {
-    background-color: rgba(3, 9, 15, 0.02) !important;
+    background-color: ${rgba(3, 9, 15, 0.02)} !important;
 }
 #ak-main-content {
     background-color: var(--darkreader-neutral-background) !important;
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -1557,6 +1557,9 @@ code:first-of-type {
     box-shadow: 0 0 4px ${rgba(3,9,15,0.15)} !important;
     background-color: ${rgba(3, 9, 15, 0.02)} !important;
 }
+.sc-186t4q4-1 {
+    background-color: rgba(3, 9, 15, 0.02) !important;
```
Gusted, line 1561: Could this like the others ones in the form of `${original color}`?

terminalmage: I don't understand the question. Are you asking if it should be wrapped in `${}`?

To be honest, I'm not even sure what that does, CSS is not one of my strengths.

Gusted: > Are you asking if it should be wrapped in `${}`?
> 
> To be honest, I'm not even sure what that does, CSS is not one of my strengths.

Yes I'm asking that. I'm sorry to not be upfront by this, but this is something Dark Reader-specific. It allows to invert the color on-the-fly given the user's setting. Thereby we don't rely on hard coded values(which won't work on light mode + custom settings).

terminalmage: That wasn't the original color, that was one I picked, using the value from the similarly-named class above it.

Here is the CSS for this DIV:

image

The DIV has a class of `sc-186t4q4-1 kTpmQd`.

Gusted (approved):

PR #8915, opened 2022-05-19
gamerant.com is dark

FadeMind:
https://gamerant.com/stranger-things-duffer-brothers-movies-homework/


Gusted:
This website isn't dark by default for devices that prefer light mode:
image

PR #8924, opened 2022-05-21
THLW Tierlist charname invert fix

DavedZrod:
in https://gamepress.gg/lostword/list/touhou-lostword-tier-list, character names would get inverted from its original color #000 to white but the white background doesn't change color, so i made it that the text is not inverted similar to the background

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -6929,6 +6929,15 @@ CSS
 
 ================================
 
+gamepress.gg
+
+CSS
+.TLW-tier-charname {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -6929,6 +6929,15 @@ CSS
 
 ================================
 
+gamepress.gg
+
+CSS
+.TLW-tier-charname {
+    color: #000 !important;
```
Gusted, line 6936:
```suggestion
    color: ${white} !important;
```

Gusted (approved):

PR #8950, opened 2022-05-25
Remove lollilol.cf & lollilol.ml from dark sites

ghost:
After trying to visit these websites in the browser, I have determined that they are not real. They are just malicious redirects and also possible adult content. I believe they need to be removed from this list and any other lists for Dark Reader. I entered both websites into virustotal.com and below are the reports for both the original URLs and the malicious redirect URLs.

**NOTE:** I have blacked out certain portions of the screenshots to hide my IP address, any tracking info, any other identifiable information, etc...

VirusTotal URL Report for lollilol.cf:

lollilol cf-adult-link

VirusTotal URL Report for malicious redirect from lollilol.cf:

lollilol cf-redirect-malicious

VirusTotal URL Report for lollilol.ml:

lollilol ml-adult-link

VirusTotal URL Report for malicious redirect from lollilol.ml:

lollilol ml-redirect-malicious

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -555,8 +555,6 @@ listen.moe
 livesplit.org
 loadout.tf
 lolesports.com
-lollilol.cf
-lollilol.ml
 lollilol.xyz
 lookmovie.ag
 looskie.com


Gusted (approved):
Thank you for your analysis.

PR #8965, opened 2022-05-27
Add rentry.co and rentry.org

Overimagine1:
Rentry.co and it's mirror, rentry.org now have native dark modes.
image


Gusted:
Hi @Overimagine1!

These sites don't provide their native dark theme by-default, which is a requirement for sites to be listed in this list.

PR #8986, opened 2022-05-31
Motivostiy.com - Make unchecked checkmarks match the background.

Word66:


src/config/dynamic-theme-fixes.config
```diff
@@ -11537,6 +11537,15 @@ INVERT
 
 ================================
 
+motivosity.com
+
+CSS
+.mv-checkbox input + i {
+    color: rgb(24, 26, 27);
```
Gusted, line 11544: This is not the correct fix, Dark Reader allows different user settings(like filter, brightness, constrast etc.) and thus this fix wouldn't be correct for those with different user settings. Could you link a example URL where the checkbox is broken?

PR #8993, opened 2022-06-02
Adds funtionality for the color picker to read local css variables

D0otDo0t:
This allows the user to enter locally defined CSS variables in addition to the original hex color values for the background and text color selector. 
Possibly helps solve #2168 for Firefox users as they can import their Pywal CSS into their profile/chrome folder and then enter their variable name into the extension.


Gusted:
Hi @D0otDo0t!

Did you tested this change? It seems to me that this wouldn't work, as extensions like paywal cannot modify other extension's scripts/popup.

D0otDo0t:
Hey @Gusted, I'll try my best to explain. 

I found that Firefox is able to read a CSS file placed within ~/.mozilla/firefox/user/chrome as long as toolkit.legacyUserProfileCustomizations.stylesheets has been enabled in Firefox's about:config. It works quite wells and I've also tested using non-existent variables, empty strings and semi colons etc. to try break it but I haven't been able to.

https://user-images.githubusercontent.com/106523981/172031880-8d86a3ec-0a0d-4267-84fb-55fb23465201.mp4

To explain the video a little bit more, when pywal executes it generates a number of templates using a color scheme picked from the input image. The "-o css" at the end of the command I enter is just a basic copy script used to transfer the created color.css template into my  ~/.mozilla/firefox/user/chrome folder. In firefox instead of entering a hex value I enter a variable from the template (i.e. --background, --foreground, --color[0-15]) which is then parsed into a valid color.

At the moment the user will have to reenter the variables each time the template is change but if you like the idea I'll try to add a manual and automatic fetching feature.

BelkaDev:
How would this detect changes? You'd need to rely on javascript
I feel like there should be a system colorscheme with a ln to .xresources or ~/.cache/wal/colors.css because all colors are set at runtime

D0otDo0t:
> How would this detect changes?

It doesn't at the moment. It just finds the hex value of a variable and stores that instead. I think to fetch it automatically you would have to store the last entered background/text variable name and then on browser startup/restart reparse colors.css and compare the new value with the old ones (or just replace the values regardless). I'm very new to javascript and typescript so I don't know how easy this is to implement.

Thanks for the hint about using ln instead of the copyscript it makes it much easier to automatically change my firefox's colors.css.

Gusted:
> Otherwise, darkreader could maybe reparse on each new tab as firefox can detect a change in colors.css but I don't know how much it would affect load times. I'm very new to javascript and typescript so I don't know how easy this is to implement.

It's possible, but it will require some more code than it currently has. Technically, dark reader could send the color of that variable to the background script, each time it's on a website. That approach is possible and not hard, but feels like it could be inaccurate(e.g. variable override or weird iframe situations etc.)/unstable. Currently you also just could copy paste the color of the variable you want to use and *poof* have the same effect.

D0otDo0t:
True that it has the same effect. However, this is still much faster than copying and pasting.

Gusted:
> True that it has the same effect. However, this is still much faster than copying and pasting.

Given it doesn't automatically update, it's a one-time thing. I don't know if it's that beneficial...

D0otDo0t:
It's really just faster currently.

I'm working on the automatic updating at the moment. I've found that firefox only loads the css file from when it first opens. For some reason the style editor will show changes to the file but none of the changes will be implemented until the browser is restarted.

Currently the automation fetches and applies a variable defined in defaults file on initial execution. I'm currently trying to figure out how to pass a user entered variable from the popup into the usersettings.

D0otDo0t:
@Gusted Automation is all done and works well with Firefox when set up and while it is not functional with Chrome I tested it a little it to ensure it didn't break anything. I'm not sure what caused package-lock.json to be modified so dramatically as I never touched it but I can re-commit with the original file from Darkreader if it's an issue.

To get it functioning Firefox's toolkit.legacyUserProfileCustomizations.stylesheets has to be set to true in about:config and then colors.css has to be copied/symlinked into your userprofile/chrome folder.
It is also required that this sheet is imported into the userContent.css. If the userContent.css doesn't contain a :root with at least one variable inside as well then color.css will not be read (I'm not sure why).

example userContent.css:
```css
@import "colors.css";

:root {
  --required = "#ffffff";
}
```   
Colors are only fetched on startup if a valid variable has been passed one of the previous sessions. Resetting individual colors to default or entering a normal hex color turns the automation off for that specific setting. Resetting the entire theme/extension results in automation being turned off completely as well.

src/background/extension.ts
```diff
@@ -537,6 +539,28 @@ export class Extension implements ExtensionState {
...
 
+    //------------------------------------
+    //
+    //    Fetch colors from local CSS
+    //
+    //
+
+    private cssUpdate() {
+        const rootVariables = getComputedStyle(document.documentElement);
+        if (this.user.settings.theme.cssVariableBg && /^#([0-9a-fA-F]{3}){1,2}$/.test(rootVariables.getPropertyValue(this.user.settings.theme.cssVariableBg).trim())){
```
Gusted, line 550: Please use the `isValidHexColor` util helper to check if it's a valid hex color.

Gusted (commented):
Please revert the changes in package-lock.json

src/definitions.d.ts
```diff
@@ -67,6 +67,11 @@ export interface Theme {
     lightColorScheme: string;
     darkColorScheme: string;
     immediateModify: boolean;
+    cssVariableBg: string;
+    cssVariableText: string;
+    cssVariableScrollBar: string;
+    cssVariableSelection: string;
```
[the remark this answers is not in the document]
Gusted: Could you please answer this?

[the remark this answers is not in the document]
D0otDo0t: To be honest at the time I kind of forgot that darkreader could have a light theme. I have changed it now so that there's background/text values for both light and dark. I've left scrollbar and selection as global because I think there corresponding color values are also global.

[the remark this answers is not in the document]
Gusted: So if I used a theme for github.com which uses custom scrollbar variable but on my default theme uses another custom scrollbar value. How would this code handle that case? It seems that this code would only handle the one that was set last.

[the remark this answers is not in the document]
D0otDo0t: Fair point. I've just tested, having custom themes works to an extent as darkreader still stores the color associated with the css variable. However, if the pywal theme is changed the colors for the custom theme do not. I'll try figuring it out.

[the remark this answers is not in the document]
D0otDo0t: I think i've fixed the issue. It wasn't due to the variables but where the cssUpdate function had been placed and called. It's works with presets, site specific and the all-site themes.

D0otDo0t:
It's now working with preset/custom themes I just had to change where the function was called. I created two presets and two custom themes and ran it through multiple pywal schemes without error. It should now work with as many preset/customThemes as darkreader allows.

src/ui/controls/color-dropdown/index.tsx
```diff
@@ -12,6 +12,8 @@ interface ColorDropDownProps {
     hasAutoOption?: boolean;
     onChange: (value: string) => void;
     onReset: () => void;
+    cssValue: string;
+    cssChange: (cssValue: string) => void;
```
Gusted, line 16: I don't think we should have a separate function for the variables one. It can reuse the `onChange` function. It does needs some logic to change the value for the correct property.

Gusted (commented):
The code seems fine to me. Could you add some tests to `tests/inject/dynamic/color.tests.ts`?

src/ui/popup/theme/controls/background-color.tsx
```diff
@@ -9,6 +9,7 @@ interface BgColorEditorProps {
     onChange: (value: BgColorValue) => void;
     canReset: boolean;
     onReset: () => void;
+    cssValue: string;
```
Gusted, line 12: Where are we still using cssValue for? Currently it's not being modified or used.

D0otDo0t: It only get used in the color-picker index.tsx if the entered hex color or cssvariable are invalid

Gusted: If the entered hex color or cssvariable is incorrect you don't have to call `props.onChange(...)`.



> Could you please also give some examples of the tests required

Well, if you define a define a variable under `:root { --ex-variable: red }` and then specify this variable in the settings to be used as background-color variable. Will the code be able to convert the variable into a value and then change the colors accordingly.

src/ui/popup/theme/page/index.tsx
```diff
@@ -73,25 +76,25 @@ function ColorsGroup({theme, change, colorSchemes}: ColorsGroupProps) {
...
-                onChange={(v) => change({scrollbarColor: v})}
-                onReset={() => change({scrollbarColor: DEFAULT_SETTINGS.theme.scrollbarColor})}
+                onChange={(v) => change(isValidColor(v)?{scrollbarColor: v}:{cssVariableScrollBar: v})}
+                onReset={() => change({scrollbarColor: DEFAULT_SETTINGS.theme.scrollbarColor, cssVariableScrollBar: DEFAULT_SETTINGS.theme.cssVariableScrollBar})}
             />
             <SelectionColorEditor
                 value={theme.selectionColor}
-                onChange={(v) => change({selectionColor: v})}
-                onReset={() => change({selectionColor: DEFAULT_SETTINGS.theme.selectionColor})}
+                onChange={(v) => change(isValidColor(v)?{selectionColor: v}:{cssVariableSelection: v})}
```
Gusted, line 96: Use spacing between the `?` and `:`

Gusted (approved):

Gusted:
Please add a item to CHANGELOG.md

README.md
```diff
@@ -103,6 +103,29 @@ However, this can be enabled using the following steps:
 - Click on the Preview new design button.
 - Enable the `Synchronize site fixes` setting under `Settings -> Manage Settings`.
 
+## Pywal colorscheme integration for Firefox
```
Gusted, line 106: I think this better fit at the wiki page for dark reader.

README.md
```diff
@@ -103,6 +103,29 @@ However, this can be enabled using the following steps:
...
+
+1. In FireFox's <code>about:config</code> set <code>toolkit.legacyUserProfileCustomizations.stylesheets</code> to <code>true</code>.
```
Gusted, line 108:
```suggestion
1. In Firefox's <code>about:config</code> set <code>toolkit.legacyUserProfileCustomizations.stylesheets</code> to <code>true</code>.
```

README.md
```diff
@@ -103,6 +103,29 @@ However, this can be enabled using the following steps:
...
+7. Click on the Preview new design button.
+8. Navigate to 'See all options -> Colors'.
+9. In place of hex color code(s) enter a valid css variable name (E.g <code>--background</code>, <code>--foreground</code>, <code>--color4</code>).
+
+To detect a change in your Pywal scheme restart firefox.
+
+This will also work with other colorscheme generators as long as template is in a valid css format:
+```css
+:root{
+--example: #123456;
```
Gusted, line 125:
```suggestion
    --example: #123456;
```

README.md
```diff
@@ -103,6 +103,29 @@ However, this can be enabled using the following steps:
...
 - Enable the `Synchronize site fixes` setting under `Settings -> Manage Settings`.
 
+## Pywal colorscheme integration for Firefox
+
+1. In FireFox's <code>about:config</code> set <code>toolkit.legacyUserProfileCustomizations.stylesheets</code> to <code>true</code>.
+2. Find your profile's root directory in <code>about:profiles</code> and (if it doesn't already exist) create a new folder inside it called <strong>chrome</strong>.
+3. Locate [Pywal's](https://github.com/dylanaraps/pywal) <strong>colors.css</strong>. By default this should be located at <code>~/.cache/wal/colors.css</code>
+4. Hard link this file to inside the <code>chrome</code> directory you just created using either:
+   - <strong>userContent.css</strong> as your link name if this file does not already exist the folder.
+   - <strong>colors.css</strong> as your link name if <strong>userContent.css</strong> already exists. Additionally, add <code>@import "colors.css";</code> to the top of your <strong>userContent.css</strong>.
```
Gusted, line 113: I think this uses vague wording. What's hardlinking?

Gusted (commented):
Lets first agree on the wording before I port it into the wiki.

README.md
D0otDo0t: Would putting in the command required be better?
`ln /location/of/colors.css /location/of/chrome/colors.css`
As well as a similar windows command?

Gusted: I think it's better to say "create a symlink" and for those that don't know can simply search "how to create symlink <platform>" and find a dozen tutorials

D0otDo0t: A lot of the tutorials I saw treated symlinks and hard links different things (i.e. `ln -s` vs `ln` respectively). Using `ln -s` creates a softlink which I can't get firefox to read.

Gusted: Maybe don't let dark reader be the one telling this? We simply just read whatever Firefox gives us. If a user wants to link their existing css generated by whatever tool, 🤷🏽.

D0otDo0t: I guess I could create my own lil guide or add something to the pywal wiki. I did create this with its main intention to have it work with pywal.

Gusted: Yeah that sounds like the best solution, this is a bit out-of-league for Dark Reader IMO.

acxz:
@D0otDo0t @Gusted is there anything holding this PR up, now that it has been approved?

Gusted:
> is there anything holding this PR up,

Conflicts and Ideally a new Dark Reader release had happen(which should happen next week) as I prefer to "spread" features over releases, so it's easier to iron out bugs and not have a huge pile of bugs :P

aflyingpumpkin:
@D0otDo0t @Gusted Is there any reason this is still being held up, the only merge conflict that exists right now is in `CHANGELOG.md` and the release you were talking about seems to have already happened.

bershanskiy:
> Is there any reason this is still being held up

The PR needs to be updated a to fix CI and I would like to see some new CI tests for this functionality, since it seems pretty fragile.

bershanskiy:
I'd like to make sure I understand this code: the intent here is to instruct Firefox to load a CSS file into the background context (by setting a few flags and placing a file or symlinks in exact correct places), then query the computed style of the background context, then communicate the style to content scripts. Is that correct?

I feel like this is rather fragile, inconvenient, and introduces plenty of technical debt. If something does not work, how does user troubleshoot the problem? What do you think of the following alternative solutions:
1. Run HTTP server (on localhost or and serve this one file for Dark Reader to be consumed? 
3. Give Dark Reader permission to open local files and point Dark Reader to this one file?

Option 1 requires no special permissions, can reuse our existing config "sync" infrastructure (basically, we just add one extra URL), enables people to share these configs, etc.
Option 2 requires special flags (like this PR does) but would enable Dark Reader to validate the CSS upon first load/subsequent reloads, point user to the exact problem ("hey, you asked me to load this file, but it does not exist" or "this file exists, but it is not valid CSS" or "I see a named CSS variable which I do not recognize"), would enable Dark Reader to poll for style updates much more frequently than on Firefox reload.

src/background/user-storage.ts
```diff
@@ -147,6 +153,32 @@ export default class UserStorage {
         UserStorage.saveStorageBarrier = null;
     });
 
+    private static rootVariables = !__TEST__ && getComputedStyle(document.documentElement);
```
bershanskiy, line 156: Perhaps, `isForefox` would be more fitting.

aflyingpumpkin:
@bershanskiy Correct me if I'm wrong, but doesn't #3519 basically do what your suggesting?

bershanskiy (changes requested):

bershanskiy:
> @bershanskiy Correct me if I'm wrong, but doesn't #3519 basically do what your suggesting?

It's actually more advanced than what I was suggesting: that PR adds external API for other extensions to communicate to DR (in a "push" manner) while here I'm suggesting a simpler/more primitive approach (ability for DR to "pull" changes from external resources).

src/ui/controls/color-picker/index.tsx
```diff
@@ -40,8 +40,15 @@ function ColorPicker(props: ColorPickerProps) {
         const value = rawValue.trim();
         if (isValidColor(value)) {
             props.onChange(value);
+            props.onChange('');
```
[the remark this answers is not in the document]
D0otDo0t: It's used to clear/reset the cssVariable when a valid color has been entered. Otherwise, every time the user reopens Firefox the colors would be instead be scraped from the CSS and change the theme they set.

niksingh710:
is this PR is being considered?

ThatOneCalculator:
> Give Dark Reader permission to open local files and point Dark Reader to this one file?

This seems like a sensible approach

ThatOneCalculator:
Can I ask why this was closed?

array-in-a-matrix:
Why closed?

alexanderby:
Too long to explain. I will keep in mind that `getComputedStyle(document.documentElement)` can work in Firefox background page.

niksingh710:
@alexanderby reading from a file is better option or using pywalfox (another plugin) to get colors?

as it already imports colors from pywal to the browser.

Issue #8351, opened 2022-03-04
[Bug] Dropdown menus are styled poorly


RheingoldRiver:
# Bug Report

## Website Address
https://www.goodreads.com/review/edit/52379336?report_event=true&survey=1

## Bug Description
Dropdown menus are illegible when using dark reader, even when there's no applicable userChrome.css interfering. (Relevant selectors: menulist, menupopup, menuitem.)

## Steps To Reproduce
1. Go to a website with dropdown menus
2. Expand a dropdown menu.

## Expected Behavior
Dropdown menus should be dark themed

## Actual Behavior
Dropdown menus are illegible

## Screenshots
(Note, I deleted all my userchrome.css for these screenshots)

With dark reader:
![image](https://user-images.githubusercontent.com/18037011/156832710-bb5711d6-e316-4aed-b487-1e4e1cab7d2a.png)

Without dark reader, using [this Firefox theme](https://addons.mozilla.org/en-US/firefox/addon/matte-black-red/?utm_content=addons-manager-reviews-link&utm_medium=firefox-browser&utm_source=firefox-browser)
![image](https://user-images.githubusercontent.com/18037011/156832788-8a9d891f-b9d6-4df7-b188-e78de2abe98b.png)

----

The same two screenshots if I use default Light theme:

With dark reader:
![image](https://user-images.githubusercontent.com/18037011/156832892-666f1b4c-54c1-4de0-a5bf-f5a31f1ebdf9.png)

Without:
![image](https://user-images.githubusercontent.com/18037011/156832848-63b4dbdc-efa7-493d-921a-f733b694d521.png)

## System Information

- OS: Windows
- Browser: Firefox 96.0.3
- Dark Reader Version: 4.9.45

## Additional Context
I think this is a regression within the last 10 or so Firefox versions, I'm *pretty* sure this used to be fine, and then I did a bunch of updates all at once, and now it's not fine.

Gusted:
Could you try disabling all firefox themes? Also you might want to disable dark reader styling such elements:


    Click Dark Reader icon.
    Click Dev tools (in the bottom-right corner).
    Click Preview new design.
    under Theme click See all options, Font's & more -> Style system controls

RheingoldRiver:
Hmm, even with System theme - auto, and that preference changed (so styling system controls is "No") (and again no userChrome.css) it looks like this:

[...]

PR #8994, opened 2022-06-02
Fixes for blocket.se

rappos:
Blocket is Swedens biggest online second hand shop, with millions of visitors every week

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2281,6 +2281,21 @@ svg
 
 ================================
 
+blocket.se
+
+CSS
+.gnHyj.template-index h3,
+#__next > div > div > div,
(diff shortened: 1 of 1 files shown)


Gusted:
Hi @rappos

Could you list a example URL where this fix is used?

rappos:
> Hi @rappos
> 
> Could you list a example URL where this fix is used?

For example the drop-down on the main page has weird colorings without this fix. https://www.blocket.se/

Press the button that says "Hela Sverige", and you'll see hopefully

Gusted:
> For example the drop-down on the main page has weird colorings without this fix. https://www.blocket.se/
> 
> Press the button that says "Hela Sverige", and you'll see hopefully

The dropdown is another issue, it's rendered by the OS and isn't local to just this site(e.g. see #8351)

rappos:
OS problem, not darkreader problem

PR #9004, opened 2022-06-03
Add RETRAC to invert fixes

lefte:
Added a RETRAC website background color fix

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -14408,8 +14408,8 @@ body {
 retracmirrors.com
 
 CSS
-div.children {
-    background-color: #101010;
+body, #app > div {
+    background-image: none !important;
 }
 
 ================================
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -14405,6 +14405,15 @@ body {
 
 ================================
 
+retracmirrors.com
+
+CSS
+div.children {
+    background-color: #101010;
```
Gusted, line 14412:
```suggestion
#app > div {
    background-image: none !important;
```

Gusted (commented):
It's better to remove the image instead of covering it up.

src/config/dynamic-theme-fixes.config
lefte: @Gusted thanks for the suggestion! I totally missed the image in the background. I added a new commit, with an additional image I found while editing.

Gusted (approved):

PR #9039, opened 2022-06-08
Fix for overleaf.com

abpaudel:
It seems like overleaf.com has made some changes due to which the PDF preview is no longer inverted. This is a simple fix for that.


Gusted:
Hi @abpaudel,

Could you show a before and after screenshot?

Gusted (approved):

PR #9071, opened 2022-06-12
Add vle.mathswatch.co.uk

Enchanted-Games:
Made some text that was blue be white when dark mode is enabled.
Made question images be inverted to match the dark theme when its enabled


Gusted:
Hi @Enchanted-Games!

Could you give a example where I can observe these changes?

PR #9081, opened 2022-06-15
Add zhihu.com config

zabOnO:
Add zhihu.com config，Set to pure black and white


Gusted:
image

Doesn't seem to be dark-by-default to me.

PR #9083, opened 2022-06-15
Update dark-sites.config

mystfox:
Add inskip.me


Gusted:
Hi @mystfox!

It seems like the site isn't dark-by-default:
image

mystfox:
Right. I wanna make this site look right, but I'll do it the right way (dynamic/filter fixes) later. tysm, sorry

PR #9115, opened 2022-06-21
Change to CONTRIBUTING.md

redtrillix:
Atom editor will no longer be supported by the developer and will be archived on December 15, 2022. You can find the original article here: https://github.blog/2022-06-08-sunsetting-atom/.

I assumed that since it will be discontinued, it should be removed as a useable editor.


Gusted (approved):
Thanks!

PR #9126, opened 2022-06-23
Add Tria.ge to dark sites

Still34:
## Summary

Tria.ge has its own dark mode, and the site is designed around it by default.


Gusted:
image
It doesn't seem to be dark-by-default.

Still34:
Unfortunately yeah, I was wrong about that. I'll close the PR in response.

PR #9129, opened 2022-06-24
Use XSS-safe method of injecting stylesheet proxy script with MV3

bershanskiy:
MV3 removes the ability to execute scripts by inserting test into `<script>` tag like so:
```js
const script = document.createElement('script');
script.append('console.log("This is considered unsafe")');
```
This is because `script.append()` is very similar to `eval()` which is explicitly forbidden.

Instead, we need to use a `<script>` tag with `src` attribute pointing to the file we wish to execute:
```js
const script = document.createElement('script');
script.src = chrome.runtime.getURL('path/to/file.js');
```

This PR does just that and will work on all platforms. (And make Dark Reader a bit safer in the process!)

~~This will be ready to be merged, once we resolve a few comments I have and I fix failing tests.~~ Done!


src/inject/dynamic-theme/index.ts
```diff
@@ -133,10 +141,11 @@ function createStaticStyleOverrides() {
     const rootVarsStyle = createOrUpdateStyle('darkreader--root-vars');
     document.head.insertBefore(rootVarsStyle, variableStyle.nextSibling);
 
-    const proxyScript = createOrUpdateScript('darkreader--proxy');
-    proxyScript.append(`(${injectProxy})(!${fixes && fixes.disableStyleSheetsProxy})`);
+    const proxyScript = createOrUpdateScript('darkreader--proxy', 'inject/proxy.js', !(fixes && fixes.disableStyleSheetsProxy));
     document.head.insertBefore(proxyScript, rootVarsStyle.nextSibling);
-    proxyScript.remove();
```
bershanskiy, line 139: Calling `proxyScript.remove();` immediately might be technically a bit racy (it's not in practice). I moved it into `onload` handler to avoid the race. However, I would prefer to actually move it into the injected file `src/inject/dynamic-theme/stylesheet-proxy.ts` instead.

This will simplify the code as well.

src/inject/dynamic-theme/index.ts
```diff
@@ -44,13 +43,22 @@ function createOrUpdateStyle(className: string, root: ParentNode = document.head
     return element;
 }
 
-function createOrUpdateScript(className: string, root: ParentNode = document.head || document) {
+/**
+ * Note: src must be included in web_accessible_resources manifest key.
+ */
+function createOrUpdateScript(className: string, src: string, args: any = undefined, root: ParentNode = document.head || document) {
     let element: HTMLScriptElement = root.querySelector(`.${className}`);
```
bershanskiy, line 50: I'm not sure why we attempt to preserve the same `<script>` element if we end up removing it anyway... Can we just not do this?

bershanskiy (commented):
Here are my comments.

alexanderby:
Wonderful! I didn't know it can still work in MV3.

Let's not merge it till Monday-Tuesday, I will make a stable release and then we can merge it.

src/manifest.json
```diff
@@ -62,5 +62,8 @@
         "switchEngine": {
             "description": "__MSG_theme_generation_mode__"
         }
-    }
+    },
+    "web_accessible_resources": [
```
alexanderby, line 66: Can it be avoided for MV2? I'm afraid it can trigger yet another precise review.

bershanskiy: Yes, of course we can make it exclusive to MV3.

src/inject/dynamic-theme/index.ts
bershanskiy: I fixed this already.

src/inject/dynamic-theme/index.ts
bershanskiy: I fixed this already.

src/manifest.json
bershanskiy: I fixed this already.

bershanskiy:
I believe this PR is ready to be merged (any time after the next stable release).

alexanderby (approved):

bershanskiy:
@alexanderby I saw that you approved the changes. Would you like to wait until after the next release anyway? I think we should wait, since there is no real benefit in rushing it.

src/inject/dynamic-theme/index.ts
```diff
@@ -54,6 +57,19 @@ function createOrUpdateScript(className: string, root: ParentNode = document.hea
...
 
+/**
+ * Note: This function is used only with MV3.
+ * String passed as src parameter must be included in web_accessible_resources manifest key.
+ */
+function createScriptMV3(src: string, args: any = undefined) {
+    const element = document.createElement('script');
+    element.src = chrome.runtime.getURL(src);
+    if (args !== undefined) {
+        element.setAttribute('args', JSON.stringify(args));
```
alexanderby, line 68: The `args` attribute seems to be non-standard. Maybe let's use `element.dataset.args = ...`? (It will create `dataset-args` attribute)

src/inject/dynamic-theme/stylesheet-proxy-mv3-injector.ts
```diff
@@ -0,0 +1,11 @@
+import {injectProxy} from './stylesheet-proxy';
```
alexanderby, line 1: The file name seems long. Maybe it can be shortened like `mv3-proxy.ts` or similar?
4xk

alexanderby:
Yes, I can merge it as soon the releases will be reviewed successfully. A new file can trigger a long review.

src/inject/dynamic-theme/index.ts
```diff
@@ -54,6 +57,19 @@ function createOrUpdateScript(className: string, root: ParentNode = document.hea
     return element;
 }
 
+/**
+ * Note: This function is used only with MV3.
+ * String passed as src parameter must be included in web_accessible_resources manifest key.
+ */
+function createScriptMV3(src: string, args: any = undefined) {
+    const element = document.createElement('script');
+    element.src = chrome.runtime.getURL(src);
```
Gusted, line 66: So, we had to revert earlier changes to this to be MV3-compliant because the loading of the script was too slow and "not blocking". Is that the same case here? Otherwise this isn't a valid way of injecting the proxy. Styles can change between the initial run of going trough styles and this script actually being loaded.

bershanskiy: > the loading of the script was too slow and "not blocking". Is that the same case here?

Unfortunately, I think so. I believe this is the limitation of current API in general. I am aware of multiple extensions which would benefit from injecting a script into the real/non-isolated page context early on in the page load process, but as far as I know, this feature does not exist right now. The last I checked, multiple widely used commercial and non-commercial extensions will cease to run in January 2023 once MV3 becomes mandatory. A larger number of extensions will regress in some way.

I see two potential solutions:
 - creating a separate CS solely for injecting this proxy. The race condition will still be there, but the proxy will have a much greater chance of winning.
 - I remember there was a discussion about adding an extra parameter to content scripts registrations in manifest. This proposed parameter would have two values "isolated world" (default) and "the real page world". This probably would be the perfect solution, but I can't find the relevant discussion.

bershanskiy: Converted to draft until I we resolve this.

Gusted: > A larger number of extensions will regress in some way.

Yup a big unfortunate for the little benefit that MV3 will add. But hey we must somehow conform to it, so just let get over with it 😄.

So I think we can use a variation of your potential solution. 




> * creating a separate CS solely for injecting this proxy. The race condition will still be there, but the proxy will have a much greater chance of winning.

I think this is the "best solution" to mitigate the timing issue, but I think we should also add mechanism to fire a event for when we injected the style(and thus from there-on were watching for style modifications) and when we catch that event we check if we already did the initial rounds of style checking and then just blatantly again go trough all styles. It's really inefficient but the best thing to do in this case.

bershanskiy: I prototyped the proposed solution and had hit a blocker: `injectScript` requires an argument which is known only by content script. We could default to `true` and then just request this data from CS and "undo" the extra proxies. This seems like the most logical solution.

Also, we can attempt to inject proxy from both places (a dedicated CS and the current CS) and just see whichever wins.

Gusted: > (a dedicated CS 

We probably should re-use the fallback.js, Otherwise the timing between executing content scripts becomes too big. We don't want to delay the first paint too much.

bershanskiy:
I'm pretty happy with how this PR turned out, I think it can be merged now. The performance characteristics of MV3 build are identical to those of MV2 build, as far as I can measure.

bershanskiy:
In a follow-up PR I might add ability for "dedicated path" to call `injectProxy(true)` when argument is not known and then downgrade it to `false` once it becomes available. But I do not think this is necessary.

bershanskiy:
@alexanderby Should I do anything to advance this PR? Or is it safe to merge now?

alexanderby:
If I understand it correctly, the MV2 version keeps working as it was and no new files will be generatedThen you can merge.

One note, I think `stylesheet-proxy-mv3-injector.ts` file is unused anymore.

Issue #8404, opened 2022-03-09
[Broken Website] forum.manjaro.org


q1800:
# Broken Website Report

## Website Address
https://forum.manjaro.org/

## Steps To Reproduce
Visit the website with Dark Reader enabled.
Scroll to the bottom of the webpage.
Notice footer image is partially off color with the rest of the webpage and also the blinding white part.

## Problematic Observation
1. Footer image is partially off color with the rest of the webpage.
2. Blinding white part.

## Screenshots
![image](https://user-images.githubusercontent.com/95879668/157499076-cc3fa612-ba93-439f-8f6b-070e281ef98a.png)

## System Information

- OS: Windows 10
- Browser: Firefox 98
- Dark Reader Version: 4.9.45

PR #9131, opened 2022-06-24
Cleanup Manjaro website fixes

q1800:
Remove discover.manjaro.org as website is no longer valid.
Remove manjaro.org as website has been completely overhauled and rewritten from scratch.
Update web page footer fix to be specific to forum.manjaro.org as was originally intended (Issue: #8404 / #8408).


Gusted (approved):
Thanks!

PR #9144, opened 2022-06-28
Add emojipedia.org to global dark list

Xaephia:


Gusted:
This site is not dark-by-default.

Issue #9080, opened 2022-06-15
[Broken Website] ikea.ee


saltedcoffii:
# Broken Website Report

## Website Address
https://www.ikea.ee

## Steps To Reproduce

- Switch language to English (if desired, bug appears either way)
- Search for anything in the search bar, I searched "bed"
- Observe bug

## Problematic Observation

Example URL: https://www.ikea.ee/en/rooms/bedroom/beds-and-sofa-beds/bed-frames?size[]=140x200%20cm&order=RECOMMENDED

Dark Reader does not properly darken some page elements

## Screenshots

![image](https://user-images.githubusercontent.com/55339220/173710523-3dffa0b4-4a4f-4195-9e0d-92d9f2a8e2a6.png)

## System Information
- OS: Arch Linux rolling
- Browser: Mozilla Firefox 101.0
- Dark Reader Version: 4.9.51

Gusted:
#9148 was merged.

PR #9148, opened 2022-06-29
Fix for ikea.ee (Issue #9080)

tony1661:
This is to fix Issues #9080 

The issues is that there is a white background by default applied to the .card css selector.

This hopes to fix that.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -8793,14 +8793,10 @@ CSS
 .hnf-svg-icon {
     fill: var(--darkreader-neutral-text) !important;
 }
-.card {
-  background-color: var(--darkreader-neutral-background) !important;
-}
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -8793,6 +8793,15 @@ CSS
...
 }
+.card {
+  background-color: var(--darkreader-neutral-background) !important;
+}
+.card-header {
+  background-color: var(--darkreader-neutral-background) !important;
+}
+.card-footer {
+  background-color: var(--darkreader-neutral-background) !important;
+}
```
Gusted, line 8804:
```suggestion
.card,
.card-header,
.card-footer {
    background-color: var(--darkreader-neutral-background) !important;
}
```

Gusted (approved):

PR #9156, opened 2022-06-30
Add runtothefinish.com

gardamork:
Some elements' (most notably article content) text changed shade correctly while background remained unaffected, making reading difficult.

Tested on front page and several articles, e.g. https://www.runtothefinish.com/half-marathon-training/

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -15150,6 +15150,15 @@ INVERT
 
 ================================
 
+runtothefinish.com
+
+CSS
+body {
+    background-color: var(--darkreader-neutral-background) !important;
+}
+
+================================
+
 rynekzdrowia.pl
 
 INVERT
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -15150,6 +15150,15 @@ INVERT
 
 ================================
 
+runtothefinish.com
+
+CSS
+body {
+    background-color: var(--darkreader-neutral-background) !important;
+}
```
Gusted, line 15158:
```suggestion
:root {
    --body-background-color: var(--darkreader-neutral-background) !important;
}
```

Gusted (commented):
Nicer to use the variable for it.

src/config/dynamic-theme-fixes.config
gardamork: Thanks for the advice!

PR #9157, opened 2022-06-30
Optimize config index encoding

bershanskiy:
This PR changes the encoding config index stores offsets as a string. It has no functional changes.

It saves about 15% to 20% of space (depending on which config is being indexed). In practice, savings are not very significant, but every bit helps.

### DIFF
--- a/src/generators/utils/parse.ts
+++ b/src/generators/utils/parse.ts
@@ -103,7 +103,7 @@ function decodeOffset(offsets: string, index: number): [number, number] {
     const length = parseInt(offsets.substring(base + offsetLength, base + offsetLength + lengthLength), 36);
     return [
         offset,
-        offset + length,
+        length,
     ];
 }
 
@@ -172,8 +172,8 @@ export function indexSitesFixesConfig<T extends SiteProps>(text: string): SitePr
     return {offsets: encodeOffsets(offsets), domains, domainPatterns, cache: {}};
 }
 
-export function parseSiteFixConfig<T extends SiteProps>(text: string, options: SitesFixesParserOptions<T>, recordStart: number, recordEnd: number): T {
-    const block = text.substring(recordStart, recordEnd);
+export function parseSiteFixConfig<T extends SiteProps>(text: string, options: SitesFixesParserOptions<T>, recordStart: number, recordLength: number): T {
+    const block = text.substring(recordStart, recordStart + recordLength);
     return parseSitesFixesConfig<T>(block, options)[0];
 }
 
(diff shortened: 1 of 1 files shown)


src/generators/utils/parse.ts
```diff
@@ -65,33 +65,62 @@ function getDomain(url: string) {
...
+function encodeNumber(n: number, length: number | null) {
+    let s = '';
+    for (let c = n; c; c = Math.floor(c / radix.length)) {
+        s = radix[c % radix.length] + s;
+    }
+    if (length !== null) {
+        s = radix[0].repeat(length - s.length) + s;
+    }
+    return s;
+}
```
Gusted, line 79: Isn't it efficient to use the full-range utf-8 by using `String.fromCharCode(...)`? Instead of a fixed amount of characters. `String.fromCharCode(...)` encodes it into a valid utf-8 string which can one to five bytes of strings(depending on the number) and javascript already decodes the utf-8 strings by default.

src/generators/utils/parse.ts
```diff
@@ -65,33 +65,62 @@ function getDomain(url: string) {
...
- *  - 3 bytes for record length (end offset - start offset)
- * Both values are stored in base 36 (radix 36) notation.
- * Maximum supported numbers:
- *  - start offset must be no more than parseInt('zzzz', 36) = 1679615
- *  - length must be no more than parseInt('zzz', 36) = 46655
- *
+ * Encode all offsets into a string, with a fixed-length header and records of equal size:
+ *  - header:
+ *    - 1 char for length of record offset (offsetLength)
+ *    - 1 char for length of record length (lengthLength)
```
Gusted, line 93: So with the current implementation they can't be larger than 92? As it would otherwise take multiple characters of space.

Gusted (commented):
Interesting change.

src/generators/utils/parse.ts
bershanskiy: If numbers/records have fixed length, we can trivially find the offset for a particular entry:
```js
const base = 1 + 1 + (offsetLength + lengthLength) * index;
```
 If we use UTF-8, we will need to iterate over the entire string every time. It is just a tradeoff O(1) search time with more space vs. O(n) search time with a bit less space. In either case, space usage is roughly the same.

src/generators/utils/parse.ts
bershanskiy: This means encoded length of record offset and encoded length of record length can not be larger than 92. This means offsets and lengths themselves can not be larger than 92<sup>92</sup> - 1. That should be plenty, right?

bershanskiy:
Also, we are talking about only few KBs of saved space, not much in the grand scheme of things.

Gusted:
> Also, we are talking about only few KBs of saved space, not much in the grand scheme of things.

Well if we make an optimization, we better make the best out of it 😄

src/generators/utils/parse.ts
Gusted: Yeah that should be plenty.

bershanskiy:
When I was preparing this patch, I already knew that storage saving would be rather slim, but didn't expect them to be this miniscule. I don't think the savings from this PR are worth the extra complexity.

PR #9164, opened 2022-07-01
Refactor: make makeChromiumHappy() consistent

bershanskiy:
This removes an error message logged by Chromium on MV3 builds.

### DIFF
--- a/src/background/make-chromium-happy.ts
+++ b/src/background/make-chromium-happy.ts
@@ -1,10 +1,14 @@
 import {MessageType} from '../utils/message';
 import type {Message} from '../definitions';
+import {isChromium} from '../utils/platform';
 
 // This function exists to prevent Chrome from logging an error about
 // closed conduit. It just sends a dummy message in response to incomming message
 // to utilise open conduit. This response message is not even used on the other side.
 export function makeChromiumHappy() {
+    if (!isChromium) {
+        return;
+    }
     chrome.runtime.onMessage.addListener((message: Message, _, sendResponse) => {
         if (![
             // Messenger
--- a/src/background/make-chromium-happy.ts
+++ b/src/background/make-chromium-happy.ts
@@ -7,9 +7,6 @@ import type {Message} from '../definitions';
 export function makeChromiumHappy() {
     chrome.runtime.onMessage.addListener((message: Message, _, sendResponse) => {
         if (![
-            // TabManager
-            MessageType.CS_FRAME_CONNECT,
-
             // Messenger
             MessageType.UI_GET_DATA,
             MessageType.UI_APPLY_DEV_DYNAMIC_THEME_FIXES,
--- a/src/background/tab-manager.ts
+++ b/src/background/tab-manager.ts
@@ -127,7 +127,6 @@ export default class TabManager {
                         frameURL: frameId === 0 ? null : senderURL,
                     });
                     this.stateManager.saveState();
-                    sendResponse({type: '¯\\_(ツ)_/¯'});
                     break;
                 }
                 case MessageType.CS_FRAME_FORGET: {


alexanderby:
Did you check if it breaks anything in non-MV3 Chrome and Firefox?

bershanskiy:
Yes, of course! I built and tested other platforms too.

This message was `{type: '¯\\_(ツ)_/¯'}` and it never used (literally nothing is listening for `'¯\\_(ツ)_/¯'`) and it was ever relevant only to Chromium builds.
The difference for Chromium builds: the message is sent right away instead of waiting for after another message is sent in the other direction, so error message is not logged.
The difference for non-Chromium builds: no message is being sent at all (the incoming message listener is not even created).

Gusted:
It seems to be fixed in recent version of Chromium, but wouldn't surprise me if we get bug reports again due to this PR.

bershanskiy:
> This was a fix for some strange bug, maybe it's no longer there, but maybe it is.

This is a Chromium-exclusive bug and it is still present in Chrome 103.0.5060.66. The origin of this bug is following:
1. CS calls `chrome.runtime.sendMessage()` creating port 1.
2. BG receives this message and handles it asynchronously and within the handler calls `chrome.runtime.sendMessage()` creating port 2 and closing port 1.
3. CS receives information about closure of port 1 "before a response was received" and logs an error.

There are multiple possible fixes, each just making sure to close the port 1 before opening port 2:
 - BG can register an async handler and make sure to call `sendResponse()` before calling `chrome.runtime.sendMessage()`
 - BG can register a handler and make sure to call ("schedule") async function and `return false` immediately
 - BG can register an async handler which will call `chrome.runtime.sendMessage()` and a "dummy" sync handler which will call `sendResponse()`. This works because the "dummy" calls `sendResponse()` prior to async handler calling `chrome.runtime.sendMessage()` (this is exactly what our current "fix" is)

> It seems to be fixed in recent version of Chromium

I still can reproduce it in Chrome 103.0.5060.66.

> but wouldn't surprise me if we get bug reports again due to this PR.

I'm actually currently working on a different fix which will prevent logging of this error even if it does occur.

src/background/tab-manager.ts
```diff
@@ -127,7 +127,6 @@ export default class TabManager {
                         frameURL: frameId === 0 ? null : senderURL,
                     });
                     this.stateManager.saveState();
-                    sendResponse({type: '¯\\_(ツ)_/¯'});
```
bershanskiy, line 130: `reply()` is called on line 125 which calls `chrome.tabs.sendMessage` on line 90 prior to this call to `sendResponse()`. I fixed this bug by calling `sendResponse()` in `makeChromiumHappy()`, but we could do the same by moving this `sendResponse()` to before `reply()`. It really does not matter much.

src/background/make-chromium-happy.ts
```diff
@@ -1,15 +1,16 @@
 import {MessageType} from '../utils/message';
 import type {Message} from '../definitions';
+import {isChromium} from '../utils/platform';
 
 // This function exists to prevent Chrome from logging an error about
 // closed conduit. It just sends a dummy message in response to incomming message
 // to utilise open conduit. This response message is not even used on the other side.
 export function makeChromiumHappy() {
+    if (!isChromium) {
```
bershanskiy, line 9: This workaround was ever present in Chromium so it does not make much sense to do it on other platforms...

bershanskiy:
@alexanderby @Gusted Sorry for merging this prior to letting you look over it, I didn't consider it a significant change worth discussion.

bershanskiy:
I compiled Chromium from source and the error message appears, so I don't think Chromium fixed it yet.

bershanskiy:
Also, I think I came a bit closer to diagnosing the real cause of this warning: I think it occurs when a tab context gets unloaded (tab navigation or reload) prior to port receiving the message. I reached this conclusion after I set console to preserve log and reloaded a tab and saw that "Navigated to" message appeared before "Unchecked runtime.lastError: The message port closed before a response was received."

Still looking into this...

bershanskiy:
I found out that simply using promise-based API fixes this error without using dummy messages because "for a promise-based channel, not receiving a response is fine". The source is the following [comment in Chromium source code](https://github.com/chromium/chromium/blob/main/extensions/renderer/one_time_message_handler.cc#L447):
```
  // Set the error for the message port. If the browser supplies an error, we
  // always use that. Otherwise, the behavior is different for promise-based vs
  // callback-based channels.
  // For a promise-based channel, not receiving a response is fine (assuming the
  // listener didn't indicate it would send one) - the extension may simply be
  // waiting for confirmation that the message sent.
  // In the callback-based scenario, we use the presence of the callback as an
  // indication that the extension expected a specific response. This is an
  // unfortunate behavior difference that we keep for backwards-compatibility in
  // callback-based API calls.
```

PR #9165, opened 2022-07-01
Fix message error handling

bershanskiy:
THIS IS WIP UNTILL WE DISCUSS IT

Prior to this PR, Dark Reader was not handling `chrome.runtime.sendMessage` errors correctly: it used try-catch when in reality it should have checked `chrome.runtime.lastError`.

HOWEVER, this commit also makes DR clean up all styles on error. I believe we fixed that bug in #9164, but since this is high risk change, I wanted to check up with you guys.

I see few ways to make this change safer:
 - make new behavior debug/beta/development-time only for some time
 - retry to deliver the message multiple times and bail out and clean up only after multiple fails
 - add reporting for this error to track frequency of its occurrences

CC @alexanderby @Gusted

### DIFF
--- a/src/inject/index.ts
+++ b/src/inject/index.ts
@@ -34,25 +34,21 @@ function sendMessage(message: Message) {
     if (unloaded) {
         return;
     }
-    try {
-        chrome.runtime.sendMessage<Message>(message, (response) => {
-            // Vivaldi bug workaround. See TabManager for details.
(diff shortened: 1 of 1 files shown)


Gusted:
Can this be closed given the recent PR?

bershanskiy:
Yes, closing this as superseded by #9174.

PR #9166, opened 2022-07-01
Fix: use chrome.tabs.onRemoved to clear TabManager data

bershanskiy:
Prior to this PR, TabManager was notified only by CS to get information about frame unload/closure based on `pagehide` event. This is not very reliable, since technically `pagehide` event handler is executed in a context which is actively being unloaded so any async operations can be thrown out arbitrarily. This commit adds an alternative path: `TabManager` listens for `chrome.tabs.onRemoved` event and deletes data about the whole tab.

Later, I plan to add similar logic for sub-frames: CS can listen for removals of `<iframe>` elements from the "parent" frame.

### DIFF
--- a/src/background/tab-manager.ts
+++ b/src/background/tab-manager.ts
@@ -63,6 +63,22 @@ export default class TabManager {
         this.tabs = {};
         this.getTabMessage = getTabMessage;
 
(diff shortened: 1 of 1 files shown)


Gusted (approved):
LGTM.

PR #9167, opened 2022-07-01
Add DevTools panels for MV3 builds

bershanskiy:
This embeds the whole Dark Reader Developer Tools into the browser DevTools UI: 
image

This change is MV3-exclussive since it requires modification to manifest. We could easily extend it to other builds if we wanted.

### DIFF
--- a/src/ui/devtools/loader.ts
+++ b/src/ui/devtools/loader.ts
@@ -1,0 +1,35 @@
+import type {DevToolsPanelSettings} from 'definitions';
+import {DEVTOOLS_PANEL_SETTINGS_STORAGE_KEY} from '../connect/connector';
+
+enum PanelState {
+    NONE,
+    CREATING,
+    EXISTS
+}
+
+let panelState: PanelState = PanelState.NONE;
+
+function createPanelIfNeeded(data: DevToolsPanelSettings | undefined) {
+    if ((data && data.enabled === false) || panelState !== PanelState.NONE) {
+        return;
+    }
+    panelState = PanelState.CREATING;
+    chrome.devtools.panels.create(
+        'Dark Reader',
+        'ui/assets/images/mode-dark-32.svg',
+        'ui/devtools/index.html',
+        () => panelState = PanelState.EXISTS
+    );
+}
+
+chrome.storage.local.get(DEVTOOLS_PANEL_SETTINGS_STORAGE_KEY).then((storage) => {
+    createPanelIfNeeded(storage.DEVTOOLS_PANEL_SETTINGS_STORAGE_KEY);
+});
+
+chrome.storage.onChanged.addListener((changes) => {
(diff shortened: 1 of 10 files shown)


Gusted:
> This change is MV3-exclussive since it requires modification to manifest. We could easily extend it to other builds if we wanted.

Feel free to include it to the other builds. If it doesn't use MV3 specific code, it shouldn't be behind that flag. *Also kindly update the CHANGELOG.md respectively*

bershanskiy:
> Feel free to include it to the other builds. If it doesn't use MV3 specific code, it shouldn't be behind that flag.

It changes `manifest.json` which we avoided doing in the past. As far as I know, this should not disable Dark Reader for existing installs on modern-ish browsers (because there is no longer a special `devtools` permission). But CWS might consider it a change worthy of a slow in-depth review.

Gusted:
> It changes `manifest.json` which we avoided doing in the past. As far as I know, this should not disable Dark Reader for existing installs on modern-ish browsers (because there is no longer a special `devtools` permission). But CWS might consider it a change worthy of a slow in-depth review.

Eh, I don't think that will trigger it, we can always revert it(at least we know what caused it). You want @alexanderby approval for it as well.

src/ui/devtools/loader.ts
```diff
@@ -0,0 +1,18 @@
+function handleShown() {
+    console.log('panel is being shown');
+}
+
+function handleHidden() {
+    console.log('panel is being hidden');
+}
```
Gusted, line 7: use `logInfo` for this

Gusted:
Is it possible to "remove" or "hide" the panel, so it won't be displayed in the dev tools. I don't think everyone would like this change.

bershanskiy:
> Is it possible to "remove" or "hide" the panel, so it won't be displayed in the dev tools. I don't think everyone would like this change.

Yes, we can do this. Where should the setting toggle be? Should we just put it on DevTools next to "Reset changes", "Apply" and "Switch to old design"?

Gusted:
> Yes, we can do this. Where should the setting toggle be? Should we just put it on DevTools next to "Reset changes", "Apply" and "Switch to old design"?

Yeah within the DevTools.

src/ui/devtools/loader.ts
```diff
@@ -0,0 +1,14 @@
+async function main() {
+    const DEVTOOLS_PANEL_ENABLED_LOCAL_STORAGE_KEY = 'DevTools_panel_enabled';
+
+    const panelEnabled = (await chrome.storage.local.get(DEVTOOLS_PANEL_ENABLED_LOCAL_STORAGE_KEY)).DEVTOOLS_PANEL_ENABLED_LOCAL_STORAGE_KEY;
```
bershanskiy, line 4: Should DevTools opt-out be `local` or `sync`?

PR #9180, opened 2022-07-03
MV3: 'system' Automation

bershanskiy:


src/background/tab-manager.ts
```diff
@@ -286,6 +292,7 @@ export default class TabManager {
             .forEach((tab) => {
                 const frames = this.tabs[tab.id];
                 Object.entries(frames)
+                    .filter((frame) => Boolean(frame))
```
Gusted, line 295: Seems like a separate PR worth.

src/inject/utils/watch-color-scheme.ts
```diff
@@ -1,20 +1,32 @@
...
+        query.removeListener(onChange);
+    }
+    query = null;
+    onChange = null;
+}
+
+export function isSystemDarkScheme(): boolean {
+    const q = query || matchMedia('(prefers-color-scheme: dark)');
+    return q.matches;
 }
```
Gusted, line 32:
```suggestion
export const isSystemDarkScheme = () => (query || matchMedia('(prefers-color-scheme: dark)')).matches;
```

PR #9187, opened 2022-07-04
MV3: settings UI notifies background of system color scheme

bershanskiy:
In MV3, BG does not have access to system color scheme so we have to rely on regular page contexts to supply this information. We recently landed #9180 which uses the main page CS context to obtain this info whenever needed (during "page connect" and "page resume"). This means 'system' automation will work whenever there is an open tab. However, there is a small edge case: if there is no active tabs, the _icon_ might be incorrect. We can partially fix this by querying this data via extension popup UI page.

This PR attempts to fix the following bug: user opens Chrome with only internal pages (which DR does not have access to), opens settings UI and sets Automation to 'system'. Since there is no CS script running, BG does not get color scheme data and keeps the old icon.

There are few more bugs:
 - user has two devices with synced settings and enables 'system' automation on device A, then device B receives synced setting but does not have an up-to-date system color data

To reiterate, the icon will be fixed as soon as user opens any content tab which Dark Reader has access to.


Gusted:
So now it will also have a check to update the current user's preferred color-scheme when you change the automation settings?

bershanskiy:
> So now it will also have a check to update the current user's preferred color-scheme when you change the automation settings?

Yes, exactly.

src/ui/popup/automation-page/index.tsx
```diff
@@ -128,7 +131,15 @@ export default function AutomationPage(props: ViewProps) {
                         'automation-page__system-dark-mode__button': true,
                         'automation-page__system-dark-mode__button--active': isSystemAutomation,
                     }}
-                    onclick={() => props.actions.changeSettings({automation: isSystemAutomation ? '' : 'system'})}
+                    onclick={() => {
+                        if (isMV3) {
+                            chrome.runtime.sendMessage<Message>({
+                                type: MessageType.CS_COLOR_SCHEME_CHANGE,
```
Gusted, line 137: Use the UI variant for this

bershanskiy: Thanks, that was a typo.

Gusted (approved):

PR #9198, opened 2022-07-05

PyramidSectionPastryVoyage544:


PyramidSectionPastryVoyage544:
Hello, sorry for the lack of detail here. There's a usability issue when the Dynamic theme is applied to purerave.com where the journal comment form is rendered difficult to see / invisible due to the field background and border/box shadow. This PR changes the background to `darkreader-neutral-background` which seems to fix appearance and then I also added a more appropriate colour to the border/box shadow.

PyramidSectionPastryVoyage544:
Also, here're a couple more screenshots for your reference - the element in dynamic light mode:

Screen Shot 2022-07-05 at 15 31 39

And with the extension off:
Screen Shot 2022-07-05 at 15 31 55

src/config/dynamic-theme-fixes.config
```diff
@@ -14346,6 +14346,26 @@ CSS
...
+}
+
+div.form-buttons textarea {
+    background-color: var(--darkreader-neutral-background) !important;
+    border-color: #005a9b !important;
+}
+
+textarea {
+    background-color: var(--darkreader-neutral-background) !important;
+}
```
Gusted, line 14365:
```suggestion
div.form-buttons textarea {
    background-color: var(--darkreader-neutral-background) !important;
    border-color: #005a9b !important;
}
textarea {
    background-color: var(--darkreader-neutral-background) !important;
}
```

src/config/dynamic-theme-fixes.config
```diff
@@ -14346,6 +14346,26 @@ CSS
 
 ================================
 
+purerave.com
+
+CSS
+:root {
+    --border: 1px solid rgba(0, 90, 155, 1) !important;
+    --box-shadow: 0 1px 8px 0 rgba(0, 90, 155, 1) !important;
```
Gusted, line 14354: Those variables should be overwritten by dark reader already(to use the Dark reader modified version) on the elements that uses this variables. Are these modifications necessary?

PyramidSectionPastryVoyage544: You're correct, thank you! I was confused that I needed it too, I must have made a mistake when testing because it works without now.

PyramidSectionPastryVoyage544: Fixed in https://github.com/darkreader/darkreader/pull/9198/commits/a52b1496f736fcb5e17236016e7ea4771d03e629

Gusted (approved):

PyramidSectionPastryVoyage544:
Thank you for maintaining my favourite browser extension!

PR #9222, opened 2022-07-07
Refactor: build JS for every platform individually

bershanskiy:
Use proper rollup plugins/methods for building extension. This will permit rollup to eliminate some dead code from builds. Also, this makes building for a single target faster. There is a speed regression in building for all platforms at once, but this is compensated with new flags which let us pick which exact platforms we want to build for: `--chrome`, `--chrome-mv3`, `--firefox`, `--thunderbird` (choose one or more, choose none to build all).

### DIFF
--- a/tasks/build.js
+++ b/tasks/build.js
@@ -12,6 +12,7 @@ const zip = require('./zip');
 const {runTasks} = require('./task');
 const {log} = require('./utils');
 const {fork} = require('child_process');
+const {PLATFORM} = require('./paths');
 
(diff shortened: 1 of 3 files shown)


bershanskiy:
This PR will enable a follow-up which will make live re-builds faster by automatically rebuilding only for the browser which has an active autoreloader connected.

bershanskiy:
This PR does not modify build output for Chrome, Firefox, and Thunderbird.

PR #9225, opened 2022-07-07
add my sites to the config

karolpeszek:


Gusted:
Hi @karolpeszek!

Thank you for your PR, unfortunately none of these sites are dark-by-default(which is [a requirement](https://github.com/darkreader/darkreader/blob/main/CONTRIBUTING.md#adding-a-website-that-is-already-dark))

PR #9236, opened 2022-07-09
Optimize: reuse Rollup plugins

bershanskiy:
Prior to this PR, build script was constructing a new instance of every Rollup plugin for every JS target. This is very inefficient, since this creates copies of every plugin with almost exact configs for every JS file. These plugin instances were not collected in a timely manner and resulted in memory bloat.

It's hard to measure exact memory usage, since it is runtime dependent, but in testing this PR leads to more than halving of peak memory usage on my system. In particular, prior to this PR node would easily surpass 1000 MB, and after it never reaches 500 MB.

The memory footprint is less important for local builds, but it is blocking #9222 because CI simply can run out of memory.

Also, we could reduce the memory footprint a bit further by marking `bundle` for memory collection before `entry.postBuild` completes.

### DIFF
--- a/tasks/bundle-js.js
+++ b/tasks/bundle-js.js
@@ -133,33 +133,68 @@ const jsEntries = [
     },
 ];
 
+const rollupPluginCache = {};
+
+function getRollupPluginInstance(name, key, create) {
+    if (!rollupPluginCache[name]) {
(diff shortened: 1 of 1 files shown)


bershanskiy:
Of course, this is just a refactoring PR and does not affect build putput.

PR #9241, opened 2022-07-09
add my sites to dark-sites.conf

karolpeszek:


Gusted:
I'm not sure what their intentions are. But this is clearly wrong, feel free to contact me at postmaster@gusted.xyz. But for now banning from the darkreader org in order to stop this spam.

PR #9246, opened 2022-07-10
Refactor build scripts: build and write minimum required number of files

bershanskiy:
This PR fixes the following inefficiencies:
 - `bundle-css` used to re-build all LESS files even when dependencies of only one CSS target were modified. Also, it used to copy output to directories of all build targets, even when building only for only one target.
 - `copy` task used to copy files to directories of all build targets, even when building only for only one target

This PR also fixes MV3 build.


bershanskiy:
The built output does not change after this PR.

PR #9280, opened 2022-07-12
Fix more icons on developer.apple.com

nyuszika7h:
Such as the App ID capability icons.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4817,6 +4817,9 @@ code {
 
 developer.apple.com
 
+INVERT
+.icon
+
 CSS
 .disabled {
     color: ${gray} !important;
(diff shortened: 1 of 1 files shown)


Gusted:
Do you have an example page?

src/config/dynamic-theme-fixes.config
```diff
@@ -4833,6 +4833,9 @@ IGNORE IMAGE ANALYSIS
 .screen-completion
 .screen-swiftui
 
+INVERT
+.icon
```
Gusted, line 4837: Move this to line 4820

nyuszika7h:
https://developer.apple.com/account/resources/identifiers/bundleId/add/bundle but it requires a paid Apple Developer membership ($99/year).

Gusted (approved):

PR #9286, opened 2022-07-13
added cloudways.com button font color override

fischgeek:


src/config/dynamic-theme-fixes.config
```diff
@@ -3665,6 +3665,13 @@ img.mega-menu-logo
 
 ================================
 
+cloudways.com
+
+INVERT
+.md-button.md-primary.md-raised
```
Gusted, line 3671: Where can I find this element?

fischgeek: I think there's one here: [Log in](https://platform.cloudways.com/login?ref_id=web_navbar), but definitely while you're logged in they are the main buttons. The font is a light purple color on a light green background and really hard to make out.

Gusted (approved):

PR #9298, opened 2022-07-15
MV3: add registered CSS proxy injector

bershanskiy:
This is a third method of injecting CSS stylesheet proxy. It is available only for Chromium 102 and above, but in testing it appears to win the race consistently.


bershanskiy:
This PR does not change build output for platforms other than MV3, so it should be safe to merge.

PR #9314, opened 2022-07-16
Move some package-lock.json entries to SHA512

bershanskiy:


bershanskiy:
Help npm use SHA512 with old packages. All other entries will be upgraded automatically when we upgrade dependencies.

Gusted:
which command did you execute to achieve this?

bershanskiy:
Sorry, GitHub lags for me for some reason and does not display comments live.

I used the following script with newly-generated `package-lock.json`:
```js
const p1 = require('./package-lock.json');
const p2 = require('./package-lock-new.json');
const fs = require('fs');

const versions = {};

const addVersion = (name, version, integrity) => {
    if (!versions[name]) {
        versions[name] = {};
    }
    versions[name][version] = integrity;
}

const walk = (packages) => {
    for (const key in packages) {
        const val = packages[key];
        if (val.integrity && val.version) {
            addVersion(key, val.version, val.integrity);
        }
        if (val.dependencies) {
            walk(val.dependencies);
        }
    }
};

walk(p2.packages);

console.log(Object.keys(versions).length);

const replace = (packages) => {
    for (const key in packages) {
        const val = packages[key];
        if (val.version && versions[key] && versions[key][val.version]) {
            val.integrity = versions[key][val.version];
        }
        val.dependencies && replace(val.dependencies);
    }
}

replace(p1.packages);

const n = JSON.stringify(p1, null, 2);
fs.writeFileSync('./new.json', n + '\n');
```

Issue #9300, opened 2022-07-15
[Bug] DarkReader injects(?) HTML code into ProtonMail text field, which is then sent as part of the email


3ship:
# Bug Report

## Website Address
mail.proton.me

## Bug Description
Today I checked two important emails that I sent a week ago and noticed that in my ProtonMail app they appeared as very light grey text on a white background. After opening the mails in a web browser and clicking "Show HTML code", I noticed that the underlying code was a mess filled with references to DarkReader.

An excerpt:
`<div style="font-family: Arial; font-size: 14px; color: rgb(232, 230, 227); --darkreader-inline-color: #d8d4cf;" data-darkreader-inline-color=""><span data-darkreader-inline-color="" style="color: rgb(232, 230, 227); --darkreader-inline-color: #d8d4cf;">`

In another recent mail, that I assume was written while DarkReader was deactivated, it was entirely devoid of any HTML tags, when I clicked "Show HTML code"

## Steps To Reproduce

1. Log into mail.proton.me
2. Create a new message while DarkReader is activated
3. Send the message
4. Deactivate DarkReader and look at the mail in your sent messages
5. Click the three-dot menu and show the mail as HTML code

## Expected Behavior

I expect the add-on to not manipulate text fields.

## Actual Behavior

The add-on manipulated a text field.

## Screenshots

Not sure how to replicate this, without disclosing sensitive info about myself.

## System Information

- OS: EndeavourOS with GNOME 42
- Browser: LibreWolf 102.0.1
- Dark Reader Version: 4.9.52

bershanskiy:
@3ship Thanks for reporting, I will look into this today.

Gusted:
Duplicate of #8908 ~ Dark Reader is faulty but also cannot prevent it.

bershanskiy:
For now, as a quick immediate fix, I recommend disabling Dark Reader on Protomail website. To do this:
 - click on Dark Reader icon next to browser omnibox (URL/search bar) to open popup with DR settings
 - click on "ON" if extension is disabled
 - click on "Site list"
 - click on "Not invert list"
 - add "mail.proton.me" to the list

I can not reproduce this with default settings

bershanskiy:
[...]

PR #9320, opened 2022-07-17
Fix: ignore contenteditable nodes

bershanskiy:
This is one possible fix for #9300. Let's discuss if we like it.


bershanskiy:
I will fix the test if we decide that this is a good way forward.

Gusted:
Problem with fixing edge cases within the inject code. It can break a thousand other legitimate websites. Problem with this site it just copy paste the `innerHTML`, which is okay, but a problem when extensions like DarkReader will inject their own code into it

PR #9321, opened 2022-07-17
Add mail.proton.me as dark

bershanskiy:
This is one possible fix for https://github.com/darkreader/darkreader/issues/9300. Let's discuss if we like it. This site is not dark by default, but it has a very good native dark theme which can be activated pretty easily.


Gusted:
The dark-sites.config shouldn't be abused to fix problems.

PR #9323, opened 2022-07-17
Implement StateManager.addChangeListener() and add tests

bershanskiy:
Previously, StateManager with a particular storage key was usable only within one JS context at a time, so we had to, for example, pass all data to background script which had a "canonical" StateManager instance which would `saveState()` centrally. This is tedious and inefficient, since we had to perform lots of messaging and occasionally we had to resurrect the whole `Extension` class just to change a few values in storage. After this PR, every context is entitled to its own copy of StateManager and all instances of it will automatically sync data with all other StateManagers with the same key. All contexts will be made aware of all data writes as soon as they happen and all contexts will sync up the data in case there is ever a data race (which is a very unlikely case, but still theoretically possible and very unpleasant if it ever happens).

This PR just updates the diagram, adds two StateManager states, ands StateManager.addChangeListener(), and updates tests.

### DIFF
--- a/src/utils/state-manager-impl.ts
+++ b/src/utils/state-manager-impl.ts
@@ -16,43 +16,59 @@ import {PromiseBarrier} from './promise-barrier';
  *  - data saved with the browser always wins (because JS typically has only
  *    default values and to ensure that if the same class exists in multiple
  *    contexts every instance of this class has the same values)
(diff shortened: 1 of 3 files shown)


Gusted (approved):
Overall LGTM

PR #9324, opened 2022-07-17
Upgrade dependencies

bershanskiy:


bershanskiy:
This PR changes a fairly large number of dependencies, but surprisingly the built output does not change at all on my machine. I think this indicates how stable the dependencies are and how extensive tree shaking is.

This should be safe to merge, I delayed all breaking/potentially problematic dependencies for later.

Gusted (approved):
Dark Reader's core part doesn't really depends on external dependencies, so it's really unlikely to get different build results.

bershanskiy:
Yes, but here we upgraded a few packages which participate in bundling process like TS and Rollup. It's nice that they are pretty stable and in general are predictable.

Issue #9349, opened 2022-07-19
[Broken Website] icloud.com/notes showing only light icons of automatic switch light/dark drawings(on apple notes). check the pictures


yazan1107:
# Broken Website Report

## Website Address

icloud.com/notes
## Steps To Reproduce

have a drawn note on iCloud notes using ipad or iphone.
go to icloud.com/notes from your browser
voila you have an icon showing dark drawing on white back ground meanwhile the drawing is white drawing on black background
## Problematic Observation

you have an **icon** showing d**ark drawing on white back ground** meanwhile the drawing is white drawing on blackground instead **it should show icon showing white drawing on blackground(Check the photos for more details)** btw on apple notes drawings changes per light/dark mode accordingly, so should the icons.
## Screenshots

better quality example:
<img width="948" alt="image" src="https://user-images.githubusercontent.com/78726445/179824104-235331d2-1794-4124-aab9-0e91dc0e1853.png">
my drawing emphasizing the problem:
<img width="949" alt="image" src="https://user-images.githubusercontent.com/78726445/179824418-1c75d392-44b4-402c-897a-e5169d08a4f5.png">

## System Information

- OS: windows 11
- Browser:  chromeVersion 103.0.5060.114 (Official Build) (64-bit)
- Dark Reader Version:  4.9.5.2 latest

yazan1107:
@c-nagy check this issue with icloud.com/notes.

direrFlood:
https://github.com/darkreader/darkreader/pull/10216

Issue #9345, opened 2022-07-19
[Broken Website] iCloud.com/reminders white foreground text boxes on black background


yazan1107:
# Broken Website Report

## Website Address

icloud.com/reminders
## Steps To Reproduce

- Go to icloud.com/reminders
- go to any reminder or create a new one
- white foreground on black background

## Problematic Observation

it should be switched to dark theme, as their native app on iphone/ipad.
## Screenshots

broken reminders website
![image](https://user-images.githubusercontent.com/78726445/178321036-c3db4096-8bc1-4b4f-8966-f5809b222ddf.png)

## System Information

- OS:  windows 11 latest
- Browser:  chrome version: Version 103.0.5060.114 (Official Build) (64-bit)
- Dark Reader Version: 4.9.5.2 latest

direrFlood:
https://github.com/darkreader/darkreader/pull/10216

PR #9358, opened 2022-07-20
iCloud.com site fixes

yazan1107:
solved issue #9345 on icloud.com/reminders
solved issue #9349  on icloud.com/notes
experienced another issue preview images/emojis on icloud.com/notes + icloud.com/reminders are converted which should not. tried to fix it i could not i need some help.
For example when having (dark images/emojis/any image) && darkreader dark mode is on -> their colors on the preview are inverted which should not -> i dont know how to fix this issue as it is only one element and not the entire canvas. need some help.
please @c-nagy provide additional information that can help us solve this issue.
screenshots:
Now with this fix preview images color are converted which should not.
image
original image
image
btw only the preview images are converted when clicking the images the images are there.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -9038,6 +9038,8 @@ CSS
 icloud.com
 
 INVERT
+.rm-list-body
+.note-list-item-attachment
 .sc-iujRgT.jxLrRl
 .icloud-text.image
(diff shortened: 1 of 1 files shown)


yazan1107:
> 

yea i experienced a bug. i am not that experienced with websites development.

Gusted (approved):

PR #9373, opened 2022-07-21
Move to ES module imports

bershanskiy:
Use ECMAScript module imports like `import`/`export` instead of CommonJS `require()`.

Note: due to the way ESLint works, this required some non-obvious changes and there is a performance regression apparently caused by a bug in `eslint-plugin-import`. I'm still working on it, but any feedback is welcome.


bershanskiy:
I decided to split this PR into a few smaller and more manageable pieces. First one is #9379.

bershanskiy:
Update: right now, these components still run in CommonJS mode:
 - the top-level `package.json`. This is necessary for maintaining compatibility with existing version published to NPM. Alternatively, instead of publishing this project's `package.json` file directly to NPM, we can create a different minimal one called `build/api/package.json`, save `darkreader.js` as `build/api/darkreader.js`, copy `LICENSE`, `index.d.ts`, and `README.md` and upload from there. This is simple change, but would require change to publishing process.
 - the eslint command which is run via `package.json` and uses plugin called `local` which is literally just one [require](https://github.com/taskworld/eslint-plugin-local/blob/master/index.js) statement: `module.exports = require('../../.eslintplugin')`. This is pretty easy to fix by changing top-level `package.json` to ES modules and publishing a one-liner to NPM which is just ESM equivalent of that.
 - `test:project` which relies on custom tsconfig file with commonjs export. Each file in the folder can get its mode from a different place (top-level `package.json`, file name ending in `.cjs` or `.mjs`, etc.) This issue will go away on its own if we change top-level `package.json` to ES modules. Also, `test:project` is currently broken and has always been broken since we don't use it...

PR #9403, opened 2022-07-24
Refactor: remove migration logic from localStorage to chrome APIs

bershanskiy:
This code was migrating custom DevTools configs from page-centric `localStorage` API to `chrome.storage`. This logic was introduced in #6429 in October 2021, and was run on the first extension startup after update. It's safe to assume that all active installs were updated already, so there should be no harm in removing this code now... On the other hand, there is no harm in keeping this logic around until January 2022 when Chrome removes MV2 support...


bershanskiy:
Superseded by #10555.

PR #9413, opened 2022-07-26
Fix for microsoft.com

NguyenASang:
Before 
image

After
image


NguyenASang:
Close, cuz I can't find any better solution to fix  (will reopen if found)

PR #9417, opened 2022-07-26
fix for strava.com

Myshor:
One more icon-sm fixes as I found problem in other place after last fixes.


Myshor:
Closing as it broke without latest fixes fetched by DR.

PR #9436, opened 2022-07-29
Fix for mail.google.com

Gusted:
- Fix used space progress bar.


haitham506:
Hello,


image
image

Top Controls Buttons are not Visible.

PR #9442, opened 2022-07-31
Make ConfigManager attributes private

bershanskiy:
This change enables ConfigManager to (later) load data only as needed and not during initialization.


bershanskiy:
Obsolete.

PR #9444, opened 2022-07-31
Added site to blocklist

xerosic:
DarkReader breaks the site aesthetics


Gusted:
Not sure what the page should represent, it almost looks to me that it's a WIP page. Which aren't allowed to be included.

PR #9468, opened 2022-08-04
Add stonecss.com

smith558:
https://stonecss.com has dark-by-default


Gusted:
Thank you for your PR, however this seems to be a WIP page, terms page doesn't exist & privacy policy isn't dark-by-default. Not adding to the list.

PR #9479, opened 2022-08-06
Added gianmarco.ga and dl.gianmarco.ga to dark-sites.config

gianmarcogg03:
Added my gianmarco.ga and dl.gianmarco.ga sites since they're already dark by default

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -210,6 +210,7 @@ discuss.noisebridge.info
 disneyplus.com
 distro.tube
 djflame.tech
+dl.gianmarco.ga
 dlive.tv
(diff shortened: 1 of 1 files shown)


src/config/dark-sites.config
```diff
@@ -210,6 +210,7 @@ discuss.noisebridge.info
 disneyplus.com
 distro.tube
 djflame.tech
+dl.gianmarco.ga
```
Gusted, line 213: Subdomains are matched by-default, please remove.

Gusted (changes requested):

Gusted (approved):

PR #9488, opened 2022-08-07
add mechapower sites to darksites config

ddomino007:


ddomino007:
i realised https://github.com/hadialqattan/no-darkreader exists

PR #9511, opened 2022-08-10
Fix for ncontrol.nfon.com

daef:
Fixes the most glaring issues on the control panel

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -13026,6 +13026,16 @@ CSS
 
 ================================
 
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -13026,6 +13026,22 @@ CSS
...
+ncontrol.nfon.com
+
+CSS
+.slider-control,
+.overview > li,
+.button,
+.label,
+.static-label,
+input
+{
```
Gusted, line 13038:
```suggestion
input {
```

PR #9521, opened 2022-08-10
Fix inline background-image when `mask` is disabled

### LINKED ISSUE #9512: [Bug] Wrong darkreader-inline-bgcolor
# Bug Report

## Website Address
https://tiddlywiki.com/

## Bug Description
When editing a new entry, the background color of the edit text box is wrong.

ShareX_4Xv9ZeAWPK

Note: TiddlyWiki is a web program that has been deployed by many people on many sites, so this problem occurs on more than just this one site.

And it was running fine in the previous version (4.9.52), so it is suspected that this is a bug introduced in the new version that is universal.

## Steps To Reproduce
- Go to https://tiddlywiki.com/
- Create a Tiddler.
chrome_9C7jv3p9Ix
- Observe its color.
chrome_T7maiDkLgf

## System Information

- OS: Windows
- Browser: Chrome 104, Firefox 104
- Dark Reader Version: Only 4.9.53

Gusted:
- Make sure that the check includes that the value is not set to `none` and instead has a real value that will conclude that the background is used as a foreground.
- Resolves #9512

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Fix Google calendar.
 - Fix Opera/Vivaldi sidebar's getting modified.
+- Fix incorrect inline background colors when `mask` is explicitly disabled.
 
 ## 4.9.54 (Aug 10, 2022)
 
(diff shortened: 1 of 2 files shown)


FSpark:
All have been updated to 4.9.55, Firefox is working fine, but Edge and Chrome are still the same.

FSpark:
Sadly, still reproducible on Edge and Chrome.

PR #9526, opened 2022-08-10
Fix: run onColorSchemeChange on all builds

bershanskiy:
This fixes the following bug:
 - open browser on a page where extension can not inject scripts
 - set extension to "system" automation
 - change system color scheme and observe that icon did not update (1)
 - open a new tab where extension can inject
 - notice that page has the wrong color (2)

This fixes (2) but does not fix (1) since I believe there is no way to fix it as far as I know.

### DIFF
--- a/src/background/tab-manager.ts
+++ b/src/background/tab-manager.ts
@@ -70,9 +70,7 @@ export default class TabManager {
         chrome.runtime.onMessage.addListener(async (message: Message, sender, sendResponse) => {
             switch (message.type) {
                 case MessageType.CS_FRAME_CONNECT: {
-                    if (__MV3__) {
-                        onColorSchemeChange(message.data.isDark);
-                    }
+                    onColorSchemeChange(message.data.isDark);
                     await this.stateManager.loadState();
                     const reply = (options: ConnectionMessageOptions) => {
                         getConnectionMessage(options).then((message) => message && chrome.tabs.sendMessage<Message>(sender.tab.id, message, {frameId: sender.frameId}));
@@ -127,9 +125,7 @@ export default class TabManager {
                     break;
                 }
                 case MessageType.CS_FRAME_RESUME: {
-                    if (__MV3__) {
-                        onColorSchemeChange(message.data.isDark);
-                    }
+                    onColorSchemeChange(message.data.isDark);
                     await this.stateManager.loadState();
                     const tabId = sender.tab.id;
                     const frameId = sender.frameId;


MahdiNazemi:
@bershanskiy, would you please explain what `__MV3__` is? I am curious to see how removing it could be the potential cause for bugs like #9542 (and maybe #9540 and #9541).

bershanskiy:
> @bershanskiy, would you please explain what `__MV3__` is?

`__MV3__` is a flag for Chromium Manifest V3 build and here it guarded a workaround for Chromium MV3 background context limitation which recently became relevant for Chromium MV2 as well. I plan to do two things:
 - check if we can reinstate this flag (remove this code) on Firefox and Thunderbird (and use this code only on Chromium Manifest V2 and Chromium Manifest V3)
 - add a check in `Extension.onColorSchemeChange` which will avoid extra tab invocations: `this.wasLastColorSchemeDark === isDark`

Either solution will fix Forefox and Thunderbird regression if it was caused by this PR.

bershanskiy:
I implemented both fixes in #9554.

MahdiNazemi:
> > @bershanskiy, would you please explain what `__MV3__` is?
> 
> `__MV3__` is a flag for Chromium Manifest V3 build and here it guarded a workaround for Chromium MV3 background context limitation which recently became relevant for Chromium MV2 as well. I plan to do two things:
> 
>     * check if we can reinstate this flag (remove this code) on Firefox and Thunderbird (and use this code only on Chromium Manifest V2 and Chromium Manifest V3)
> 
>     * add a check in `Extension.onColorSchemeChange` which will avoid extra tab invocations: `this.wasLastColorSchemeDark === isDark`
> 
> 
> Either solution will fix Forefox and Thunderbird regression if it was caused by this PR.

Thank you for your clear explanation. I will pull the latest version and let you know how it goes. Is there any way to test the extension on Firefox without signing it or installing Firefox Developer?

bershanskiy:
> Is there any way to test the extension on Firefox without signing it or installing Firefox Developer?

@MahdiNazemi To test on Firefox you just need a regular Firefox. Just go to `about:debugging`, select "This Firefox", then click "Load temporary addon" and find `manifest.json` file of the extension you wish to download.

Just in case, the full process is this (assuming you have NodeJS and NPM already):
 1. `git clone https://github.com/darkreader/darkreader.git`
 2. cd `darkreader`
 3. `npm install`
 5. `npm run build`
 6. Load extension in Firefox:
    - go to `about:debugging`
    - select "This Firefox"
    - click "Load temporary addon"
    - file picker will open and you need to select `manifest.json` _file_ at `build/release/firefox/manifest.json`
 7. Load extension on Chrome:
    - Go to `chrome://extensions/`
    - Toggle "Developer mode" in the upper right corner
    - In the upper left, three buttons will appear, click on "load unpacked"
    - file picker will open and you need to select `build/release/chrome` _folder_

> Additionally, does `isDark` refer to DarkReader being active in either light or dark mode, or does it only include dark mode?

`Extension.isDark` (also referred to as `this.isDark`) records whether the system color scheme is dark. Some functions have `isDark` argument, it is the new value which will be written to `Extension.isDark`.

> I am curious if these new pieces of code cover cases where the user has automation enabled with the "Toggle dark/light" option.

Not sure what you mean, could you clarify?

PR #9536, opened 2022-08-11
Fixed many dark images and some background color

JJ96869:
Changes:
- Highlighted comment section is changed into dark mode
- Most icons with black outline on dark mode have changed into white color
- Background color on stream page is smoother

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -17337,6 +17337,36 @@ INVERT
 .widget-ihsg-chart img
 .konvajs-content canvas
 .flotcompanychart canvas
+.ant-modal-content img
+.analyst-title img
+.card-calendar-title img
+.calendar-more span
+.save-orderbook img
+.company-navigation-more span
+.dropmenu-more span
+.navmore span
+.stream-more-wrap span
+.stream-more span
+.trading-menu-img
+.securities-input-pass-wrapper img
+.styled__PasswordContainer-w16pvr-2 img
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -17337,6 +17337,36 @@ INVERT
...
+}
+body {
+    background-color: var(--darkreader-neutral-background) !important;
+}
+.highlight {
+    background-color: var(--darkreader-neutral-background) !important;
+}
+.rightcard {
+    background-color: var(--darkreader-neutral-background) !important;
+}
```
Gusted, line 17369: Please merge these selectors together.

src/config/dynamic-theme-fixes.config
```diff
@@ -17337,6 +17337,36 @@ INVERT
...
+.searchstream
+.ssnlogo
+
+CSS
+html {
+    background-color: var(--darkreader-neutral-background) !important;
+}
+body {
+    background-color: var(--darkreader-neutral-background) !important;
+}
```
Gusted, line 17363: Are you sure about this? Darkreader should already do this by default.

Gusted: unfortunately that's a side-effect of Dark Reader. I personally prefer to not allow such fix, as otherwise every other domain that has this problem will have the same fix, while it actually should be fixed within the code instead of custom fixes.

Gusted (approved):

PR #9552, opened 2022-08-12
my.nextdns.io fixes

FadeMind:
https://my.nextdns.io/ `<configid>` /settings

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -12820,6 +12820,7 @@ img[src*="/static/media/sonos"]
 .d-flex > .text-break > img
 .d-flex > .flex-grow-1 > div > img
 img[src*="handshake-logo"]
+.d-flex.align-items-center.flex-wrap a img
 
 CSS
 .text-right[style*="opacity: 0.3"] {
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -12829,6 +12829,9 @@ CSS
 g.rsm-geographies {
     filter: invert(1) hue-rotate(180deg) brightness(130%) !important;
 }
+img[src*="/misc/flags/"] {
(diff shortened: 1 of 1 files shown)


FadeMind:
Don't touch flags in analytics tab

src/config/dynamic-theme-fixes.config
```diff
@@ -12820,6 +12820,7 @@ img[src*="/static/media/sonos"]
 .d-flex > .text-break > img
 .d-flex > .flex-grow-1 > div > img
 img[src*="handshake-logo"]
+.d-flex.align-items-center.flex-wrap a img
```
Gusted, line 12823: What about appending `:not(img[src*="/misc/flags/"] )`?

FadeMind: `:not(img[src*="/misc/flags/"]` breaks icons

Gusted: Which selector is breaking them? Try to fix that selector instead of manually reverting the filter.

FadeMind: You mixed things IMHO.
`.d-flex.align-items-center.flex-wrap a img` is intended to invert directly logos in Web3 section. 

CSS rule is for revert to default generic colors flags. 

Your idea breaks all site (added in new line standalone)

Gusted: > CSS rule is for revert to default generic colors flags.

Is there a filter by the site that is affecting these flags?

FadeMind: Mixing with 
```
.d-md-inline.ml-md-3.mt-1.mt-md-0
span.notranslate
```
invert part of text and breaks things visible. My workaround with CSS hack do this properly, but infact it is ugly.

Gusted (approved):

PR #9639, opened 2022-08-21
Get correct frameId in `sendMessage`

### LINKED ISSUE #9575: [Broken Website] Box Notes -- page stays dark when darkreader is turned off
# Broken Website Box Notes  -- page stays dark when darkreader is turned off

## Website Address

box.com

## Steps To Reproduce

1) go to box.com
2) turn on dark reader
3) log in with username password
4) click into notes
5) turn off dark reader
NOTE: this can be done in reverse i.e not have dark reader on and then turn it on while in notes, and it will still be buggy

## Problematic Observation

If you have dark reader on while you go into notes, it will work correctly. However, if you try to turn it off, not only will it stay dark, the only way to fix it is to reload the page. Similarly, if dark reader gets turned on while in notes, the page will stay light, except for the top bar on the page.

## Screenshots

When dark mode is on when you enter notes then turned off
image
When dark mode is off when you enter notes then turned on
image

## System Information

- OS:  Windows 10
- Browser:  Google Chrome
- Dark Reader Version:  4.9.55

Gusted:
- The current frameId was actually the index of the item within the `frames` array. This patch fixes that by changing the position of `frameId`, so it will get the actually stored frameId variable from the array, instead of the index.
- Resolves #9575

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+- Fix iframes not getting modified when settings were changed.
+
 ## 4.9.56 (Aug 16, 2022)
 
 - Fix browser theme not changing when automation + scheme behavior was enabled.
--- a/src/background/tab-manager.ts
+++ b/src/background/tab-manager.ts
@@ -302,7 +302,8 @@ export default class TabManager {
                 const frames = this.tabs[tab.id];
                 Object.entries(frames)
                     .filter(([, {state}]) => state === DocumentState.ACTIVE || state === DocumentState.PASSIVE)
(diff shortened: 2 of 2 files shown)


bershanskiy:
Wow, it appears that this code was first introduced in #460 merged in 2018... I wonder how we did not find this bug sooner.

Gusted:
FYI I saw this multiple times over the years, but just assumed my weird & unstable setup was the issue. I normally don't act on bugs I encounter unless I see something suspicious in the code(incorrect argument order isn't something you spot)

bershanskiy:
We should add tests for stuff like this. I'll draft one when I have more time.

PR #9653, opened 2022-08-23
Fix for vjudge.net

KerakTelor86:
Background is a grayish noise-like image, need to set it to none for proper dark theming.

Before:
image

After:
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -19887,6 +19887,15 @@ INVERT
 
 ================================
 
+vjudge.net
+
+CSS
+body {
+    background: none !important;
+}
+
+================================
+
 vk.com
 
 CSS


src/config/dynamic-theme-fixes.config
```diff
@@ -19885,6 +19885,22 @@ INVERT
...
+CSS
+body {
+    background: none !important;
+}
+#contest-rank-table td.accepted {
+    background-color: rgb(0, 96, 0);
+}
+#contest-rank-table td.prob.fb {
+    background-color: rgb(0, 72, 0);
+}
```
Gusted, line 19901: These are cosmetic fixes, please remove them.

src/config/dynamic-theme-fixes.config
```diff
@@ -19885,6 +19885,22 @@ INVERT
 .hero-cover
 .hero-cover p
 
+================================
+
+vjudge.net
+
+CSS
+body {
+    background: none !important;
```
Gusted, line 19894:
```suggestion
    background-image: none !important;
```

src/config/dynamic-theme-fixes.config
KerakTelor86: The "darkness" of these two elements are not supposed to be inverted. The convention for most online judges (including ICPC) is darker = earlier. Should I use `INVERT` instead or would that also be considered a cosmetic fix?

Gusted: > The "darkness" of these two elements are not supposed to be inverted. Please see the unaltered scoreboard screenshot for an example of what they should look like before Dark Reader inverts them.

I understand the green one to be fixed, not sure about the red one to be "less noticeable". That really is on the line of cosmetic fixes for me.



> Maybe there's a way I could force Dark Reader to not invert the colors of these two elements without resorting to CSS fixes?

Unfortunately not, you could use:

```css
#contest-rank-table td.accepted {
    background-color: ${lightgreen} !important;
}
#contest-rank-table td.prob.fb {
    background-color: ${darkred} !important;
}

Comment (KerakTelor86) [reply src/config/dynamic-theme-fixes.config:19901]: I didn't change the reds, both CSS overrides are for the green elements. `td.prob.fb` needs to be a darker than `td.accepted`. I guess I'd just need to use `${lightgreen}` and `${darkgreen}` instead of hardcoding the RGB values?

Comment (Gusted) [reply src/config/dynamic-theme-fixes.config:19901]: >  I guess I'd just need to use ${lightgreen} and ${darkgreen} instead of hardcoding the RGB values?

Yes

Comment (KerakTelor86) [reply src/config/dynamic-theme-fixes.config:19901]: Hmm, `${lightgreen}` results in poor contrast with the timestamp and # of tries, is there a list of other colors I can try?

image

Comment (KerakTelor86) [reply src/config/dynamic-theme-fixes.config:19901]: Perhaps this could work? They're sort of "hardcoded" though, if that's what you're trying to avoid.

```css
#contest-rank-table td.accepted {
    background-color: ${#b8ffb8} !important;
}
#contest-rank-table td.prob.fb {
    background-color: ${#c8ffc8} !important;
}
```

image

Gusted: > Perhaps this could work? They're sort of "hardcoded" though, if that's what you're trying to avoid.

If it's using `${...}` it should be fine.

KerakTelor86: Alright, committed. Thank you for the review!

Gusted (approved):

src/config/dynamic-theme-fixes.config
```diff
@@ -19885,6 +19885,22 @@ INVERT
...
+body {
+    background-image: none !important;
+}
+#contest-rank-table td.accepted {
+    background-color: ${#b8ffb8} !important;
+}
+#contest-rank-table td.prob.fb {
+    background-color: ${#c8ffc8} !important;
+}
+
```
Gusted, line 19902:
```suggestion
```

Gusted (approved):

PR #9655, opened 2022-08-23
fix calendar icon not readable in Google Mail

Primajin:
This fixes the calendar icon being not readable when opening a calendar invitation in Google Mail.

Before:
Screenshot 2022-08-23 at 11 10 34

After:
Screenshot 2022-08-23 at 11 10 52

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -11641,6 +11641,9 @@ div[class*="bym"][role="navigation"] {
 ::-webkit-scrollbar {
     background-color: transparent !important
 }
+.aRg, .aRj {
+  color: var(--darkreader-neutral-background);
+}
 
 IGNORE INLINE STYLE
 .at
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -11641,6 +11641,9 @@ div[class*="bym"][role="navigation"] {
 ::-webkit-scrollbar {
     background-color: transparent !important
 }
+.aRg, .aRj {
+  color: var(--darkreader-neutral-background);
```
Gusted, line 11645:
```suggestion
.aRg,
.aRj {
  color: ${white} !important;
```

Primajin: Hmm but it is white before, we don't want it to be white, then it becomes unreadable 🤔

Gusted: `${...}` is Dark Reader specific CSS syntax. Give it a try 😉

Gusted (approved):

PR #9656, opened 2022-08-23
Add lastfmstats to dark-sites

zuranihenry:


Gusted:
Hi @zuranihenry!

Thank you for your contribution, however this domain doesn't seem to be dark-by-default:
image

PR #9657, opened 2022-08-23
Update dark-sites.config

abdurrahmanseyidoglu:
add one more natively dark website


src/config/dark-sites.config
```diff
@@ -992,3 +993,4 @@ zoomquilt.org
 zoomquilt2.com
 ztdp.ca
 zunivers.zerator.com
+
```
Gusted, line 996:
```suggestion
```

PR #9670, opened 2022-08-25
fixed duolingo.com

wlky:
css class names seem to change relatively often, but with setting the #root background color at least tha main part of the website gets darkend

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5983,14 +5983,16 @@ INVERT
 duolingo.com
 
 INVERT
-.Z392z
-._24NNT
+._1kJpR
+._2P12E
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -5983,14 +5983,16 @@ INVERT
...
+._1kJpR
+._2P12E
+._2TTO0
 [data-test="skill-icon"] + div img
 
 CSS
 button[aria-disabled="true"] {
     opacity: 0.3 !important;
 }
+#root{background:var(--darkreader-neutral-background);}
```
Gusted, line 5995:
```suggestion
#root {
    background-color: var(--darkreader-neutral-background) !important;
}
```

PR #9672, opened 2022-08-25
added yaroflasher.com to dark-sites.config

SuperS-1:


Gusted:
Certificates of the site expired. Not adding for now.

Gusted:
> Can this issue be closed?

Yup closing.

> When going through old commits in the backlog, such as this one here, is there an exact time frame (like 30 days) when commits can be or should be closed? (For example, commits that are either non-actionable, when the PR author never returns to fix suggestions or corrections needed on an existing PR, or when someone opens a PR, and you give feedback or ask a question to the PR author, but they never return to respond? I hope some time frame already exists or can be agreed upon to help clean up / close old commits like this one and other examples I mentioned above that literally sit open for months or more than a year.

As for now, it really depends when I have the time to go trough the backlog, but I'd usually wait a liberal 14-21 days before taking any actions, which should be more than enough for even who are busy to make a response.

Gusted:
Feel free to do that, I don't have any permissions to give you write permissions to issue/PRs so you can add labels instead of making comment noise, so please try to minimize the "noise" for now 😅.

Gusted:
> If I don't create a comment to flag you (using the @Gusted method), how do I flag or mark the issue for your attention? Is there another method or way?
>
> When you say "comment noise", do you mean about spamming you with 20 different things in a day? If so, I understand.
>
> I guess my main question is how to do this without creating comments on every single PR to alert you and therefore spamming you?

Just a few per day, it's possible that it can become one big chaos of people suddenly commenting, which is a mess to oversee.

PR #9680, opened 2022-08-27
Use rollup-plugin-typescript2 for debug builds

bershanskiy:
`rollup-plugin-typescript2` is a bit different from `@rollup/plugin-typescript`:
 - `@rollup/plugin-typescript` is faster and uses less memory
 - `@rollup/plugin-typescript` has fewer dependencies
 - `@rollup/plugin-typescript` is officially supported by Rollup/Svelte team
 - `rollup-plugin-typescript2` prints out TypeScript compilation errors

The first three points make `@rollup/plugin-typescript` more suitable for release builds, while `rollup-plugin-typescript2` is more convenient for development (since having a detailed error message is _very_ helpful during development).

Related issues: #7858 (in which `rollup-plugin-typescript2` was replaced with `@rollup/plugin-typescript` due to excessive memory usage) and #9236 (which might have fixed the memory issue?).

### DIFF
--- a/tasks/bundle-js.js
+++ b/tasks/bundle-js.js
@@ -7,6 +7,7 @@ import rollupPluginNodeResolve from '@rollup/plugin-node-resolve';
 import rollupPluginReplace from '@rollup/plugin-replace';
 /** @type {any} */
 import rollupPluginTypescript from '@rollup/plugin-typescript';
+import rollupPluginTypescript2 from 'rollup-plugin-typescript2';
 import typescript from 'typescript';
 import paths from './paths.js';
 import * as reload from './reload.js';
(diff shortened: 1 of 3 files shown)


tasks/bundle-js.js
```diff
@@ -132,8 +134,12 @@ async function bundleJS(/** @type {JSEntry} */entry, platform, debug, watch, log
                     inlineSources: debug ? true : false,
                     noEmitOnError: watch ? false : true,
                     cacheDir: debug ? `${fs.realpathSync(os.tmpdir())}/darkreader_typescript_cache` : undefined,
-                })
-            ),
+                };
+                if (debug) {
+                    config.verbosty = 3;
```
alex-kinokon, line 139: did you mean `verbosity`

PR #9733, opened 2022-09-04
Fixes for Google Photos

GlowingScrewdriver:
On the page to create/edit an album, the text input area for the title of the album is a different shade of grey than the rest of the background. This addresses that issue by making its background transparent.
Before:
2022-09-04-121643_1366x768_scrot
After:
2022-09-04-121620_1366x768_scrot

AFAIK, most other text input fields are `<input>` elements; the only other `<textarea>` that I can find doesn not behave badly with this change (this is the input box for commenting on an album).

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -14653,6 +14653,15 @@ INVERT
 
 ================================
 
+photos.google.com
+
+CSS
+textarea {
+    background-color: #00000000 !important;
+}
+
+================================
+
 phys.nagoya-u.ac.jp
 
 INVERT


src/config/dynamic-theme-fixes.config
```diff
@@ -14653,6 +14653,15 @@ INVERT
 
 ================================
 
+photos.google.com
+
+CSS
+textarea {
+    background-color: #00000000 !important;
```
Gusted, line 14660: Could you try if:
```suggestion
    background-color: var(--darkreader-neutral-background) !important;
```
or
```suggestion
    background-color: ${white} !important;
```
works as well?

GlowingScrewdriver: Neither of them works. `${white}` resolves to `black`, which doesn't match the background.

Weirdly enough, even `var(--darkreader-neutral-background)` doesn't work; it's darker than it should be. I checked the style for the original webpage, and the background color is white; So I can't figure out why this does not work.

Just out of curiosity, is this a matter of coding style and readability? Or is there a performance penalty with using transparency? Or is it something else altogether?

Gusted: > Just out of curiosity, is this a matter of coding style and readability? Or is there a performance penalty with using transparency? Or is it something else altogether?

Overall transparent is a hack and not an real fix.

Gusted:
```suggestion
    background-color: transparent !important;
```

GlowingScrewdriver: Alright. Thanks!

Gusted: For now transparent should be good, but more or less because Google has some funky CSS that is causing this.

Issue #9659, opened 2022-08-23
[Broken Website] web.whatsapp.com


ShalokShalom:
# Broken Website Report

## Website Address

https://web.whatsapp.com/

## Problematic Observation

WhatsApp Web requires a QR code from you. In the last 3 years or so, since I use Darkreader, is this an issue. I appreciate your effort, to make all websites Darkreader compatible. 

Again and again has been tried, to make the site work with its QR codes and Darkreader. 

web.whatsapp.com has, in the past and at the present, again and again and again shown, that it makes the QR code on the site, which is needed to login to the site, completely unreadable. 

**THIS IS NOT POSSIBLE** _WITH WHATS APP_. I and others have reported this several times, and I ran into this already a good dozen times. On every device that I setup for me or my friends, is this an issue and I am sick of it. I honestly suggest, to take Whatsapp and if possible, every page with QR code OUT of the list of compatible sites. Or on the list, who does not change the webcontent. Its tiring.

Thanks

## System Information

- OS:  Linux
- Browser:  Firefox
- Dark Reader Version:  current

ShalokShalom:
I also see a solution, that detects QR codes and does _not_ render dark around and especially **INTO** it (as it does now) and such a fancy feature could provide the best user experience. 😄

Gusted:
I'm not sure if Whatsapp should be used as an argument here. I've it mentioned before, it works as intended and since #2748 I never had any problems with their QR code again (nor anybody reported one since) and it isn't even inverted (or touched) in the first place.

The problem is the white borders around it (the icon is left out of the equation as that's allowed per QR's code specification as part of fault tolerance) which is required for whatsapp's implementation to recognize the QR code.

At best this is an separate issue where the current fix/implementation is messing up with the colors for the QR code.
But Dark Reader cannot touch the QR Code itself and the white borders around it are already hardcoded white. I would be really surprised if the icon is an issue here as that indicates a problem with Whatsapp's qr scanning code.

Gusted:
> I'm not sure if Whatsapp should be used as an argument here.

There are certainly sites I've seen where Dark Reader f*cks them up. However they are all very isolated cases and no generic checker in dark reader would be able to detect them (and it's just ridiculous to spend a lot of CPU time on a edge-case of an edge-case IMO).

ShalokShalom:
Idk if its the browser, or something else. I had issues with it again and again, its really tiresome

Gusted:
[...]

PR #9747, opened 2022-09-04
Fix for whatsapp.com

Gusted:
- Try to restore the QR as much as possible (likely increase stability to scan this QR code?)
- Attempt to fix the issue #9659

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -20580,7 +20580,6 @@ whatsapp.com
 INVERT
 span[data-icon="audio-download"]
 .landing-main .invisible-space > span
-.landing-main div[data-ref] span > svg > path
 div.landing-header > span > svg
 #wafaq_search_input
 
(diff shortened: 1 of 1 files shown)


ShalokShalom:
Thanks a lot, that sounds great!

I will immediately test with all my devices, once I can.

ShalokShalom:
I guess this will take a while, until a new version comes out. 

Is there any way, how I cam get there early? :)

Gusted:
I will merge it for now, so you can use [the synchronization feature](https://github.com/darkreader/darkreader#site-fixes) to fetch the latest fixes.

ShalokShalom:
Thanks 😊 

 Did you consider Fosshost as CDN?

https://fosshost.org/

ShalokShalom:
Wonderful, that fixes that!

I have tried it with just one mobile, and I am confident it works now

Might have been related to the API that you and I were using

I had an Android 10, maybe you are using iOS devices and they are better to read QR codes

thanks a ton :D

Gusted:
> Did you consider Fosshost as CDN?
> 
> https://fosshost.org/

No the syncronization was rather a "hack"/"workaround" for those that needed fixes before a new Dark Reader release. But I'll happily look into other options to make it more modernized.

PR #9768, opened 2022-09-06
Updated dynamic theme for overreacted.io

PyroGenesis:
The background of the `div` was set using an inline style with a CSS variable. I have overridden the CSS variable with the value of the one used by Dark Reader

Test Link: https://overreacted.io/optimized-for-change/

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -14382,7 +14382,7 @@ INVERT
 overreacted.io
 
 CSS
-div[style*="background: var(--bg)"] {
+:root {
     --bg: var(--darkreader-neutral-background) !important;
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -14379,6 +14379,15 @@ INVERT
 
 ================================
 
+overreacted.io
+
+CSS
+div[style*="background: var(--bg)"] {
```
Gusted, line 14385:
```suggestion
:root {
```

No need to specify this variable that deeply.

PyroGenesis: Understood.

Gusted (approved):

PR #9787, opened 2022-09-08
Bump dependencies

bershanskiy:
This does not cause any change in built output (extensions or API build). The following packages were not upgraded because `ts-jest` supports only `jest` version 28: `@types/jest`, `jest`, `jest-extended`.


bershanskiy:
For reference, `ts-jest` version 29 will support `jest` 29 relatively soon, prerelease versions already exists ([changelog](https://github.com/kulshekhar/ts-jest/blob/main/CHANGELOG.md)).

PR #9824, opened 2022-09-14
Add support for the Official Terraria Wiki

RobbyCBennett:
terraria.wiki.gg

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -18714,6 +18714,22 @@ IGNORE INLINE STYLE
 
 ================================
 
+terraria.wiki.gg
+
+CSS
+:root {
+    --darkreader-bg--theme-content-background: transparent;
+    --darkreader-bg--ooui--menuselect-background-color: #111111;
+}
+body, .navbox, #mw-head, #footer, #simpleSearch #searchButton, #simpleSearch #searchButton:hover, #searchInput:hover ~ #searchButton, #searchInput:focus ~ #searchButton, #mw-panel .portal, #mw-panel #p-logo + .portal {
+    background: none;
+}
+.suggestions {
+    background: var(--darkreader-bg--ooui--menuselect-background-color);
+}
+
+================================
+
 tesco.com
 
 CSS


src/config/dynamic-theme-fixes.config
```diff
@@ -18714,6 +18714,22 @@ IGNORE INLINE STYLE
...
+terraria.wiki.gg
+
+CSS
+:root {
+    --darkreader-bg--theme-content-background: transparent;
+    --darkreader-bg--ooui--menuselect-background-color: #111111;
+}
+body, .navbox, #mw-head, #footer, #simpleSearch #searchButton, #simpleSearch #searchButton:hover, #searchInput:hover ~ #searchButton, #searchInput:focus ~ #searchButton, #mw-panel .portal, #mw-panel #p-logo + .portal {
+    background: none;
+}
```
Gusted, line 18726: What does this fix? The backgrounds are not breaking anything for Dark Reader.

src/config/dynamic-theme-fixes.config
```diff
@@ -18714,6 +18714,22 @@ IGNORE INLINE STYLE
...
+.suggestions {
```
Gusted, line 18727: Where can I find this element?

src/config/dynamic-theme-fixes.config
```diff
@@ -18714,6 +18714,22 @@ IGNORE INLINE STYLE
 
 ================================
 
+terraria.wiki.gg
+
+CSS
+:root {
+    --darkreader-bg--theme-content-background: transparent;
+    --darkreader-bg--ooui--menuselect-background-color: #111111;
```
Gusted, line 18722: IMO the generated colors seems to be correct, given the non-modified site's colors.

Gusted (changes requested):
Hi, thank you for your PR. However I currently don't see what this PR is trying to fix on this site, please explain these changes.

RobbyCBennett:
Light mode:
light


Default dark mode (dynamic theme):
dark

The default dark mode is worse than light mode. The background is still white, and the tabs at the top become unreadable (Page, Discussion, Read, View source, and History)

My fix:
modified

src/config/dynamic-theme-fixes.config
RobbyCBennett: See my comment below. This gets rid of the mid brown background colors and images. The mid brown has a value of 35% brightness. It's incredibly bright compared to the rest of the web. This also gets rid of the white background image.

src/config/dynamic-theme-fixes.config
RobbyCBennett: `--darkreader-bg--theme-content-background: transparent;`
This is to get rid of a lot of the mid-brown background colors. Here's an alternative that has the same functionality.

```css
.mw-body, .infocard, .infocard .main-heading .hgroup {
    background: none;
}
```

`--darkreader-bg--ooui--menuselect-background-color`
This change is for the popup in when typing in the search bar to make it match the rest of the site background.

src/config/dynamic-theme-fixes.config
Gusted: Removing them is not the solution in this situation, you are better of by using this CSS rule instead:

```
```

To reduce the brightness of the image.

PR #9829, opened 2022-09-15
Add delete all sites button

FonecaBot2:
Added delete all sites button


src/ui/popup/site-list-page/index.tsx
```diff
@@ -28,6 +29,7 @@ export default function SiteListPage(props: ViewProps) {
                 onChange={onSiteListChange}
             />
             <label class="site-list-page__description">Enter website name and press Enter</label>
+            <RemoveAllButton {...props} />
```
Gusted, line 32: I'm not sure if we should add another button here, it's already quite crowded in my opinion. What about making this a trash icon, which would be added on the top-{left,right} of the site-list?

FonecaBot2: That sounds great but about the exact position, do you think it's okay to add it next to the 'Disable on these websites' label or somewhere else? Maybe bellow it?

Gusted: > do you think it's okay to add it next to the 'Disable on these websites' label or somewhere else? Maybe bellow it?

I really would prefer if this icon will be on the site-list, my idea was roughly this:
image
Whereby the icon would be either on one of the two red spots.

FonecaBot2: I updated the button. Let me know if there's anything I should change c:

Gusted:
Hi @FonecaBot2!

It is getting there, but I was thinking more about something like this:
image

(The button should like be round or have no background)

```diff
diff --git a/src/ui/controls/delete-all-button/style.less b/src/ui/controls/delete-all-button/style.less
index 6c24c5ea..cad843c5 100644
--- a/src/ui/controls/delete-all-button/style.less
+++ b/src/ui/controls/delete-all-button/style.less
@@ -6,4 +6,11 @@
     display: inline-block;
     height: 1rem;
     width: 1rem;
-}
\ No newline at end of file
+}
+
+.delete-all-icon-group {
+    position: absolute;
+    z-index: 1;
+    left: -1em;
+    top: -0.25em;
+}
diff --git a/src/ui/popup/remove-all-sites-button/remove-site-list-button.tsx b/src/ui/popup/remove-all-sites-button/remove-site-list-button.tsx
index f11175df..1352dafe 100644
--- a/src/ui/popup/remove-all-sites-button/remove-site-list-button.tsx
+++ b/src/ui/popup/remove-all-sites-button/remove-site-list-button.tsx
@@ -32,7 +32,7 @@ export default function RemoveAllButton(props: ViewProps) {
     ) : null;
 
     return (
-        <ControlGroup>
+        <ControlGroup class="delete-all-icon-group">
             <ControlGroup.Control>
                 <DeleteAllButton onClick={showDialog}>
                     {dialog}
```

FonecaBot2:
Hi @Gusted!
I updated everything, but not sure if the position is right. Let me know anything.

Gusted:
Hi @FonecaBot2!

Positioning is good. Some small enhancements that could be added in this PR, before it's ready to merge:
- Only show the icon when siteList isn't empty.
- Change the trash icon's color on hover, so it feels more like an interactive button.

Gusted (approved):
LGTM! Many thanks!

Gusted:
Please add a note in [the readme](https://github.com/darkreader/darkreader/blob/main/CHANGELOG.md), then it can be merged.

FonecaBot2:
Done!

PR #9840, opened 2022-09-17
Partial fix non-inverted menu images at inet.se

AnonymusElephant:
Fixed non-inverted menu images in sidemenu and submenues. Could not track down why some images unloads (problem existed before this fix).

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -9722,6 +9722,19 @@ IGNORE INLINE STYLE
 
 ================================
 
+inet.se
+
+INVERT
+.h1yh8nho
+.i1gv55ha
+
+CSS
+:root {
+--svg-icon-default-color: #e8e6e3
+}
+
+================================
+
 infinitysearch.co
 
 INVERT
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -9730,7 +9730,7 @@ INVERT
 
 CSS
 :root {
---svg-icon-default-color: #e8e6e3
+    --svg-icon-default-color: #e8e6e3 !important;
 }
 
 ================================
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -9726,7 +9726,7 @@ inet.se
 
 INVERT
 .h1yh8nho
-.i1gv55ha
+a.item-button span[class*="icon"]
 
 CSS
 :root {


src/config/dynamic-theme-fixes.config
```diff
@@ -9722,6 +9722,19 @@ IGNORE INLINE STYLE
 
 ================================
 
+inet.se
+
+INVERT
+.h1yh8nho
+.i1gv55ha
```
Gusted, line 9729: Why does this element needs to be inverted?

src/config/dynamic-theme-fixes.config
```diff
@@ -9722,6 +9722,19 @@ IGNORE INLINE STYLE
 
 ================================
 
+inet.se
+
+INVERT
+.h1yh8nho
```
Gusted, line 9728: How can I find this element? This seems to be hidden by default.

src/config/dynamic-theme-fixes.config
```diff
@@ -9722,6 +9722,19 @@ IGNORE INLINE STYLE
...
+.i1gv55ha
+
+CSS
+:root {
+--svg-icon-default-color: #e8e6e3
```
Gusted, line 9733:
```suggestion
    --svg-icon-default-color: #e8e6e3 !important;
```

src/config/dynamic-theme-fixes.config
AnonymusElephant: It's the icons for the different categories in menus on the site, that are black if not inverted which makes them difficult to see.

src/config/dynamic-theme-fixes.config
AnonymusElephant: It's the compact menu selector that only appears when site is in a narrow window to back or show/hide main categories. Same issue here as with icons tha it's black by default and hard to see

src/config/dynamic-theme-fixes.config
Gusted:
```suggestion
a.item-button span[class*="icon"]
```

src/config/dynamic-theme-fixes.config
AnonymusElephant: Oh yes! I forgot the "important!" line, thanks!

src/config/dynamic-theme-fixes.config
AnonymusElephant: That works better, thanks! Still find that when turning on darkmode in dark reader some icon in main side menu first loads in correctly to then unload and be replaced by none or broken icon elements (was a problem before making corrections to dark theme)

AnonymusElephant:
Fixed push with sugestions form Gusted

Gusted (approved):

PR #9847, opened 2022-09-17
add masterclass.com, remove my note about guess-css.app

Pdzoc:


Gusted:
Hi @Pdzoc!

Thank you for your PR, however it seems that some subdomains of this site is not dark-by-default(e.g. https://learn.masterclass.com/security, https://www.masterclass.com/articles#quiz-1) and therefor cannot be added to the list.

PR #9849, opened 2022-09-18
Add dark sites

Tiagoquix:
- dbdicontoolbox.com/web - Dark Reader hides the background when it's on
- loldle.net - no effect, already dark
- splatoon.nintendo.com - removes some green colors of some buttons


Gusted (approved):
Thanks!

PR #9867, opened 2022-09-20
Added 2 Sites

torma616:
Added trainblackjack.com and dule.one. Both respect system-set dark mode.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -230,6 +230,7 @@ dsa-drakensang.de
 dsc.gg
 duckychannel.com.tw
 duelingnexus.com
+dule.one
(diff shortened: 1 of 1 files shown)


Gusted:
Hi @torma616!

Thank you for your PR, unfortunately sites must be dark-by-default[^0] in order to become part of this list. 

[^0]: https://github.com/darkreader/darkreader/blob/main/CONTRIBUTING.md#adding-a-website-that-is-already-dark

Gusted:
> is there any list to add these to?

No we don't have a list for that.

Gusted:
> They both look worse with darkreader enabled than with it disabled.

Please open a issue about it then.

PR #9881, opened 2022-09-21
Add support for the Official Terraria Wiki

RobbyCBennett:
This commit only makes three changes: make the header, header link arrows, & search icon more readable.

Restore the header background image which Dark Reader removed

#mw-head

Restore the header link arrows which Dark Reader made hard to see

#mw-panel ...

Restore the search icon which Dark Reader made hard to see

#simpleSearch ...


Gusted (approved):
LGTM! Thanks!

PR #9895, opened 2022-09-22
hp.com fixes

FadeMind:
https://support.hp.com/pl-pl/contact

- fix hp support assistant question mark 
- fix service center icon geo


FadeMind:
HP using prefixes like `https://h30434.www3.hp.com`

Gusted (approved):

Issue #9893, opened 2022-09-22
[Broken Website] ubisoft.com


Tiagoquix:
# Broken Website Report

## Website Address
`https://www.ubisoft.com/en-us/`

## Steps To Reproduce
- Go to ubisoft.com.
- Mouse-over top boxes/buttons ("Play", "Shop" etc.)

## Problematic Observation
White background.

## Screenshots
With Dark Reader enabled:
![image](https://user-images.githubusercontent.com/30274161/191672566-068af995-e2fa-4bea-9367-d4ecebe64856.png)
![image](https://user-images.githubusercontent.com/30274161/191672585-df9be1e2-ac7b-461b-a6fa-888a3098f868.png)

## System Information
- OS: Windows 10 Pro 21H2
- Browser: Firefox 105.0 (64-bit)
- Dark Reader Version: 4.9.57

## Additional Context
No additional details.

Issue #4682, opened 2021-01-06
[Broken Website] www.virustotal.com


JadenJSJ:
**Website**
www.virustotal.com

**How to Reproduce**
An example of this is
> Visit the site
> The site dosen't fully turn dark
> The logo just dissapears

**Expected behavior**
Expected behaviour is the logo existing, and the upload box dark

**Actual behavior**
Logo not there and the upload box being light themed

**Screenshots**
![image](https://user-images.githubusercontent.com/44781632/103733376-34330f80-501c-11eb-85ad-65f49ebf8f2c.png)

**Browser Info**

Google Chrome | 89.0.4379.0 (Official Build) canary (64-bit) (cohort: Clang-64)
Revision | de20bcd3d57dae2e47f65e5323e617273233af40-refs/branch-heads/4379@{#1}
OS | Windows 10 OS Version 2009 (Build 19042.685)
JavaScript | V8 8.9.204

**System info:**

Device name	JadenJSJ-Laptop
Processor	Intel(R) Core(TM) i5-7200U CPU @ 2.50GHz   2.70 GHz
Installed RAM	8.00 GB (7.85 GB usable)
System type	64-bit operating system, x64-based processor
Edition	Windows 10 Home Single Language
Version	20H2
Installed on	‎10/‎28/‎2020
OS build	19042.685
Experience	Windows Feature Experience Pack 120.2212.551.0

**Additional context**
But the weird thing is that the dark theme works time to time then randomly stops working.
![image](https://user-images.githubusercontent.com/44781632/103733652-d0f5ad00-501c-11eb-9c96-2a6101434ad4.png)

AndrewDaGuy:
thats a lot of info you got there 0_o

Gusted:
Note to myself: It might be a browser specific issues, Firefox didn't show any reproduce at all with `layout.css.constructable-stylesheets.enabled` enabled which should give the same web envirement as chrome for the API's virustotal CSS uses.

AndrewDaGuy:
the logo seems to  disappear when the upload part loads

AndrewDaGuy:
More info: it seems to always break when you do a hard reset, going to check on other broswers

rakleed:
[...]

PR #9898, opened 2022-09-23
Fix ubisoft menu and global-navigation

badgerlocke:
Fix issue #9893 
Screenshot from 2022-09-22 20-55-38


Gusted (approved):
Thanks!

Gusted:
The fixes aren't correctly being applied due to the same issue as outlined in #4682

Issue #9892, opened 2022-09-22
[Broken Website] infograficos + arte (estadao.com.br)


Tiagoquix:
# Broken Website Report

## Website Address
```
https://infograficos.estadao.com.br/public/jogos/2017_01_27/crosswords/

https://arte.estadao.com.br/jogos/palavras-cruzadas/?d=2020-02-17
```

## Steps To Reproduce
- Go to "infograficos" or "arte" of "estadao.com.br".
- Click anywhere in the crosswords for the alphabet to appear.
- Click in any letter of the alphabet.

## Problematic Observation
The letters should be black.

## Screenshots
Without Dark Reader:
![image](https://user-images.githubusercontent.com/30274161/191669942-bf1bac80-c37c-4264-9a3b-0170a031bab9.png)

With Dark Reader:
![image](https://user-images.githubusercontent.com/30274161/191669994-1ac6750d-877f-45e3-891c-1e0758291cea.png)

## System Information
- OS: Windows 10 Pro 21H2
- Browser: Firefox 105.0 (64-bit)
- Dark Reader Version: 4.9.57

## Additional Context
The problem happens in both websites.

PR #9899, opened 2022-09-23
Fix estadao.com.br

badgerlocke:
Fix issue #9892

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -6614,7 +6614,8 @@ CSS
 estadao.com.br
 
 INVERT
-#puzzleHolder 
+#puzzleHolder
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -6611,6 +6611,13 @@ CSS
 
 ================================
 
+estadao.com.br
+
+INVERT
+#puzzleHolder
```
Gusted, line 6617: Please add the following suggestion, so the letters won't turn dark.
```suggestion
#puzzleHolder
.puzzleField
```

badgerlocke: Thanks!

Gusted (approved):

PR #9917, opened 2022-09-25
Fix: two race conditions during background startup

bershanskiy:
This fixes two bugs in early BG startup:
 - extension attempted to do some work before loading DevTools and other components
 - DevTools attempted to access ConfigManager data before it was loaded

Possibly related to recent bug when Dark Reader automation did not work right after Firefox update and sometimes on Chrome. (I do not have any means to test this hypothesis.)

Repro steps:
 - install out MV3 build on Chromium (easiest to reproduce), enable DR (just set to ON, no automation)
 - close browser to shut down BG
 - repeat until bug occurs (up to 5 times): open browser and before opening anything else, open Popup (will fail to load) or open example.com (will be white)

### DIFF
--- a/src/background/config-manager.ts
+++ b/src/background/config-manager.ts
@@ -8,6 +8,7 @@ import type {ParsedColorSchemeConfig} from '../utils/colorscheme-parser';
 import {ParseColorSchemeConfig} from '../utils/colorscheme-parser';
 import {logWarn} from './utils/log';
 import {DEFAULT_COLORSCHEME} from '../defaults';
+import UserStorage from './user-storage';
 
 const CONFIG_URLs = {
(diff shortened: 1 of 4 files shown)


bershanskiy:
Note: this commit may add extra latency due to more sync points (more awaits), but I plan to remove some of these during refactoring later today (e.g., popup UI does not need complete config strings, thee is no need to re-parse and then re-stringify configs on every load, etc.)

bershanskiy:
[I implemented the data split](https://github.com/darkreader/darkreader/compare/main...bershanskiy:darkreader:experiment-split-devtoolsdata) but it did not lead to an appreciable speedup. May be, I'll come back to it later.

PR #9926, opened 2022-09-26
[WIP] Run browser tests on CI

bershanskiy:
WIP


bershanskiy:
Looks like this produces flakes: https://github.com/darkreader/darkreader/actions/runs/3129677915/jobs/5079055993

Let's see how often they occur.

PR #9931, opened 2022-09-26
CI: use NodeJS 16 everywhere

bershanskiy:
NodeJS 15 reached EOL on 2021-06-01, NodeJS 14 is in maintenance mode, Node 16 is the current LTS. We should move NodeJS 15 tasks to either 14 or 16. Let's go with 16, since it is the most common version in other tasks.

### DIFF
--- a/.github/workflows/test-browser.yml
+++ b/.github/workflows/test-browser.yml
@@ -73,7 +73,7 @@ jobs:
       - name: Node.js setup
         uses: actions/setup-node@v2
         with:
-          node-version: ${{ matrix.node }}
+          node-version: 16.x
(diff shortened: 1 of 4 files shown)


.github/workflows/test-browser.yml
```diff
@@ -73,7 +73,7 @@ jobs:
       - name: Node.js setup
         uses: actions/setup-node@v2
         with:
-          node-version: ${{ matrix.node }}
```
bershanskiy, line 76: `matrix.node` is not defined in this workflow, it was defaulting to NodeJS 16. So this change is effectively a noop.

PR #9954, opened 2022-09-29
Message about local file access permission

bershanskiy:
When Dark Reader does not have local file access permission, it displays a confusing message that extension is "Enabled for PDF files" or some other local file type but obviously does not work on these files. This commit adds a message "Not allowed to access local files" instead.

Few questions since this is UI change:
 - Does the message text sound good?
 - Should the text be red or green? I made it red to call attention to it.
 - I'm considering making it a clickable which opens the page with the permission toggle. Would it be intuitive enough for users or would it just be noise? E.g., how will they now what to click on the page?

## Before: 
image

## After
image

## UI with granted permission remains unchanged
image

@Gusted @alexanderby


bershanskiy:
CI fails because I added localization only for English, Ukrainian, and Russian. I'll fix it later.

Gusted:
> CI fails because I added localization only for English, Ukrainian, and Russian. I'll fix it later.

TBH translations could be done just for some languages and always fallback to English. Quite annoying currently to add translation strings to the new UI.

src/background/extension.ts
```diff
@@ -336,6 +336,14 @@ export class Extension {
         setShortcut(command, shortcut);
     }
 
+    private static async isAllowedFileSchemeAccess() {
+        try {
+            return await chrome.extension.isAllowedFileSchemeAccess();
+        } catch (e) {
+            return new Promise<boolean>((resolve) => chrome.extension.isAllowedFileSchemeAccess(resolve));
```
Gusted, line 343: I cannot find documentation of why this function should accept a resolve object. Is this due some browser compatibility?

bershanskiy: It's just a bit of syntactic sugar: `chrome.extension.isAllowedFileSchemeAccess()` can be used with callbacks and on some browser versions with promises. This whole function just wraps callback-style call in a Promise to make `collectData()` easier to read.

bershanskiy:
I added other locales using Google Translate, so I will appreciate if anyone can proof-read them. I'll merge this later today if there are no objections raised.

PR #9958, opened 2022-09-29
MV3: remove tabs permission

bershanskiy:
tabs permission generates the following permission message during installation:
"Read your browsing history"

### DIFF
--- a/src/manifest-chrome-mv3.json
+++ b/src/manifest-chrome-mv3.json
@@ -41,7 +41,6 @@
         "alarms",
         "fontSettings",
         "scripting",
-        "tabs",
         "storage",
         "unlimitedStorage"
     ],


Gusted:
> tabs permission generates the following permission message during installation:
> "Read your browsing history"

Are you completely sure this still works? We are using the tabs API in the background script, does MV3 allow for more API usages without permissions?

bershanskiy:
> > tabs permission generates the following permission message during installation:
> > "Read your browsing history"
> 
> Are you completely sure this still works? We are using the tabs API in the background script, does MV3 allow for more API usages without permissions?

Yep, there are three things to consider:
 - despite the confusing naming, `tabs` permission is not required to use `chrome.tabs` API. Instead, it controls a few attributes on `chrome.Tab` objects returned by `chrome.tabs.query()`. The main thing is `Tab.url`, but I refactored that out in #9956. Also, `Tab.url` and similar attributes are also accessible on recent Chrome versions if extension has `activeTab` permission. Technically, we could remove `tabs` permission in MV2 builds as well, but at this time I'm not sure what is the compatibility situation there and I would like to do more testing.
 - We do use frame URLs, but those come from message sender objects. In other words, they do not require `tabs` permission.
 - Later, we might even be able to remove wide-reaching required host permission `*://*/*` and replace it with an optional `*://*/*` host permission and an `activeTab` permission (neither of which generate warnings). But this would require onboarding permission flow and some changes to content script registration of color scheme watcher.

As a side note, Apple Safari automatically provides a decent onboarding flow even making "required" tabs and host permissions optional while Chrome and Firefox have no onboarding flow and are a pain to develop for with minimal permissions. I hope Chrome and Firefox will catch up to Safari at some point.

Gusted:
Alright then you got my LGTM for the MV3 build. 

> but at this time I'm not sure what is the compatibility situation there and I would like to do more testing.

As per usual, changing permissions is always tricky behavior on browsers, never been consistent across versions or vendors.

bershanskiy:
> As per usual, changing permissions is always tricky behavior on browsers, never been consistent across versions or vendors.

That's the point. Also downgrading permissions is trivial (automatic during update install), but obtaining permissions again is tricky (could disable extension during install for existing users). So I don't see much benefit to changing MV2 build manifest. I limited MV3 builds to Chrome 102+ because of API availability, and I think we should make MV2 builds as backwards-compatible as possible while making MV3 build "future-proof".

bershanskiy:
Rebased to ensure this PR is affecting exactly one line in the built output.

Issue #9823, opened 2022-09-14
[Dependency] @rollup/plugin-typescript does not produce good warnings


bershanskiy:
It is possible to make `@rollup/plugin-typescript` print out descriptive TS error warnings by editing `node_modules/@rollup/plugin-typescript/dist/index.js`: inside of `diagnosticToWarning` function inside of `if (diagnostic.file)` bock add `console.log(diagnostic.file.fileName)` to print out file name (and any other relevant data).

Perhaps, this should be fixed in the upstream?

PR #9989, opened 2022-10-03
Fix: increase NodeJS memory limit for all build tasks

bershanskiy:


Gusted:
I find it slightly concerning that this is required, this implies that it becomes practically impossible to do any development on low-end machines.

bershanskiy:
This is needed only when TypeScript compiler in Rollup plugin hits a type error. I believe that Rollup should simply abort compilation once the first type error is found, but it does not. For reference, there is a related #9823. Also, I observed that if I manually patch Rollup plugin, then it generates multiple (sometimes ~10) warnings about the same error. That could indicate an error in the plugin since I would expect it to abort on the first error and not hog lots of memory and then abort. Unfortunately, often it is more convenient to let it hog RAM for 10 seconds and get the error location than to manually search for that error for 10 minutes.

Gusted:
> I believe that Rollup should simply abort compilation once the first type error is found, but it does not.

I'm pretty sure they want to report **all** type errors it found. Which is something you might want to expect from a full-fledged compiler.



> That could indicate an error in the plugin since I would expect it to abort on the first error and not hog lots of memory and then abort. Unfortunately, often it is more convenient to let it hog RAM for 10 seconds and get the error location than to manually search for that error for 10 minutes.

Seems like we should report it upstream and might want to look into [other alternatives](https://github.com/evanw/esbuild) otherwise, that are lighter on memory/CPU.

PR #9990, opened 2022-10-03
Descriptive message in settings import

bershanskiy:
- Add descriptive messages during settings import
- use standard validateSettings() util instead of buggy custom getValidatedObject()


bershanskiy:
These messages will show up on mobile and in full tabs on desktop, but not in the popup on desktop. I'm not sure how to prevent popup from closing to be able to display the message. We could display message in-page or open a dedicated tab instead.

src/ui/controls/message-box/index.tsx
```diff
@@ -20,7 +21,7 @@ export default function MessageBox(props: MessageBoxProps) {
                     <Button class="message-box__button message-box__button-ok" onclick={props.onOK}>
                         OK
                     </Button>
-                    <Button class="message-box__button message-box__button-cancel" onclick={props.onCancel}>
+                    <Button class="message-box__button message-box__button-cancel" onclick={props.onCancel} style={props.hideCancel ? 'display: none' : ''}>
```
Gusted, line 24: This is JSX, we can simply not render this button. We don't need a hacky `display: none` CSS hack.

PR #9996, opened 2022-10-03
Fix for esbuild.github.io

bershanskiy:
Temporary fixes for gradient backgrounds since Dark Reader does not support resolving variables yet.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -6671,6 +6671,15 @@ INVERT
 
 ================================
 
+esbuild.github.io
+
+CSS
+.progress {
+   background: linear-gradient(to right, rgba(255, 255, 255, 0) 0%, #181a1b 25%, #181a1b 75%, rgba(255, 255, 255, 0) 100%);
+}
+
+================================
+
 eshop-switch.com
 
 INVERT
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -18103,6 +18103,12 @@ body,
 .topbar .icon-site-switcher-bubble {
     background-repeat: no-repeat !important;
 }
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -18094,6 +18103,12 @@ body,
 .topbar .icon-site-switcher-bubble {
     background-repeat: no-repeat !important;
 }
+div.h100.bar-lg.p-bs-wrapper:has(a[href="/questions"]) {
```
Gusted, line 18106: FYI don't do multiple sites fixes in a PR, unless indicated in the commit title.

bershanskiy: Yeah, the PR title was accidental (I created a PR and then remembered that I saw a very similar bug elsewhere). Also, a proper fix would be to implement support for lowering expressions with `background` set to `linear-gradient` dependent on some `var`.

Gusted: Unfortunately that's not possible. The browser doesn't show a `var(...)` value for the `background` property. The browsers itself doesn't know how to resolve it at a CSS level.

bershanskiy: stackoverflow.com has the following CSS (for example):
```css
background: linear-gradient(0deg, var(--orange-100) 30%, var(--white))
```
So we might want to lower that and substitute our vars in place of these?

Gusted: > So we might want to lower that and substitute our vars in place of these?

Yeah that actually should be the case, dark reader should be capable of modifying that CSS rule.

wan7teen-ping:

https://youtu.be/aIaqr4eq18g

Pada Sel, 4 Okt 2022, 4:25 AM Anton Bershanskiy ***@***.***>
menulis:

PR #9998, opened 2022-10-03
Fix: hide "System" Automation option in Chromium on Linux

bershanskiy:
Chromium currently fetches color preferences from GTK environments like Gnome, but does not relay them to the page content. Chromium is not even aware of preferences in KDE environments like KDE at all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Fixes a issue where `darkreader-fallback` wasn't removed from the DOM, when Dark Reader finds a `<meta name="darkreader-lock">` element.
 - Be stricter when the user specifies a last slash for a URL in the sitelist.
+- Hide "System" Automation on Chromium on Linux because Chromium on Linux does not support Media Queries.
 
 ## 4.9.58 (Sep 22, 2022)
 
--- a/src/ui/controls/message-box/index.tsx
+++ b/src/ui/controls/message-box/index.tsx
@@ -21,9 +21,11 @@ export default function MessageBox(props: MessageBoxProps) {
                     <Button class="message-box__button message-box__button-ok" onclick={props.onOK}>
                         OK
                     </Button>
-                    <Button class="message-box__button message-box__button-cancel" onclick={props.onCancel} style={props.hideCancel ? 'display: none' : ''}>
-                        Cancel
-                    </Button>
+                    {props.hideCancel ? null :
+                        <Button class="message-box__button message-box__button-cancel" onclick={props.onCancel}>
+                            Cancel
+                        </Button>
+                    }
                 </div>
             </div>
         </Overlay.Portal>
--- a/src/ui/popup/automation-page/index.tsx
+++ b/src/ui/popup/automation-page/index.tsx
@@ -7,6 +7,7 @@ import type {Message} from '../../../definitions';
 import type {ViewProps} from '../types';
 import type {Automation} from 'definitions';
 import {AutomationMode} from '../../../utils/automation';
+import {isChromium, isLinux} from '../../../utils/platform';
 
 declare const __CHROMIUM_MV3__: boolean;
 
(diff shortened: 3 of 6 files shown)


Gusted:
I'm really hesitant to just hide things from the user, as they have caused confusion over and over again in the issue tracker of why machine X does have this and machine Y doesn't have this. Some might even say "But hey, this works for me, so please revert". 

 I'd personally prefer to have a *big* warning label when the user selects this automation option in Linux(so including Firefox) that it this automation mode can be inaccurate.
 
 Also `isLinux` might be true for Android mobile devices(IIRC chromium browsers does this), not sure about Firefox.

bershanskiy:
These are good points.

Regarding warning vs hide option: I agree that a detailed and descriptive warning would be better than a hidden setting. I just decided to hide this option since this UI (and the feature it controls) is not officially launched on desktop yet.

If we show this UI and also add a warning, we need to decide a few things:
1. What should be the design of such a warning? Just a static message underneath the option? Or a modal which opens when user selects this option on an incompatible system? Or both?
2. What should we do when user enables "System" Automation in an environment which supports media queries, then enables settings sync and installs Dark Reader in an environment which does not support media queries?
3. Some people use workarounds like [this one](https://github.com/qcasey/chromium-patches/blob/main/patches/gtk-dark-mode-switch-fix.patch), but it is very rare use case. Should we attempt to document these solutions somewhere?

For reference, currently only Chromium does not support media queries while Firefox and Gnome Web and other browsers on both Gnome and KDE do. So I'm not sure why Chromium is lagging behind so much on such a basic feature on both Gnome and KDE.

Gusted:
> I just decided to hide this option since this UI (and the feature it controls) is not officially launched on desktop yet.

I've been _advocating_ to users to use this UI, so even though I don't know the official numbers, there's quite a user base using this new UI. It's small, but it's not _experimental_ anymore and users will (unfortunately) notice.




> 1. What should be the design of such a warning? Just a static message underneath the option? Or a modal which opens when user selects this option on an incompatible system? Or both?

Static message should be fine, it should be colored. I'd prefer to avoid _intrusive_ modals after selecting a option.

  

> 2\. What should we do when user enables "System" Automation in an environment which supports media queries, then enables settings sync and installs Dark Reader in an environment which does not support media queries?

Is there a environment that doesn't support media queries? I am unaware of that currently.



> 3\. Some people use workarounds like [this one](https://github.com/qcasey/chromium-patches/blob/main/patches/gtk-dark-mode-switch-fix.patch), but it is very rare use case. Should we attempt to document these solutions somewhere?

To be honest, yeah we should.
I've not been a big fan of [the wiki](https://github.com/darkreader/darkreader/wiki) and prefer to be everything a file, in the root directory. We likely want to generalize it a bit and have such file be called "LIMITATIONS.md", so it can be used to document other problems with Dark Reader which aren't trivial to fix (if Dark Reader is able to do anything at al).

bershanskiy:
I started implementing this change, so far I have this change, will finalize a PR tomorrow. Turns out, there are many browsers which pretend to support reporting system color scheme, but they actually do not do anything practical for our use case. Here is the UI prototype:
image

Any feedback is welcome (text, spacing, colors, etc.)

bershanskiy:
> Is there a environment that doesn't support media queries? I am unaware of that currently.

Few examples:
1. Chromium on Linux listens to GTK theme, but applies it only to Chromium's own UI but not Media Queries
2. Chromium on Linux currently does not listen to Qt theme at all, but there has been [a lot of progress lately](https://bugs.chromium.org/p/chromium/issues/detail?id=1317782)
3. Chrome on Android listens to the system theme, but does not relay it to Media Queries. Some forks (e.g., Kiwi and few browsers which do not support extensions) override Media Queries with their own values.

wan7teen-ping:

https://youtu.be/aIaqr4eq18g

Pada Sel, 4 Okt 2022, 6:26 AM Anton Bershanskiy ***@***.***>
menulis:

Gusted:
> Few examples:
> 
>     1. Chromium on Linux listens to GTK theme, but applies it only to Chromium's own UI but not Media Queries
> 
>     2. Chromium on Linux currently does not listen to Qt theme at all, but there has been [a lot of progress lately](https://bugs.chromium.org/p/chromium/issues/detail?id=1317782)
> 
>     3. Chrome on Android listens to the system theme, but does not relay it to Media Queries. Some forks (e.g., Kiwi and few browsers which do not support extensions) override Media Queries with their own values.

Aha I see, so they do support it, but not in the way we expect it.

bershanskiy:
> Aha I see, so they do support it, but not in the way we expect it.

Yes, they act in the most unfortunate way imaginable: they act as if they support it, but the actual value is not accurate and the events are never delivered. I'm not aware of any method of feature detection for these media queries, hence we have to fall back to these silly user agent detection hardly distinguishable from user agent sniffing from early 2000-s.

PR #10007, opened 2022-10-05
MV2: backport CSS Stylesheet proxy injection

bershanskiy:
This backports to MV2 builds CSS Stylesheet proxy injection via `<script src="">` instead of inline script. I believe that the inline injection breaks on Safari causing some sites to have bad theme, but I can not verify that this actually fixes the problem. This is a draft until:
 - we decide if we want to launch this on Chromium, Firefox, and Safari all at once or stretch it out
 - I add a test in a separate PR
 - test for a bit to make sure no regressions appear
 - we discuss this PR since it changes `manifest.json` (the original reason to delay this change to MV3)


wan7teen-ping:

https://youtu.be/aIaqr4eq18g

Pada Rab, 5 Okt 2022, 4:22 PM Anton Bershanskiy ***@***.***>
menulis:

PR #10026, opened 2022-10-07
Fix: settings export in Kiwi browser

bershanskiy:
Kiwi browser has a bug which prevents files from being downloaded from background page: Kiwi initiates file download and requests unique file name from the system, but never actually completes the download.


wan7teen-ping:

https://youtu.be/aIaqr4eq18g

Pada Jum, 7 Okt 2022, 4:34 PM Anton Bershanskiy ***@***.***>
menulis:

PR #10044, opened 2022-10-09
Fix for my.account.sony.com

rakleed:
- Invert background image.

-----

### Steps to reproduce

1. Go to https://store.playstation.com/en-us/pages/latest
2. Click on `Sign in` in the top right corner


Gusted (approved):
LGTM, thanks!

PR #10045, opened 2022-10-09
Fix for fivethirtyeight.com

c-nagy:
Fix text and outline on charts in homepage sidebar and chart forecast pages.


Gusted (approved):
Thanks, LGTM!

PR #10047, opened 2022-10-09
Fix: add workaround for extensions blocking Page Lifecycle API

### LINKED ISSUE #10004: [Bug] Unfocused tabs glitch out on dark reader global toggle
# Bug Report

When I toggle dark reader globally (either manually or via automation), all of my unfocused tabs glitch out and I have to reload them to make them display correctly.

## Website Address
Affects every website

## Steps To Reproduce
1. Have a few tabs open
2. Toggle dark mode (via dark reader's gui) on one tab
3. Go to an unfocused tab

## Expected Behavior
Websites look normal in unfocused tabs.

## Actual Behavior
Websites look broken in unfocused tabs.

## Screenshots
Here's an example of a page that was unfocused when I toggled dark reader:
image

## System Information
- OS: macOS, Linux
- Browser: Firefox 105
- Dark Reader Version: 4.9.58

## Additional Context
This has happened to me for ever on multiple mac and linux computers. I just turned off all automations and never toggled the extension because it is maddening.

bershanskiy:
Fixes #10004. Please see comments in code for explanation.


bershanskiy:
@Gusted This code is not very complicated, but if you have time I would appreciate your feedback on it. This fixes the bug by listening for `focus` event in case `visibilitychange` event is blocked.

Of course, this does not support the optimizations which rely on proper `visibilitychange` delivery. In theory, we could add those optimizations even for users which have these extensions installed in by listening for `pagehide` and `freeze` or [`tabs.onActiveChanged`](https://developer.chrome.com/docs/extensions/reference/tabs/#event-onActiveChanged) or [`tabs.onActivated`](https://developer.chrome.com/docs/extensions/reference/tabs/#event-onActivated), but I don't really see a lot of benefit in this. Lots of complexity for a very few actual users who happen to both block Page Lifecycle API and use Dark Reader. I [guess]timate that at most 10000 users block Page Lifecycle API.

bershanskiy:
Testing with:
 - Disable Page Visibility API Marvin Schopf
   - [CWS](https://chrome.google.com/webstore/detail/disable-page-visibility-a/eecfoibnnhheckhfokpihgefmlnenofb?hl=en) 9000+ users
   - [AMO](https://addons.mozilla.org/en-US/firefox/addon/disable-visibility-api/) 79 users
   - [GitHub](https://github.com/marvinschopf/disable-page-visibility-api)
 - Disable Page Visibility API by Frank Dreßler
   - [AMO](https://addons.mozilla.org/en-US/firefox/addon/disable-page-visibility/) 2,653 users
 - Video Background Play Fix by JanH
    - [AMO](https://addons.mozilla.org/en-US/firefox/addon/video-background-play-fix/) 131,031 users
 - Disable Page Visibility API (Android compatible) by futpib
    - [AMO](https://addons.mozilla.org/en-US/firefox/addon/disable_page_visibility_api/) 9 users

Not tested:
 - https://github.com/wpears/dont

src/inject/dynamic-theme/index.ts
```diff
@@ -318,28 +318,54 @@ function onDOMReady() {
     logWarn(`DOM is ready, but still have styles being loaded.`, loadingStyles);
 }
 
-let documentVisibilityListener: () => void = null;
+/**
+ * The following code contains a workaround for extensions designed to prevent page from knowing when it is hidden
```
Gusted, line 322:
```suggestion
 *  Begin of Page Lifecycle API workarounds
 *
 * The following code contains a workaround for extensions designed to prevent page from knowing when it is hidden
```

Gusted (commented):
So just making sure, we are essentially adding extra event listeners? What if multiple event listeners are being triggered? The callback code would be called multiple times, we likely want a safe-guard for this.

bershanskiy:
> So just making sure, we are essentially adding extra event listeners?

Yes, adding listeners for `pageshow` and `focus` which will be triggered only if `visibilitychange` listeners are blocked. (Since `visibilitychange` listener removes `pageshow` and `focus` listeners.)

> What if multiple event listeners are being triggered? The callback code would be called multiple times, we likely want a safe-guard for this.

The `documentVisibilityCallback` can be invoked only in `onDocumentVisible()` and every invocation of `onDocumentVisible()` resets the `documentVisibilityCallback` (in `stopWatchingForDocumentVisibility()`). Since code is synchronous and JS is single-threaded, this should happen within the same task on event loop. However, for readability and to avoid regressions later during refactors, we can move `documentVisibilityCallback = null` closer to call-site of `documentVisibilityCallback`.

bershanskiy:
@Gusted Thanks for your feedback, I incorporated it. I'll retest this tomorrow just in case and merge unless you have any more comments.

src/inject/dynamic-theme/index.ts
```diff
@@ -318,28 +318,57 @@ function onDOMReady() {
...
+    if (!document.hidden || document.visibilityState !== 'hidden') {
+        onDocumentVisible();
     }
 }
 
+function watchForDocumentVisibility(callback: () => void) {
+    documentVisibilityCallback = callback;
+    document.addEventListener('visibilitychange', documentVisibilityListener);
+    window.addEventListener('pageshow', documentVisibilityListener);
+    window.addEventListener('focus', onDocumentVisible);
```
Gusted, line 359: Why does this call directly `onDocumentVisible`? we likely want to call `stopWatchingForDocumentVisibility` if this happens.

bershanskiy: > Why does this call directly `onDocumentVisible`?

When page has focus, it is by definition in "active" state (user is interacting with the page, which implies page is visible). We can call `documentVisibilityListener` instead, of course and just move the body of `onDocumentVisible` into `documentVisibilityListener`.

> we likely want to call stopWatchingForDocumentVisibility if this happens.
We do, `onDocumentVisible()` calls `stopWatchingForDocumentVisibility()`

Gusted (approved):
Alright LGTM.

PR #10052, opened 2022-10-09
Fix Wikipedia - math (dynamic)

Gitoffthelawn:
Example page: https://en.wikipedia.org/wiki/Schlegel_diagram (first sentence)

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -21512,6 +21512,7 @@ wikipedia.org
 wikiless.org
 
 INVERT
+.mwe-math-element
 .mwe-math-fallback-image-inline
 .mwe-math-fallback-image-display
 .mw-ext-score


Gusted:
Are you using a custom extension for Wikipedia or MathML? This selector was previously removed, see: #6520

Gitoffthelawn:
Hmmm... I'm not using anything special for Wikipedia or MathML.

Gusted:
Alright, could you check which element is rendering that symbol?
It should be `mwe-math-fallback-image-inline`:
image

Gitoffthelawn:
Yes, it's `mwe-math-fallback-image-inline`.

Gusted:
@Gitoffthelawn Alright that's weird, then that element should already be inverted(see my previous screenshot). Does it show for you that the `filter: ....` CSSRule is applied as well?

Gitoffthelawn:
Yes, the element in question has the `filter: ...` CSS rule applied.

That does seem to be the problem, though.  If I disable that CSS rule, `mwe-math-fallback-image-inline` becomes visible.

Gusted:
Alright that's weird. The rendered image should be rendered black. Are you sure you don't have an extension that enables native MathML or use a custom theme for Wikipedia?

Gitoffthelawn:
I'm sure of that, but I'm wondering if there could be a difference caused by one (or more) of the following:
 1. Firefox theme
 2. Firefox settings
 3. OS theme

I have everything set to dark, so I hypothesize that Firefox knows to render those images correctly, but then Dark Reader is inverting them, which makes them almost invisible.  Of course, that's just a hypothesis at this point. ;)

Another thought is that Wikipedia is correctly identifying the browser's dark theme for me, but not for you, and so it is presenting those images correctly for me without any need for Dark Reader to intervene.  (And then Dark Reader intervenes, which causes the issue.)

Gusted:
I checked wikipedia with system's color scheme set to dark. No changes and everything is fine, so the influence of system's color scheme is not at play here. 

> I have everything set to dark, so I hypothesize that Firefox knows to render those images correctly

Don't expect too much from a browser.

> but then Dark Reader is inverting them, which makes them almost invisible. Of course, that's just a hypothesis at this point. ;)

Yeah we are definitely inverting them, but that's because the image should be rendered black, the URL on my side is:
image

> Another thought is that Wikipedia is correctly identifying the browser's dark theme for me, but not for you, and so it is presenting those images correctly for me without any need for Dark Reader to intervene

Easy to check, if you disable Dark Reader does Wikipedia have a dark theme? If not, they don't indentify browser's dark theme.

Gitoffthelawn:
> Easy to check, if you disable Dark Reader does Wikipedia have a dark theme? If not, they don't indentify browser's dark theme.

Here's what I see:

_dr

When I close my eyes, both of them go completely dark, but I think that's how it's supposed to work. ;)  Hey, it's the weekend.

Gusted:
I think this is a problem on your device, the image is rendered white even with Dark Reader disabled (thus on stock Wikipedia). Not sure why, but that isn't the behavior you should expect.

Gitoffthelawn:
I might take a look into this.  It will be interesting to discover what is really going on.

PR #10061, opened 2022-10-10
Fix for drive and mail.google.com

rakleed:
- Invert PDFs and docs.

-----

I checked, this rule only works for `.pdf`, `.docx`, `.doc`. It doesn't work for images because they have `div[role="img"]`, and it doesn't work for tables, presentations, `.txt` and unknown files.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -6031,6 +6031,7 @@ div[data-label="nd"] > div > div > svg > path[fill="#000000"]
 div[role="document"] > div[role="button"] .a-b-c
 div > svg > circle[fill="white"]
 img[src*="empty_state_trash"]
+div[role="dialog"] div[role="document"]
 
 CSS
 span[data-type="spelling"] {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -6031,6 +6031,7 @@ div[data-label="nd"] > div > div > svg > path[fill="#000000"]
 div[role="document"] > div[role="button"] .a-b-c
 div > svg > circle[fill="white"]
 img[src*="empty_state_trash"]
+div[role="document"]
```
Gusted, line 6034: It feels bit too general to include this. Should we include `div[role="dialog"]`?

image

rakleed: I did not see any problems with my rule, but just in case, I clarified the rule on your recommendation.

Gusted: I've never had a great experience with using "general"-looking selectors.

Gusted (approved):

PR #10071, opened 2022-10-11
Update dark-sites.config, added inet.se

hueindahaus:


Gusted:
Hi @hueindahaus,

Thank you for your Pull Request, unfortunately this site isn't dark-by-default:
image

Gusted

hueindahaus:
The site is default dark if prefers-color-scheme is dark. <meta name="darkreader-lock" /> does not seem to do anything for some. Maybe because they have an old version of darkreader?

Gusted:
> The site is default dark if prefers-color-scheme is dark.

Please see [the contribution guide](https://github.com/darkreader/darkreader/blob/main/CONTRIBUTING.md#adding-a-website-that-is-already-dark).




> `<meta name="darkreader-lock" /> `does not seem to do anything for some. Maybe because they have an old version of darkreader?

It doesn't seem they are using Dark Reader.

Gusted:
> Sorry, I'm not trying to be unpolite. The meta tag should be working for everyone using dark reader?

I'm sorry I wasn't aware the site was using `<meta name="darkreader-lock">`. But unfortunately they are misusing it... as those without `(prefers-color-scheme: dark)` will not have a light mode site... 

In any case, adding the site to this list doesn't have much effect.

hueindahaus:
Ok thank you!

PR #10089, opened 2022-10-13
Fix: in Chromium detect PDF viewer via script instead of URL

bershanskiy:
Dark Reader relies on `isPDF(url: string)` function to distinguish PDF files. This function fails if files have an incorrect file extension or lack it entirely or if we did not include a particular URL pattern in `isPDF()`. This PR introduces `detectPDF()` which is run within page context and finds the actual `<embed type="application/pdf">` node. The old code is kept in place as a backup in case Chromium changes the way it wraps PDF files into a viewer.

### DIFF
--- a/src/background/extension.ts
+++ b/src/background/extension.ts
@@ -38,6 +38,7 @@ interface SystemColorState {
     wasLastColorSchemeDark: boolean | null;
 }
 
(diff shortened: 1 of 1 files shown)


bershanskiy:
This PR fixes only commands and not the appearance of popup (that will be a separate PR).

PR #10114, opened 2022-10-16
Refactor: create RuntimeMesseageListener

bershanskiy:
Previously, we used a dedicated port for sending messages which require response. This was not pretty, it was just a workaround for Firefox and Chromium incompatibility when there are multiple message listeners. This PR registers a single "wrapper" event listener which internally calls all real event listeners.

The difference is:
 - when listener completes prior to calling `sendResponse`, Chromium checks whether listener returns truthy or falsy value, and if value is falsy closes messaging port, the actual value is ignored, so `undefined` is considered falsy
 - Firefox considers a Promise to be truthy and if a listener returns a Promise, Firefox awaits that and returns the value Promise resolves to. If listener returns `Promise<undefined>` (which Dark Reader did), Firefox treats it as a message `undefined`.

Having a single listener alleviates Firefox problem. Also, it will probably improve Safari compatibility helping us eventually release slightly more modern/advanced Dark Reader for Safari, if we decide to do that.

### DIFF
--- a/src/background/index.ts
+++ b/src/background/index.ts
@@ -7,6 +7,7 @@ import {makeChromiumHappy} from './make-chromium-happy';
 import DevTools from './devtools';
 import {logInfo} from './utils/log';
 import {sendLog} from './utils/sendLog';
+import Messaging from './utils/messaging';
 
 type TestMessage = {
     type: 'getManifest';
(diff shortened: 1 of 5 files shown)


bershanskiy:
I'll test this tomorrow on MacOS Safari and merge it if it works.

bershanskiy:
I'll postpone this until after the upcoming release because if this commit breaks something, the breakage could be substantial. Also, I'm going to be a bit less time to work on Dark Reader within the next two months.

PR #10124, opened 2022-10-17
Fix jira status style

aokellermann:
found while using widgets on jira cloud dashboard

before
image

after
image


Gusted (approved):
LGTM, thanks!

Issue #7130, opened 2021-10-19
[Broken Website] Side-bar Add-on + twitter.com


takasismomo:
# Broken Website Report

## Website Address

[twitter.com](https://twitter.com)

## Steps To Reproduce

- install addon: [Side View](https://addons.mozilla.org/ja/firefox/addon/side-view/?utm_source=addons.mozilla.org&utm_medium=referral&utm_content=search)
- Go to twitter.com.
- Right click on the tab, Open in sidebar

## Expected Behavior

- Side bar has Same colors as on the tab.

## Actual Behavior

- Lost colors of icons.
- Lost user icons. 

## Screenshots

- Expected Behavior(Dark Reader is disabled on addon manager)
![ホーム _ Twitter — Mozilla Firefox 2021_10_19 16_56_04](https://user-images.githubusercontent.com/41175008/137867593-25487e20-024b-42d9-a21a-e856fb78400b.png)

- Actual Behavior(Dark Reader is enabled and turned off)
![ホーム _ Twitter — Mozilla Firefox 2021_10_19 16_55_10](https://user-images.githubusercontent.com/41175008/137867575-1498f9e4-12e8-483b-ad51-7abf0074b349.png)
## System Information

- OS: Windows10
- Browser: Firefox93.0
- Dark Reader Version: 4.9.39

nivurek:
Try this:

https://github.com/darkreader/darkreader/issues/6882#issuecomment-939128800

marlop352:
This seems to happen with any website I open in the sidebar, regardless of what addon I'm using to show a website on the sidebar, tested with:
https://addons.mozilla.org/en-US/firefox/addon/side-view/
https://addons.mozilla.org/en-US/firefox/addon/tweetdeck-for-sidebar/
https://addons.mozilla.org/en-US/firefox/addon/twitter-as-a-sidebar/
https://addons.mozilla.org/en-US/firefox/addon/open-link-in-sidebar/

Any tests I can do to help debug it?
Any about:config options that could be relevant?

### System Information
- OS: Windows11
- Browser: Firefox97.0
- Dark Reader Version: 4.9.45

( I'm using https://github.com/black7375/Firefox-UI-Fix/ )

Gusted:
Only thing that is reproducible is that it's not getting modified, which isn't the problem that was originally reported:

![image](https://user-images.githubusercontent.com/25481501/195725998-6ae64658-c741-433b-834e-a0a9499ba494.png)

[...]

PR #10130, opened 2022-10-19
Fix: disable Dark Reader in Firefox sidebar

bershanskiy:
This is a temporary fix for Firefox Sidebars, which simply disables Dark Reader in Firefox Sidebar to prevent Dark Reader "fallback" style from breaking tabs. I would like to include this fix in Dark Reader 4.9.59, so it is intentionally designed to be as trivial as possible to avoid regressions. Relevant issues: #7130 and #10105.

This PR does not cause any changes in diff for any built output except for Firefox and Thunderbird.

I would like to include proper support for Firefox and Opera Sidebars in Dark Reader 4.9.50. Possibly, I will also support Vivaldi sidebars. That effort is tracked as #10114.

### DIFF
--- a/src/background/make-firefox-happy.ts
+++ b/src/background/make-firefox-happy.ts
@@ -1,0 +1,12 @@
+import {MessageType} from '../utils/message';
+import type {Message} from '../definitions';
+import {isFirefox} from '../utils/platform';
+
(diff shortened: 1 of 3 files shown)


bershanskiy:
I tested for a bit and it works as expected, merging now. I'll finalize #10114 some time before Dark Reader 4.9.60.

Issue #10126, opened 2022-10-18
[Broken Website] Photopea, arrow on button, almost invisible


duianto:
# Broken Website Report

## Website Address

https://www.photopea.com/

## Steps To Reproduce

- Go to https://www.photopea.com/
- Click on the button `New Project` or choose `File > New...`
- Near the top left of the New Project window, click on the button: `Create`

## Problematic Observation

Near the top middle of the window, below the red `Account` button.
The arrow on the button (tooltip text: "Save selected layers") is almost invisible.

## Screenshots

- Dark reader enabled

![save_selected_layers_after](https://user-images.githubusercontent.com/13420573/196441559-37bb3fbe-366f-4166-900b-949ac89c73ba.png)

- Dark reader disabled

![save_selected_layers_before](https://user-images.githubusercontent.com/13420573/196441618-5300587a-dedf-4777-8073-9e99460421c2.png)

## System Information

OS: Windows 21H2
Browser: Microsoft Edge 106.0.1370.42 (Official build) (64-bit)
Dark Reader Version: 4.9.58

### Possible source
The following css rule shows the arrow on the button when it's unchecked.
- Right-click on the "Save selected layers" button
- Choose `Inspect` from the right-click menu
This opens the Edge DevTools window with this html element selected:
```
/html/body/div[2]/div[1]/div[3]/div[2]/div[2]/div/div[2]/button/svg
```
In the DevTools right panel, on the Styles tab
- Scroll down to the rule:
```
[data-darkreader-inline-fill] {
    fill: var(--darkreader-inline-fill) !important;
}
```
- Uncheck it to see the arrow on the button.

Issue #10125, opened 2022-10-18
[Broken Website] Photopea, invisible close buttons


duianto:
# Broken Website Report

## Website Address

https://www.photopea.com/

## Steps To Reproduce

- Go to https://www.photopea.com/
- Click on the button `New Project` or choose `File > New...`
- Near the top left of the New Project window, click on the button: `Create`

## Problematic Observation

Look at the Image tab: `New Project.psd` near the top left corner under the `Auto-Select` setting.

The right side of the image tab is empty.

Disable dark reader to see that the tab is supposed to have a visible "x" button on the right side.

## Screenshots

- Dark reader enabled

![image_tab_after](https://user-images.githubusercontent.com/13420573/196437349-5afe1e3a-57c1-4d80-acf6-b4f15e819aaf.png)

- Dark reader disabled

![image_tab_before](https://user-images.githubusercontent.com/13420573/196437388-457285e6-41ff-44eb-82f2-64a2243a2b5d.png)

## System Information

- OS: Windows 21H2 
- Browser: Microsoft Edge 106.0.1370.42 (Official build) (64-bit) 
- Dark Reader Version: 4.9.58 

## Additional Context

The same thing is observed in the top right corner of the Photopea preferences window (Edit > Preferences...)

- Dark reader enabled

![preferences_after](https://user-images.githubusercontent.com/13420573/196437416-357a2cb6-bbd4-4dea-b0b4-d0720accfca5.png)

- Dark reader disabled

![preferences_before](https://user-images.githubusercontent.com/13420573/196437457-ac0ef1ac-1ed7-4a89-9b99-13200a3e5855.png)

## Possible source

The issue might be caused by a css variable not being defined:
```
--darkreader-bgimg--icon_cross is not defined
```

### Image tab

The css rule for the Image tab seems to be found like this:  
- Right-click to the right of the image tab  
or to the right of the full screen button at the top of the window.
- Choose Inspect at the bottom of the right-click menu. This opens the browser's DevTools window.
- Click on the top left button in the DevTools window (the mouse over tooltip says: "Select an element in the page to inspect it")  
- Click on the top right side of an image tab. This selects the html element:
```
/html/body/div[2]/div[1]/div[4]/div[2]/div/div[1]/div[1]/span[2]
```

Its first rule is:
```
.block .panelhead div .cross {
    background-image: var(--darkreader-bgimg--icon_cross);
}
```
- Unchecking the rule to show image tabs close button.

[...]

PR #10131, opened 2022-10-19
Fixes for www.photopea.com

### LINKED ISSUE #10127: [Broken Website] Photopea, border color around panels
# Broken Website Report

## Website Address

https://www.photopea.com/

## Steps To Reproduce

Go to https://www.photopea.com/
Click on the button `New Project` or choose `File > New...`
Near the top left of the New Project window, click on the button: `Create`

## Problematic Observation

Look at the right side panels.
The borders around the panels don't really match the UI.
They have the color: #807769 (https://www.colorhexa.com/807769 calls it: `Dark grayish orange`)
It looks like there's a light behind the panels, shining through the cracks.

## Screenshots

- Dark reader enabled

right_panel_borders_after

- Dark reader disabled

right_panel_borders_before

## System Information

- OS: Windows 21H2
- Browser: Microsoft Edge 106.0.1370.42 (Official build) (64-bit)
- Dark Reader Version: 4.9.58

### Possible source

The following css rule restores the default border color when it's unchecked.

- In the right panel.
- Right-click on the border below the top left buttons: Info and Properties
- And choose `Inspect` from the right-click menu
- This opens the Edge DevTools, and selects the html element:
```
/html/body/div[2]/div[1]/div[4]/div[3]/div[1]/div[2]/div[2]
```
In the DevTools right panel, on the Styles tab.
- uncheck the following rule, to show the default darker border color.
```
* {
    border-color: var(--darkreader-border--brdr);
}
```

### Note
The same rule also restores the default border color around a color swatch,
in the Preferences window (Edit > Preferences...)

- Before

preferences before

- After

preferences_default_border_color

bershanskiy:
Fixes #10127, #10126, #10125.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -22451,6 +22451,21 @@ INVERT
 
 ================================
 
+www.photopea.com
+
+CSS
+.block .panelhead div .cross {
+  background-image: var(--icon_cross);
+}
+[data-darkreader-inline-fill] {
+    fill: ${white} !important;
+}
+* {
+    border-color: var(--darkreader-bg--bg-panel) !important;
+}
+
+================================
+
 www.physics.utoronto.ca/~aephraim/aephraim.html
 
 CSS


bershanskiy:
@duianto Thank you for detailed reports, this PR should fix the issues. The fixes should be bundled with the upcoming Dark Reader release 4.9.59. If you would like, I can tell you how to enable these fixes on your current install.

bershanskiy:
@duianto I added www.photopea.com to the global dark list, so you will need to manually enable Dark Reader on the site after update. The fixes will be enabled automatically.

duianto:
>If you would like, I can tell you how to enable these fixes on your current install.

Sure, it would be useful to be able to try to fix stuff manually.

Maybe there's another way, but I tried adding the fix to the Dark Reader Dev tools, dynamic theme editor, and clicked apply.
And it seemed to fix the:
- image tab close button
- arrow on the, save selected layers button
- and the borders around the right side panels

### Issue
But the close button on the Edit > Preferences window is still invisible.
And I noticed that the same issue occur in most or all popup windows,
File
- New...
- Print..
- etc.
Edit
- Fill...
- Stroke...
- Preset Manager...
- Preferences...
- etc.

As mentioned at the bottom of: [[Broken Website] Photopea, invisible close buttons](https://github.com/darkreader/darkreader/issues/10125#top) #10125

The css rule for the preferences close button seems to be:
`.window .whead .cross`

A quick check on a couple of the other pop windows, shows the same rule.

When that rule is added to the fix, then the close button appears on the popup windows.
```diff
www.photopea.com

CSS
-.block .panelhead div .cross {
+.block .panelhead div .cross,
+.window .whead .cross {
  background-image: var(--icon_cross);
}
[data-darkreader-inline-fill] {
    fill: ${white} !important;
}
* {
    border-color: var(--darkreader-bg--bg-panel) !important;
}

================================

```

PR #10153, opened 2022-10-22
Fix searchbar background on OpenStax textbooks

embeddedt:
Before:

image

After:

image

Example URL: https://openstax.org/books/calculus-volume-1/pages/3-7-derivatives-of-inverse-functions


bershanskiy:
@embeddedt Thanks for PR, I saw that this selector can be mobile-search-input and updated the selector accordingly.

PR #10158, opened 2022-10-23
Fixed getImageData warning

alexanderby:
- Use `willReadFrequently` option.

### DIFF
--- a/src/inject/dynamic-theme/image.ts
+++ b/src/inject/dynamic-theme/image.ts
@@ -124,7 +124,7 @@ function analyzeImage(image: HTMLImageElement) {
     context.clearRect(0, 0, width, height);
 
     context.drawImage(image, 0, 0, naturalWidth, naturalHeight, 0, 0, width, height);
-    const imageData = context.getImageData(0, 0, width, height);
+    const imageData = context.getImageData(0, 0, width, height, {willReadFrequently: true} as any);
     const d = imageData.data;
 
     const TRANSPARENT_ALPHA_THRESHOLD = 0.05;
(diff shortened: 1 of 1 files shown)


src/inject/dynamic-theme/image.ts
```diff
@@ -124,7 +124,7 @@ function analyzeImage(image: HTMLImageElement) {
     context.clearRect(0, 0, width, height);
 
     context.drawImage(image, 0, 0, naturalWidth, naturalHeight, 0, 0, width, height);
-    const imageData = context.getImageData(0, 0, width, height);
+    const imageData = context.getImageData(0, 0, width, height, {willReadFrequently: true} as any);
```
Gusted, line 127: You need to add `{willReadFrequently: true}` to the `getContext` function. But this is optimistic, as there aren't that many sites that will likely see a lot of images that need toe be analyzed. Not sure what browsers does with this option, but it shouldn't negatively impact other sites with just a few images.

Gusted (changes requested):

src/inject/dynamic-theme/image.ts
alexanderby: I was seeing this warning on almost every website, including Gmail, Linkedin...

Gusted: Might be me, but I don't visit "the major" sites from day-to-day 😅.

alexanderby: Ah, you are right, it's the wrong place to put the flag 🤦

Gusted (approved):

PR #10171, opened 2022-10-25
Fixes for Wikipedia

GlowingScrewdriver:
Same fixes as PR #3943; selectors needed changing due to a changes in the Wikipedia webpage.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -21816,7 +21816,7 @@ img[alt="audio speaker icon"]
 img[src*="Wiktionary-logo"]
 
 CSS
-.mwe-popups-discreet > svg,
+.mwe-popups-discreet > img,
 div .thumbimage[src$=".png"],
 div .thumbimage img[src$=".png"],
 .image-carousel .image-loaded{


src/config/dynamic-theme-fixes.config
```diff
@@ -21816,7 +21816,7 @@ img[alt="audio speaker icon"]
 img[src*="Wiktionary-logo"]
 
 CSS
-.mwe-popups-discreet > svg,
+.mwe-popups-discreet > img,
```
Gusted, line 21819: This will invert every popup image, that isn't correct behavior.

GlowingScrewdriver: But this isn't an image inversion... All this does is to make the background color of the images in the pop-up box white, so that it matches the behaviour of the original webpage. That way it shouldn't cause any problems as long as the images themselves aren't inverted (and that part is already taken care of).

Gusted: Ah oops, sorry. it looked like this was the invert section.

Gusted (approved):

Issue #10178, opened 2022-10-26
Cannot scroll down on settings since upgrade to 4.9.59


021kw497:
Cannot scroll down on Dark Reader's settings page since installed upgrade to (Dark Reader) V4.9.59.

Some of the settings at the bottom are not visible unless the font size is greatly reduced..  I have poor vision and it's a problem for me.

Running Dark Reader on Firefox 106.1.0 (Build #2015909131) and Android 10, Kernel 4.9.190.   

I'm a newbie on GitHub.  Hope my post is OK.

Thanks.
JD

bershanskiy:
@021kw497 Thanks for this report, I managed to reproduce this on PC (in a wonky way) and just merged a revert for changes which caused this regression. I'm not sure when we will be able to publish a patch/hotfix release to AMO and when it will be approved by Mozilla, so if you need this fix immediately I recommend building extension on GitHub Actions and side-loading it. It's really easy after you do it once, ask me if you need help. I hope to get patch/hotfix release out this week.

PR #10179, opened 2022-10-26
Fix: revert Popup UI CSS width and height changes after v4.9.58

bershanskiy:
This PR reverts Popup CSS changes the following commits:
 - 4c226cfb99b51ffa6172b76b1ca51bdaa502875f
 - 964967d780c031d87bb1d8ce0b28d53c4419e55f
 - 2118ba29fc5f092f1965a7f2785831e90cbe1618

### DIFF
--- a/src/ui/popup/style.less
+++ b/src/ui/popup/style.less
@@ -26,10 +26,6 @@ body {
     padding: @popup-content-padding;
     position: relative;
     width: @popup-content-width;
-    // Prevent popup from entering a feedback loop when it displays scrollbars,
-    // adjusts popup size to be larger, then hides scrollbars, adjusts popup size to be smaller,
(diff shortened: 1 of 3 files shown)


src/ui/popup/style.less
```diff
@@ -26,10 +26,6 @@ body {
     padding: @popup-content-padding;
     position: relative;
     width: @popup-content-width;
-    // Prevent popup from entering a feedback loop when it displays scrollbars,
-    // adjusts popup size to be larger, then hides scrollbars, adjusts popup size to be smaller,
-    // then displays scrollbars, etc.
-    overflow: hidden;
```
bershanskiy, line 32: This line was causing #10178, but I decided to revert all CSS changes which could in theory cause other regressions too.

PR #10184, opened 2022-10-26
make caret more visible in notion.so + remove trailing whitespace

DeWarner:


src/config/dynamic-theme-fixes.config
```diff
@@ -14212,6 +14212,9 @@ CSS
 body {
     background-color: var(--darkreader-neutral-background) !important;
 }
+body * {
```
Gusted, line 14215:
```suggestion
* {
```

src/config/dynamic-theme-fixes.config
```diff
@@ -14212,6 +14212,9 @@ CSS
...
+    caret-color: #898989 !important;
```
Gusted, line 14216: Is it possible to use `${lightgray}`? Or any other fix that uses the following form: `${...}`

DeWarner:
```suggestion
    caret-color: ${gray} !important;
```

PR #10188, opened 2022-10-26
Build: adjust archive timestamps for user time zone

bershanskiy:
Arch build script for Dark Reader [requires `strip-nondeterminism`](https://github.com/archlinux/svntogit-community/blob/packages/dark-reader/trunk/PKGBUILD#L12) to remove timestamps from ZIP outputs:
```
makedepends=('nodejs-lts-gallium' 'npm' 'strip-nondeterminism')
```

The timestamps differ from system to system because `yazl` parameter `mtime` takes in `Date` which includes user system time zone. This commit adds the time offset to counter-balance the time zone and provide accurate timestamp identical to commit timestamp in UTC.

CC @polyzen

### DIFF
--- a/tasks/zip.js
+++ b/tasks/zip.js
@@ -11,14 +11,14 @@ const {getDestDir, PLATFORM} = paths;
  * @param {object} details
  * @returns {Promise<void>}
  */
-function archiveFiles({files, dest, cwd, date}) {
+function archiveFiles({files, dest, cwd, date, mode}) {
     return new Promise((resolve) => {
         const archive = new yazl.ZipFile();
         files.sort();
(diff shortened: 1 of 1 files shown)


bershanskiy:
This is odd, but even after this patch the built archive on my local machine is slightly different from what GitHub Action produce: I see literally one-bit differences somewhere in file index (`A4` vs `B4`) for some files:
image

I'm not sure how to remove these differences as well.

bershanskiy:
After this, my local build output perfectly matches the GitHub Actions output.

polyzen:
Thank you for this!

PR #10190, opened 2022-10-27
Updated dynamic theme fixes for mesenger.com

suitangi:
- Remove random white blocks around the send message button. 
- Removed dark border/padding near some messages with gradient backgrounds 

Before:
before

After:
after


Gusted (approved):
LGTM

PR #10201, opened 2022-10-28
Fix: use correct key codes in command update flow

bershanskiy:
This fixes shortcut update logic (exclusive to Firefox) to handle unusual keyboard layouts like accented languages or non-default key mappings like Dvorak instead of QUERTY.


bershanskiy:
CC @Helmoes

PR #10206, opened 2022-10-30
png optimization

pea-sys:
I used zopfli and reduced the image size.
I used WinMerge when checking the image loss less
same


Gusted (approved):
Alright this LGTM.

PR #10212, opened 2022-10-31
Invert font color of presented answers

Katzenwerfer:
I only tested with the [solver tool](https://www.geogebra.org/solver).

I am not sure if it might break any of the other tools.

By default, the font's color of answers is dark when using dark mode. This should set it to invert into white.


Gusted (approved):
LGTM, thanks!

PR #10216, opened 2022-10-31
Fix for beta.icloud.com

rakleed:
- Invert logo.
- Make background dark.

I added a semi-transparent dark background to keep the image in the background. Or will such a decision greatly affect performance and is it easier to just remove the image in the background?


direrFlood:
https://github.com/darkreader/darkreader/issues/10218 after merging your fix i am still facing this issues: https://github.com/darkreader/darkreader/issues/10218

PR #10271, opened 2022-11-07
Add sites to dark list

FireMasterK:
- kavin.rocks
- nitter.kavin.rocks

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -524,6 +524,7 @@ niebezpiecznik.pl
 nitter.42l.fr
 nitter.dark.fail
 nitter.fdn.fr
+nitter.kavin.rocks
 nitter.net
 nitter.nixnet.services
 nitter.pussthecat.org


src/config/dark-sites.config
```diff
@@ -524,6 +525,7 @@ niebezpiecznik.pl
 nitter.42l.fr
 nitter.dark.fail
 nitter.fdn.fr
+nitter.kavin.rocks
```
Gusted, line 528: Subdomains doesn't have to be listed.

FireMasterK: Why's this? I don't want all my subdomains to be listed (since not all of them are compliant, for being listed globally)
I just want the root of my domain and specific subdomains

Gusted: Because currently matching the subdomain to root domain is the behavior, see #373 #660

FireMasterK: I see, according to https://github.com/darkreader/darkreader/pull/830#issuecomment-439964977, I can't add a rule for this either :/

I guess I will remove the root domain for now.

Gusted: Yeah it's a problem that I still want to solve when I can find enough time and can put some better thought into it, but almost feels like solving an P = NP problem with the amount of feedback that I have gathered by now.

Gusted (approved):

PR #10287, opened 2022-11-11
Fix for appstoreconnect.apple.com

jordanbaird:
Don't invert app icons in analytics


jordanbaird:
I just realized that App Store Connect generates a random selector for each icon, so this fix only works for the specific one I chose. Do not merge.

PR #10298, opened 2022-11-13
Update for codeberg.org

tylerwolf35:
Fix header and footer

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4170,6 +4170,12 @@ main header .header-background-image {
 main header .header-title {
     z-index: 1 !important;
 }
+#navbar {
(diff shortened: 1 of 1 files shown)


Gusted:
Hi @tylerwolf35,

Could you elaborate what this should fix?

tylerwolf35:
> Hi @tylerwolf35,
> 
> 
> 
> Could you elaborate what this should fix? 

The current Codeberg theme does not alter the navbar or footer of the Codeberg website, and the color of these sticks out greatly. Dark Reader only alters the background of the website. This change fixes that, keeping style consistent with the way that Dark Reader behaves on similar websites.

Gusted:
Which theme have you specified? The blue color of the footer/navbar is part of the site's identity(Blue in this case), which should be preserved in the dynamic themes.

PR #10304, opened 2022-11-15
feat: add klikindomaret.com

barraIhsan:
I have a few questions.

1. Do I need to flip the verisign secured logo? (I think some companies don't like their logo color to be changed)
original image
image
inverted image
image

2. How to remove white background on images?
image

3. Do I need to remove the white background from the product page
image
or just the logo?
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -11192,6 +11192,18 @@ INVERT
 
 ================================
 
+klikindomaret.com
+
+INVERT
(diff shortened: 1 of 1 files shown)


Gusted:
> 1. Do I need to flip the verisign secured logo? (I think some companies don't like their logo color to be changed)
>        original image

Yes, companies also don't like if everyone can see their logo instead of being hidden away.

> 2. How to remove white background on images?

That's not really possible, as those white background are within image.

PR #10355, opened 2022-11-21
Update dynamic-theme-fixes.config for GCP

flyinmryan:
Fixing black on black text inputs in the Google Cloud Platform Admin Console

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4528,6 +4528,9 @@ CSS
 .cfc-color-text-secondary {
     color: ${#333} !important;
 }
+input[type=text] {
+    color: white;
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -4528,6 +4528,9 @@ CSS
 .cfc-color-text-secondary {
     color: ${#333} !important;
 }
+input[type=text] {
+    color: white;
```
Gusted, line 4532:
```suggestion
    color: var(--darkreader-neutral-text) !important;
```

MikeAtNavagis: That works.  Thanks.

PR #10357, opened 2022-11-22
Invert for mathematical formulas on WikiData

rappos:
Formulas written certain places (see the examples on https://www.wikidata.org/wiki/Property:P2534 ), are black on a dark background. This fix should invert it correctly.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -21929,6 +21929,7 @@ wikidata.org
 INVERT
 .wd-mp-headerimage
 .mwe-math-fallback-image-inline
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -21928,6 +21928,7 @@ wikidata.org
 
 INVERT
 .wd-mp-headerimage
+.mwe-math-fallback-image-inline
```
Gusted, line 21931:
```suggestion
.mwe-math-fallback-image-inline
.mwe-math-fallback-image-display
```

Gusted (approved):

PR #10364, opened 2022-11-23

asteele0:
When editing a page on Confluence, assigning custom text color is overwritten because--for some reason--DarkReader creates a `--darkreader-text` override of `--custom-text-color`, but it's not defined. 

This is just defining the override back to the original custom text color, which allows manual text color changes to work again.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1910,6 +1910,10 @@ div.zgsxji-0.jMIyWM > div > div > div > div {
 .jira-issue-status-lozenge {
     background-color: var(--darkreader-neutral-background) !important;
 }
+.fabric-text-color-mark {
+ --darkreader-text--custom-text-color: var(--custom-text-color);
+}
+
 
 IGNORE INLINE STYLE
 .ak-editor-content-area *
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -1910,6 +1910,10 @@ div.zgsxji-0.jMIyWM > div > div > div > div {
 .jira-issue-status-lozenge {
     background-color: var(--darkreader-neutral-background) !important;
 }
+.fabric-text-color-mark {
+ --darkreader-text--custom-text-color: var(--custom-text-color);
```
Gusted, line 1914:
```suggestion
    --darkreader-text--custom-text-color: var(--custom-text-color);
```

src/config/dynamic-theme-fixes.config
```diff
@@ -1910,6 +1910,10 @@ div.zgsxji-0.jMIyWM > div > div > div > div {
 .jira-issue-status-lozenge {
     background-color: var(--darkreader-neutral-background) !important;
 }
+.fabric-text-color-mark {
+    --darkreader-text--custom-text-color: var(--custom-text-color);
+}
+
```
Gusted, line 1916:
```suggestion
```

PR #10391, opened 2022-11-27
Add leafly.com w/ TODO list

progenitor-amborella:
IMPLEMENTED:
- Made the "level bars" on strain pages visible.
- Made cookies notice visible.

TODO:
- [1] On any strain page: There are a handful of w3.org SVGs that I didn't know the solution for however. A couple until you scroll down to the "effects" section, [2] where there are nine grouped up together. 
- [3] Near the bottom of a strains page, there's a bit where lines are drawn that are not visible in dark mode, however on a different strain page there can be varying amounts of lines. 
- [4] The search filters are difficult to read in dark mode. Interestingly, on rare occasions it works fine. I don't know how to reproduce this in either direction, as this is only the second time I've seen it as readable in a few days.
- [5] Social media buttons (found on articles) are difficult to read in dark mode. Interestingly, "copy link" does not follow this explicitly.

I've been looking for a solution to all of these, but my programming skills are not very fleshed out. If someone does have the solution for these, I would be interested in learning about it!

[1]
image
image
image
image
[2]
image
image
[3]
image
image
image
[4]
image
image
image

[5]
image
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -11563,6 +11563,18 @@ CSS
 
 ================================
 
+leafly.com
+
+CSS
+.bg-light-grey {
+    background-color: var(--darkreader-neutral-background) !important;
+}
+.bg-default {
+    background-color: var(--darkreader-neutral-text) !important;
+}
+
+================================
+
 leagueoflegends.com
 
 INVERT
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -11565,6 +11565,9 @@ CSS
 
 leafly.com
 
+INVERT
+div.pl-lg
+
 CSS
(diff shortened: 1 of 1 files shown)


Gusted:
Hi @nitrousgranola,

Do you have a link for those TODO problems? I can take a look at it.

nitrousgranola:
When you say a link to them, do you mean the pages on which the issues are encountered? If so, 

Here is a strain page: https://leafly.com/strains/clementine

Here is an article: https://www.leafly.com/news/health/how-to-use-cbd-for-anxiety

As for the search function, just put anything into the search bar up top on that website.

Gusted:
Seems like this could help *partially*

```css

INVERT
a[href^="/strains"] .absolute svg

IGNORE INLINE STYLE
button[data-testid="strain-quiz-cta"] path
```

PR #10407, opened 2022-11-29

Error-Eric:
Fixed the bugs that (1) the colour of the check following a user's name may be misrepresented, (2) the colour of some records may be misrepresented.


Gusted (approved):
LGTM, thanks!

PR #10416, opened 2022-11-30
Fixes for rewards.microsoft.com and rewards.bing.com

INeedTogepi:
Added support for rewards.microsoft.com and rewards.bing.com. Right now rewards.microsoft.com redirects to rewards.bing.com, but they are the same site, Microsoft just changed the URL to bing in the last day or so. The IGNORE INLINE STYLE is there to make sure the various svg icons and animations they have on the page stay their original color, since they can look weird when modified (see the example images below).
Before:
before path change
After:
after path change

 The INVERT is to make some text remain dark in so that it can be read in front of a bright image behind it (see the example images below).
Before:
before text change
After:
after text change

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -17053,6 +17053,26 @@ path[d^="M121.865 50.29c0"]
 
 ================================
 
+rewards.bing.com
+
+IGNORE INLINE STYLE
+path
+INVERT
+#promo-item > section > div > div > h1
+#promo-item > section > div > div > p
+
+================================
+
+rewards.microsoft.com
+
+IGNORE INLINE STYLE
+path
+INVERT
+#promo-item > section > div > div > h1
+#promo-item > section > div > div > p
+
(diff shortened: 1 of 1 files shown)


INeedTogepi:
I apologize for the workflow thing. I'm a beginner when it comes to collaborating on Github, and I didn't know that what I clicked on was something that was going to do an action for the repository. I thought it would just show me if it passed the config check you talked about in the contributing page.

INeedTogepi:
So I discovered an issue with the INVERT rule I added. I thought that it would always make the text the same color as the original color when darkreader is disabled, though it seems like that's not the case:

Light text when darkreader is disabled:
Light text when darkreader is disabled

Dark text when darkreader is enabled:
dark text when darkreader is enabled


The goal would be for those specific selectors I used to be unmodified and remain their original color. Also, it doesn't seem like IGNORE INLINE STYLE does this, since I tried putting the selectors under that when the original text color was dark, and that made the text light when darkreader was enabled.

PR #10426, opened 2022-12-02
Add missing check for ownerNode in proxyDocumentStyleSheets

TomasRiker:
In the function `proxyDocumentStyleSheets`, there was one place where it wasn't checked if a stylesheet actually has an `ownerNode` before trying to access its `classList`. At all other places, there is a check. The missing check has been added. I've been observing JavaScript errors originating from the missing check. This pull request fixes issue #10427.


bershanskiy:
@TomasRiker Thanks for your fix, I adjusted types just a bit and merged #10434.

PR #10446, opened 2022-12-07
Fix for transum.org

Matr1x-101:
Fixes https://www.transum.org/software/Online_Exercise/Circle_Theorems/Default.asp?Level=2 and various other excersises

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -20485,6 +20485,13 @@ INVERT
 
 ================================
 
+transum.org
+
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -20485,6 +20485,13 @@ INVERT
 
 ================================
 
+transum.org
+
+INVERT
+img[src*=Level]
```
Gusted, line 20491:
```suggestion
img[src*="Level"]
```

Issue #7946, opened 2022-01-13
[Broken Website] Duolingo characters broken


noxan-dev:
# Broken Website Report

## Website Address

Duolingo (Japanese if that matters)

## Steps To Reproduce

- Go to Duolingo
- Select Japanese
- Play until complete a lesson
- It May take a bit to get the faces that it affects, the endings are random

Here is an example:
- Go to example.com.
- Hover over the example button.
- Notice that when hovering over the example button, it does not change color.
-->

## Expected Behavior

Faces to be light

## Actual Behavior

Faces get darkened

## Screenshots

face in dark mode
![image](https://user-images.githubusercontent.com/59417659/149383133-37eee333-1477-443a-86cb-2e10684a09c7.png)

faces in light mode
![image](https://user-images.githubusercontent.com/59417659/149383449-59f49f23-266b-4f3f-abc1-85bb3def3f24.png)

## System Information

- OS:  Windows 10
- Browser:  Opera 82.0.4227.43
- Dark Reader Version:  4.9.34. I copied the updated config files from the more recent updates. I cannot update due to the sidebar on my browser injecting dark mode into Twitter, Instagram, and the rest. except for WhatsApp and Messenger.

Gusted:
We don't support outdated versions. Please use the latest(next version when opera fix is included).

noxan-dev:
> We don't support outdated versions. Please use the latest(next version when opera fix is included).

ok, I can do that. I also tested on Edge using the latest version and it's the same effect

Gooberpatrol66:
Could this be reopened? I'm getting this on the latest version (4.9.57)

Gusted:
@Gooberpatrol66 Please open a new issue, this had invalid information.

PR #10450, opened 2022-12-07
Update dynamic theme for Duolingo.com

genderquery:
* Remove old rules that no longer apply to new site
I couldn't find the existing class names on the new site, so I removed them.

* Correct the background color for skill path icons
I'm not sure the approach I used is the best, but it seems to work. 

* Do not apply inline styles to SVGs
The `svg *` selector might be too liberal, but it doesn't seem to cause any problems.

* Resolves issue #7946

There are some other outstanding minor cosmetic and readability issues, but they are difficult to develop a long-term fix for because the class names are auto-generated on each new build of the website.

I only tested this with Firefox 107.0.1 on Windows 10 with Dark Reader version 4.9.60. I also checked these changes didn't break anything in the Light Mode.

I'm new to this, so any feedback is welcome!

Before: 
Screenshot 2022-12-07 153858
Screenshot 2022-12-07 150326

After:
Screenshot 2022-12-07 153925
Screenshot 2022-12-07 150409

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -6370,19 +6370,16 @@ INVERT
 
 duolingo.com
 
-INVERT
-.Z392z
-._24NNT
-[data-test="skill-icon"] + div img
-
 CSS
 button[aria-disabled="true"] {
     opacity: 0.3 !important;
 }
+[data-test^="skill-path"] {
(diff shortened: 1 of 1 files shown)


genderquery:
Whoops, I shouldn't have removed
```css
button[aria-disabled="true"] {
    opacity: 0.3 !important;
}
```
Sorry. I'll try to fix it.

PR #10466, opened 2022-12-12
Fix `flaggenlexikon.de`

Myzel394:


src/config/dynamic-theme-fixes.config
```diff
@@ -7630,6 +7630,14 @@ CSS
 
 ================================
 
+flaggenlexikon.de
+
+INVERT
+body
+center
```
Gusted, line 7637:
```suggestion
CSS
body {
    background-image: none !important;
}
```

Myzel394: improved fix

Myzel394:
`img[src="https://www.flaggenlexikon.de/pix/flaggenlexikon.gif"]` is inverted because it contains no background and so the text wouldn't be readable if it was shown normally

Gusted (approved):

PR #10470, opened 2022-12-12
Add dark site: eveonline.com

steven-noorbergen:
As per the checklist:

- The site is dark-by-default, regardless of system's preferred color scheme
- There are no redirects
- The website is complete and finished.

Not 100% sure if subdomains need adding as well, but we use the following subdomains, each of which also uses the same dark-by-default design:

- www.eveonline.com
- secure.eveonline.com
- universe.eveonline.com
- community.eveonline.com
- forums.eveonline.com

Looking at dark-sites.config I am assuming that matching is done on a partial domain basis. If not, please let me know and I'll add the other subdomains as well.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -266,6 +266,7 @@ esportal.com
 essentialsp.tk
 etke.cc
 etlegacy.com
+eveonline.com
 evewho.com
 evokemusic.ai
 evowars.io


Gusted:
Hi @steven-noorbergen,

Thank you for your pull request and efforts, unfortunately currently all sub-domains needs to be dark as well and https://support.eveonline.com/ is currently not dark by default.

steven-noorbergen:
Hi @Gusted 

Would a separate PR specifying the subdomains that _are_ dark-by-default work then?

Gusted:
That's fine!

steven-noorbergen:
Opened #10483 as discussed.

PR #10491, opened 2022-12-15
Fix for guildwars2.com

brybrant:
Fixed the header + navigation ; it was too bright

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -9342,6 +9342,27 @@ CSS
 
 ================================
 
+guildwars2.com
+
+CSS
+body {
+    background-image: none !important;
+}
+body:before {
+    width: 100% !important;
+}
+.main-hd .sub-menu {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -9342,6 +9342,27 @@ CSS
...
 
+guildwars2.com
+
+CSS
+body {
+    background-image: none !important;
+}
+body:before {
+    width: 100% !important;
+}
```
Gusted, line 9353: I'm not sure about this particular change, seems like this isn't necessary. It doesn't obstruct any text.

brybrant: Yes, it's not necessary but that background image on <body> is very bright for dark mode

Gusted: Then we should remove the background image.

brybrant: Do I need to make any changes to this pr?

PR #10492, opened 2022-12-15
Add dark sites

sauce420:


src/config/dark-sites.config
```diff
@@ -413,6 +415,7 @@ jaxcore.app
 jbzd.com.pl
 jekyllrb.com
 jellymar.io
+jeremypetch.com
```
Gusted, line 418: This page is still work in progress, please remove it. WIP pages aren't allowed on the list.
```suggestion
```

src/config/dark-sites.config
```diff
@@ -8,7 +8,8 @@
 2bqueue.info
 314n.org
 5stardata.info
-9anime.id
+9anime.gs
+9anime.pl
```
Gusted, line 12: So many changes for 9anime the last few year... I'm tempted to exclude them from this list as I'm not keeping up with this.

Gusted (commented):
Thanks!

PR #10545, opened 2022-12-30
Second update for petco.com

hjpoe:
Changes text color in moving banner on [petco.com](https://www.petco.com/shop/en/petcostore) to be visible


hjpoe:
I'm not 100% happy with this fix, because I think someone who knows more magic could do it better, but it's better than the text being barely readable.

PR #10546, opened 2022-12-30
Update to logo of petco.com

hjpoe:
For some reason the previous fix for [petco.com](https://www.petco.com/shop/en/petcostore) no longer works. This works, but similar to my previous pull request, someone else might be able to work out something cleaner.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -15642,7 +15642,7 @@ img[src*="logo"]
 petco.com/shop/en/petcostore
 
 INVERT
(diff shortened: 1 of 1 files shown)


Gusted:
Could you please check if this is correct? The change has no added benefit.

hjpoe:
They seem to update the website so often that it would be impossible to try to keep up with them. Perhaps there's some other way to fix the logo so it will do the thing even if they update the website? 

<sub>(I'm sorry this may not be so coherent, I currently have a headache, and I'm trying to distract myself from it by doing stuff)</sub>

PR #10547, opened 2022-12-30
Fix for icloud.com

rakleed:
- Invert logos.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -9897,6 +9897,8 @@ INVERT
 #cloudos-loading-spinner
 .editor-container
 .light > svg
+.apple-icloud-logo > g > path:first-child
+.apple-icloud-logo.dark > g > path:last-child
 
 CSS
 .cw-pane-container,


akirataguchi115 (changes requested):
I think you could just say
`.icloud-logo`
also these would be a nice addition
```
.tags-cloud-toggle
.tt-input-field
.network-badge-symbol
```

akirataguchi115:
In the meantime I also tried to ask Apple Developers why they decided to implement the center parts with canvas instead of editable text: https://developer.apple.com/forums/thread/722684

rakleed:
@akirataguchi115 when using the `.icloud-logo` selector, the Apple icon is not inverted on the main [icloud.com](https://www.icloud.com/), and it will no longer be possible to make a working rule for it that would not break inverting on pages with Notes, Disk, etc.

> also these would be a nice addition
> 
> ```
> .tags-cloud-toggle
> .tt-input-field
> .network-badge-symbol
> ```

I didn't find where you got those selectors from. Maybe it's better to do a separate PR for them?

akirataguchi115:
I'll test your first note about .icloud-logo as soon as I can!

Also yeah sure I guess I'll do a separate PR for the three other selectors. Thanks for letting me know!

akirataguchi115 (approved):
Works as intended.

Gusted (approved):
Thanks!

PR #10572, opened 2023-01-05
dark-sites.config: add splatoon3.ink

AAGaming00:
Site is dark by default, and its background also breaks if Dark Reader is on.


Gusted (approved):
Thanks!

PR #10580, opened 2023-01-06
Fix librivox.org (Dynamic mode)

Gitoffthelawn:
Fixes white blocks around "Read by volunteers from around the world." string (in dark dynamic mode).
Fixes statistics area.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -11865,8 +11865,6 @@ CSS
 .homepage-main-wrap,
 .homepage-main-block p span {
     background-color: var(--darkreader-neutral-background) !important;
-}
-.homepage-main-block p span {
     border-color: transparent !important;
 }
 .box_1 {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -11866,6 +11866,9 @@ CSS
 .homepage-main-block p span {
     background-color: var(--darkreader-neutral-background) !important;
 }
+.homepage-main-block p span {
+    border-color: transparent !important;
+}
```
Gusted, line 11871:
```suggestion
.homepage-main-block p span {
    background-color: var(--darkreader-neutral-background) !important;
    border-color: transparent !important;
}
```

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
Gitoffthelawn: Looks good.  Thank you!

PR #10587, opened 2023-01-07
Update fix for steamdeck.com

pnaf:
* Fixes "Available now" section on home page.
* Deletes outdated fixes introduced in #10033.

Before:

After:

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -19001,18 +19001,18 @@ i.ml-1
 steamdeck.com
 
 INVERT
-section#available-now div.col_4.copy
-section#available-now h1
+section#available-now div.col_9
 section#experience div.col_8
 section#experience h2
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -19001,19 +19001,19 @@ i.ml-1
...
-section#available-now div.col_4.copy
-section#available-now h1
+section#available-now div.col_9
 section#experience div.col_8
 section#experience h2
 section#topfeature div.col_5
 
+CSS
+section#available-now {
+    background-color: #f0f0f0 !important;
```
Gusted, line 19011: Can this be made in the format of `${color}`, see the table at https://github.com/darkreader/darkreader/blob/main/CONTRIBUTING.md#how-to-use-the-dev-tools.

pnaf: Fixed. Good catch!

PR #10590, opened 2023-01-08
add my sites (reoccur.cat/xyz) to global dark list

reoccurcat:


bershanskiy:
Hi @reoccurcat could you please add the following HTML to your site to signal Dark Reader that the site is dark?

```html
<head>
  <meta name="darkreader-lock"></meta>
</head>
```

This is equivalent to being added to the dark list, but gives you full freedom to mark individual pages as dark at any point, without waiting for Dark Reader to be updated.

PR #10593, opened 2023-01-08
Fix for bitrix24.ru

rakleed:
- Invert table.


bershanskiy:
Can't test unfortunately, so will have to merge as is.

PR #10598, opened 2023-01-09
Fix for boardgamearena.com

branflaker:
Added boardgamearean.com section to dynamic-theme-fixes.config.
Updated card text to be darker against light card background for Memoir 44 game.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2831,6 +2831,11 @@ INVERT
 boardgamearena.com
 
 CSS
+.nex_metropad {
+    background-image: url("https://x.boardgamearena.net/data/themereleases/current/games/nextstation/220831-1241/img/mainpad_back.jpg");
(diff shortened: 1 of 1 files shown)


branflaker:
@bershanskiy Why add the wildcard?  I don't see that used for other domains in the config.

bershanskiy:
@branflaker Eventually, we plan to migrate away from the current behavior of treating `example.com` like a wildcard meaning "`example.com` and all its subdomains". But this is some time off.

bershanskiy:
Other than that, I think it's fine. I can merge whenever it is convenient for you.

branflaker:
Okay, thank you -- I just noticed that it failed the alphabetical sorting check because of that and so when I inspected for other instances of using a wildcard, I didn't see any.

Happy for it to merge whenever.  There may be more changes with this site/domain in the future because there's a lot of custom CSS for each game on the site.

PR #10601, opened 2023-01-10
Bugfix: Cannot read properties of undefined (reading 'colorScheme')

dderiso:
I moved a test into a try brace.


bershanskiy:
Hi, @dderiso thanks for taking your time creating the PR. I understand the fix (and it looks mostly OK), but I'm having hard time understanding the circumstances when this bug occurs. Could you please provide repro steps? I'll try repro this as well. Thanks.

For code cleanliness, I'll change the patch to this:
```diff
- if (typeof el.style.colorScheme === 'string') {
+ if (el && typeof el.style === 'object' && typeof el.style.colorScheme === 'string') {
```

dderiso:
Hey @bershanskiy, happy new year. I noticed this bug when I was working on my web app (angular  + material design), and although it may appear on other sites, I haven't been looking for it outside my own app. My sense is that the issue should appear everywhere. Here's my reasoning: The current procedure is to: (1) creates an element and then (2) inspects its attribute looking for a "style.colorScheme" of a certain type (string). A safer procedure is: (1) create the element AND add the attribute of the required type (and then (2) is not necessary).

dderiso:
Hey @bershanskiy, happy new year. I noticed this bug when I was working on my own web app, and although it may appear on other sites, I haven't been looking for it. My sense is that the issue should appear everywhere. Here's my reasoning: The current procedure is to: (1) creates an element and then (2) inspects its attribute looking for a "style.colorScheme" of a certain type (string). A safer procedure is: (1) create the element AND add the attribute of the required type (and then (2) is not necessary).

bershanskiy:
> I noticed this bug when I was working on my own web app,

Am I correct in understanding that this error was logged into the page console when you were debugging/developing your own website (as opposed to Dark Reader background context)? Also, which browser were you using? Is the page public and would you be able to link to the website?

dderiso:
> Am I correct in understanding 
yes (and the error's source were the lines I modified and submitted in the PR, fixing these on my local resolved the error)

> which browser were you using? 
Chrome (latest)

> Is the page public
no (but I'd gladly share if it was)

It comes down to making sure the required style tag is added when the element is created. Otherwise, attempting to access the style tag may not work, and so moving the accessor into try/catch scope is a workaround.

PR #10614, opened 2023-01-12
Remove fix for mesa3d.org

q1800:
Remove fix for mesa3d.org as it's no longer needed.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -13063,13 +13063,6 @@ CSS
 
 ================================
 
-mesa3d.org
-
-INVERT
-[data-bs-toggle="collapse"] > a > svg
-
-================================
-
 messages.android.com
 
 INVERT


bershanskiy:
This fix still matches elements on the page, however, the fix makes dark mode deviate from the stylistic intent (original page makes "chevrons" color match text color), so I'm in favor of removing it.

DR Off:
image

DR On, before PR:
image

DR On, after PR:
image

bershanskiy:
I'm not sure what was broken about these chevrons before and I'm not aware of any changes between 4.9.60 and 4.9.62 which would fix anything about them (the  only guess I have is a color string using `/` without spaces, but that's rare). Just to make sure, I made a few screenshots (below). Based on these screenshots, I think the fix makes the page look worse, not better, if anything.

Dark Reader 4.9.60 on Chromium 109 with fix (before PR):
image


Dark Reader 4.9.60 on Chromium 109 without fix (after PR):
image

Dark Deader 4.9.62 on Firefox 108.0.2 with fix (before PR):
image

Dark Deader 4.9.62 on Firefox 108.0.2 without without fix (after PR):
image

Dark Reader 4.9.60 on Firefox 108.0.2 with fix (before PR):
image

Dark Reader 4.9.60 on Firefox 108.0.2 without without fix (after PR):
image

PR #10623, opened 2023-01-13
Fix: move entire isCSSColorSchemePropSupported() into try-catch

bershanskiy:
Fixes #10601.

Credit for original bug discovery goes to @dderiso.


dderiso:
Credit for an epic plugin goes to you guys :)

PR #10624, opened 2023-01-14
Fix NatureMade articles (Dynamic)

Gitoffthelawn:
Example:
https://www.naturemade.com/blogs/health-articles/how-much-fish-oil-should-you-take-epa-and-dha-omega-3s-are-what-counts


Gitoffthelawn (commented):
Even better.  Thanks!

PR #10625, opened 2023-01-14
Fix for nagatabi-p.jimdofree.com

34j:
This PR fixes a problem in nagatabi-p.jimdofree.com where formula images are displayed inverted.

Before (Dark): 
image
After (Dark):
image
After (Light):
image


Gusted (approved):
Thanks.

PR #10630, opened 2023-01-14
Fix for positiveintelligence.com

Gitoffthelawn:
- Fixes primary top header link (logo)
- Fixes footer logo


Gitoffthelawn (commented):
WFM. Thanks!

BTW, why do you prefer to explicitly specify `img`?

PR #10631, opened 2023-01-14
Update dark-sites.config

nsde:


src/config/dark-sites.config
```diff
@@ -557,6 +559,7 @@ oldfag.org
 omcar.pl
 omgwtfnzbs.me
 onionplay.co
+onlix.me
```
Gusted, line 562: This isn't a dark by default site, please remove:

image

Gusted (changes requested):

PR #10633, opened 2023-01-14
Add kika.de

Myzel394:
This site seems to be completely dark. There are some white buttons, but in overall it's pretty dark

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -434,6 +434,7 @@ kbhgames.com
 keezersquest.nl
 kernel.fish
 kfocus.org
+kika.de
 kilonova.ro
 kimbatt.github.io/js-Z
 kingdom-leaks.com


bershanskiy:
Hi, thanks for this PR, but it appears that only the landing page supports dark mode well. The rest of the site is mostly light, for example:
https://www.kika.de/erwachsene/index.html

Myzel394:
sorry about that. I'll create a PR instead to fix the small issues on the kids section. @bershanskiy do you know how to exclude Dark Reader on specific elements?

bershanskiy:
@Myzel394 We have some documentation explaining how to use Dark Reader Developer Tools [here](https://github.com/darkreader/darkreader/blob/main/CONTRIBUTING.md). You can use them right from your browser without building Dark Reader yourself. Also, if you would provide a list of broken elements (preferably with screenshots) I would be able to look into this as well.

Myzel394:
@bershanskiy on kika.de we can simply exclude the whole "nav" element and its children.

Navbar with DarkReader (incorrect)
image

Navbar without DarkReader (correct)
image

Myzel394:
The selector is:

```
nav[data-testid="navbar"]
```

PR #10641, opened 2023-01-16
Add some pages on www.kika.de to dark sites

### LINKED ISSUE #10640: [Broken Website] www.kika.de lading page
### Prerequisites

### Website Issue Description

Copied from #10633:

> This site seems to be completely dark. There are some white buttons, but in overall it's pretty dark

> @bershanskiy on kika.de we can simply exclude the whole "nav" element and its children.

> The selector is:
>
> ```
> nav[data-testid="navbar"]
> ```

### Website Address

kika.de

### Steps To Reproduce

1. Open the site, observe navbar

### Screenshots

>
> Navbar with DarkReader (incorrect)
image
>
> Navbar without DarkReader (correct)
image

### Operating System

Any

Chromium 109

### Dark Reader version

Any (4.9.62)

### Additional Context

Originally reported by @Myzel394 in #10633.

bershanskiy:
Fixes #10640.

CC @Myzel394

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -848,6 +848,15 @@ www.canalplus.com
 www.cc.com
 www.digikam.org
 www.eveonline.com
+www.kika.de/$
+www.kika.de/kika-live/*
+www.kika.de/kummerkasten/*
+www.kika.de/live/*
+www.kika.de/selbermachen/*
+www.kika.de/spiele/*
+www.kika.de/suche/*
+www.kika.de/ueber-kika/*
+www.kika.de/videos/*
 www.mechapower.eu
 www.mrdemoapple.uk
 www.mtv.com


bershanskiy:
Testing:
The following page is not dark and should be modified by Dark Reader:
https://www.kika.de/erwachsene/index.html
https://kommunikation.kika.de/presse-startseite-100.html

The following pages are dark and do not need to be modified by Dark Reader:
https://www.kika.de/
https://www.kika.de/videos/kindervideos-videos-fuer-kinder-100
https://www.kika.de/spiele/kinderspiele-online-kostenlos-100
https://www.kika.de/live/live-100
https://www.kika.de/selbermachen/basteln-mit-kindern-100
https://www.kika.de/suche/suche104
https://www.kika.de/kummerkasten/kummerkasten-112

bershanskiy:
I'll give this a few days to get comments (mostly from @Myzel394) and then merge (but no earlier than Jan 18).

Myzel394:
@bershanskiy the site is generally dark, except for `/erwachsene/*`, I think we should declare everything except for this path (and subpaths) as dark. The sites contains series for children (for example: https://www.kika.de/das-dschungelbuch/das-dschungelbuch-100) and the paths all start with the series name. If they add another series, we would have to add that exception too and always check the site for new series. Instead, let's simply assume the whole site is dark, except for `/erwachsene/*` (I couldn't find any other light sites)

bershanskiy:
I'd prefer to explicitly specify the dark pages and not use exclusions, since that is easier to maintain. Do you happen to be affiliated with this site, if you don't mind me asking?

Myzel394:
No I'm not affiliated with it, sorry. I'd strongly still suggest to assume that the site is dark in general and add only some exclusions, since we will otherwise need to update DarkReader whenever the site releases a new series

bershanskiy:
Thanks for your work, I appreciate it.

Unfortunately, at the moment we do not support exception patterns (but I plan to add support at some point). For now, it's better to list dark pages explicitly and within a few releases I'll rework the dark site list processing and add support for negative patterns.

PR #10650, opened 2023-01-18
Fix for 9game.cn

felixonmars:
Invert logos in inner pages.

Before:
Screenshot_20230118_081547

After:
Screenshot_20230118_081535


bershanskiy:
Thanks for the fix!

-------

Note to self:
Clarification about meaning of "inner pages": this bug does not occur on the landing page (`https://www.9game.cn/`), it only occurs on other pages (like `https://www.9game.cn/chunqiufengshen/`).

PR #10655, opened 2023-01-19
Fix for psprices.com

rakleed:
- Remove bright background.


bershanskiy:
This is somewhat arbitrary change, but the page does look slightly better after this change, so I'll merge it.

PR #10664, opened 2023-01-20
fix: getExtensionPageTabMV3config function performance

anavalo:
This PR fixes the `TODO` of `getExtensionPageTabMV3()` in UI utils.

The `chrome.tabs.query` method was replaced by  `chrome.windows.getAll` and `chrome.tabs.sendMessage`.

This way, you can get the active tab of the current window and send the message to it, without the need to go through all the pop-up windows.

closes: #10665

### DIFF
--- a/src/ui/utils.ts
+++ b/src/ui/utils.ts
@@ -150,29 +150,23 @@ export async function getFontList() {
 
 type page = 'devtools' | 'stylesheet-editor';
 
-// TODO(Anton): There must be a better way to do this
-// This function ping-pongs a message to possible DevTools popups.
-// This function should have reasonable performance since it sends
-// messages only to popups and not regular windows.
 async function getExtensionPageTabMV3(): Promise<chrome.tabs.Tab | null> {
     return new Promise((resolve) => {
-        chrome.windows.getAll({
-            populate: true,
-            windowTypes: ['popup'],
-        }, (w) => {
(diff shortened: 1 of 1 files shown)


bershanskiy:
Hi @anavalo, thanks for the PR. However this code is buggy.

This is because:
1. Dark Reader Devtools actually have a separate window, since it is a `"popup"`
image
2. By definition, if user opens Dark Reader popup, the regular window is active at the moment
image

This means we can't use `currentWindow` set to `true` so we basically have to get all active tabs... and iterate over them... which means we are better off actually filtering by window type anyway. Hence the current solution is the best one at this time.

The perfect solution would be to use URL filter in `tabs.query()` and look up the known URL, which is exactly what we do in `getExtensionPageTab()`. Unfortunately, this requires `tabs` permission and I would like to remove it at some point, at least in MV3 builds. The proper solution overall would be modify browser to accept URL filter and just by default skip all tabs whose URLs the extension does not have access to (which is what we do). This is not a big deal for us really, since we filter by window type `"popup"` which is very uncommon. In fact I'm not aware of a practical case when user would have multiple `"popup"` windows.

bershanskiy:
Demo of the bug caused by PR:

[buggy pr.webm](https://user-images.githubusercontent.com/45960703/213885526-949feb56-aa31-4e5f-ba3c-bb0d87fd3f4f.webm)

PR #10670, opened 2023-01-21
Update fix for Jira sites

varp:
Before
image
image
After
image
image


bershanskiy:
Hi @varp, thanks for this PR. Does `stash.*.*` host Jira and Bitbucket? Would it be appropriate to simply fold `stash.*.*` fix into the other one?

varp:
HI @bershanskiy. No it doesn't. It's two separate domains and hosts. Could you explain me how `.*` works? Is that a regexp or it's just a dot and glob pattern?

varp:
@bershanskiy any changes to that PR should I do to move it by the flow?

PR #10673, opened 2023-01-22
Ports brightspace.com theme to brightspace.algonquincollege.com

mpsparrow:
Ports brightspace.com theme to brightspace.algonquincollege.com

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3049,6 +3049,7 @@ CSS
 
 ================================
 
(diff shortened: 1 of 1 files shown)


bershanskiy:
Thanks for this fix. Please note that since the site requires an account, I can not test it, so I'm just merging this as is.

mpsparrow:
Thank you! All three URLs under this theme run an installation of the Brightspace software. Therefore, there styles will be universal to any organizations standard Brightspace installation.

PR #10675, opened 2023-01-22
Fix instructables.com

Gitoffthelawn:
Fix main header logo, which is also a link.

Example page:
https://www.instructables.com/how-to-recover-a-dry-printer-head/


Gitoffthelawn:
You bet!

PR #10679, opened 2023-01-23
Fix for service-now.com

sKravciv:
Fix for code editor in service-now instances

*.service-now.com

CSS
.cm-s-snc span.cm-string {
  color: #cf947b !important;
}
.cm-s-snc span.cm-def {
  color: #60a4d8 !important;
}
.cm-s-snc span.cm-property {
  color: #ffffff !important;
}
.cm-s-snc_readonly span.cm-string{
  color: #cf947b !important;
}


bershanskiy:
Hi @sKravciv, how may I access this editor to test the fix? Thanks.

bershanskiy:
And the CI might fail, I think.

sKravciv:
Hi @bershanskiy, you cannot access instance of servicenow because it is enterprise solution. You can request developer instance [here](https://developer.servicenow.com/dev.do#!/home) - but it is a little bit complicated. I can provide screeshots if that would be helpful. What can I do to make the CI pass? Thanks

bershanskiy:
My main concern is the arbitrary color values like `#cf947b`. They might not play well with custom color settings.

sKravciv:
I was using colors from different servicenow editor where the code is readable
screen

bershanskiy:
Would it be possible to use Dark Reader CSS variables like these?
```CSS
--darkreader-neutral-background
--darkreader-neutral-text
--darkreader-selection-background
--darkreader-selection-text
```

sKravciv:
Unfortunately no. I have tried using all of them but it makes the code unreadable

sKravciv:
Hi @bershanskiy, is it possible to merge this with base branch or is there something missing? Thank You

PR #10696, opened 2023-01-27
Fix for gitlab.host

MrHeliX:
Added gitlab.host; include button "View eligible approvers" in dark style


bershanskiy:
This resource requires authorization, so I'm unable to test the fix. The source looks fine, so I'll merge it as is.

PR #10718, opened 2023-01-31
Fix for astro.build

Kornelijus:
Fixes logo and gradient for the hero card for [astro.build](https://astro.build/).

astro build_ (5)
astro build_ (4)


Gusted (approved):
Thanks.

PR #10743, opened 2023-02-05
Fixes for uottawa.brightspace.com

GentlyTech:
- Fixed uOttawa logo on homepage
    - (amended) ~~Used filter in CSS instead of INVERT because the changes should only apply when dark~~ Just used normal INVERT as requested
- Fixed dropdown backgrounds
- Fixed course section cards on homepage
- Fixed borders on some buttons and headers
- Fixed certain button backgrounds on hover/focus

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -21293,6 +21293,44 @@ CSS
 
 ================================
 
+uottawa.brightspace.com
+
+INVERT
+.d2l-navigation-link-image-container
+
+CSS
+:host {
+    --d2l-button-icon-background-color-hover: var(--darkreader-bg--d2l-color-gypsum) !important;
+}
+d2l-card {
+    background: var(--darkreader-neutral-background) !important;
+    border-color: var(--darkreader-border--d2l-color-gypsum) !important;
+}
+d2l-dropdown-content > div,
+d2l-menu-item {
+    background-color: var(--darkreader-neutral-background) !important;
+    border-radius: 10px !important;
+}
+d2l-empty-state-simple {
+    border-color: var(--darkreader-border--d2l-color-mica) !important;
+}
+.d2l-button-filter > ul > li > a.vui-button {
+    border-color: var(--darkreader-border--d2l-color-mica) !important;
+}
+.d2l-label-text:has(.d2l-button-subtle-content):hover,
+.d2l-label-text:has(.d2l-button-subtle-content):focus,
+.d2l-label-text:has(.d2l-button-subtle-content):active {
+    background-color: var(--darkreader-bg--d2l-color-gypsum) !important;
+}
+.d2l-navigation-centerer {
+    color: inherit !important;
+}
+.d2l-tabs-layout {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -21293,6 +21293,45 @@ CSS
 
 ================================
 
+uottawa.brightspace.com
+
+CSS
+button:hover, button:focus, :host([active]) button {
```
Gusted, line 21299:
```suggestion
button
:host([active]) button {
```

src/config/dynamic-theme-fixes.config
```diff
@@ -21293,6 +21293,45 @@ CSS
...
+.d2l-label-text:has(.d2l-button-subtle-content):active {
+    background-color: var(--darkreader-bg-d2l-color-gypsum) !important;
+}
+.d2l-navigation-centerer {
+    color: inherit;
+}
+.d2l-navigation-link-image-container {
+    filter: invert(1);
+    mix-blend-mode: lighten;
+}
```
Gusted, line 21316: Hmm this doesn't look sound, what does this specifically fix?

Gusted (changes requested):
Please add `!important;` after the CSS declarations.

src/config/dynamic-theme-fixes.config
GentlyTech: Hmm, removing the hover/focus selectors makes the changes not apply to the button when you hover over it.

Gusted: The `!important` on the CSS declaration should take care of that..

src/config/dynamic-theme-fixes.config
GentlyTech: The filter and mix-blend-mode is for the university logo in the nav bar. I tried using invert but it wouldn't look quite right.

However I did realize a problem with this (apart from the slight "jankiness" of it), which is that you can't see the logo if you set the scheme to light.

So I guess just remove it for now?

Fix with INVERT:

WithINVERT

Fix with CSS:

WithCSS

When using light scheme (when applying the CSS):

WithCSS_Light

src/config/dynamic-theme-fixes.config
GentlyTech: Oh, changing the selector to :host works (the site has an obsession with shadow dom).

GentlyTech: Ok, nonetheless I have added the `!imporatant` statements. Sorry about that.

src/config/dynamic-theme-fixes.config
Gusted: Yeah normally we just do `INVERT` and _accept_ the jankiness.

GentlyTech: Sorry by jankiness I was referring to my CSS patch. But ok, will just do invert.

GentlyTech:
Ok I have updated the commit with your requested changes.

Gusted (approved):
Thanks.

PR #10744, opened 2023-02-05
add own blog which is already dark

peterge1998:


bershanskiy:
Thanks for this PR,

since this is your own blog we would prefer if you could instead modify your site to instruct to disable itself. You can do this by adding a `<meta>` tag with name `darkreader-lock` to the page `<head>` element:

```html
<head>
  <meta name="darkreader-lock"></meta>
</head>
```

This way you can configure Dark Reader bu yourself without the need to wait for Dark Reader release.

peterge1998:
> Thanks for this PR,
> 
> since this is your own blog we would prefer if you could instead modify your site to instruct to disable itself. You can do this by adding a `<meta>` tag with name `darkreader-lock` to the page `<head>` element:
> 
> ```
> <head>
>   <meta name="darkreader-lock"></meta>
> </head>
> ```
> 
> This way you can configure Dark Reader bu yourself without the need to wait for Dark Reader release.

Did not know this method, thanks!

PR #10751, opened 2023-02-07
Fix for derStandard.at

R0M4N0F5K7:
This is an Austrian newspaper that uses many colors on their homepage. Fixed sidebar, several versions of title bars, some buttons, pop-ups, etc.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5287,6 +5287,27 @@ body {
 
 ================================
 
+derStandard.at
+
+CSS
+:root {
+    --theme-background: var(--darkreader-bg--theme-background) !important;
+    --theme-sidebar: var(--darkreader-bg--theme-sidebar) !important;
+    --theme--primary: var(--darkreader-bg--theme--primary) !important;
+    --theme--secondary: var(--darkreader-bg--theme--secondary) !important;
+    --theme--secondary: var(--darkreader-bg--theme--secondary) !important;
+    --theme--on-secondary: none !important;
+    --theme--surface-2dp: var(--darkreader-bg--theme--surface-2dp) !important;
+    --theme--on-surface: var(--darkreader-text--theme--on-surface) !important;
+    --theme--on-base: var(--darkreader-text--theme--on-base) !important;
+    --theme--on-base--hls: var(--darkreader-text--theme--on-base--hls) !important;
+    --theme--surface-2dp-border: var(--darkreader-text--theme--surface-2dp-border) !important;
+    --mut-header-compact-desktop-base: var(--darkreader-neutral-background) !important;
+    --darkreader-bg--mut-header-compact-desktop-base: var(--darkreader-neutral-background) !important;
+}
+
+================================
+
 designobserver.com
 
 INVERT


src/config/dynamic-theme-fixes.config
```diff
@@ -5287,6 +5287,27 @@ body {
...
+    --theme--secondary: var(--darkreader-bg--theme--secondary) !important;
+    --theme--on-secondary: none !important;
+    --theme--surface-2dp: var(--darkreader-bg--theme--surface-2dp) !important;
+    --theme--on-surface: var(--darkreader-text--theme--on-surface) !important;
+    --theme--on-base: var(--darkreader-text--theme--on-base) !important;
+    --theme--on-base--hls: var(--darkreader-text--theme--on-base--hls) !important;
+    --theme--surface-2dp-border: var(--darkreader-text--theme--surface-2dp-border) !important;
+    --mut-header-compact-desktop-base: var(--darkreader-neutral-background) !important;
+    --darkreader-bg--mut-header-compact-desktop-base: var(--darkreader-neutral-background) !important;
+}
```
Gusted, line 5307: If it needs to define that many variables, there's something wrong on Dark Reader's side. Is there an example page(that wouldn't require an subscription)?

R0M4N0F5K7: Yeah, I feel like Dark Reader has issues if colours inherited and calculated several level deep from combined variables, the theming is quite complex on this hp.
To answer your question, none of that page requires a subscription. Sometimes it prompts you to agree to continue with ads when you visit the page, but as long as it doesn't detect your ad blocker it is no issue.

The page uses different color themes for different news categories, which weirdly react in different ways on dark reader.
What I tried to fix, as it is missed by default dark mode:

The pill in the header (this is ALWAYS broken, in chromium browsers it is visible that it works for a moment, but then seems to be inverted again):
Screenshot 2023-02-09 at 22-47-27 derStandard at Nachrichten Kommentare   Community

The floating header that appears in desktop mode when you scroll down a bit and up again (this is ALWAYS broken):
Screenshot 2023-02-09 at 22-48-09 derStandard at Nachrichten Kommentare   Community
Weirdly even though the mut-header-compact-desktop-base should fix that it doesn't load the fixed variable in some articles, which is why I overwrote it a second time which made it always work (line 5305 and 5306)

The floating header in mobile mode that appears if you scroll down and then up again, works in most themes but when it doesn't it is really blinding:
Screenshot 2023-02-09 at 23-06-55 EU-Abgeordnete wollen mehr Klimaschutz im Lieferkettengesetz

The collapsed sidebar (which in some themes sometimes takes like 10 seconds to be darkend, sometimes isn't at all):
Screenshot 2023-02-09 at 22-48-45 Landbauers Sager zu Erdbebenopfern verschiebt die Grenzen des Akzeptablen

Opened sidemenu (this gets really bright in some color themes, very annoying):
Screenshot 2023-02-09 at 22-49-06 Landbauers Sager zu Erdbebenopfern verschiebt die Grenzen des Akzeptablen

Weirdly, after the first half of variables are defined these "surfaces" that pop-up when you click on the profil icon for example are all of a sudden bright, even thoug they were not before, so most of the second half of my lines fix them again:
Screenshot 2023-02-09 at 22-51-58 Austria-Trainer Wimmer Ich kann auch explodieren

So yeah, I get your confusion. I suppose it is just terrible and complicated mixed/calculated theming that seems to have broken inheritance sometimes on this homepage that requires these fixes.
Less than half of those lines are required to make it work on most articles, but sometimes it will take some long seconds to apply without my configs or doesn't at all, like on that page which seems to have some issues for this one seems to have a tricky theme:
https://www.derstandard.at/story/2000143396531/eu-abgeordnete-wollen-mehr-klimaschutz-im-lieferkettengesetz
image

PR #10752, opened 2023-02-08
Swapped descriptions and uses in CSS variables

felipecock:
It seems to be a swapped description and use in:
   --darkreader-selection-background
and
   --darkreader-selection-text
CSS variables.


Gusted (approved):
Good catch, thanks!

felipecock:
You're welcome!
Thanks for sharing this project with the world.

PR #10756, opened 2023-02-09
Fix multilanguage site-toggle.tsx

JNylson:


Gusted:
Unfortunately it isn't that easy to do fix this, as currently it would require translations for all the other locales to pass the CI.

JNylson:
I'll try it later

PR #10759, opened 2023-02-09
Fix for kaava.mafynetti.com

Hypernova234:
Fix for a math input site used in almost every finnish high school. 

Fixed so that you can see the text on the math input images, as it was black on a really dark background before which made this extension unusable on this website. It was the first time I tried changing anything on a website, and I didn't manage to change the color of the text in the image, so I managed to change the background in the images instead. 

Someone else could probably do this a lot better than me, but I don't think that is going to happen and this is much better than completely unusable :D

Screenshot of what it looked like before:
image

Screenshot of what it looks like now:
image

This is also my first time using GitHub so give me some slack :)

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -11115,6 +11115,10 @@ CSS
 .rich-text-editor img[src*="/math.svg"] {
     background-color: white !important;
 }
+.rich-text-editor-toolbar-equation {
+background-color: white;
+color: white;
+}
 
 ================================
 
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -11109,6 +11109,21 @@ CSS
 
 ================================
 
+kaava.mafynetti.fi
+
+CSS
+.rich-text-editor img[src*="/math.svg"] {
+    background-color: white;
+}
+.rich-text-editor:focus img[src*="/math.svg"] {
+    background-color: white; 
+}
+.rich-text-editor.rich-text-focused img[src*="/math.svg"] {
+    background-color: white;
+}
+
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -11109,6 +11109,21 @@ CSS
...
+CSS
+.rich-text-editor img[src*="/math.svg"] {
+    background-color: white;
+}
+.rich-text-editor:focus img[src*="/math.svg"] {
+    background-color: white; 
+}
+.rich-text-editor.rich-text-focused img[src*="/math.svg"] {
+    background-color: white;
+}
```
Gusted, line 11123:
```suggestion
.rich-text-editor img[src*="/math.svg"] {
    background-color: white !important;
}
```

Gusted (changes requested):

src/config/dynamic-theme-fixes.config
Gusted: Specifying `!important` is enough to ensure this would happen for the other states as well.

Hypernova234: Sure thing, I committed the suggestion, but I found another place with the exact same problem up in the toolbar that I also changed and now I don't know how to add it to this pull request. 

Here's a screenshot of what I changed:
image

And here's the CSS code for it:

.rich-text-editor-toolbar-equation {
    background-color: white;
    color: white;
}

This one looks even worse than the other one but It's functional.

Hypernova234: There we go, I didn't find it earlier as the website didn't come up in the code when I Ctrl + F:ed it. Seems like you have to do this inside the code window to find it.

PR #10760, opened 2023-02-09
Fix for all mathbooks on digi.sets.fi

Hypernova234:
The pictures that covered almost the whole page werent being inverted at all and made the bright text above them really hard to read so I inverted them the exact amount so that it blends into the background.

Before:
image

After:
image

This could probably also be done better, but it's better than being completely unusable :)

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5624,6 +5624,36 @@ CSS
 
 ================================
 
+digi.sets.fi/makortrep/ma-kort-repetition-1/
+digi.sets.fi/ma8kort/
+digi.sets.fi/ma7kort/
+digi.sets.fi/ma6kort/
+digi.sets.fi/ma5kort/
+digi.sets.fi/ma4kort/
+digi.sets.fi/ma3kort/
+digi.sets.fi/ma2kort/
+digi.sets.fi/ma1kort/
+digi.sets.fi/ma13lang/
+digi.sets.fi/ma12lang/
+digi.sets.fi/ma11lang/
+digi.sets.fi/ma10lang/
+digi.sets.fi/ma9lang/
+digi.sets.fi/ma8lang/
+digi.sets.fi/ma7lang/
+digi.sets.fi/ma6lang/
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -5624,6 +5624,36 @@ CSS
...
+digi.sets.fi/ma5lang/
+digi.sets.fi/ma4lang/
+digi.sets.fi/ma3lang/
+digi.sets.fi/ma2lang/
+digi.sets.fi/ma1lang/
+
+CSS
+.bi {
+    filter: invert(0.936);
+}
```
Gusted, line 5653: I think this would better fit:
```suggestion
INVERT
.bi 
```

Hypernova234: I tried a lot to get this to work and I thought the same way, but for some reason it just doesn't work. The exact amount of invertion on the original technique also looks better as it blends in with the background and doesn't leave half of the book another hue of black than the rest.

src/config/dynamic-theme-fixes.config
```diff
@@ -5624,6 +5624,36 @@ CSS
...
+digi.sets.fi/ma10lang/
+digi.sets.fi/ma9lang/
+digi.sets.fi/ma8lang/
+digi.sets.fi/ma7lang/
+digi.sets.fi/ma6lang/
+digi.sets.fi/ma5lang/
+digi.sets.fi/ma4lang/
+digi.sets.fi/ma3lang/
+digi.sets.fi/ma2lang/
+digi.sets.fi/ma1lang/
```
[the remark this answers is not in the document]
Hypernova234: I don't think you can do this, as then It'll also change all the pictures on the homescreen and all of the other books that this website provides that also use the .bi selector for the images. I didn't bother looking in other books on this website and didn't think it was worth the risk. 

The uppermost link also didn't change anything on the page unless the "ma-kort-repetition-1/" was there but the second link worked fine, so I think It's kind of hit or miss and I didn't want to take the chance.

PR #10800, opened 2023-02-18
Remove IMAP & POP3 protocol code

Gitoffthelawn:
I think this code is likely unused.  If I'm mistaken, please reject this PR.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -10249,16 +10249,6 @@ body {
 
 ================================
 
-imap:
-pop3:
-
-CSS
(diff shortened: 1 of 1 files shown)


bershanskiy:
Hi @Gitoffthelawn, this code is used, but only on Thunderbird builds as far as I remember. We could probably move this code to into JS in `getDynamicThemeFixesFor()` and compile it out on all other platforms.

Gitoffthelawn:
@bershanskiy Thanks.  That sounds good.

Gitoffthelawn:
@bershanskiy Was the resolution recommended by @bershanskiy implemented?  Thanks.

PR #10801, opened 2023-02-18
Fix isitdownrightnow.com

Gitoffthelawn:
Removes the additional gray boxes to the left of each host name.

Kept code formatting consistent, for better or worse.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -10774,6 +10774,52 @@ div.grey-gui-bg,
 
 ================================
 
+isitdownrightnow.com
+
+CSS
+.ts1,
+.ts2,
+.ts3,
+.ts4,
+.ts5,
+.ts6,
+.ts7,
+.ts8,
+.ts9,
+.ts10,
+.ts11,
+.ts12,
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -10774,6 +10774,52 @@ div.grey-gui-bg,
 
 ================================
 
+isitdownrightnow.com
+
+CSS
+.ts1,
```
alexanderby, line 10780: What about `[class^="ts"]`?

alexanderby (changes requested):

src/config/dynamic-theme-fixes.config
Gitoffthelawn: > What about `[class^="ts"]`?

I like that too.  It's more broad, so could potentially have false positives down the line, but the site isn't very complex, so might be worth the risk.

Also, could then have false negatives if they specify other classes before the `.ts`... classes for those elements.

What's your preference?  I'm good either way.  The way in the PR has less risk but more bulk.  The way you mentioned has more risk but less bulk.  Tradeoffs.

Issue #10716, opened 2023-01-31
[Bug] Detection of the active tab (exposed in the toggle button) is ignoring app/shortcut windows


mhochk:
### Prerequisites

- [X] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this bug to avoid opening a duplicate.
- [X] I can reproduce this bug in a new, unmodified web browser profile with Dark Reader installed as the only extension.
- [X] I understand I need to use the [Broken Website Report template](https://github.com/darkreader/darkreader/issues/new?assignees=&labels=Broken+Website&template=broken-website-report.yml&title=%5BBroken+Website%5D+Replace+with+title) if this bug I am reporting occurs on a single website.

### Bug Description

The detection of the "current" active tab is scoped to only "normal" tabs, so is excluding Progressive Web Apps (PWAs) and sites installed as apps/shortcuts. The core functionality still works in these cases (e.g. the site list is still correctly respected), but the helper UI for manipulating the "current" URL applies to the wrong URL.

### Website Address

https://en.wikipedia.org/wiki/Fish

### Steps To Reproduce

1. Navigate to a website (e.g. https://en.wikipedia.org/wiki/Fish)
2. Install the site as a shortcut/app:
    If in Chrome, use ... > More tools > Create shortcut. Check the 'Open as window' option and Create button.
    If in Edge, use ... > Apps > Install this site as an app. Install button.
3. In the newly opened window, select the Extensions icon and the Dark Reader menu item to open the Dark Reader UI.
4. Look at the URL shown as the 'current' site.

### Expected Behavior

The URL shows the web page currently being shown in the window (e.g. https://en.wikipedia.org/wiki/Fish)

### Actual Behavior

The URL shows the last active tab in the main browser window.

### Screenshots

![image](https://user-images.githubusercontent.com/45496878/215844467-5f71cab2-618b-4a84-87ba-c023a6acfd11.png)

### Operating System

Windows

### Web Browser name and version

Chrome 112.0.5570.0

### Dark Reader version

4.9.62

### Additional Context

May have been introduced as part of the work to exclude Dev Tools.
https://github.com/darkreader/darkreader/commit/30214490f9c76c102878be82201ff17b0925c343

bbaros:
I see this issue happening again on version 4.9.70 while using Outlook PWA.

zeroepoch:
[...]

PR #10857, opened 2023-03-02
fix app/shortcuts active tab detection

DovieW:
You can't use more than one value for the windowType property.

You can use multiple values with chrome.windows.getAll() though

Might be able to make that work somehow, maybe it'll make this a bit cleaner.

### DIFF
--- a/src/background/tab-manager.ts
+++ b/src/background/tab-manager.ts
@@ -358,12 +358,25 @@ export default class TabManager {
             windowType: 'normal',
         }))[0];
         if (!tab) {
+            tab = (await TabManager.queryTabs({
+                active: true,
+                lastFocusedWindow: true,
+                windowType: 'app',
+            }))[0];
+        }
+        if (!tab) {
             // When Dark Reader's DevTools are open, last focused window might be the DevTools window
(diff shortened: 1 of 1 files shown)


DovieW:
This is my first PR ever so not sure if I'm doing this right.

This PR is for issue #10716

alexanderby (approved):
Looks great, let's see if it works, thanks!

MikeMazz23:
When I download from the web store this bug still exists in v.4.9.62. However, interestingly, when I download the latest from source - same version 4.9.62 - then build it and load the build output as unpacked, the bug is fixed!

When I load unpacked, I choose the mv3 output folder, but the webstore appears to be installing mv2 version? I'm not sure if that affects the TabManager API calls.

DovieW:
@MikeMazz23 I don't think this change is in a release yet. 
4.9.62 was released on Jan 8.

MikeMazz23:
Ahh I see. I assumed a new version would be created right after the last version was released, that way I can see what version source is up to vs the last released version. It seems development keeps going and we only increment once the release is ready.

PR #10876, opened 2023-03-06
Fix learn.inside.dtu.dk

MikaelFangel:
This fixes the very bright site learn.inside.dtu.dk and makes unreadable content readable again.


MikaelFangel:
Thought I would add two screenshots showing the difference. Can provide more if necessary (:

Before:
before

After:
after

PR #10880, opened 2023-03-06
Update dark-sites.config

TylerPalko:


src/config/dark-sites.config
```diff
@@ -503,6 +503,7 @@ marvil.co
 mastercomfig.com
 mastodon.online
 mastodon.social
+https://messages.google.com/web
```
alexanderby, line 506: Is it dark by default? If not, it shouldn't be in this list. Also, please don't use https:// and preserve the alphabetical order.

alexanderby (changes requested):

PR #10898, opened 2023-03-10
Fix for namecheap.com

felixonmars:
Invert whois guard logo on domain management page.

Before:
Screenshot_20230310_051533

After:
Screenshot_20230310_051601


bershanskiy:
Thank you!

PR #10901, opened 2023-03-11
Fix for source.android.com

harshil21:
They share the same code as the rest, so we can just add the domain


bershanskiy:
Thanks!

PR #10904, opened 2023-03-12
allegro.pl fix

Jckpt:
Fix for auction list hover and some auctions had a light background.
Before:
before
After:
after


Jckpt:
Also fixed the missing padding
before:
before
after:
after

PR #10917, opened 2023-03-15
Updated the Bing Chat theming

TheBrenny:
This updates the Bing Chat theming when using Bing. Specifically, it hides the background image, inverts the top and bottom fades, and recolours the buttons and links while keeping the branding in mind.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2551,7 +2551,7 @@ cib-tooltip,
 .attributions > .attribution-item > .background {
     background: var(--darkreader-neutral-background) !important;
 }
-cib-message[attributions]:not([serp-slot="right-rail"]),
+cib-message[type="text"]:not([serp-slot="right-rail"]),
 button.container > .item-content,
 button.container > .item-content:before,
 .container > .options-list-container,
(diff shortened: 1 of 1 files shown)


bigplayer-ai:
How about the chat prompt/text field it's still bright white.?

TheBrenny:
Lgtm. It's coloured by `var(--darkreader-neutral-background)`, so if that's not styled correctly for the user, it might look off. Otherwise, I've noticed inconsistencies around toggling dark mode while on the page, so refreshing *could* also fix it.

image

bigplayer-ai:
How about the bing chat on bing.com search page (the side panel when searching)? It is still white :(

xilopaint:
@TheBrenny What am I doing wrong? I used your template, but the look ended up like this:

Screenshot 2023-03-23 at 13 11 23

TheBrenny:
@xilopaint Well that was awkward. I 100% reckon that the Bing team made some changes to how they styled things. I remember being confused about how I should target the response chats specifically, and _now_ there's a `type="text"` attribute sitting on the messages. Weird.

Up front though, I don't actually make use of the custom tags, and specific attributes. They're doing this really funky way of having embedded shadow roots or something, which means that I can't target the children items of a `cib-message`.

PR #10947, opened 2023-03-21
Added thepigeon.org to dark-sites.config

CamoCatX:


bershanskiy:
Hi, what is the purpose of this site? It seems to host only a single photo of a bird.

bershanskiy:
The site appears to be defunct (Internet Archive shows some potentially useful content prior to circa 2017, but only placeholder images since then). We generally do not accept fixes for placeholder content since they would not be useful to anyone while they are fixing placeholder content and will break once placeholder content is replaced with some actual content.

CamoCatX:
@bershanskiy This website has no clear purpose, it is just fun.

PR #10948, opened 2023-03-22
Tests: support Firefox browser tests

bershanskiy:


src/inject/index.ts
```diff
@@ -224,4 +225,69 @@ if (__TEST__) {
...
+
+        socket.onmessage = (e) => {
+            function respond(data: any) {
+                socket.send(JSON.stringify({id, data}));
+            }
+
+            const {id, data, type} = JSON.parse(e.data);
+            switch (type) {
+                case 'firefox-eval': {
+                    const result = eval(data);
```
alexanderby, line 268: This should be probably wrapped into
```typescript
if (__DEBUG__) {
    const result = eval(data);
}
```
or
```typescript
if (__TEST__) {
    const result = eval(data);
}
```

I'm afraid it will not pass the review if included into build.

PR #10957, opened 2023-03-23
Added nexangelus.github.io/dbd_PerkFinder/ to dark-sites.config

andre-paulo98:
This is a website a friend and I made.
It's dark by default (there's no option to choose theme either way)

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -549,6 +549,7 @@ netzhack.de
 neundex.com
 neurocore.xyz
 newgrounds.com
+nexangelus.github.io/dbd_PerkFinder/
 nexusmods.com
 nfs.fandom.com
 nfspolska.pl


bershanskiy:
Thanks for the PR!

> This is a website a friend and I made.

We prefer that websites add a "Dark Reader lock" to the page if they can instead of using the config. To add the "lock" please just add `<meta>` tag with `name` attribute `darkreader-lock` to `<head>` tag. That use the following HTML as a reference:
```
<head>
    <meta name="darkreader-lock">
</head>
```

andre-paulo98:
I'll do that then. Didn't know about that feature!

bershanskiy:
For reference:
 - Migrated to https://github.com/nexangelus/dbd_PerkFinder/issues/7
 - Fixed by commit https://github.com/nexangelus/dbd_PerkFinder/commit/c427d980fae06b9a450689df99de0c1d9cd9a9aa

PR #10962, opened 2023-03-25
Update dark-sites.config

TylerPalko:


bershanskiy:
Hi, this site does not appear to be dark:
image

PR #10983, opened 2023-03-28
Fix for needymeds.org

Gitoffthelawn:
Fixes unreadable links and text in header

Example URL:
https://www.needymeds.org

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -14643,6 +14643,15 @@ INVERT
 
 ================================
 
+needymeds.org
+
+CSS
+#header {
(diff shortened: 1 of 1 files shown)


bershanskiy:
@Gitoffthelawn Sorry for the delay (I missed your message). In this particular case the background behind the header appeared to have a better contrast than the header, hence I removed the entire `background` property.

Gitoffthelawn:
> @Gitoffthelawn Sorry for the delay (I missed your message). In this particular case the background behind the header appeared to have a better contrast than the header, hence I removed the entire `background` property.

No problem at all.  What you wrote makes good sense... thanks for the explanation.

PR #10984, opened 2023-03-28
Fix vcalc.com

Gitoffthelawn:
Fixes:

1. Primary logo link in upper-left corner
2. Large bright background in middle of page

Example URL:
https://www.vcalc.com


Gitoffthelawn (commented):
WFM

PR #10986, opened 2023-03-28
Fix for t.bilibili.com

aka-commit:


bershanskiy:
@aka-commit Thanks for the PR. I can't find what was broken before this fix, could you please clarify?

bershanskiy:
I'm not able to verify the fix, unfortunately. I'll merge as is and will fix issues if they arise.

PR #10991, opened 2023-03-28
Update fix for uteka.ru

rakleed:
- Update logo inversion.


bershanskiy:
I can't verify the fix, but it looks reasonable. Merging as is.

PR #10993, opened 2023-03-28
Add fix for www.freeriderhd.com

andreynering:
This game render in a canvas, and the container needs to be white so the result is correct.

## Before

Screenshot 2023-03-28 at 18 53 02

## After

Screenshot 2023-03-28 at 18 53 16

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -23596,6 +23596,15 @@ img
 
 ================================
 
+www.freeriderhd.com
+
+CSS
+#game-container {
+    background-color: white;
+}
+
+================================
+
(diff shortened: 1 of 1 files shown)


bershanskiy:
Thanks for the fix, I just changed `white` to `gray` since white is too high contrast (too bright).

andreynering:
Thanks @bershanskiy.

I would honestly choose a lighter grey, though. `grey` is a bit too dark and changes the characteristics of the game.

What about `#dadada` or `#cdcdcd`? Looks like a good middle term to me.

bershanskiy:
Looks like I changed the PR and merged it prematurely, I updated changed the color to `#cdcdcd` in #11010.

PR #10995, opened 2023-03-29
Fix for mail.google.com

0xallie:
Profile pictures in the "Search mail" dropdown were previously inverted, this uninverts them.


bershanskiy:
I'm unable to verify the presence of the issue, but the fix makes sense.

PR #10997, opened 2023-03-30
Fix for app.tray.io in Filter modes

M108:


bershanskiy:
Hi, @M108 thanks for the PR, I don't immediately see what was broken before, could you please provide some screenshots?

M108:
Issue mostly lies with the connectors themselves in Filter/Filter+ modes.  This is before the fix
image

This is after
image

bershanskiy:
Thanks for explanation!

PR #11005, opened 2023-03-31
Fix for omnicalculator.com

wyrmiyu:
Inversion of .GenericText caused the text to blend into background completely.
Changing it to use --darkreader-neutral-text fixes the issue.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -15428,8 +15428,10 @@ CSS
 
 omnicalculator.com
 
-INVERT
-.GenericText
+CSS
+.GenericText {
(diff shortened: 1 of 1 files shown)


wyrmiyu:
I have tested this with Firefox and Google Chrome (111.0.5563.147) and Firefox (111.0.1) with the following two calculators:

https://www.omnicalculator.com/health/car-heat
https://www.omnicalculator.com/chemistry/average-atomic-mass

To reproduce the issue, you need to enter values to the calculator to get results on the left hand side column, where the .GenericText class is used.

bershanskiy:
Thanks for the explanation about calculator, without it I wouldn't be able to find the issue. This looks great after the fix.

PR #11013, opened 2023-04-01
Add linku.la to dark sites list

Drogobo:
nimi.li, linku.la, drogobo.com

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -250,6 +250,7 @@ dotapicker.com
 draculatheme.com
 dragonage-game.de
 drakewars.com
+drogobo.com
 drunkenslug.com
 drvortex.dev
 dsa-drakensang.de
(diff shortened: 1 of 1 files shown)


bershanskiy:
Hi, I see that one of these is your own site. If I understand correctly, you own the repository https://codeberg.org/Drogobo/pages

Could you please add the following to your page to tell Dark Reader to disable itself?
```html
<head>
    <meta name="darkreader-lock">
</head>
```

bershanskiy:
Hi, I removed two sites because:
 - `drogobo.com` is better served by adding a `<meta>` tag to it
 - `nimi.li` is not dark be default

bershanskiy:
@Drogobo I created a PR to your site'a repository here: https://codeberg.org/Drogobo/pages/pulls/1

Drogobo:
Thank you very much!

PR #11018, opened 2023-04-02
Update fix for lightning.force.com

Gitoffthelawn:
Fixes obsolete syntax for pseudo-element.

Also simplifies hex color.

BTW, might be an opportunity to change `#333` to `${#333}` for `.slds-card`, but I'll leave that decision to others.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -12309,7 +12309,7 @@ lightning.force.com
 
 CSS
 .slds-brand-band,
-.slds-brand-band:after,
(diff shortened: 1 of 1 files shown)


bershanskiy:
`#333` and `#333333` are two different colors, so I reverted that change. The other change looks good.

Gitoffthelawn (commented):
@bershanskiy 
Just trying to learn your style... why do you prefer the long form?  Browser sees it the same way, no? The long form takes up 3 more bytes in extension's storage... I was an assembly language programmer, so I'm always optimizing for bytes! :-)

PR #11021, opened 2023-04-02
Add airconsole.com to dark sites

Yeah-Jack:
Added 2 sites


bershanskiy:
Thanks! wisehosting.com appears to be down, I'll add the other site first and then make a separate commit for wisehosting.com

bershanskiy:
Follow-up PR: #11024

PR #11024, opened 2023-04-02
Add wisehosting.com to dark sites

bershanskiy:
Split out from #11021


bershanskiy:
The site went back online and I was able to verify that it is dark by default.

PR #11043, opened 2023-04-04
Update fix for antagning.se

rappos:
Some selection boxes had text in the same color as the background, this should fix that


rappos:
For me it was this dropdown menu on the main page
image

PR #11052, opened 2023-04-06
Update inversion-fixes.config

UsernameIsAlreadyTaken12345:
Fix for iframes with surveys on different sites

### DIFF
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -15,6 +15,7 @@ twitterwidget
 .sr-reader
 .sr-backdrop
 iframe:fullscreen
+[class="opinary-iframe"]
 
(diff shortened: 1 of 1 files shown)


bershanskiy:
Hi, thanks for this PR. How frequently does this occur?

UsernameIsAlreadyTaken12345:
Hi, I saw this on different sites that includes a simple survey (a single question), e. g. here (at the end of the article - the question itself is as black as the background): [](https://utopia.de/news/bekannter-werbesatz-soll-nun-gleichberechtigung-bringen-doch-die-aenderung-ist-halbgar/)
I saw the same type of survey on a news site (behind a pay wall).

In every case the question was 'invisible'

PR #11063, opened 2023-04-07
UI highlight new toggle menus

alexanderby:


bershanskiy:
Hi, @alexanderby I'm sorry about CI failure: I caused it by enabling more tests on CI recently. To fix it, we should run `npm run test:project -- -u` before merging.

bershanskiy:
Or if we do not want to run these tests on CI, we could revert #11061.

alexanderby:
That's fine, thanks for adding this. I will update the snapshot.

PR #11076, opened 2023-04-08
Tests: check validity of browser tests

bershanskiy:
PR #11060 fixed a number of project tests, but left one out. This PR fixes the last remaining browser test.

### DIFF
--- a/tests/project/__snapshots__/tsconf.tests.ts.snap
+++ b/tests/project/__snapshots__/tsconf.tests.ts.snap
@@ -268,6 +268,52 @@ exports[`TypeScript project config file should parse and resolve correctly: src/
 }
 `;
 
+exports[`TypeScript project config file should parse and resolve correctly: tests/browser 1`] = `
+{
+  "compileOnSave": false,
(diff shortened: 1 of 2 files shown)


bershanskiy:
External dependency block:
```
      node_modules/@jest/environment/build/index.d.ts(401,26): error TS2430: Interface 'JestImportMeta' incorrectly extends interface 'ImportMeta'.
        The types of 'jest.replaceProperty' are incompatible between these types.
          Type '<T extends object, K extends Exclude<keyof T, keyof { [K in keyof T as Required<T>[K] extends ClassLike ? K : never]: T[K]; } | keyof { [K in keyof T as Required<T>[K] extends FunctionLike ? K : never]: T[K]; }>, V extends T[K]>(object: T, propertyKey: K, value: V) => Replaced<...>' is not assignable to type '<T, K extends keyof T>(obj: T, key: K, value: T[K]) => ReplaceProperty<T[K]>'.
            Types of parameters 'object' and 'obj' are incompatible.
              Type 'T' is not assignable to type 'object'.
```

PR #11112, opened 2023-04-13
Add OWA IGNORE IMAGE ANALYSIS for mail.eni.it/owa

aureliocirella:


src/config/dynamic-theme-fixes.config
```diff
@@ -13005,6 +13005,46 @@ CSS
     background-color: ${#cfcfcf} !important;
 }
 
+
+================================
+
+mail.eni.it/owa
+
+IGNORE IMAGE ANALYSIS
+.image-font-png
```
alexanderby, line 13014: Maybe it can be replaced with one of those 3 selectors?
`[class^="image-"]`
`[class^="-png"]`
`[class^="image-"][class^="-png"]`

aureliocirella (commented):
Class list replaced by suggested class selector.

PR #11132, opened 2023-04-16
Update zh-CN.config

nkh0472:
translated the new string @system_dark_mode_chromium_warning and, 
unified the translation of dark.


nkh0472:
conflict with earlier PR #11130

PR #11152, opened 2023-04-18
Fix for fanfiction.net

lerndmina:
The dynamic mode inverted a lot of the page content which causes it to turn an ugly yellow. I've fixed this by using the neutral background var to allow the user to customise what actually is shown.


src/config/dynamic-theme-fixes.config
```diff
@@ -23885,6 +23885,54 @@ body {
 
 ================================
 
+www.fanfiction.net
```
alexanderby, line 23888: Could you please leave just `fanfiction.net`? If so please preserve the alphabetical order.

alexanderby (changes requested):

lerndmina:
Accidentally marked as resolved. I'll fix this soon.

src/config/dynamic-theme-fixes.config
lerndmina: Sure, I've done so now.

bershanskiy (approved):
The www was removed and alphabetic order was fixed.

PR #11155, opened 2023-04-19
Fix: hide new design preview button on mobile

bershanskiy:
This button does nothing on mobile, so we should hide it. Specifically, `isMobile` overrides `previewNewDesign` in `src/ui/popup/components/body.tsx`


bershanskiy:
This added a regression (fixed in #11225). Fortunately, this regression was never released.

PR #11185, opened 2023-04-26
Fix bugs.chromium.org (with a report of an issue)

NguyenASang:
### What I fixed
Actually the whole page, a lot of modifications so I can't remember all of it.

### What I can't fix

Weird background color:

image

Ultra shiny separators:
image

...

### Problem with "shadow-root"
To fix [this website](https://bugs.chromium.org) with bunch of "shadow-root", the most thing I can do is change the root color. Although Dark Reader supports modifying "shadow-root" child elements, it actually doesn't work in some cases. 

Proof:

https://user-images.githubusercontent.com/87893636/234637639-7b3e4592-0558-4478-a9e8-0b9aa1952238.mp4

But somehow, modifying attribute of "input#searchq" still works although it's under a "shadow-root".

I hope you @Gusted can take a look at this.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3295,71 +3295,68 @@ INVERT
 
 bugs.chromium.org
 
-INVERT
-.project-logo
-.anchor
-.menu
-
 CSS
 :root {
-    --darkreader-fix--border: 1px solid ${#ccc} !important;
+    --chops-white: #161616 !important;
+    --chops-black: #ffffff !important;
+    --chops-blue-50: #032742 !important;
+    --chops-blue-75: #021524 !important;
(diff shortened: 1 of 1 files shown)


NguyenASang:
@alexanderby Hello, what about the issue I mentioned ?

bershanskiy:
> Hello, what about the issue I mentioned ?

This issue is caused by Dark Reader not handling deeply nested shadow DOM, I'll post an update once I find a good fix for this.

NguyenASang:
Thank you very much.

bershanskiy:
@NguyenASang I created a PR #11212 fixing this. I provided an explanation of the bug within it. Now I just need to add unit tests and test it for a bit more.

I expect the fix to be available in the next release, which we probably will make available this week.

PR #11196, opened 2023-04-29
Fix for bugs.chromium.org

rakleed:
- Return logo inversion.


NguyenASang:
Not working, I report the issue in #11185 but no answer.


https://user-images.githubusercontent.com/87893636/235335065-00508311-c873-4b7b-9f7d-78387291ca9a.mp4

bershanskiy:
Hi, I looked into this bug yesterday and I have a decent idea about the root cause, but I'm not sure what's the best way to fix it right now. I'll post here later today to let you know about the progress.

PR #11220, opened 2023-05-04
Fix for astroproxy.com

rakleed:
- Add background for QR code for 2FA setup.


bershanskiy:
I could not find the QR code, but the fix looks decent, so merging as is.

bershanskiy:
I found the QR code, updated URL to point to this exact page just in case the selector is not specific enough.

PR #11223, opened 2023-05-04
Fix for digitalextremes.zendesk.com

felixonmars:
Invert logo

Before:
Screenshot_20230504_170103

After:
Screenshot_20230504_170141


bershanskiy:
Thanks!

PR #11240, opened 2023-05-07
Fix for kinhmatanna.com

NguyenASang:
### What I fixed
- Invert logos
- Fixed the z-index of some elements is too low
- Changed some colors to make it easier to see


bershanskiy:
Thanks!

PR #11245, opened 2023-05-09
Update fix for stockbit.com

JJ96869:


bershanskiy:
Hi, thanks for this PR, how may I confirm that these inverts are no longer necessary?

PR #11246, opened 2023-05-09
boardgamearena.com - Fixes for Tooltips and Little Factory

branflaker:
Various boardgamearena.com fixes.
Fixes disappearing cards for Little Factory.
Fixes inverted gradient on the bottom of tooltip.


branflaker:
So the test failed for the dynamic theme formatting, but does not report what's causing it to fail.

PR #11266, opened 2023-05-14
Fix translations for Serbian language

kizule:
One translation was changed to translation on Serbian language, and one was replaced with English version/removed because it was wrong (letters were replaced with Serbian Cyrillic).


bershanskiy:
Thanks!

PR #11298, opened 2023-05-21
Refactor: better regular expressions for URL analysis

bershanskiy:


alexanderby (commented):
Looks like a good change, let's trying to merge after this release?

alex-kinokon:
If you have a previously entered rule that fails this new regular expression, the text list [silently breaks](https://github.com/bershanskiy/darkreader/blob/8e4463c593a71cfd9791814738b4b50d2fdb33b5/src/ui/popup/components/site-list-settings/index.tsx#L31) and becomes read only. If you try to edit or add a new item, nothing will happen and there is no error message.

PR #11299, opened 2023-05-21
Add blog.sourceloc.net to dark sites

sourcelocation:


bershanskiy:
Hi, I'm unable to reach blog.sourceloc.net . We generally do not add sites which are not available on internet. Also, if you are the owner of the site (since you have the same user name), I would recommend using ["Dark Reader lock" instead](https://github.com/darkreader/darkreader/blob/main/CONTRIBUTING.md#disabling-dark-reader-on-your-site).

sourcelocation:
hi, sorry for late response. It just so happened that the website was down for 1 hour back then lol. And yes, it is my website. Thanks for suggestion

bershanskiy:
Are you going to use "Dark Reader lock" thing or would you like me to reopen this issue?

PR #11307, opened 2023-05-22
Fix: work around Alarms API bug in Chromium

bershanskiy:


alexanderby:
Looks very interesting, could you please describe what does it do?

bershanskiy:
First, some background:
 - Chromium implements `chrome.alarms` with a dedicated `AlarmManager` class, which internally stores all alarms in (somewhat) persistent storage and schedules the next closest (upcoming) alarm with [`base::OneShotTimer`](https://github.com/chromium/chromium/blob/main/extensions/browser/api/alarms/alarm_manager.h#L242)
 - `base::OneShotTimer` is based on time delta and may drift on some platforms (e.g., desktop Windows and Linux) when computer suspends power. Chromium has a `base::WallClockTimer` which [is not susceptible to drift](
https://github.com/chromium/chromium/blob/main/base/timer/wall_clock_timer.h#L27). `base::WallClockTimer` wraps `base::OneShotTimer` and listens to power resume and [reschedules it accordingly](https://github.com/chromium/chromium/blob/main/base/timer/wall_clock_timer.cc#L47).
 - since `AlarmManager` has a single instance of `base::OneShotTimer`, it stores only the upcoming alarm and upon its expiry fires an event and finds the next upcoming alarm time. Therefore `chrome.alarms.create('alarm-fixer', {when: 0})` triggers the alarm check (schedules an alarm in the past and forces time check).
 - I filed a CL/PR for replacing `base::OneShotTimer` with `base::WallClockTimer`, but reviewer stalled since apparently there is a potential for API ambiguity. (Can't find the bug and CL/PR I filed).
 - Apparently (I did not conclusively validate), all extensions under the same profile share a single `AlarmManager` instance so extensions may trigger alarms of other extensions. Author of uBlock Origin noticed this bug when he attempted to port uBlock Origin to MV3 (called uBO Minus, I think)
 - Some other people reported this bug as well, e.g., [this](https://bugs.chromium.org/p/chromium/issues/detail?id=1245934)

This PR creates a dummy alarm and thus forces `AlarmManager` re-check. One day I managed to reproduce the bug and test the fix, but the next day I failed to reproduce the bug and therefore am not sure the fix works as intended.

PR #11314, opened 2023-05-24
Fix for onedrive.live.com

felixonmars:
Fix inverted command bar when a file or folder is selected.

Before:
Screenshot_20230524_235755

After:
Screenshot_20230524_235905

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -15943,6 +15943,15 @@ body {
 
 ================================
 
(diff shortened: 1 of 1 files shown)


timjobutler:
I have updated to the latest version of Dark Reader (4.9.64) but this issue is persisting. Running the latest version of Chrome (Version 114.0.5735.134).

Any clues as to why the command bar is still inverted?

DYNAMIC-MODE-OneDrive-Action-Bar-White-Out

PR #11333, opened 2023-05-29
fix: Cannot read properties of undefined (reading 'contains')

paddingme:


bershanskiy:
Hi, thanks for the PR! It looks good, however I'd like to also know if there are specific repro steps for this bug.

bershanskiy:
I looked over the PR in a greater detail, and created #11342 instead. It also checks for falsy values, but a bit differently.

PR #11334, opened 2023-05-29

jojo2massol:
Added mafreebox.free.fr (internet router local domain name):
- fixed background image
- fixed inverted icon colors
- fixed start button image color

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -13172,6 +13172,22 @@ td {
 
 ================================
 
+mafreebox.freebox.fr
+
+CSS
+* {
+filter:none !important;
+}
+body {
+    background-image: linear-gradient(rgb(14 14 14 / 50%), rgb(40 40 40 / 50%)),url(../resources/images/fbx/bg_freeboxos.svg);
+}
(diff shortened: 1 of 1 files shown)


jojo2massol:
image
Before / After

src/config/dynamic-theme-fixes.config
```diff
@@ -13172,6 +13172,31 @@ td {
...
+
+CSS
+* {
+filter:none !important;
+}
+.desktop-icon-outer {
+    color: #fff;
+}
+body {
+    background-image: linear-gradient(rgb(14 14 14 / 50%), rgb(40 40 40 / 50%)),url(../resources/images/fbx/bg_freeboxos.svg);
```
bershanskiy: I believe an absolute URL would be more reliable

bershanskiy (changes requested):

src/config/dynamic-theme-fixes.config
jojo2massol: Strangely, the full path doesn't work.
The default value on the website is already a relative path.

bershanskiy (approved):

PR #11347, opened 2023-05-31
Add codeberg.org to dark sites

dwaxe:
Codeberg is a popular site, and its theme is dark by default.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -159,6 +159,7 @@ clashofstats.com
 clt.gg
 cncnet.org
 co.wukko.me
+codeberg.org
 codepen.io
 coder.com
 codesandbox.io


bershanskiy:
The site landing page appears light for me. Is there a way to add dark theme to the landing page? It's supposedly OSS, so it should be possible, but I can't find the repo.

dwaxe:
https://codeberg.org/ is dark from what I can see. Is there a different landing page you reached from Google or something?

bershanskiy:
I'm talking about the logged-out landing page (the one you see if you open a new incognito tab without logging in).

dwaxe:
@bershanskiy Could you share a screenshot? This is what it looks like when I open codeberg.org
image

dwaxe:
Is this still a valid PR?

archeite:
> @bershanskiy Could you share a screenshot? This is what it looks like when I open codeberg.org image

I believe codeberg is setting it based on your `prefers-color-scheme` (browser theme), so the landing page is light/dark depending on your current theme

dwaxe:
Do we have an existing policy for websites that already respond appropriately to `prefers-color-scheme`?

PR #11364, opened 2023-06-04
Update fix for wanikani.com

notpushkin:


bershanskiy:
Do I need to create an account to see the fix? Could you please provide a screenshot?

PR #11370, opened 2023-06-05
Fix for postcrossing.com

dedif:
All the white parts of country flags on Postcrossing.com were inverted to black. This fix prevents that, so that the site displays correct flags.

(The change with westlaw.com wasn't my intention; I think Git wasn't properly synchronised? I don't really understand Git extremely well, could you tell me what happened?)

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -20559,6 +20559,20 @@ INVERT
 
 ================================
 
+westlaw.com
+
+CSS
+.co_hl.blue,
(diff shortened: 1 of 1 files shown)


bershanskiy:
I do not see broken flags (they appear fine on the home page). Could you please clarify what this fixes? Thanks!

dedif:
Hello, it seems that currently the flags are displayed correctly now. They used to be incorrect when I wrote this patch quite a while ago, but this patch isn't necessary anymore

PR #11422, opened 2023-06-19
Update inversion-fixes.config - Snowflake.com

paperlinguist:
Added inversion for user avatar in Filter mode for snowflake snowsight UI.


paperlinguist:
image
 vs
image

PR #11425, opened 2023-06-20
Fix for drive.google.com

NguyenASang:
## What I fixed
- Remove unnecessary fixes
- Fix some weird color inversion
- Fix some elements


src/config/dynamic-theme-fixes.config
```diff
@@ -6712,77 +6712,56 @@ INVERT
...
-img[src*="empty_state_trash"]
+div[style*="add_black"]
+div[aria-label="Show filter toolbar"]
+div.settings-view-general div[style]:last-child
+nav div[style]:last-child:not(div[role="group"])
 
 CSS
-span[data-type="spelling"] {
-    mix-blend-mode: normal !important;
+* {
```
alexanderby, line 6721: Did you try `:root` instead of `*`? Custom CSS properties are inherited down the elements tree.

NguyenASang: I tried it, I use `*` because the color is set under a unique id, also I can't guarantee that id is the same for everyone

image

NguyenASang:
@alexanderby I also fixed the `xiph.org` site wasn't sorted alphabetically

NguyenASang:
@bershanskiy take a look at this.

PR #11438, opened 2023-06-25
Finalize the fix for bugs.chromium.org

NguyenASang:
## What I fixed
Rest of things that can't be fixed in #11185 

## Feature request

Take a look at this image:

image

Let say `dialog-content` isn't a unique class, the path to that element should be: `chops-dialog .dialog-content`

But that's only true when `.dialog-content` is a unique class in `shadow-root`. Why don't we add a way to make the path easier ?

Like this: `chops-dialog > shadow-root > dialog > .dialog-content`

Or even: `chops-dialog > dialog > .dialog-content`

That would be easier to work on some **_complex_** cases

## Additional context
Thank you @bershanskiy , everything works smoothly. Also pls close #11196


NguyenASang:
@bershanskiy Do you think my request is possible ?

PR #11453, opened 2023-06-28
fix the wrong search result background in Bing

WitherZuo:
Bing's search result, now on sometimes, will have light and white background when use dynamic theme mode, like this:  

Screenshot 2023-06-28 at 22-20-52 Bing - 必应

This pull request may fix it.


bershanskiy:
Is this bug specific to 4.9.64?

WitherZuo:
> Is this bug specific to 4.9.64?

No, it looks like a bug (or a feature?) with Bing.

Issue #11446, opened 2023-06-27
[Broken Website] Pushbullet.com


bossanovaorca:
### Prerequisites

- [X] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this website issue to avoid opening a duplicate.
- [X] I can reproduce this website issue in a new, unmodified web browser profile with Dark Reader installed as the only extension.

### Website Issue Description

Website is blank and will not load with Dark Reader enabled

### Website Address

https://www.pushbullet.com/

### Steps To Reproduce

1. Enable Dark Reader
2. Visit Pushbullet.com

### Operating System

Windows 10, Windows 11, macOS

### Web Browser name and version

Firefox 116

### Dark Reader version

4.9.64

### Additional Context

```
error message:can't access property "insertRule", document.styleSheets[0] is undefined stack:onecup.css@https://www.pushbullet.com/js/pushbullet.js:7989:9
@https://www.pushbullet.com/js/pushbullet.js:11890:8
onecup.css@https://www.pushbullet.com/js/pushbullet.js:7982:5
@https://www.pushbullet.com/js/pushbullet.js:11889:6
@https://www.pushbullet.com/js/pushbullet.js:12385:4
 type:unhandled
```

PR #11455, opened 2023-06-29
Fix: broken `document.styleSheets` proxy (#11454)

### LINKED ISSUE #11454: [Broken Website] Dropbox Paper
### Prerequisites

### Website Issue Description

Loading any doc from Dropbox Paper (https://paper.dropbox.com / https:paper.new) with Dark Reader enabled is broken as of the last release.

### Website Address

https://paper.dropbox.com

### Steps To Reproduce

1. Sign in to Dropbox Paper
2. Create a doc (or edit an existing doc)
3. 💥

### Screenshots

_No response_

### Operating System

macOS

Chrome (stable and beta), Brave (stable), Arc (stable), etc

### Dark Reader version

v4.9.64

### Additional Context

[v4.9.63](https://github.com/darkreader/darkreader/releases/tag/v4.9.63) works just fine!

bobthecow:
A logic error introduced in bec13d4d1 breaks the `document.styleSheets` proxy, filtering out _all_ stylesheets and style tags, preventing sites from dynamically accessing and altering their own styles.

Fixes #11454 and #11446

Before:

After:

### DIFF
--- a/src/inject/dynamic-theme/stylesheet-proxy.ts
+++ b/src/inject/dynamic-theme/stylesheet-proxy.ts
@@ -107,7 +107,7 @@ export function injectProxy(enableStyleSheetsProxy: boolean, enableCustomElement
             const docSheets: StyleSheetList = documentStyleSheetsDescriptor!.get!.call(this);
 
             const filteredSheets = [...docSheets].filter((styleSheet) =>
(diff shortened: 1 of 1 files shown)


bobthecow:
IMO the proximate cause here is TypeScript. The mess of type coercion in the offending change made it too hard to reason about the logic, so a set of parenthesis was dropped. It was meant to be `A && !(B && B.C)` but instead was written `A && !B && B.C` (which is always false 🙃). I initially added the missing parenthesis, but that was still too hard to read, so I landed on this fix.

bershanskiy:
Thanks for the PR, I confirmed the bug and the fix for it, I will add a unit test to the fix and merge, then we will probably publish `4.9.65` with this fix.

PR #11500, opened 2023-07-11
Fixes underlines for theverge.com

stromfeldt:
It's virtually impossible to see underlines of links within articles on theverge.com (they're black on the dark background).

Xnapper-2023-07-11-19 24 46

This matches the underline colour to the text colour.

Xnapper-2023-07-11-19 25 02

### DIFF
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -2394,6 +2394,15 @@ img
 
 ================================
 
+theverge.com
+
+CSS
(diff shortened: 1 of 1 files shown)


src/config/inversion-fixes.config
```diff
@@ -2394,6 +2394,15 @@ img
 
 ================================
 
+theverge.com
+
+CSS
+.\[\&_a\]\:shadow-underline-black a {
```
alexanderby, line 2400: What an interesting selector.

PR #11511, opened 2023-07-15
Fix dynamic theme for Reddit comments

Nanobiologist:
- Fix comment-tree HTML element being grey text on white background
- Edit coloring of buttons near post
## Before:
image
## After:
image


Nanobiologist:
Please only merge the additions

PR #11517, opened 2023-07-16
Add checkout.steampowered.com to the global list of dark sites

Tiagoquix:


Myshor:
Maybe it was removed because when I enter checkout.steampowered.com I am redirected to https://store.steampowered.com/

Is there something necessary to be done to have it opened as checkout.steampowered.com?

PR #11532, opened 2023-07-22
Bump dependencies

bershanskiy:
This does not result in changes to built output.


package-lock.json
```diff
@@ -3049,6 +3057,15 @@
         "node": "^4.5.0 || >= 5.9"
       }
     },
+    "node_modules/basic-ftp": {
```
bershanskiy, line 3060: I'm not really happy about inclussion of this module in the lock (and FTP in general), but we are not actually using it anywhere in our code. It is depended upon by Puppeteer used only in testing and it should not be reaching this code at all. (`basic-ftp` <- `get-uri` <- `pac-proxy-agent` <- `proxy-agent` <- `@puppeteer/browsers`)

PR #11609, opened 2023-08-16
Fix for podcast.adobe.com

rappos:
Added fixes for the play button:
image
and for the logos in navbar and footer:
image


bershanskiy:
Original title: Button and logo fixes for podcast.adobe.com

PR #11619, opened 2023-08-20

poxynam:
Fixed math and symbols not showing properly on sanomapro.fi. Inverts all images because of lazyness, but it's fine.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -19266,7 +19266,15 @@ CSS
 sanomapro.fi
 
 INVERT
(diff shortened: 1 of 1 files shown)


bershanskiy:
Inverted images look pretty bad. There is definitely a better way to fix this. Could you please link to a page with which has this problem? I might be able to find a better selector.

Screenshot from 2023-08-31 11-18-21

poxynam:
To see any of the actual math editor stuff you need to login with a student account. I could probably find a better selector myself, but I really just wanted to do my homework without burning my eyes.

PR #11620, opened 2023-08-20
Update fix for discord.com

aliyss:
Discord.com Status Symbols were messed up. This fixes them.


aliyss:
Oh just for some information, I'm using * purposely to match everything in the mask, since those are svgs, so the things below are circles, ellipses and other such things, that need to render correctly. Most of these are circles and ellipses, but can also be rects, due to mobile users.

PR #11642, opened 2023-08-28
Update dark-sites.config

Andrew-J-Larson:
Website has a dark theme option in settings.


bershanskiy:
This PR erroneously ignored blog pages[1] so I created #11685 to fix this. See also #11683 for original bug report.

[1] https://www.freecodecamp.org/news/

Issue #10853, opened 2023-03-01
[Broken Website] bing.com/chat is not in dark mode.


bigplayer-ai:
### Prerequisites

- [X] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this website issue to avoid opening a duplicate.
- [X] I can reproduce this website issue in a new, unmodified web browser profile with Dark Reader installed as the only extension.

### Website Issue Description

bing.com/chat is broken, It does not show dark theme when viewing it. I tried using other extensions, I am experiecning the same issue. I tried to change different filtering settings in dark reader with no luck fixing the issue.
Check the photo attached.
![bing.com/chat is not in dark mode.](https://user-images.githubusercontent.com/119108387/222200691-963fc5fd-62b5-4a16-b1bd-dc1e4db7ca4b.png)

![Clip_20230301_203232](https://user-images.githubusercontent.com/119108387/222232805-60c03b47-e9b2-46f7-b3cc-7550e886d7a6.png)

### Website Address

bing.com/chat

### Steps To Reproduce

1) Enable dark reader
2) Open bing.com/chat
3) See bing.com/chat is not in dark mode.

### Screenshots

![bing.com/chat is not in dark mode.](https://user-images.githubusercontent.com/119108387/222200691-963fc5fd-62b5-4a16-b1bd-dc1e4db7ca4b.png)

### Operating System

Windows 11 22H2 22621.1265

### Web Browser name and version

Microsoft Edge Version 110.0.1587.57 (Official build) (64-bit)

### Dark Reader version

 Dark Reader Version: 4.9.62

alexanderby:
Works good now.

PR #11648, opened 2023-08-29
Fix bing.com ai chat tab and clean up bing.com css

Amnesiac9:
Fixes issues with the chat tab on bing.com
-black text on dark background,
-issues with invisible text
-invisible boxes
-fixes the tooltips.

Also cleaned up the css grouping to fix a few typos and reduce repeated lines.


Amnesiac9:
Would fix issue https://github.com/darkreader/darkreader/issues/10853

PR #11649, opened 2023-08-29
Fix for laurelroad.key.com

grablair:
Fixes the account settings page, which has a white background and white text while in dark mode

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -12776,6 +12776,9 @@ CSS
 
 laurelroad.key.com
 
+INVERT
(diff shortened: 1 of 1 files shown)


bershanskiy:
Thanks for the fix. I had also fixed the site logo on the login page. Unfortunately, I am not able to test either change on the internal page since I do not have an account on the site. I would appreciate if you could check the changes on the internal pages.

I'll merge the changes once you approve them.

image

grablair:
Thanks for that!

Checked on the internal site and they use a different image path, so I added the class of the logo element as well:

Screenshot 2023-08-30 at 15 42 32

That image shows the internal logo inverted and the accounts settings page.

Issue #11625, opened 2023-08-21
[Bug] Possible incorrect colouring of SVG images


Gitoffthelawn:
### Prerequisites

- [X] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this bug to avoid opening a duplicate.
- [X] I can reproduce this bug in a new, unmodified web browser profile with Dark Reader installed as the only extension.
- [X] I understand I need to use the [Broken Website Report template](https://github.com/darkreader/darkreader/issues/new?assignees=&labels=Broken+Website&template=broken-website-report.yml&title=%5BBroken+Website%5D+Replace+with+title) if this bug I am reporting occurs on a single website.

### Bug Description

***(Please note that this isn't intended as a report for a particular website, but is intended as a report regarding a possible larger issue.)***

I have noticed that sometimes Dark Reader has issues properly colouring SVG images.  Finding a reliable example to present in an issue report has been a challenge... until now.

Take a look at https://keepa.com/ while using Dark Reader in dynamic dark mode.

Notice the letter `p` in the logo in the upper-left corner is dark while all the other letters are light.

Taking a look at the upstream HTML and CSS, once can see that each letter in the word `Keepa` is composed using individual paths within 2 `g` elements within a parent SVG element.  What's helpful here is that, unless I'm not seeing something, each letter is created using the same colours.  Thus, DR should use the same colours for its output.  But DR uses a different colour for the letter `p` than for all the other letters.

### Website Address

https://keepa.com/

### Steps To Reproduce

1. Enable DR dynamic dark mode
2. Go to https://keepa.com/
3. Notice the logo in the upper-left corner

### Expected Behavior

All letters in the logo to be the same colour.

### Actual Behavior

The letter `p` is a different colour.

### Screenshots

![up](https://github.com/darkreader/darkreader/assets/7672848/acab0b55-6fe6-4dc7-90f8-98c4c2c3e7cb)

### Operating System

Linux

### Web Browser name and version

Firefox 115

### Dark Reader version

4.9.65

romiljain5:
Let me check

romiljain5:
Any idea on how do we select colors for dark mode for SVG, You can refer inline-style.ts file

Gitoffthelawn:
@romiljain5 Romil, I'm thinking about the possible side-effects of increasing the SVG size limit.  Do you have any thoughts on the topic?

romiljain5:
[...]

PR #11664, opened 2023-09-03
#11625: Fix SVG issue

romiljain5:
- Update SMALL_SVG_LIMIT to 38 as the size of p SVG icon in keepa was more then given SMALL_SVG_LIMIT
- Also tested build, Working fine after this change

#11625


Gitoffthelawn (commented):
Thanks!  Any thoughts on impact of this change (besides the intended positive effect)?

Issue #11669, opened 2023-09-05
[Broken Website][Dynamic] stackexchange.com - Inline code-blocks


TriMoon:
### Prerequisites

- [X] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this website issue to avoid opening a duplicate.
- [X] I can reproduce this website issue in a new, unmodified web browser profile with Dark Reader installed as the only extension.

### Website Issue Description

The background of **inline-code blocks** is too dark to notice when using the dynamic-theme.

### Website Address

unix.stackexchange.com

### Steps To Reproduce

1. Visit https://unix.stackexchange.com/a/755807/550161
1. It should show as the attached "Before" image in screenshots below.
1. Open "Developer Tools" in Darkreader menu.
1. Press <kbd>Ctrl</kbd>+<kbd>F</kbd>, then type <kbd>stackexchange</kbd>.
1. Insert the below code just above "IGNORE INLINE STYLE" leaving a blank line in between that line and the code below:
   ```css
   .s-prose :not(.s-code-block) > code {
   	background-color: #404040 !important;
   }
   ```
1. Refresh the page.
1. It should show as the attached "After" image in screenshots below.

### Screenshots

- Before:
![image](https://github.com/darkreader/darkreader/assets/1723378/d56ce7a9-1d8f-4fd5-980c-a942df1ccbe6)

- After: 
![image](https://github.com/darkreader/darkreader/assets/1723378/842b27f3-b11d-4b0e-a884-6bae19a19c5f)

### Operating System

Kubuntu 23.04

### Web Browser name and version

Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/118.0 ID:20230904002824

### Dark Reader version

4.9.65

### Additional Context

My propsed fix:
```diff
diff --git a/src/config/dynamic-theme-fixes.config b/src/config/dynamic-theme-fixes.config
index 79594e0d..8fde9f68 100644
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -20772,6 +20772,9 @@ div.h100.bar-lg.p-bs-wrapper:has(a[href="https://stackoverflow.co/teams"]) {
 ::-webkit-scrollbar-thumb {
     background: #313537 !important;
 }
+.s-prose :not(.s-code-block) > code {
+    background-color: #404040 !important;
+}
 
[...]

PR #11670, opened 2023-09-05
Update fix for stackexchange.com

TriMoon:
Fixes: #11669

 - Fix inline code-blocks


bershanskiy:
Hi, this PR actually reduces the contrast making the text harder to read. Is there any specific benefit to this change?

TriMoon:
@bershanskiy  See the screenshots in the linked issue in OP: #11669

TriMoon:
@bershanskiy 
> Hi, this PR actually reduces the contrast making the text harder to read. Is there any specific benefit to this change?

How does it **REDUCE** the contrast, when the current uses `var(--darkreader-bg--black-075)` = `#262a2b`, which clearly is lower as `#404040` which this PR uses.
Eg. this PR **INCREASES** contrast instead...

PR #11678, opened 2023-09-07
Update fix for lightningmaps.org

LeviWald:
- corrected colors of radar map (optional setting for US only)


bershanskiy:
Hi, I can't find the issue this is fixing, could you please upload screenshots before and after the fix? Thanks!

PR #11681, opened 2023-09-09
Update fix for web.whatsapp.com

aliyss:
Without Edit:
image

With Edit:
image

Same is with incoming tails


paniash:
Good to see this change but there does exist a native dark theme for whatsapp web FYI.

aliyss:
> Good to see this change but there does exist a native dark theme for whatsapp web FYI.

I know, but I prefer the same theme throughout all websites. And my firefox is semi transparent, so if I have different types of dark modes for each website, then the transparency doesn't look good. I prefer having contrast to 150%.

PR #11683, opened 2023-09-09
Removed `freecodecamp.org` from global dark sites list

paniash:
https://freecodecamp.org is not dark by default

Eg: https://www.freecodecamp.org/news/how-to-sync-your-fork-with-the-original-git-repository/


bershanskiy:
Thanks for the report, I would like to make sure Dark Reader knows that actual coding pages are dark by default, while changing the blog pages. I created #11685 with this info. Would you be interested in testing that change (e.g., finding any other pages which should be inverted or ignored)?

PR #11685, opened 2023-09-09
Add freecodecamp.org with some restrictions to dark sites

bershanskiy:
Related to #11642 and #11683.


bershanskiy:
Possible site to consider: https://forum.freecodecamp.org/

PR #11691, opened 2023-09-11
Add fix for yandex.eu

illtellyoulater:
Fixes captcha sequence not visible when dark mode on (black on black).
Yes it's yandex.eu not yandex.ru.


illtellyoulater (commented):
all good to me :) can this be merged or is there anything else I should do?

PR #11702, opened 2023-09-15
Add sky.shiiyu.moe to dark sites

0xallie:


bershanskiy:
Thanks for the PR, however https://sky.shiiyu.moe/api does not appear to be dark.

PR #11716, opened 2023-09-20
Fix for www.kleinanzeigen.de

sp00n:
The modal dialoge when opening the user stats was still white

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -25309,6 +25309,15 @@ INVERT
 
 ================================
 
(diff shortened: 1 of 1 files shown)


bershanskiy:
I was not able to find the issue this PR fixes, but the structure of the fix look OK, so merging as is.

sp00n:
Yeah, you need to be logged in and a) check the ratings other people gave you, and b) when giving a rating to anther user after you sold/bought something.
So nothing that was easily checked.

This was the first instance, the second one looks basically the same (just another CSS selector):

image

PR #11798, opened 2023-10-09
Added Selenized Color Scheme

TheWanderer1983:
Selenized light and dark color scheme.

Fixed missing separator.

### DIFF
--- a/src/config/color-schemes.drconf
+++ b/src/config/color-schemes.drconf
@@ -90,6 +90,18 @@ text: #586e75
 
 ================================
 
+Selenized
+
+DARK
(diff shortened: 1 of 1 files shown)


TheWanderer1983:
Solarized has been added under colors schemes. Thank you. There is another color scheme that is based of Solarized called Selenized. https://github.com/jan-warchol/selenized

Request adding Selenized as another color scheme.

Why would you want it? Here are details on its improvements over Solarized. https://github.com/jan-warchol/selenized/blob/master/whats-wrong-with-solarized.md

PR #11808, opened 2023-10-11
Change API to allow multiple dynamic theme fixes

PascalPixel:


alexanderby:
Hi @PascalPixel! Could you please show an example why do you need this?

PR #11859, opened 2023-10-27

trobblob:
Some text colors on Dynamics CRM websites stay black when turning on dark mode, so I fixed that.


alexanderby:
Hi, did you check that it works?

src/config/dynamic-theme-fixes.config
```diff
@@ -99,6 +99,15 @@ INVERT
 
 ================================
 
+*.crm*.dynamics.com
```
alexanderby, line 102: Does this URL pattern work?

src/config/dynamic-theme-fixes.config
```diff
@@ -99,6 +99,15 @@ INVERT
...
+
+CSS
+.ms-Label {
+    color: var(--dark-neutral-text);
```
alexanderby, line 106: Maybe it should look like
```
```

alexanderby (approved):

src/config/dynamic-theme-fixes.config
alexanderby: Made an Approve by accident

alexanderby (changes requested):

src/config/dynamic-theme-fixes.config
trobblob: This URL pattern worked when I tested it with the dark reader dev tools.

src/config/dynamic-theme-fixes.config
trobblob: I had no problem without `!important`. But, if it's better to use `!important`, I can make that change.

alexanderby: Please update the variable and the !important, check if it works and I will fix the conflict (probably similar lines were edited in another Pull Request by someone else).

alexanderby (approved):

PR #11883, opened 2023-11-07
Fix for zhihuishu.com

DeepChirp:
Before:
图片

After:
图片


Myshor:
Can you share some URL with the example? The CSS fix looks fine but it would be nice to check it live before merge.

DeepChirp:
> Can you share some URL with the example? The CSS fix looks fine but it would be nice to check it live before merge.

But this webpage must register an account

Myshor:
Merging only because it seems to be nicely class selected.

PR #11887, opened 2023-11-08
Invert sheet music images

wolf20482:
Add rule to invert sheet music images (any image with URL beginning with https://musescore.com/static/musescore/scoredata/)


wolf20482:
Preview:
image

PR #11888, opened 2023-11-09
Update dynamic-theme-fixes.config for Web of Science

sohailssangha:
Web of Science search function dynamically marks search text with yellow background color. This creates bad contrast with default color of article headings in purple text color. Which makes reading the highlighted text difficult to read in dark mode.

This fix is to solve the visibility problem. 
Specifically changes the color of text that is dynamically marked to darkreader-neutral-text.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -25844,7 +25844,7 @@ www.webofscience.com
 CSS
 .hh, .labl, mark{
     color: var(--darkreader-neutral-text) !important;
-    background-color: ${COLOR} !important;
+    background-color: gray !important;
 }
 
 ================================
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -25839,6 +25839,16 @@ IGNORE INLINE STYLE
 
 ================================
 
+www.webofscience.com
+
+CSS
+.hh, .labl, mark{
+    color: var(--darkreader-neutral-text) !important;
+    background-color: ${COLOR} !important;
```
alexanderby, line 25847: Please replace `COLOR` with the actual source color (before the inversion), like white or black, or use `var(--darkreader-neutral-background)`

alexanderby (changes requested):

src/config/dynamic-theme-fixes.config
sohailssangha: Thanks, changed color to gray. Any white-black, or user var do not provide the correct contrast. Please refer to images below where the search text is "artificial intelligence"

image

image

alexanderby (approved):

PR #11926, opened 2023-11-20
Fix for dice.com

KairuByte:
Multiple fixes to get dice.com to display properly.


KairuByte:
One thing to note, I am not 100% certain the site is completely themed, but I did some digging and didn't find anything amiss.

I will point out there are some graphical oddities, however these exist in the original theme, so I did not go to great lengths to handle them.

PR #11942, opened 2023-11-25
fix(typst.app): made copy/link buttons visible

Andrew15-5:
https://typst.app/docs/packages/:

Before:

image

After:

image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -23497,6 +23497,16 @@ CSS
 
 ================================
 
+typst.app
+
+CSS
+.package-list > tbody:nth-child(2) > tr > td:nth-child(4),
+.package-list > tbody:nth-child(2) > tr > td:nth-child(5) {
+    filter: invert();
+}
+
+================================
+
 ubank.com.au
 
 CSS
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -23497,6 +23497,16 @@ CSS
 
 ================================
 
+typst.app
+
+CSS
```
alexanderby, line 23502: For some reason I can't edit the file, please use the following config instead:
```
typst.app

INVERT
img[src$=".svg"]

```

alexanderby (changes requested):

src/config/dynamic-theme-fixes.config
Andrew15-5: Does `INVERT` literally add `filter: invert();`? Or is it something different?

alexanderby: Yes, this is an equivalent of `invert(1)` (not sure if it works without a number in every browser). You can leave the long selectors if they work better (but for INVERT section each selector should be on a separate line without commas).

Andrew15-5: > for INVERT section each selector should be on a separate line without commas

Yes, I've seen this before. (But I forgot about it all together and used pure CSS.)

alexanderby (approved):
Beautiful 👍

PR #11960, opened 2023-11-29
meteo.imgw.pl fixes

FadeMind:
https://meteo.imgw.pl/
map font and logos in footer

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -14870,8 +14870,7 @@ meteo.imgw.pl
 
 INVERT
 .meteo-chart > div > canvas
-.ft-copyright div div img
-#map-container span
+#map-container svg
 
 CSS
 .menu-close img {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -14870,6 +14870,8 @@ meteo.imgw.pl
 
 INVERT
 .meteo-chart > div > canvas
+.ft-copyright div div img
+#map-container span
```
Myshor, line 14874: Again I will complain... You inverted everything on the map - even nice icons are inverted.
Invert breaks Szczecin as it is now not visible because dark... I propose to invert map:
#map-container svg 😉

Myshor: obraz
doesn't look it better? :)

Myshor (changes requested):
Logos in footer can not be better fixed so no comment for that.
Fix the map and I can merge.

Myshor (approved):

PR #11981, opened 2023-12-08
chore(deps): bump rollup to 4.7.0

kxxt:
rollup 4.7.0 added riscv64 support: https://github.com/rollup/rollup/pull/5288

Updating rollup makes it possible to build darkreader on riscv64 linux.


alexanderby:
Hi @kxxt, we use our own tool for upgrading the dependencies, you can check it out now.

PR #11982, opened 2023-12-08
Applies invert to Google Drive AODocs plugin button

ThomasAllan:
Included in the invert a plugin button for AO Docs in Google Drive.


ThomasAllan:
Before 
image
After 
image

PR #11992, opened 2023-12-09
Correct battlelog.battlefield.com dark site matching

Tajfun403:
In the `battlelog.battlefield.com` domain, only `battlelog.battlefield.com/bf4` subpage is in dark mode. Other major subpages - `battlelog.battlefield.com/bf3` and `battlelog.battlefield.com/bfh` - are white theme.


Myshor:
Confirmed information. Merging fix.

PR #11998, opened 2023-12-10
Fixed consultant.ru unreadable text

filadog:
A popular website for laws and regulations in Russia, improved readability a bit.

### Before:
Web capture_10-12-2023_5917_www consultant ru

### After:
Web capture_10-12-2023_5853_www consultant ru


Myshor:
Original page has blue links:
obraz

That's the reason why Dark reader still uses blue for links.
Making all links with <a> tag the same color as normal text is unacceptable to me for a merge.

PR #11999, opened 2023-12-10
Add rusherhack.org to dark-sites.config

tycrek:


Myshor:
Not sure it should be added. When I clicked on login there was white dialog with disabled Dark Reader.

Issue #10225, opened 2022-11-01
shipstation.com tags get inverted/color shifted


Amnesiac9:
On the website shipstation.com the colorized "tags" on orders get colorized by DarkReader. Is there a way to disable these tags from being color shifted so that they stay unaltered?

![image](https://user-images.githubusercontent.com/8893451/199335473-13d4687b-1d6d-4cb4-a5dc-69a0e9a004be.png)

If there's some way to add this to the dev section of my darkreader for this website that would be amazing. 

From the inspector, the part of the code DarkReader is inverting:

`<div class="pill-18Ymbb9 tag-3f6XHbY pill-3HIsaht collapsed-1G04emq" style="background: rgb(255, 153, 204); --darkreader-inline-bgcolor: #591331; --darkreader-inline-bgimage: none;" data-darkreader-inline-bgcolor="" data-darkreader-inline-bgimage=""></div>`

Each on of the collapsed (smaller) "tags" that are displayed on the page have the same part:
` class="pill-18Ymbb9 tag-3f6XHbY pill-3HIsaht collapsed-1G04emq"`

The non collapsed "tags" have this instead:

`class="pill-18Ymbb9 tag-3f6XHbY white-text-1O4Wf2a"`

Amnesiac9:
Got it fixed by adding this to dev tools

shipstation.com

INVERT
.icon

IGNORE INLINE STYLE
.pill-18Ymbb9

Gusted:
Hi @Amnesiac9, does this fix for?

```
shipstation.com

INVERT
class[*="note-icon"]

IGNORE INLINE STYLE
*
```

Amnesiac9:
Hi Gusted,

Mostly, yes! I did discover a few other icons and elements that are not looking correct, and I fixed it with the below:

`shipstation.com

INVERT
.note-icon-3F9ryyv
.sub-item-icon-1L2L6PE

CSS
.status-text-1NbkYoX {
    background-color: var(--darkreader-neutral-background) !important;
}

IGNORE INLINE STYLE
.pill-18Ymbb9`

Amnesiac9:
Found a few additional hard to read areas and corrected:

```
shipstation.com

INVERT
.note-icon-3F9ryyv
.sub-item-icon-1L2L6PE
.quantity-3CkoKmy
.pill-content-34d0DxN

CSS
.status-text-1NbkYoX {
    background-color: var(--darkreader-neutral-background) !important;
}

[...]

PR #12006, opened 2023-12-11
Fix for Shipstation.com tag icons #10225

Amnesiac9:
Fixes issue #10225, shipstation needs a wildcard prefix otherwise the style changes aren't applied.


Myshor:
OK. Accepted.

PR #12011, opened 2023-12-12
Update dark-sites.config - added github.com to list

raspberri05:


Myshor:
Github uses detect system settings by default to enable or disable dark mode which means if you set your system to light mode then Github will open in light mode in your browser.

Sooo if you use Dark Reader with light system this change will make that Github will be light.

Summarizing. I must reject this.

PR #12014, opened 2023-12-13
StackExchange style fixes for Vote buttons and Background

artshade:
Dear Developers and Community,

Thank you for the solution!

Please consider the following changes. I have more but included a few just for public testing on your environments. What do you think?

Best and kind regards :sparkles:

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -21256,12 +21256,14 @@ INVERT
 .h-auto[alt="Role-playing Games"]
 a.js-gps-track::before
 img[alt="The Stack Exchange Network"]
+button[class^="js-vote"]
 
 CSS
 body,
 #content {
     background-image: none !important;
     background-color: var(--darkreader-neutral-background) !important;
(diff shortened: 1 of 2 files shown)


Myshor:
Can you share some URL that can be nicely compared what are the differences? Fast checking - I just found that Vote buttons arrows changed from white to little darker color using filter mode on StackExchange.

Myshor:
Wait... the changes are not for "Dynamic", but to "Filter". But ok. I will try to see again what is changed as I didn't press vote buttons at all. :)

Myshor:
@artshade I just want to confirm... Are these fixes for Filter mode or Dynamic mode... 
Because for dynamic filter the fixes should be addet to [dynamic-theme-fixes.config](https://github.com/artshade/darkreader/blob/patch-1/src/config/dynamic-theme-fixes.config).

And to tell the truth. We could add them in both places as I already checked how it works in filter mode and dynamic (with additional code fixes). 🙂

If you allow me to mess in your branch I can edit the dynamic theme fixes as I already tried to include your fixes in my local dynamic-theme-fixes.config. 😉

PR #12015, opened 2023-12-13
Fix for *.gamer.com.tw

### LINKED ISSUE #11974: [Broken Website] https://forum.gamer.com.tw/ In case the website uses a tartan background
### Prerequisites

### Website Issue Description

The background color is replaced by dark, but the image is still white, so the background makes people dizzy when scrolling.

### Website Address

https://forum.gamer.com.tw/

### Steps To Reproduce

1. Visit the website. (or `home.gamer.com.tw` subdomain)
2. Use Dynamic generation mode.

### Screenshots

Disabled
Enabled

### Operating System

Arch Linux

Firefox 120.0.1

### Dark Reader version

4.9.70

### Additional Context

I think the same issue also happens on other pages with similar backgrounds, the problem might be how to determine if the background image should be inverted.

lumynou5:
Closes #11974.

The only rest problem is home.gamer.com.tw/homeindex.php can have user-customized themes, but it seems impossible to apply dark theme without breaking the themes.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -25785,7 +25785,7 @@ CSS
 
 ================================
 
-www.gamer.com.tw
+*.gamer.com.tw
 
 CSS
 body {


lumynou5:
Example: https://home.gamer.com.tw/homeindex.php?owner=viperroxas
With dark theme enabled: screenshot with dark theme enabled
After this PR: screenshot after this PR
With dark theme disabled: screenshot with dark theme disabled

src/config/dynamic-theme-fixes.config
```diff
@@ -25785,7 +25785,7 @@ CSS
 
 ================================
 
-www.gamer.com.tw
+*.gamer.com.tw
```
Myshor, line 25788: Changing www. to * would need to move that site fix up to have it sorted alphabetically.

I propose to leave first www.gamer.com.tw line and just add *. alias then it should work fine and moving won't be needed.

Myshor (approved):

PR #12021, opened 2023-12-14
Fix for wayfarer.nianticlabs.com

felixonmars:
Invert logo.

Before:

After:
image


Myshor:
I can not check. The page is not loading to me. Maybe it will other time or anyone else will confirm and merge.

PR #12028, opened 2023-12-16
Add recipesage.com to dark-sites.config

julianpoy:
Hi there! I'm the maintainer for [RecipeSage](https://github.com/julianpoy/recipesage). It automatically respects dark mode for the whole site/app, and darkreader can add some funky coloration.


julianpoy:
Re-read the contributing description and saw that it mentions that the entire site must be "dark by default, regardless of the system's preferred color scheme". That is not the case here, as the site automatically detects the user's preferred color scheme. Closing this PR out in that case.

PR #12030, opened 2023-12-16
config: Globablly invert img.Wirisformula

arielbeje:
It always contains an image with black text on a transparent background, which can be seen in https://demo.wiris.com/integrations/ckeditor/

Prompted by https://opal.openu.ac.il having the same issue as https://studysmarter.de


Myshor:
Checked that and confirm it can be useful.

PR #12052, opened 2023-12-22
Fix gmail search text for filter[+] mode

alexalekseyenko:
The search text in Gmail is black [again] in Filter and Filter+ modes, making it pretty much impossible to see. This is a best effort attempt at inverting the text element in a way that [hopefully] won't break too soon.

Fixes #12051

Issue #12051, opened 2023-12-22
[Broken Website] Gmail search text back in black


alexalekseyenko:
### Prerequisites

- [X] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this website issue to avoid opening a duplicate.
- [X] I can reproduce this website issue in a new, unmodified web browser profile with Dark Reader installed as the only extension.

### Website Issue Description

Gmail search text black [again] and difficult to read

When using Filter and Filter+ modes, the text in Gmail's search bar is basically invisible.

### Website Address

mail.google.com

### Steps To Reproduce

1. Make sure Dark Reader is set to Filter or Filter+ mode
2. Navigate to mail.googl.com
3. Type something into the search bar and [try to] observe the color

### Screenshots

<img width="744" alt="Screenshot 2023-12-22 at 10 57 12" src="https://github.com/darkreader/darkreader/assets/17463710/03faa77a-d9cb-41d4-bf92-849436f1f69f">

### Operating System

macOS 14.2.1

### Web Browser name and version

Brave 1.60.125; Chromium: 119.0.6045.199 (Official Build) (arm64)

### Dark Reader version

4.9.73

### Additional Context

This is similar to #11275 and #11267

I'll submit a PR and link it to this issue

PR #12059, opened 2023-12-23
Fix search text/results in Gmail

alexalekseyenko:
This seems to properly invert the search text and results menu for me in Filter and Filter+ modes

Following up on #12052 and the discussion therein, this is the set of changes which seems to fix the issue on my machine (#12051)

Would appreciate getting another set of eyes on this, because it seems like this might not be as reproducible as I'd hoped:
https://github.com/darkreader/darkreader/pull/12052#issuecomment-1868378302

### DIFF
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -1505,8 +1505,11 @@ CSS
 mail.google.com
 
 INVERT
-#aso_search_form_anchor [role='presentation']
 #aso_search_form_anchor [role='presentation'] tr td div
+#aso_search_form_anchor [role='presentation'] [role='listbox'] [role='option'] span
+
+NO INVERT
+#aso_search_form_anchor [role='presentation'] [role='listbox']
 
 ================================
 


Myshor:
For me there is still black subjects after applying this fixes:
obraz

And I confirm that previous fixes still makes entered search text black.

Most satysfying fix for my machine is:
```
mail.google.com

INVERT
#aso_search_form_anchor [role='presentation'] tr td div:not([class*="asor_"])
#aso_search_form_anchor [role='presentation'] [role='listbox'] [role='option'] span

NO INVERT
#aso_search_form_anchor [role='presentation'] [role='listbox']
```

but it makes white buttons with black text:
obraz

Everytime I try to do something with that buttons mail subjects change to dark not readable color.

Still I can not accept this fix as it will be working bad on my machine. It can end bad on any other people too.

alexanderby:
Fixed by https://github.com/darkreader/darkreader/commit/582b079044b80d0fe5b4f2cca92e1c91ba3fda15

PR #12067, opened 2023-12-27
Fix for *.webex.com

EnricoTirri:
add *.webex.com

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -154,6 +154,18 @@ CSS
 
 ================================
 
+*.webex.com
+
+CSS
+body {
+    background: var(--darkreader-neutral-background) !important;
+}
+.wxp-chapter-bar-item-progress{
+    background: var(--mds-color-decorative-violet-50) !important;
+}
+
+================================
+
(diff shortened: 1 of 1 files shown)


Myshor:
Can you share some screenshots what exactly you fixed here? I signed in went through all free webex things and all was already dark so I do not know what exactly you want to fix.

EnricoTirri:
When I open a recording of my organization the body of the page is not inverting the color, the "body" CSS part is to ensure it to invert dark, the ".wxp-chapter-bar-item-progress" part is to set an accent color to the recording bar progress, that otherwise would be of a dark blue-ish color that is not so visible 
before
after

EnricoTirri:
I think that this is the same for all organizations that use webex recording database, so I've set the CSS to use all sites that uses the ".webex.com" address

EnricoTirri:
If you want to find out what is not working I think I can provide you a copy of HTML file of the page if it can be useful

Myshor:
I was more aware of forcing background into body but I didn't find any issues for that.

Merging. 😉

EnricoTirri:
Perfect thanks

PR #12069, opened 2023-12-27
feat: Add blackarch.org as already dark site

Myzel394:


Myshor:
Confirmed and merged.

PR #12073, opened 2023-12-29

WritheAndShine:
Added goproblems.com and  ignored inline style on the playing field so that the board is readable and the piece color is not inverted.


Myshor:
Added invert for logo and merging. 😉

PR #12091, opened 2024-01-02
Fixes for usosweb

Tajfun403:
USOS is a Polish university platform usually accessible through a subdomain of university's website - for example usosweb.uj.edu.pl in case of UJ whose main domain is uj.edu.pl.
As of right now, the background of many elements in the platform is not reversed.

firefox_LafaL41Hn8

This pull request corrects the background of main panels, tables, and a few others.

firefox_vhGMiwez7v

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -24021,12 +24021,12 @@ timetable-entry
 
 CSS
 :root {
-    --background: var(--darkreader-neutral-background) !important;
+    --background: var(--darkreader-bg--background) !important;
     --background-secondary: var(--darkreader-bg--background-secondary) !important;
     --font-color: var(--darkreader-neutral-text) !important;
 }
 #page-body {
-    background-color: var(--darkreader-neutral-background) !important;
+    background-color: var(--darkreader-bg--background) !important;
 }
 usos-module-link-tile:hover {
     background-color: var(--darkreader-border--border) !important;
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -24005,6 +24005,60 @@ nav .menu-icon .i
...
+}
+timetable-entry {
+    color: #111 !important;
+}
+#titlebar {
+    color: var(--darkreader-neutral-text) !important;
+    background-color: var(--darkreader-neutral-background) !important;
+}
+body {
+    background: #333 !important;
```
alexanderby, line 24057: I'm afraid the colour like `#333` may look not good with light colour scheme. Could you check if something like `background: ${#dddddd} !important;` will work? (where `#dddddd` is preferred to be a source colour).

Tajfun403: Source color is around `#e8e8e8`, but with it, the outer background color is sandwiched between left and right panels (in vanilla, the brightness goes right > left > outer. With this color, it goes right < outer < left). Meanwhile `${#dddddd}` blends too closely with the left panel (only 2 units of difference for me, compared to 8 in vanilla).
I found `${#d8d8d8}` to do the trick and be rather close to the vanilla style, while still looking decent in the light color scheme.

alexanderby:
Great, thank you for the changes. Unfortunately it didn't get into the latest release, that was pushed yesterday, so this fix will be available in 1-2 weeks.

PR #12118, opened 2024-01-07
Add firebase.google.com to dark-sites.config

geofferb:
Website supports dark mode or following system theme.


Myshor:
I would merge it when site opens default as dark but I end with light version when I open it with Dark Reader disabled.

PR #12120, opened 2024-01-07
Update dynamic-theme-fixes.config (Add community.mimecast.com/s/knowledge-hub)

zfast25:
Background at the top of community.mimecast.com/s/knowledge-hub is not properly inverted, and is largely white with gray text, "Welcome to the Mimecast Knowledge Hub" 

Added INVERT for .cb-section_background 

This does also invert the image in the background, but I wasn't able to find a source for the image to fix that and it seems better than having a whole background in white while the rest of the page is properly inverted in black.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5389,10 +5389,12 @@ CSS
 
 ================================
 
-community.mimecast.com/s/knowledge-hub
+community.mimecast.com
 
-INVERT
-.cb-section_background
+CSS
+.cb-section_background {
+    background-image: none !important;
+}
 
 ================================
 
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -5389,6 +5389,13 @@ CSS
 
 ================================
 
+community.mimecast.com/s/knowledge-hub
+
+INVERT
+.cb-section_background
```
Myshor, line 5395: That totally inverts (make negative) the photo. Which is BAAAD!

I propose CSS fix instead of INVERT:
CSS
.cb-section_background {
    mix-blend-mode: multiply !important;
}

The picture will be dark but at least not inverted.

Myshor (changes requested):

PR #12132, opened 2024-01-11
Fix japscan.lol

Egsagon:
Removed the white background image for japscan.lol.


Myshor:
As I checked the site - there is already darkened background created by Dark Reader (if you compare it with light theme or extension disabled.

Is there really a big need to force remove it?

Egsagon:
The background is only a bit darkened, it still looks gray-white.

Its of course not a big need to remove it but it would make the website a better looking.

PR #12136, opened 2024-01-12
feat: ignoreImageSelectors support regexp

xachary:
Vue SPA, Scoped attribute that will be automatically assigned to the elements. 
String type conditions cannot be compared, because scoped attribute is generated by vue.
For example:

image

image

So, support RegExp can handle this situation:

```ts
    enableDarkMode(
      {},
      { ignoreImageAnalysis: [/.xxx > section > header\[data-v-[^[\]]+\]/] },
    )
```

### DIFF
--- a/src/inject/dynamic-theme/modify-css.ts
+++ b/src/inject/dynamic-theme/modify-css.ts
@@ -47,7 +47,7 @@ export function getModifiableCSSDeclaration(
     value: string,
     rule: CSSStyleRule,
     variablesStore: VariablesStore,
-    ignoreImageSelectors: string[],
+    ignoreImageSelectors: Array<string | RegExp>,
     isCancelled: (() => boolean) | null,
 ): ModifiableCSSDeclaration | null {
     if (property.startsWith('--')) {
@@ -293,7 +293,7 @@ function getColorModifier(prop: string, value: string, rule: CSSStyleRule): stri
 const imageDetailsCache = new Map<string, ImageDetails>();
 const awaitingForImageLoading = new Map<string, Array<(imageDetails: ImageDetails | null) => void>>();
 
-function shouldIgnoreImage(selectorText: string, selectors: string[]) {
+function shouldIgnoreImage(selectorText: string, selectors: Array<string | RegExp>) {
     if (!selectorText || selectors.length === 0) {
         return false;
     }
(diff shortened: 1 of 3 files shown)


alexanderby:
The idea is good. But the problem with your solution is that the RegExp is not created anywhere. It could check if the string starts with `/` and ends with `/`. But instead I propose a simpler solution: If the selector ends with `*`, let's do `x.startsWith(selector.substring(0, selector.length - 1))` instead of `===`.

src/inject/dynamic-theme/modify-css.ts
```diff
@@ -303,7 +303,7 @@ function shouldIgnoreImage(selectorText: string, selectors: string[]) {
     const ruleSelectors = selectorText.split(/,\s*/g);
     for (let i = 0; i < selectors.length; i++) {
         const ignoredSelector = selectors[i];
-        if (ruleSelectors.some((s) => s === ignoredSelector)) {
+        if (ruleSelectors.some((s) => typeof ignoredSelector === 'string' ? s === ignoredSelector : ignoredSelector.test(s))) {
```
alexanderby, line 306: I suggest you to use strings and check if the `ignoredSelector` ends with `*` here. If so, make another code branch in `if...else` block.

alexanderby (changes requested):

src/inject/dynamic-theme/modify-css.ts
xachary: > I suggest you to use strings and check if the `ignoredSelector` ends with `*` here. If so, make another code branch in `if...else` block.

"*" can not handle these:

```
// :: selector
.xxx > section > headerp[v-data-xxxxxxxx] :: before { /* ...  */ }
// deep not scoped selector, like some UI components
.xxx > section > headerp[v-data-xxxxxxxx] > .yyy > .zzz { /* ...  */ }
```

@alexanderby

alexanderby: Both selectors in your example start with `.xxx > section > header[data-v-`. The selector is split into parts on line 303. That means that the config can have `header[data-v-*` selector and easily compare the part of the selector with `s.startsWith(ignoredSelector.substring(0, ignoredSelector.length - 1))`.

src/inject/dynamic-theme/modify-css.ts
```diff
@@ -303,7 +303,12 @@ function shouldIgnoreImage(selectorText: string, selectors: Array<string | RegEx
     const ruleSelectors = selectorText.split(/,\s*/g);
     for (let i = 0; i < selectors.length; i++) {
         const ignoredSelector = selectors[i];
-        if (ruleSelectors.some((s) => typeof ignoredSelector === 'string' ? s === ignoredSelector : ignoredSelector.test(s))) {
+        if (ruleSelectors.some((s) => {
+            if (ignoredSelector && ignoredSelector.constructor === RegExp) {
```
alexanderby, line 307: The regular expression is not created anywhere, you will just have strings like `/abc/`. I suggest you run the code and put a breakpoint. Anyway, I suggest you to look into a simple string solution, as far as regular expressions in our config file will make it harder to maintain.

alexanderby (changes requested):

PR #12142, opened 2024-01-14
Added fix for sbb.ch

Root-Core:
The site is currently completely broken with Darkreader. This patch fixes most of it and tries to preserve the corporate identity.

There is a lot of content, so I didn't test against all pages - but the most important page is the start page and it is usable now.

--------------

grafik
grafik

--------------

grafik
grafik

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -20306,6 +20306,10 @@ CSS
 #sbb-logo__signet, .mod_header_logo_content > * > path:not(path[fill='#EC0000']) {
   fill: white !important;
 }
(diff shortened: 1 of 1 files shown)


Root-Core:
@Myshor I think your addition is buggy - at least on Firefox.
It works if you apply the styles via the DR dev tools, but not if you load the page normally.

image

The clock looked fine to me, but if you want to fix it, go ahead.

Myshor:
Thanks for the info... After switching from light theme to dark it works fine but it's like you showed when loaded first time...
I will do the fix with making fill with gray color so it will work fine with white circle or dark or even light one. :)

PR #12151, opened 2024-01-15
fix for sbb.ch

Myshor:
Another fix for clock colors.


Root-Core:
It still does not work for me, as it loads with a white dial. :/
The gray ticks look okay in light mode, but a bit blunt in dark mode (on the white dial).

If you switch between dark and light mode, it works again. Just like before.

image

Swiss people are very delicate about their clocks. ;)

Myshor:
I found how to fix it with light circle and dark hands in light scheme and dark circle with white hands in dark. 😁

Myshor:
As #12152 will be merged you can confirm how it looks now 😉

PR #12152, opened 2024-01-15
fix for sbb.ch

Myshor:
Final clock colors fixes.


Root-Core:
image

image

It seems to work now. ;)

PR #12153, opened 2024-01-16
increase icon visibility in nicovideo.jp

htsign:
Before:
image

After:
image


Myshor:
Can you share some URL where the icons can be seen to check fix onsite?

htsign:
You can see the fix here.
https://www.nicovideo.jp/watch/so43193670

PR #12156, opened 2024-01-16
Fix for id.fedoraproject.org

felixonmars:
Invert logo

Before:
image

After:
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -11538,6 +11538,13 @@ CSS
 
 ================================
 
+id.fedoraproject.org
+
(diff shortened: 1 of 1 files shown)


Myshor:
I am not sure if this fix is really needed - logo is readable without invert.
Comparing to accounts.fedoraproject.org which shows the same logo this fix will do some uncostistency.

felixonmars:
> I am not sure if this fix is really needed - logo is readable without invert. Comparing to accounts.fedoraproject.org which shows the same logo this fix will do some uncostistency.

It is readable but not very well I think. I guess the same fix would apply to the latter domain as well.

PR #12163, opened 2024-01-17
Fix for celum.ssg-service.com

GlowingScrewdriver:
On some webpages under celum.ssg-service.com, all text is displayed in dark colors when the page is hovered over. Colors are fine when the cursor is not over the page.

This change causes the affected text to be displayed in light colors at all times.

Admittedly, there is a subtle difference in text color between the hovered and non-hovered states of the page; however, similar behaviour is displayed even on the original webpage without Dark Reader.

Note: This CSS rule was found on the webpage. When disabled, the problem does not occur:
``` CSS
:hover {
  color: inherit;
}
```

Example page: https://celum.ssg-service.com/smartViews/view?view=bergamont-bike-b2c-manuals

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4481,6 +4481,15 @@ INVERT
 
 ================================
 
+celum.ssg-service.com
+
+CSS
+:hover {
+    color: var(--darkreader-neutral-text) !important;
+}
+
+================================
+
 ceneo.pl
 
 INVERT
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4484,7 +4484,7 @@ INVERT
 celum.ssg-service.com
 
 CSS
-:hover {
+body:hover {
     color: var(--darkreader-neutral-text) !important;
 }
 


Myshor:
Seems it's broken in chromium browsers only. I can not confirm dark text on hover in Firefox.

Second thing... When you force color on hover it breaks red color on hover in the menu on the left.

I will give a tip how to fix it in review.

src/config/dynamic-theme-fixes.config
```diff
@@ -4481,6 +4481,15 @@ INVERT
 
 ================================
 
+celum.ssg-service.com
+
+CSS
+:hover {
```
Myshor, line 4487: As I mentioned in comment this breaks hover color for hover over menu on the left 'cause selector here is too general.

Change it to `body:hover` and it will work like it should.

Myshor (changes requested):

GlowingScrewdriver:
> Second thing... When you force color on hover it breaks red color on hover in the menu on the left.
> I will give a tip how to fix it in review.

Thanks for the help!  

> Seems it's broken in chromium browsers only. I can not confirm dark text on hover in Firefox.

That's odd... I found and fixed this issue on Firefox 120.0

Myshor (approved):

PR #12170, opened 2024-01-18
Update dynamic-theme-fixes.config, add brack.ch

Nucume:
Added the Swiss retailer "brack.ch", due to the dynamic filter failing to find the background set in body, body > header & popover. Added some more adjustments to the text being more legible with the extension + fixed the search.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3693,6 +3693,22 @@ article, article section, article footer, .css-1fzo8jw, .css-efdc07, article .cs
 
 ================================
 
+brack.ch
+
+CSS
+body, body > header, .popover {
+    background: var(--darkreader-neutral-background) !important;
+    color: var(--darkreader-neutral-text) !important;
+}
+.list-group-item {
+    background: var(--darkreader-neutral-background) !important;
+}
+a {
+    color: var(--darkreader-neutral-text) !important;
+}
+
+================================
+
 brainly.com
 
 INVERT
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3703,11 +3703,11 @@ body, body > header, .popover, .popover-body {
 .list-group-item, .modal-content {
     background: var(--darkreader-neutral-background) !important;
 }
-a:not(:hover) {
+a:not(.product__category):not(:hover) {
     color: var(--darkreader-neutral-text) !important;
 }
-a:hover {
-    color: ${#323232} !important;
+.bhaal1Vl:hover, .bhauyhAt:hover {
+    color: ${black} !important;
 }
 
 ================================
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3696,15 +3696,18 @@ article, article section, article footer, .css-1fzo8jw, .css-efdc07, article .cs
 brack.ch
 
 CSS
-body, body > header, .popover {
+body, body > header, .popover, .popover-body {
     background: var(--darkreader-neutral-background) !important;
     color: var(--darkreader-neutral-text) !important;
 }
-.list-group-item {
+.list-group-item, .modal-content {
     background: var(--darkreader-neutral-background) !important;
 }
-a {
+a:not(:hover) {
     color: var(--darkreader-neutral-text) !important;
+}
+a:hover {
+    color: ${#323232} !important;
 }
 
 ================================


src/config/dynamic-theme-fixes.config
```diff
@@ -3693,6 +3693,22 @@ article, article section, article footer, .css-1fzo8jw, .css-efdc07, article .cs
 
 ================================
 
+brack.ch
+
+CSS
+body, body > header, .popover {
```
Myshor, line 3699: Setting that breaks some hover color changes. As we don't want main background switched to white on hover then I would change this line to:
`body, body > header:not(:hover), .popover:not(:hover)`

src/config/dynamic-theme-fixes.config
```diff
@@ -3693,6 +3693,22 @@ article, article section, article footer, .css-1fzo8jw, .css-efdc07, article .cs
...
+    background: var(--darkreader-neutral-background) !important;
+    color: var(--darkreader-neutral-text) !important;
+}
+.list-group-item {
+    background: var(--darkreader-neutral-background) !important;
+}
+a {
```
Myshor, line 3706: Similar situation - change to `a:not(:hover)` and it will be perfect.

Myshor (changes requested):

src/config/dynamic-theme-fixes.config
Nucume: Setting this seems to create some additional issues. Hovering over the header part of the website creates a thin white bottom line. Likewise with the popover when logged into an account, it creates a ugly white border breaking the entire cohesion. I am unfortunately bad in CSS, so I am unsure how else this could be fixed.

popover using :not(:hover)

src/config/dynamic-theme-fixes.config
```diff
@@ -3693,6 +3693,25 @@ article, article section, article footer, .css-1fzo8jw, .css-efdc07, article .cs
 
 ================================
 
+brack.ch
+
+CSS
+body, body > header, .popover, .popover-body {
```
Myshor, line 3699: Checked that and it can be as it is.

src/config/dynamic-theme-fixes.config
```diff
@@ -3693,6 +3693,25 @@ article, article section, article footer, .css-1fzo8jw, .css-efdc07, article .cs
...
+    background: var(--darkreader-neutral-background) !important;
+    color: var(--darkreader-neutral-text) !important;
+}
+.list-group-item, .modal-content {
+    background: var(--darkreader-neutral-background) !important;
+}
+a:not(:hover) {
```
Myshor, line 3706: Still won't accept as we change colors for this parts of the page:
obraz

So I'm again proposing `a:not(.product__category):not(:hover) {`

src/config/dynamic-theme-fixes.config
```diff
@@ -3693,6 +3693,25 @@ article, article section, article footer, .css-1fzo8jw, .css-efdc07, article .cs
...
+    color: var(--darkreader-neutral-text) !important;
+}
+a:hover {
```
Myshor, line 3709: With changes included above whole a:hover entry to change color when hover is not needed.

Myshor (changes requested):

src/config/dynamic-theme-fixes.config
```diff
@@ -3693,6 +3693,25 @@ article, article section, article footer, .css-1fzo8jw, .css-efdc07, article .cs
...
+    background: var(--darkreader-neutral-background) !important;
+    color: var(--darkreader-neutral-text) !important;
+}
+.list-group-item, .modal-content {
+    background: var(--darkreader-neutral-background) !important;
+}
+a:not(.product__category):not(:hover) {
+    color: var(--darkreader-neutral-text) !important;
+}
+.bhaal1Vl:hover, .bhauyhAt:hover {
```
Myshor, line 3709: Everything seems to work ok (at least on main page) but I can not find this classes there. Can you share URL (and screenshot if possible) that I could check how page works with and without this additional fix?

Myshor (commented):
One small request without need of changes.

Myshor:
@Nucume I made review already... Waiting for URL and screen with showed additional selectors which you use with hover that I can approve or suggest some nicer approach.

Nucume (commented):
Me being the absolute genius was writing all the replies while having it open as review. Apologies. There should be a reply nested somewhere with everything being mentioned. (This is what happens when a newbie tries GitHub, eh?)

Myshor (approved):

Myshor:
Thank you for cooperation. 🙂
I know i'm perfectionist when doing review but we have solution.

PR #12186, opened 2024-01-22
Unify globo.com websites fixes

Tiagoquix:
Since the change of how Dark Reader treats URL patterns, the fixes for `globo.com` websites do not work anymore.

`globo.com` by itself doesn't have any content. The contents are stored in subdomains (such as `g1.globo.com` and `ge.globo.com`).

For instance, https://github.com/darkreader/darkreader/pull/9704 used to work, but now does not.

Considering that all Globo sites use basically the same pattern for their subdomains, [and considering all the previous issues](https://github.com/darkreader/darkreader/issues?q=is%3Aissue+globo), it should not be a problem to unify the fixes.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -9872,17 +9872,6 @@ CSS
     background-color: var(--darkreader-neutral-background) !important;
 }
 
-================================
-
-ge.globo.com
-
-CSS
-.glb-grid .tabela-body {
-    background-color: var(--darkreader-neutral-background) !important;
-}
-
-================================
-
 geekflare.com
 
 INVERT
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -10388,7 +10377,7 @@ body
 
 ================================
 
-globo.com
+*.globo.com
```
Myshor, line 10380: As you added asterisk in the site you need to move whole site fix up and put it after `*.crm*.dynamics.com` before `*.maricopa.edu`.

The sitefixes must be listed alphabetically.

Alternatively you can add `globo.com` before the line `*.globo.com` then it can stay in this position and both aliases will be used for sitefixes.

How you will fix it it's your choice. 😉

Myshor (changes requested):

Myshor (approved):

PR #12187, opened 2024-01-23
Add iskdeusingqt6.org to global dark list

nathanielcwm:


nathanielcwm:
Nvm it's only a dark site with prefers-color-scheme

PR #12188, opened 2024-01-23
feat: invert unsplash logo and navbar selected

barraIhsan:
As the title says, it'll invert the unsplash logo and the color of the navbar when the item is selected
Before:
image

After:
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -24307,6 +24307,16 @@ IGNORE IMAGE ANALYSIS
 
 ================================
 
+unsplash.com
+
+INVERT
+svg[aria-labelledby="unsplash-home"]
+
+CSS
+li > a[aria-current="page"] {box-shadow: rgb(230,230,230) 0px -2px inset;}
+
+================================
+
 uokik.gov.pl
 
 INVERT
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -24307,6 +24307,16 @@ IGNORE IMAGE ANALYSIS
 
 ================================
 
+unsplash.com
+
+INVERT
+svg[aria-labelledby="unsplash-home"]
+
+CSS
+li > a[aria-current="page"] {box-shadow: rgb(230,230,230) 0px -2px inset;}
```
Myshor, line 24316: As we can use dark/light scheme in Dark Reader setting almost white color for that "underline" breaks it in light sheme.
To avoid this we use ${source_color}.

So change `rgb(230,230,230)` to `${rgb(25,25,25)}` to have the same effect in dark scheme but make it dark in light scheme. 😉

Myshor (changes requested):

src/config/dynamic-theme-fixes.config
barraIhsan: oh, ok. Didn't know that

Myshor (approved):

PR #12195, opened 2024-01-24
Add a few sites to dark-sites.config

bgirlmango:


src/config/dark-sites.config
```diff
@@ -502,6 +502,10 @@ humblegames.com
 hyper.is
 hyperbeam.dev
 i3wm.org
+iaas.store
+*.iaas.store
```
Myshor, line 506: Is this site has some subpages that you need this alias?
If yes alias must be added in the beginning of the list as it must be alphabetical ordered.

src/config/dark-sites.config
```diff
@@ -502,6 +502,10 @@ humblegames.com
...
+iaasdev.ru
```
Myshor, line 507: This page ends with some page placeholder which is white as hell when I tried it.
You sure it's correct address?

Myshor (changes requested):

src/config/dark-sites.config
bgirlmango: there are several pages (dashboard., admin. , auth.), there will most likely be more of them and they all support a dark theme. I wrote * to apply them all, is that okay?

src/config/dark-sites.config
bgirlmango: im sorry for your eyes :) , we have added your extension to the placeholder (via cdn). the address is correct and all pages on it will support a dark theme

Myshor: ok... now placeholder is dark too (hope the final page will be dark too)

src/config/dark-sites.config
Myshor: You can add *. but they must be moved into the beginning of the list or tests will reject them.
Summarizing: they can not be added just under iaas.store

P.S. Same goes to *.iaasdev.ru

bgirlmango:
I moved *.iaas.store and *.iaasdev.ru to the top of the list.

> final page will be dark
absolutely

Myshor (approved):

Myshor:
Thank you for cooperation... Proposed changes merged.

PR #12199, opened 2024-01-24
Fix for hazi.ro

felixonmars:
Fix disabled input background as well.


src/config/dynamic-theme-fixes.config
```diff
@@ -11059,6 +11059,7 @@ CSS
     --dropdown-bg: var(--darkreader-neutral-background);
     --ui-block-bg: var(--darkreader-neutral-background);
     --input-bg: var(--darkreader-neutral-background);
+    --input-disabled-bg: var(--darkreader-neutral-background);
```
Myshor, line 11062: Can you share some screenshot wher this would be used and was broken?

Myshor (commented):
One question before approve.

PR #12203, opened 2024-01-25
Invert book page on yuzu.com

gandalf3:
Invert book page on yuzu.com

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -12488,6 +12488,13 @@ img[alt="Jewish Currents"]
 
 ================================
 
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -12488,6 +12488,13 @@ img[alt="Jewish Currents"]
 
 ================================
 
+jigsaw.yuzu.com
```
Myshor, line 12491: Can't check what exactly is fixed without account. Can you share at least screens before and after fix?

Myshor (changes requested):

PR #12209, opened 2024-01-26
Fix for mapy.cz and landscan.ornl.gov

Tinh1000000:
Invert map tiles used on 2D maps
Fix map markers with improper appearances while planning directions (.icon polygon & CSS)

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -14699,6 +14699,20 @@ INVERT
 
 ================================
 
+mapy.cz
+*.mapy.cz
+
+INVERT
+img[src*="mapserver.mapy.cz"]:not([src*="bing"])
+.icon polygon
+
+CSS
+.marker span.text {
+  color: black !important;
+}
+
+================================
+
 marginalrevolution.com
 
 INVERT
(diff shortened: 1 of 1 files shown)


Tinh1000000:
For landscan.ornl.gov:

Invert OSM tiles used for map
Adjust population density legend to use original key colours (manually specifying them individually for now, I do not immediately know of a way to specifically ignore changes to the legend)

src/config/dynamic-theme-fixes.config
```diff
@@ -13357,6 +13357,60 @@ INVERT
 
 ================================
 
+landscan.ornl.gov
```
Myshor, line 13360: As for maps.cz it seems it works nice but for this map is not.

Have you zoomed map to some city when default landscan is enabled?
Without your fix it's like this:
obraz

but after updates it is more eye-painfull:
obraz

Not sure if I would like to touch this map at all.

Myshor (changes requested):
I reject for the part of landscan.ornl.gov

Myshor (approved):

Myshor:
Now it looks darker... Merging. 😉

PR #12211, opened 2024-01-26
Fix for *.instructure.com

Bren2024:
Inverted text box background


Myshor:
Can you share URL when this fix is applied?
At least can you show screenshots before and after fix?

Bren2024:
Although I can't really share the URL because viewing the page requires being logged in with an account (among other reasons), here are the before and after screenshots: 
Before
After

PR #12212, opened 2024-01-26
Improve compatibility with Unpaywall extension

arthomnix:
[Unpaywall](https://unpaywall.org/products/extension) is an extension that gives links to legal open-access versions of journal articles. When an open-access version of an article is available, it adds an iframe to the page with an icon that links to the open-access version:
20240126_11h24m50s_grim

However, when using Dark Reader in dynamic mode, there is a white border around the icon:

(dark mode)
20240126_11h25m05s_grim

(light mode)
20240126_11h25m17s_grim

With this fix applied:

(dark mode)
20240126_11h25m48s_grim

(light mode)
20240126_11h26m01s_grim

I've applied the fix to all iframes with an id of `unpaywall` and a `src` ending in `unpaywall.html`. This is applied to all sites as the extension could add the iframe to any site, and (correct me if I'm wrong) there is no way to create fixes that apply directly to extension-local pages.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -98,6 +98,9 @@ mitid-authenticators-code-app > .code-app-container {
     padding-top: 1rem;
     background-color: white !important;
 }
+iframe#unpaywall[src$="unpaywall.html"] {
+    color-scheme: light !important;
+}
 
 IGNORE INLINE STYLE
 .sr-wrapper *


Myshor:
> and (correct me if I'm wrong) there is no way to create fixes that apply directly to extension-local pages.

There was some issue about it and diefinitely it's not possible in Firefox, not sure how it ended with Chromium browsers but all I remember is that you need to make different approach with CSS for all extension pages in browsers directory.

PR #12228, opened 2024-01-29
Fix math formulas on Wikipedia

alexalekseyenko:
Some math formulas are not visible due to incorrect inversion, for instance:
https://en.wikipedia.org/wiki/Josephson_voltage_standard
This fixes inversion for `` tags with a `src` that starts with `https://wikimedia.org/api/rest_v1/media/math/render/svg`

### DIFF
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -2951,6 +2951,7 @@ NO INVERT
 img[src*="svg.png"]
 img[alt^="Skeletal" i]
 img[alt^="Structural" i]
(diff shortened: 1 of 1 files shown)


kizule:
What exactly isn't visible?

Myshor:
It's broken in Chromium browsers when Filter mode is enabled.
In Firefox this fix will make formulas darker but visible.

But with newest Darkreader I will need to add alias to *.wikipedia.org or it won't fix anything on en.wikipedia.org 😉

PR #12229, opened 2024-01-29
Inverted inline math in hoi4.paradoxwikis.com so it's visible

ezanri:


ezanri:
https://hoi4.paradoxwikis.com/Air_combat

PR #12249, opened 2024-02-02
fix mail.google.com: selecting contacts when composing an email

RobbyCBennett:
Light mode:
image

Dark mode before fix:
image

Dark mode after fix:
image

I removed the white backgrounds and added new background colors for the focused option and the hovered option. Just like in the light mode, I matched the hover background color with the checkmark circle background color.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -14612,10 +14612,10 @@ div[class*="bym"][role="navigation"], .afC, .agJ {
     background-color: var(--darkreader-neutral-background) !important;
 }
 .agJ:hover, .ag5 {
-    background-color: #181A1B !important;
+    background-color: ${#E8E6E5} !important;
 }
 .agJ.bjE {
-    background-color: #1D1F20 !important;
+    background-color: ${#E3E1E0} !important;
 }
 .ain .TO,
 .TO.ol {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -14611,14 +14611,20 @@ CSS
 div[class*="bym"][role="navigation"], .afC, .agJ {
     background-color: var(--darkreader-neutral-background) !important;
 }
+.agJ:hover, .ag5 {
+    background-color: #181A1B !important;
```
Myshor, line 14615: We can use light scheme in Dark Reader too and forcing dark colors in CSS can mess it a little bit.
To avoid this we use ${source_color} instead of inverted one.
So I think we should use here:
`background-color: ${#E8E6E5} !important;`

src/config/dynamic-theme-fixes.config
```diff
@@ -14611,14 +14611,20 @@ CSS
...
+}
+.agJ.bjE {
+    background-color: #1D1F20 !important;
```
Myshor, line 14618: Similar as above:
`background-color: ${#E3E1E0} !important;`

Myshor (changes requested):

src/config/dynamic-theme-fixes.config
RobbyCBennett: Fixed

src/config/dynamic-theme-fixes.config
RobbyCBennett: Fixed

Myshor (approved):

PR #12250, opened 2024-02-03
fix bbc.com placeholder images

RobbyCBennett:
Light mode:
image

Dark mode before fix:
image

Dark mode after fix:
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2622,6 +2622,16 @@ div[class^='scrollableList'] {
 ================================
 
 bbc.co.uk
+bbc.com
+
+CSS
+.haCQWq, .fFEZuA {
+    filter: invert(90%) !important;
+}
+
+================================
+
+bbc.co.uk
 bbc.co.uk/weather
 bbc.com/weather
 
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -2621,6 +2621,16 @@ div[class^='scrollableList'] {
 
 ================================
 
+bbc.co.uk
```
Myshor, line 2624: I would remove this whole section as bbc.co.uk URL is already used below and you add the same CSS fix into it.
for bbc.com I would add it before bbc.com/news URL then.

It will be less lines and the same effect. 😉

Myshor (changes requested):

src/config/dynamic-theme-fixes.config
RobbyCBennett: Okay. The website still looks good after that second change.

While we're on the subject of reducing the size of the file, I noticed that it uses spaces: mostly 4 but sometimes 3 or 2. With spaces, it's 484 KB and with tabs it's 476 KB.

Myshor (approved):

src/config/dynamic-theme-fixes.config
Myshor: Most people who add manually fixes uses 4 spaces.
But sometimes it goes copy/paste from browser developer tools and there are 2 spaces.
I didn't catch 3. Not sure if Tabs would be passed through the tests before merge. 😉

PR #12251, opened 2024-02-03
Fixes for 10ten-ja-reader extension

maiself:
https://github.com/birchill/10ten-ja-reader

The extension provides a popup dictionary for quickly looking up Japanese words. These fixes exclude the extension's popup, which has it's own dark mode, from being restyled by Dark Reader.

Related issue: https://github.com/birchill/10ten-ja-reader/issues/1549

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -106,6 +106,7 @@ IGNORE INLINE STYLE
 .sr-wrapper *
 .sr-reader *
 .diigoHighlight
+div[id^="tenten-"]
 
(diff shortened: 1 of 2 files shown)


Myshor:
Can you share screens how it looks before this fix and after?

I am asking for them because ater reading the mentioned issue in description it was working with dynamic mode and you do changes in dynamic.

For the filter if it should not be touched at all I think it should be included in NO INVERT and you put it into INVERT section.

That's why I would like to see how it looks before and after fix. Would be nice to see how it looks with dark reader disabled to compare.

PR #12252, opened 2024-02-03
Don't Invert the Contents of the Video Player for Amazon

erockce:
Despite being in the "`NO INVERT`" section, the entire `video` tag was still inverted.

This change is more specific which includes the class.

### DIFF
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -112,7 +112,7 @@ NO INVERT
 .av-hero-background
 .avu-content img
 .dv-tile img
-video
+.vjs-tech video
 
 CSS
 #nav-flyout-ewc {


Myshor:
Can you share some specific URL to check it live?

Or at least share some screens before and after fix.

erockce:
You're right, sorry.  I was lazy thinking would be a simple understood fix.

Initially the fix was for the main https://www.amazon.com home page.  In the scrolling carousel header, sometimes there is a featured video. The video picture thumbnail looks normal.  However., when the video starts playing the colors on the video are inverted. The browser's Inspect feature can be used to find the "`<video ...`" HTML tag. Previously only "`video`" was set under "`NO INVERT`". On the carousel however the HTML tag would be "`<video ... class="vjs-tech"`". Now of course today, there is no feature video to show and provide an example.

But the same issue may be observed for customer reviews that have videos.  The thumbnail will be correct, but when the video plays, the colors are inverted.

Here's an example:
https://www.amazon.com/gp/customer-reviews/R2SIW593DQ5AQ6?ASIN=B0C5S4YF13

When using

```
NO INVERT
...
.vjs-tech video
```

Both the picture thumbnail _and_ the playing video look correct when `amazon.com` is set with Scheme: Dark, and Mode: Filter.

PR #12253, opened 2024-02-03
Update dynamic-theme-fixes.config (myacademy.oracle.com)

AlyaqdhanZ:
Fixed dark theme for oracle academy and the slides player

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -16372,6 +16372,7 @@ button svg
 [data-module="display-views"] .dynamic-widget__result .list-view .completion-bar .progress
 .slide-displays-root div:has(div) > div :first-child > :first-child > :nth-child(3) > img
 .logo--desktop
+.login-logo
 
 CSS
 header {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -16363,6 +16363,33 @@ body {
...
+.logo--desktop
+.login-logo
+#logo
+
+CSS
+header, #header {
+    background: #333 !important;
+}
+.message-box__icon.message-box__icon_type_question, .splide__arrow {
+    background-color: white !important;
```
Myshor, line 16383: Similar comment as above:
`background-color: ${black} !important;`

src/config/dynamic-theme-fixes.config
```diff
@@ -16363,6 +16363,33 @@ body {
...
+    border-radius: 50%;
+    border: 2px solid white !important;
```
Myshor, line 16385: `border: 2px solid ${black} !important;`

Myshor (changes requested):
Some suggestions in review comments.

One more request.
Can you share the screens to show how it was before fix and after in dark scheme?
+ check how it will look after changing to light scheme.

myacademy.oracle.com needs account to browse there and without I can only approve looking on screens.

AlyaqdhanZ:
before fix:
image
image

dark theme (fixed):
image
image

light theme (fixed):
image
image

also I think people won't use the extension light theme they would just disable it, at least that's what I do.

Myshor:
Some use light theme and some contrast + brightness changes... I know they use because I had some bugs about that already. Not many but uses. 🙂

Myshor (approved):

Myshor:
Fixes merged.

PR #12255, opened 2024-02-04
Update dynamic-theme-fixes.config (myacademy.oracle.com) fixes

AlyaqdhanZ:
reordered code by specificity and fixed some inconsistencies between dark and light theme icons

before fix (light theme):
image
image

after fix (light theme):
image
image

and dark theme have same icons thats why the CSS white color is hard coded

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -16370,8 +16370,8 @@ INVERT
 .login-logo
 .logo--desktop
 .nav-control-mobile span
-.dynamic-widget__result .list-view .completion-bar .progress
 .splide--container button
+.dynamic-widget__result .list-view .completion-bar
 .course-details__cta > a.play::before
 .course-details__cta > div > a.play::before
 .thumb__background
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -16378,7 +16378,6 @@ INVERT
 .slide-displays-root
 .slide-displays-root svg[viewBox="0 0 843 573"]
 .slide-displays-root > div:nth-child(3) > div > :last-child > div > :last-child img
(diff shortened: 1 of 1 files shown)


AlyaqdhanZ:
(dark theme)
before:
image
after:
image

AlyaqdhanZ:
(dark theme)
before:
image
image

after:
image
image

AlyaqdhanZ:
I need help to fix the quiz section, how do I invert an image that is declared in the CSS by background-image:
image
image

I don't wanna make a static background like the headers because it will get rid of every other image and it will be bland, example:
image

AlyaqdhanZ:
never mind I just need to use black without making it dynamic

AlyaqdhanZ:
quiz section after fix:
dark theme:
image

light theme:
image

AlyaqdhanZ:
(before)
light:
image
dark:
image

(after)
light:
image
dark:
image

AlyaqdhanZ:
Why I feel people will get annoyed when I like everything to be perfect? is it a bad thing to care about the little details? or I'm just overthinking?

AlyaqdhanZ:
Oracle is so ANNOYING look at this, they used an img for this table:
image
image

is this even fixable? because I inverted all imgs and it doesn't even have a unique class for all table most of them has random ids

AlyaqdhanZ:
@Myshor The problem of using the source for this one is that every other image in that place won't get the fix, but I think I fixed it for now, idk why Oracle has messed up semantics.

image
image

AlyaqdhanZ:
can I open another pull request if I found other things?

Myshor:
Do what you think will be better. I am away from PC right now so no need to do it fast. 😉

AlyaqdhanZ:
@Myshor Whenever you're free merge it, I think it's good enough, Oracle is a headache 🌚

PR #12279, opened 2024-02-09
Fix myworkday.com dark mode

danielm0hr:
Fixes the team absence calendar and the small three-dot icon which opens the context menu for an employee.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -173,6 +173,22 @@ CSS
 
 ================================
 
+*.myworkday.com
+
+INVERT
+.WIF3
+.WIF3:hover
+.WIG3
+.WIG3:hover
+.WFD3 .WGL3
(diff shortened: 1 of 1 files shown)


Myshor:
As I presume it is something you use in work then won't be able to share URL to check that live.

Can you at least show screens of the before and after fix places? (At least the parts that won't be some personal data).

When we can not confirm personally the fix we would like to see the screens before merging.

danielm0hr:
Yes, Workday is a SaaS product for which you need a subscription. See screenshots below.

Team absence calendar, before and after:
Bildschirmfoto vom 2024-02-12 09-38-05_2
Bildschirmfoto vom 2024-02-12 09-37-35_2


Three dot button, before and after:
Bildschirmfoto vom 2024-02-12 09-44-38_2
Bildschirmfoto vom 2024-02-12 09-44-53_2

PR #12296, opened 2024-02-11
Add Chrono Divide game page to globally dark sites as DarkReader breaks the UI

Metadorius:
Enabling dark mode is kind of redundant there (except one subpage for file browser) and it seems to break some of the game's UI.


Myshor:
I could not find any isues for the game but as site is dark I am merging.

Metadorius:
> I could not find any isues for the game but as site is dark I am merging.

I was having some checkboxes, dropdowns, buttons and panels not rendered or rendered incompletely. It went away after a few enable/disable cycles so it may not be consistently reproducible.

PR #12297, opened 2024-02-12
added autotask.net to dector hints

SrS2225a:
Adds dark theme detector support for autotask.net PSA software (enabled by user settings)


Myshor:
Hard to review as it needs access to enter.

Can you at least show some screens with browser developer tools with that CSS included in dark mode that can confirm your proposal?

SrS2225a:
Of course. Here are a few:

image
image

With dark reader on using the same element:
image

Without autotask dark mode enabled, which becomes `LandingDatto.min.css`:
image

Let me know if you need anymore information. Also, since the version number changed (v get parameter) since my initial pr, I will push up the fix later tonight.

src/config/detector-hints.config
```diff
@@ -90,7 +90,7 @@ TARGET
 html
 
 MATCH
-link[href*="/Mvc/FrameworkContent/Styles/LandingMidnight.min.css?v=71278"]
+link[href*="/Mvc/FrameworkContent/Styles/LandingMidnight.min.css?v=72248"]
```
Myshor, line 93: If version changes maybe change to:
`link[href*="/Mvc/FrameworkContent/Styles/LandingMidnight.min.css?"]`
`*=` means it looks for this part of text.

Myshor (changes requested):

src/config/detector-hints.config
SrS2225a: Thanks, should be good to go now

Myshor (approved):

PR #12302, opened 2024-02-12
darkreader background instead of big white image

Szomti:


src/config/dynamic-theme-fixes.config
```diff
@@ -14890,6 +14890,15 @@ div.pdf
 
 ================================
 
+manytoon.com
+
+CSS
+.profile-manga {
+   background: var(--darkreader-neutral-background) !important;
```
Myshor, line 14897: As I presume it needs to be logged in. I can not confirm this fix is needed on the page.

Second thing - if site uses some backround-image then why not use?
`background-image: none !important;`

Myshor (changes requested):
Share some pic of backgroun before and after fix.

Consider code changes I suggested.

src/config/dynamic-theme-fixes.config
Szomti: First time I did this I thought about it, but I used this tool wrong and thought it didn't work because of that...

Myshor:
I managed to find where fix go without screens and URLs. 😉

Myshor (approved):

PR #12305, opened 2024-02-13
Invert display mode icon color on google drive

Badbird5907:
Before:
brave_95Ns8vhvxy

After:
brave_Sy7kJusrh6

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -7567,6 +7567,7 @@ img[src*="empty_state_trash"]
 img[src$="google_gsuite"]
 div[tabindex="-1"][role="toolbar"] > div > div > div > svg > path[d^="M10 18h4v-2h-4v2zM3"]
 button[class="ao-search-infolder ao-hint"]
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -7567,6 +7567,7 @@ img[src*="empty_state_trash"]
 img[src$="google_gsuite"]
 div[tabindex="-1"][role="toolbar"] > div > div > div > svg > path[d^="M10 18h4v-2h-4v2zM3"]
 button[class="ao-search-infolder ao-hint"]
+div > div > div > div > div > div > div:nth-child(2) > div > div:nth-child(1) > div > div > div > div > div:nth-child(1) > span > div > div > div > button > div > svg > path
```
Myshor, line 7570: This doesn't work for me at all.

I tried to make better selectors for this checks:
obraz

and I ended with:
```
button[aria-checked="true"] > div > svg > path
button[aria-checked="true"] > div > span > svg > path
```

Check if my suggestion will work for you too. If make changes in your PR. 😉

Myshor (changes requested):

src/config/dynamic-theme-fixes.config
Badbird5907: Yup, this works.

Myshor (approved):

PR #12318, opened 2024-02-15
myacademy.oracle.com little fixes

AlyaqdhanZ:
image

image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -16505,10 +16505,10 @@ INVERT
 .course-details__cta a.play::before
 .thumb__background
 .universal-side-panel__switcher button svg
-.slide-displays-root img
+.universal img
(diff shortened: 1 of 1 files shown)


Myshor:
As I understand you invert whole slides - that inverts photos on them too as can be seen on your screenshot.

I prefer original colors than inverted photos.

On the other hand... Can you tell how often there are photos in this slides?

AlyaqdhanZ:
> As I understand you invert whole slides - that inverts photos on them too as can be seen on your screenshot.
> 
> I prefer original colors than inverted photos.
> 
> On the other hand... Can you tell how often there are photos in this slides?

I don't understand what you mean, but this what the slide looks like now
image

Because it's dark theme I think it makes more sense that the previews are inverted, I fixed some arrow tips that were barely visible, and the last fix in CSS was for the exam section to make the text clearer.

image

image

Myshor:
I meant this one:
obraz

but I think it's very small amount of slides which will have some photos in them.

I will merge this fix. 😉

PR #12326, opened 2024-02-16

liveintottenham:
kakao page update plz -> bubble_msg_txt color change

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4636,6 +4636,15 @@ INVERT
 
 ================================
 
+center-pf.kakao.com
+
+CSS
+.box_bubble .txt_chat {
+    color: rgb(255 255 255);
+}
+
+================================
+
 central.proxyvote.com
 
 INVERT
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -28353,3 +28353,13 @@ CSS
...
     background-color: var(--darkreader-neutral-text) !important;
 }
+
+================================
+
+center-pf.kakao.com
+
+CSS
+.box_bubble .txt_chat {
+    color: rgb(255 255 255);
```
Myshor, line 28363: As I can not open myself this page I would like you to check how this will work with Light scheme set to dark reader. Because I think it should use black color then.

To fix that use:
`color: ${rgb(0 0 0)};`

It will make text white in Dark scheme and black in light.

Myshor (changes requested):
Read comment and share screens for dark scheme and light scheme with the final fix.

src/config/dynamic-theme-fixes.config
liveintottenham: In a bright scheme, it appears as shown above. (default)


to fix color: `${rgb(0 0 0)}`

Myshor (approved):

Issue #12275, opened 2024-02-08
[Broken Website] White background in Generative AI on google.com


Tony20221:
### Prerequisites

- [X] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this website issue to avoid opening a duplicate.
- [X] I can reproduce this website issue in a new, unmodified web browser profile with Dark Reader installed as the only extension.

### Website Issue Description

Doing a search, the Generative AI section at the top had low contrast between the text and background and it's hard to read.

![image](https://github.com/darkreader/darkreader/assets/99626433/d8edf6ba-78d2-462e-b7f0-3a5db21495e2)

### Website Address

https://github.com/darkreader/darkreader/issues/new?assignees=&labels=Broken+Website&projects=&template=broken-website-report.yml&title=%5BBroken+Website%5D+Replace+with+title

### Steps To Reproduce

Go to the url provided

### Screenshots

![image](https://github.com/darkreader/darkreader/assets/99626433/59d8637d-036e-4b00-9b54-4b9cd3ab09f6)

### Operating System

Windows 10

### Web Browser name and version

Chrome

### Dark Reader version

latest

Myshor:
First... correct the 'cause it is link to creating issue.

Second thing... go to chrome://extensions/ and click on more details for dark reader see the extension version in your browser.
Check if it is 4.9.77

There was fix for similar Issue #12007 and there was fixes already done for that and should be included in latest 4.9.77 version.

When you confirm you use the latest 4.9.77 version then we will need to check that fix if it needs to be updated.

I can not confirm anything as option to use new AI search is not yet available for me.

Tony20221:
I don't know what your first line is about.

It's version 4.9.77. I am not using AI search specifically. It just appears for me sometimes when I do a Google search. 
Create a Bard account. Maybe then it will show up for you.

Myshor:
> I don't know what your first line is about.

I meant this link - it's not the link to the site with problem.
> https://github.com/darkreader/darkreader/issues/new?assignees=&labels=Broken+Website&projects=&template=broken-website-report.yml&title=%5BBroken+Website%5D+Replace+with+title

Myshor:
Not sure how that bard should work. Maybe it's location related. I am moved to gemini.google.com when I try to go to bard.google.com and no possibility to enable it in google search. Maybe @suhan-paradkar will take a look as author of last fixes?

Tony20221:
[...]

PR #12340, opened 2024-02-19
bugfix: Google Search Generative AI beta

MalekNecibi:
Previous attempt included two methods of inverting background, canceling or masking eachothers affects and leaving and a white background. Sticking with --darkreader-neutral-background as it seems more seamless


MalekNecibi:
Full Description in [GitHub Issue](https://github.com/darkreader/darkreader/issues/12275#issuecomment-1951539156)

PR #12342, opened 2024-02-19
Adding app.getdx.com to global dark list

abinoda:


Myshor:
I check the adress and land on white as hell login page... maybe after login it is dark but I can not merge this proposal.

PR #12344, opened 2024-02-20
fix for docs.rust-lang.org

fin-w:
fix contrast of text highlighted by search results

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -7277,6 +7277,15 @@ CSS
 
 ================================
 
+doc.rust-lang.org
+
+CSS
+mark {
+    color: var(--darkreader-neutral-text) !important;
+}
+
+================================
+
 docs.codacy.com
 
 INVERT


Myshor:
Can you share some example URL?

At least show some screen what exactly is highlighted and how it looked before fix and after fix.
I am trying to figure out it by myself with too much time. 🙂

fin-w:
When you search within doc.rust-lang.org and then click on a result, the url of the page loaded includes the search query to highlight specifically within the page.

My change would improve the contrast of words highlighted in the page (without DarkReader they are pale background / dark text, with DarkReader they currently are dark background / only slightly darker text).

Follow this link for a page with "string" highlighted for example (usually you would navigate here by searching "string" within the website and then clicking on the top result).
https://doc.rust-lang.org/book/ch04-03-slices.html?highlight=string#string-slices

The website has three built-in dark modes and two light ones, and defaults to a dark theme. If you prefer, I can add it to the dark-sites.config file?

Myshor:
Thanks for the link. I see what fix change. Merging.

PR #12360, opened 2024-02-22
Fix for sciencedirect.com

Bunando:
Various fixes for text, background, and overlay colors.

Example Webpage: https://www.sciencedirect.com/science/article/abs/pii/S1877343513001796

Before: 
Before

After: 
After


Myshor:
OK... I made some fixes for that page but didn't check Chromium based browsers. My bad. In Firefox everything looked fine with my fixes.

As I see you do not mess too much with colors so I merge. 😉

PR #12376, opened 2024-02-26
Fixes for lists.debian.org

fin-w:
Remove background image and invert logo, in line with the other debian subdomains.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -8487,6 +8487,7 @@ INVERT
 ================================
 
 excel.officeapps.live.com
+*.officeapps.live.com
 
 INVERT
 .ewr-sheettable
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -14267,6 +14267,18 @@ body {
 
 ================================
 
+lists.debian.org
```
Myshor, line 14270: Good catch but as I checked the debian.org fixes we repeat fixes from there here.
The same problem I found on wiki.debian.com

So instead of making fix for subpages with prefixes I propose you add alias `*.debian.org` under `debian.org` line in sitefixes.

Myshor (changes requested):

src/config/dynamic-theme-fixes.config
fin-w: I didn't realise that was possible! Thanks for the tip, I'll make the changes now.

PR #12380, opened 2024-02-27
Fixes for crowdin.com/editor

fin-w:
In the translation editor, most icons were dark in dark theme (i.e. invisible) and the left-hand menu background was white. INVERT the icons and make the menu have a dark background.

Also delete incorrectly INVERTed page load progress spinner.

Requires log in and affiliation with a project to see the changes.


alexanderby:
Maybe a selector like `[class*=static-icon]` could work.

PR #12381, opened 2024-02-27
Adds 2 new Sites to the Dark Sites Config

Wuemeli:


src/config/dark-sites.config
```diff
@@ -1145,3 +1145,5 @@ zt64.github.io
 ztdp.ca
 zunivers.zerator.com
 zyrenth.dev
+wuemeli.com
```
Myshor, line 1148: All pages in this file must be sorted alphabeticaly - correct that and we can merge that.

Myshor (changes requested):

src/config/dark-sites.config
Wuemeli: Sorted :)

Myshor (approved):

PR #12385, opened 2024-02-28
Add learn.microsoft.com to dark-sites.config

geofferb:
[learn.microsoft.com](https://learn.microsoft.com/en-us/) has built-in dark mode that defaults to the system setting.


Rudxain:
Unfortunately, it only works with JS.

- Firefox for Android: 123.1.0
- DR: 4.9.78
- uBlock Origin: 1.56.0
- uBO rule: `no-scripting: * true`

PR #12393, opened 2024-02-28
Simplify fixes for crowdin.com/editor, Fixes for llvm.org

fin-w:
INVERT the icons with a simpler method than listing them all out individually, as per advice https://github.com/darkreader/darkreader/pull/12380#issuecomment-1969463219

This does not change the way the page appears, only simplifies the theme-fixes file.


fin-w:
I've also included a fix for llvm.org in this merge request:

Remove background images from areas where they make text unreadable, merge the *.llvm.org rules already present in the config into general llvm.org rules.

PR #12408, opened 2024-03-03
Add developer.mozilla.org to dark-sites.config

geofferb:
[developer.mozilla.org](https://developer.mozilla.org/en-US/) has built in dark mode that defaults to system theme.


alexanderby:
Unfortunately this list is intended for completely dark websites.

PR #12415, opened 2024-03-05
Update dynamic-theme-fixes.config: Player-colour fix on chess-results…

lordnuuk:
….com

Fixing the unnecessary inversion of the box indicating color of the pieces of the player, on the games summary page for each player in a tournament, on the chess-results.com. Only needed for the white player.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -4765,6 +4765,13 @@ INVERT
 
 ================================
 
+chess-results.com
+
+INVERT
+.FarbewT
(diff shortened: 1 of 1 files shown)


Myshor:
Can you share some example URL to check how the fix works?

At least some screenshots before and after fix would be nice.

lordnuuk:
Before

Screenshot 2024-03-06 at 02 31 08

After 

Screenshot 2024-03-06 at 02 30 17

Notice the change of color of the box indicating the when the player had white pieces

PR #12431, opened 2024-03-08
Remove obsolete rule for gameinformer.com

rakleed:


Myshor:
Checked that in Windows using Firefox and I still need this fix.

Not merging.

PR #12432, opened 2024-03-09
Fix code blocks on chat.openai.com

alexalekseyenko:
The code blocks on [chat.openai.com](chat.openai.com) are already dark and prior to this change Filter/Filter+ mode would invert them.

Before:
Screenshot 2024-03-08 at 20 49 05

After:
Screenshot 2024-03-08 at 20 48 55

### DIFF
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -349,6 +349,7 @@ chat.openai.com
 
 INVERT
 .overflow-hidden:has(textarea)
+pre .dark
 
 ================================
 


src/config/inversion-fixes.config
```diff
@@ -349,6 +349,7 @@ chat.openai.com
 
 INVERT
 .overflow-hidden:has(textarea)
+pre .dark
```
Myshor, line 352: Why to INVERT something that is good.
Check how NO INVERT section looks like in this file and apply that here.

I could do that by myself but if you will make that change you will learn something new. 😉

Myshor (changes requested):

alexalekseyenko:
When the rule is in `INVERT` I get this:
Screenshot 2024-03-13 at 14 10 25

but when it's in `NO INVERT` I get this:
Screenshot 2024-03-13 at 14 10 59

I haven't looked into the dark reader source in a while so apologies if I'm misunderstanding how this is supposed to work

Myshor (approved):

Myshor:
OK thanks for the screens.

Merging.

PR #12470, opened 2024-03-18
DEL allegro.pl rules

FadeMind:
Allegro.pl using now native official dark mode IF dark theme is enabled in OS settings.

Your call.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1150,50 +1150,6 @@ INVERT
 
 ================================
 
-allegro.pl
-
-INVERT
-i[title="Smart!"]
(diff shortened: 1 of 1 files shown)


Myshor:
Maybe yes but not everyone would use Dark Mode with dark system settings. Deleting all entries for allegro will leave this people without fixes for light version.

I am not closing yet this issue because I need to check if the fixes is still needed for light version of allegro. As I checked now dark version the fixes don't mess that much and could be included in Dark Reader.

Myshor (changes requested):
No changes needed but I change this because I don't want that accidently we just merge and remove all previous fixes.

Myshor (approved):
Seems to work fine with Dark Reader applied to light mode.

FadeMind:
@Myshor Thanks

PR #12471, opened 2024-03-19
Fix dropdown boxes on netbank.sparthy.dk

jesperbakhandskemager:


jesperbakhandskemager:
Changes this:
image
to this:
image

PR #12481, opened 2024-03-21
Remove quarantined domains pref from README

Tiagoquix:
It is a security feature and should not be changed by common users.

https://support.mozilla.org/en-US/kb/quarantined-domains


Myshor:
As it actually works fine with this flag set to "true" (because I have it set to true) I merge. 😉

PR #12488, opened 2024-03-22
Updated sciencedirect.com

AnthonyMakesStuff:
Added a fix for when reading articles on sciencedirect.com

Before:
before

After:
after

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -21688,6 +21688,12 @@ INVERT
 
 sciencedirect.com
 
+INVERT
+.article-textbox
+.article-textbox > span#tbcap1 > p
+.article-textbox > div
+.article-textbox a
+
 CSS
 body,
 .gh-nav .gh-nav-action,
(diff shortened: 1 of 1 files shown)


Myshor:
Can you share the URL to Article? Would be nice to check it live.

AnthonyMakesStuff:
This is the article [https://www.sciencedirect.com/science/article/pii/S2589537022003546](https://www.sciencedirect.com/science/article/pii/S2589537022003546)

src/config/dynamic-theme-fixes.config
```diff
@@ -21688,6 +21688,12 @@ INVERT
 
 sciencedirect.com
 
+INVERT
```
Myshor, line 21691: As it works fine in Firefox then I can't merge inverts because it will fix Chrome based browsers but break Firefox.

I managed to get what is needed from Firefox.

Instead of INVERTs add this lines on the end of CSS section:
```
.article-textbox {
    background-color: ${rgb(213, 211, 207)} !important;
}
.Article .topic-link {
    color: ${rgb(46, 50, 52)} !important;
    text-decoration-color: ${rgb(46, 50, 52)} !important;
}
```

Myshor (changes requested):

alexanderby:
It's interesting. The MIME type of the CSS file is text/plain instead of text/css.

Myshor (approved):

PR #12495, opened 2024-03-24
Tom's hardware white-on-white fix

erkinalp:


alexanderby:
Let's put it on hold, the JS fix is on it's way. Could you please share a link to double check if the problem you encountered is already fixed?

erkinalp:
> Let's put it on hold, the JS fix is on it's way. 

This workaround also applies to older versions, though.

alexanderby:
@Gitoffthelawn I've checked and everything seems fine on the latest build version.

alexanderby:
Works well in v4.9.81.

PR #12502, opened 2024-03-26
Fix `wordpress.com` site stats border color

fxzxmicah:
URL: `https://wordpress.com/stats/day/***`


fxzxmicah:
Looks like wordpress updated their site. So this change is no longer valid. I'll fix it again later.

PR #12521, opened 2024-03-29
google search AI result fix

notAxion:
background-color fix for google search Generative AI Result


alexanderby:
It seems we the same fix for `www.google.*` already.

notAxion:
It seems so 
I guess its fine to close this then

It also seems like i don't have the latest config in my installed extension that's why the confusion.

PR #12570, opened 2024-04-08
Added a fix for the myorca site

kc6827:
Fixed some icons/headers on the "myorca.com*" site that were not visible.

Before:
image
image

After:
image
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -17416,6 +17416,15 @@ IGNORE INLINE STYLE
 
 ================================
 
+myorca.*
+
+INVERT
+h1
+strong
+.b
+
+================================
+
 myunidays.com
 
 INVERT


src/config/dynamic-theme-fixes.config
```diff
@@ -17416,6 +17416,15 @@ IGNORE INLINE STYLE
 
 ================================
 
+myorca.*
+
+INVERT
+h1
```
alexanderby, line 17422: While you may seem it working on a specific page, it can break on other pages. I suggest using more specific selectors like `.m-banner h1` etc. Maybe instead of `INVERT` some `CSS` like `color: black !important` will do the job.

alexanderby (changes requested):

alexanderby:
The changes were implemented in https://github.com/darkreader/darkreader/pull/12627

PR #12572, opened 2024-04-08

utopicchaotica:
fix app.univerusrec.com select options not showing up as inverted

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1902,7 +1902,9 @@ CSS
 html {
     background-color: #181a1b !important;
 }
-input, select, textarea {
+input,
+select,
+textarea {
     color: --darkreader-selection-text !important;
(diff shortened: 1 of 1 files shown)


Myshor:
Can you at least share screens before and after? When I go to the app.univerusrec.com URL I end on https://univerussportandrecreation.com/ and I can not see anything that need fixes.

utopicchaotica:
Not sure if this screenshots are even helpful cuz I had to blur out all the sensitive info.

image
The CC/BCC select options show up as the site defined dark text color instead of light, e.g. first item in the CC is "--- Carbon Copy ---".

The Email Message textarea input is also the site defined dark text colour.

After fix:
image

image
The >html> element inside the <frame> isn't getting the dark background colour that is defined in the root <html>. Not sure why, but I'm sure it's cuz iframe CSS is stupid.

After fix:
image

PR #12590, opened 2024-04-12
Fix for temu.com

nvd916:
Background of certain divs were white.

Before:
Screenshot 2024-04-12 at 02-25-16 Waterproof Mattress Protector Quilted (without Pillow Core) - Temu

After:
Screenshot 2024-04-12 at 02-26-09 Waterproof Mattress Protector Quilted (without Pillow Core) - Temu

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -24643,6 +24643,15 @@ body,
 
 ================================
 
+temu.com
+
+CSS
+._2Fkk_bmp {
+    background-color: ${#fff0} !important;
+}
+
+================================
+
 tenforums.com
 
 CSS
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -24647,7 +24647,7 @@ temu.com
 
 CSS
 ._2Fkk_bmp {
-    background-color: ${#fff0} !important;
+    background-color: transparent !important;
 }
 
 ================================


src/config/dynamic-theme-fixes.config
```diff
@@ -24643,6 +24643,15 @@ body,
 
 ================================
 
+temu.com
+
+CSS
+._2Fkk_bmp {
+    background-color: ${#fff0} !important;
```
alexanderby, line 24650: Did you mean to just use `transparent`? The 4-th `0` means zero opacity.

alexanderby (changes requested):

src/config/dynamic-theme-fixes.config
nvd916: I see the element has `background: hsla(0,0%,100%,0)`, which is equivalent to `background-color: #fff0`. Based on the contributions document, I should put that color inside `${}` since we want it inverted. 
But yes, effectively it is `transparent`. I actually did not know that keyword existed. Putting it inside `${}` or not yields the same result:
- If I do `background-color: transparent !important`, when Dark Reader is on, the result will be `background-color: transparent !important`. 
- If I do `background-color: ${transparent} !important`, when Dark Reader is on, the result will be `background-color: rgba(255, 255, 255, 0) !important`.

nvd916: It is interesting that the original CSS value of `hsla(0,0%,100%,0)` already means that the background color is transparent, but the computed value disregards the alpha component of the color. It stays the same when turning Dark Reader on, hence the problem. Forcing it to use `transparent` using the Dark Reader Dev Tools solves the problem for some reason. 

image

alexanderby: Yes, I meant `transparent` without `${}`. Those colors with 0 transparency may be useful for gradients (to fade out a white color for example), but not for usual backgrounds.

alexanderby (approved):

PR #12659, opened 2024-05-01
Amex fix

charmarkk:
Added rule for *.americanexpress.com, properly inverted dropdown select elements for payment bank accounts


Myshor:
I pressume it's somewhere after loggin in.

Can you share some before and after fix screens? (of course blurred/censored to not share what is private).

Just to be sure what we fix here. 😉

Myshor:
Thanks for screens and CSS preview. I will add a comment in review then. 😉

src/config/dynamic-theme-fixes.config
```diff
@@ -113,6 +113,15 @@ IGNORE INLINE STYLE
 
 ================================
 
+*.americanexpress.com
+
+CSS
+select.css-1ui3puo > option {
```
Myshor, line 119: change to:
`select[aria-labelledby="bankAccount-label"] > option`

this should be better than class which can be generated name and can change in the future. 😉

Myshor (changes requested):

charmarkk:
Fair enough! Working in accessibility we try not to use ARIA as a selector
for the same reason, but that makes sense!

src/config/dynamic-theme-fixes.config
charmarkk: Will go in and change later today.

charmarkk: Sorted,

Myshor (approved):

PR #12664, opened 2024-05-02
Add fixes for simracingmachines.com

ShutterQuick:
* Fixed white background color
* Inverted partner logos, which were black for better contrast against dark background

Before fix:
image

After fix:
image


src/config/dynamic-theme-fixes.config
```diff
@@ -22904,6 +22904,18 @@ body {
 
 ================================
 
+simracingmachines.com
+
+INVERT
+li[id^="Slide-multicolumn-"] img[class="multicolumn-card__image"][src*="logo"]
```
Myshor, line 22910: Can you add one more invert for partners? They are black on dark too.
`li[id^="Slide-template-"] img[class="multicolumn-card__image"]`

I just don't want to mess with your pull request.

Myshor (changes requested):

src/config/dynamic-theme-fixes.config
Myshor: I mean this part: (screen with my proposed invert added)
obraz

ShutterQuick: @Myshor Good catch, I didn't see that.

I've updated the MR.

Myshor (approved):

Issue #12655, opened 2024-04-30
[Broken Website] Revolver.news


joewinke:
### Prerequisites

- [X] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this website issue to avoid opening a duplicate.
- [X] I can reproduce this website issue in a new, unmodified web browser profile with Dark Reader installed as the only extension.

### Website Issue Description

Its too dark. I tried to fix in the developer tools, first with 'a', then added the class and parent containers. Nothing is working.

![Screenshot 2024-04-30 at 12 40 22 PM](https://github.com/darkreader/darkreader/assets/797297/d4344df5-637f-4f20-8331-0a76de11cc33)

 

### Website Address

revolver.news

### Steps To Reproduce

1. Load webpage
2. Enable dark reader

### Screenshots

![Screenshot 2024-04-30 at 12 40 22 PM](https://github.com/darkreader/darkreader/assets/797297/d41f2ad7-5e6b-410c-8aba-373ead99eacd)

### Operating System

Mac OS Latest

### Web Browser name and version

Firefox latest

### Dark Reader version

4.9.83

PR #12683, opened 2024-05-06
Fix for revolver.news

light-orth:
Article titles and category on main page unaffected by color changes as stated in #12655.

My attempt at a simple fix for this by inverting the elements instead.


alexanderby:
I need to figure out why the CSS fix didn't work (and possibly why the text is dark in the first place).

alexanderby:
The problem it doesn't work is `style="color: rgb(0, 0, 0) !important;"`. The important inline styles have the highest precedence.

PR #12686, opened 2024-05-06
Update dynamic theme fix for `porkbun.com`

oberrich:
Properly invert sales chat pop-up

*(before)*

*(after)*

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -20220,7 +20220,11 @@ porkbun.com
 
 INVERT
 img[src="/images/handshake-icon.svg"]
-iframe
+
+CSS
+iframe {
+    color-scheme: initial !important;
+}
 
 ================================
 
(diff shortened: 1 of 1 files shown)


oberrich:
I see there is a slight problem: The chat pop-up gets loaded in delayed. Is there some way to config dark-reader to be run again on DOM change or something?

alexanderby:
What browser are you using? I don't see such white corners.

oberrich:
> What browser are you using? I don't see such white corners.

Version 123.0.6312.113 (Official Build, ungoogled-chromium) (64-bit) and I have a fast connection through fiber optic cable.
Opening the site sometimes results in the image being white, if I disable and reenable it for porkbun *after* loading the site, it correctly inverts the image.

The chat widget is loaded through an iframe - possibly at a later point, maybe that causes the issue?

oberrich:
Landed in 46af1fd67e2d9629d2926547a6e394effd542e67

PR #12687, opened 2024-05-06
Add circuits.vcn.com to dynamic-theme-fixes.config

ccwaddey:
This makes images on a site that my company uses for work more readable in dark mode. It also makes selection boxes readable. Since this is a private access website, I understand if this won't be accepted, but I'm not sure how to make the changes automatically apply when I open up the site for the first time without having to go into dev tools and apply them there. Other users of the site at my company would appreciate this very much, since not everyone is technical enough to figure this out. Let me know if you want pictures of how this looks in dark mode and light mode, or if you need anything else.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5257,6 +5257,22 @@ INVERT
 
 ================================
 
+circuits.vcn.com
+
+INVERT
+img
+
+CSS
+select {
+    background-color: #121212 !important;
+    color: #e6e3ea !important;
+}
+img {
+    background-color: #e6e3ea !important;
+}
+
+================================
+
 cisce.org
 
 INVERT


ccwaddey:
Here are some before and after screenshots of what I'm talking about:
before:
image
after:
image

ccwaddey:
And for the selections:
before:
image
after:
image

ccwaddey:
And feel free to remove the img CSS selector from this. I don't know why that's in there and it looks a lot cleaner without it.

src/config/dynamic-theme-fixes.config
```diff
@@ -5257,6 +5257,22 @@ INVERT
...
 ================================
 
+circuits.vcn.com
+
+INVERT
+img
+
+CSS
+select {
+    background-color: #121212 !important;
```
alexanderby, line 5267: Please use `${initial_color} !important` or `var(--darkreader-neutral-background) !important` to make theme applied properly. See more info here https://github.com/darkreader/darkreader/blob/main/CONTRIBUTING.md

alexanderby (changes requested):
As soon as the URL is accessible through internet I see no issues accepting this PR, after non-static colors are used.

src/config/dynamic-theme-fixes.config
ccwaddey: So I changed this to use variables instead of hardcoded colors and added some styling so that you can see the arrows for a select menu in certain cases. The colors look the same as far as I can tell, but the dropdowns on some selects look better now:
normal select before and after:
image

special case select before:
image
and after: 
image

notice that the arrow is visible in the dropdown.

src/config/dynamic-theme-fixes.config
```diff
@@ -5257,6 +5257,30 @@ INVERT
 
 ================================
 
+circuits.vcn.com
+
+INVERT
+img
```
alexanderby, line 5263: Please try being more specific (refer elements by class), as far as every image on every page will be inverted.

alexanderby (changes requested):

src/config/dynamic-theme-fixes.config
ccwaddey: Gotcha. I added an attribute specifier to only invert images that are sourced from the graphing backend.

ccwaddey: In case you're wondering why I chose referencing them by src tags that start with "/rrd_graph.php", it's because there are no class attributes that I can see for these types of images. The image tag is just `` or similar.

PR #12695, opened 2024-05-08
fixed gamepress.gg background

nvd916:
Gamepress's background makes it hard to read in dark mode. It is better to remove it as removing it does not affect browsing experience and improves readability.

Before: 
Screenshot 2024-05-08 at 02-03-58 Learning With Manga Collab - Walkthrough

After: 
Screenshot 2024-05-08 at 02-05-09 Learning With Manga Collab - Walkthrough


oberrich:
[Example article to reproduce](https://gamepress.gg/grandorder/jp-summon-simulator-beta#2180) - LGTM!

PR #12691, opened 2024-05-07
[Feature] Improve Font replacement

oberrich:
In this draft PR, I plan to fix some global issues with font replacement.

Currently there are multiple issues with replacing the font globally:
- Icon fonts aren't exhaustively excluded, leading to replacement characters being commonly displayed
- Complex code viewers and editors break when using a non-monospaced font
In GitHub.com's editor for example, sometimes the code text will be using the replacement font but sometimes it won't.
Selecting text inside the editor with an active replacement font, will still have GitHub's font as the "selection" font and will result in an incorrect cursor position.

I want to implement the following:
- Improve detection of icon fonts (first progress https://github.com/darkreader/darkreader/commit/4bee8fa9ff2c329290209f4c412834a1d730ad89)
- Improve detection of monospaced fonts
- Introduce hot fixes for large/planet scale sites like GitHub (wip)

Currently this feature is implemented naively in `createTextStyle`.
I think it might be interesting to expand on this feature by:
- Detecting monospaced fonts
- Detecting emoji/icon fonts

This could be done possibly by either checking the name (e.g. is font family `ui-monospace`, `Consolas`, ...`), actually analyzing the font's glyphs programmatically or rendering it to see if it displays correctly. Fallback: Don't replace font.

This would allow to offer a much more granular control over font replacement, for example based on the standard CSS font-family aliases `ui-*`, `emoji`, `math`, etc.

PR #12696, opened 2024-05-08

shimizumasaru:
Improved disappearance of TBS NEWS banner in dark mode on Yahoo Japan site

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -18039,6 +18039,13 @@ INVERT
 
 ================================
 
+news.yahoo.co.jp
+
+IGNORE IMAGE ANALYSIS
+.QrZuR
+
+================================
+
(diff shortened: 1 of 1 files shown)


oberrich:
Adding the following fix will also stop icons from getting replaced in dynamic mode (also fixed in #12691).
```css
CSS
[class*="Icon"] {
    font-family: RiffIcons !important;
}
```

I couldn't find the banner you were mentioning as the site is in Japanese and I'm unable to navigate.

shimizumasaru:
@oberrich 
 My code worked fine after applying the global fix applied in #12691.I'm sending you an image just to be sure.I am sending you an image just in case.I am sending an image of the affected area just in case.I am sending a screenshot image of the relevant section just in case.


NG
pict_NG_001

OK Dark
pict_OK_Dark


OK Light
pict_OK_Light

PR #12697, opened 2024-05-08
Fix disabled checkbox/radio contrast

iki:
Invert disabled checkbox/radio inputs to make them readable

Sample:
- [x] Checked
- [ ] Not checked


oberrich:
**Without** patch
image

With patch
image

I think this PR will break a lot of sites and shouldn't be merged.

PR #12706, opened 2024-05-09
Fix for docusign.com

felixonmars:
Invert logos

Before:
Screenshot from 2024-05-09 19-45-20

After:
Screenshot from 2024-05-09 19-45-11


felixonmars:
It doesn't work here with * removed:

image

Adding it back works:

image

(I've noticed this for quite some time. Any ideas?)

PR #12708, opened 2024-05-10
fix: dynamic mode on `emojipedia.org`

oberrich:
The site has a regular dark mode but overriding with dynamic mode breaks the site and adds a mismatched segment.
Just changing based on `main` seems to be the best option the other stuff causes extra issues, this ones rly weird.

This also fixes the pop-up which in default dark mode is a white flash bang.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -8820,9 +8820,7 @@ body {
 emojipedia.org
 
 CSS
-div.sticky,
-main,
-aside {
+main {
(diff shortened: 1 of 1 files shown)


oberrich:
I have done some more testing

```css
.sticky,
main,
div.main {
    background-color: ${#e9e9e9} !important;
}
```

seemed to work as well at first, until it didn't.

I think this may be an issue with the core of darkreader.
I've converted this into an issue, referencing this PR for investigation.

PR #12738, opened 2024-05-18
Dynamic fixes for www1.flightrising.com

nidgithm:
If I did something wrong (like if it's too complex) just let me know.


Myshor:
I went through the pages I was able without logging in and saw some nice fixes.

As the tests are passed I merge. 😉

PR #12755, opened 2024-05-24
Update Italian translation

its-miroma:
It looks like most strings were machine-translated.


src/_locales/it.config
```diff
@@ -1,37 +1,37 @@
 @extension_description
-Tema scuro per ogni sito. Prenditi cura dei tuoi occhi, usa Dark Reader per navigare giorno e notte.
+Tema scuro per tutti i siti. Prenditi cura dei tuoi occhi, usa il tema scuro per navigare giorno e notte.
```
Myshor, line 2: Dark Reader is a product name so we won't translate it.

Myshor (changes requested):
Just one comment and request to change.

PR #12765, opened 2024-05-27

cinderblox94:
Fixed aspect for Knossos site


alexanderby (changes requested):
It seems your fix was based on an old config snapshot and that created many conflicts. Could you please make the PR from scratch only adding your fix?

PR #12768, opened 2024-05-27
Add other scp white background errored sites from scp international hub

Eveheeero:


src/config/dynamic-theme-fixes.config
```diff
@@ -22488,15 +22488,19 @@ INVERT
 
 ================================
 
+lafundacionscp.wikidot.com
```
alexanderby, line 22491: We currently sort fixes by the first URL in the sub-list. Thus the order is broken now.

alexanderby (changes requested):

src/config/dynamic-theme-fixes.config
Eveheeero: does it right?

PR #12772, opened 2024-05-28
Adding my website

BjarneVerschorre:
bjarne.verschorre.be is a dark themed website, doesn't need this extension
Used `$ cat aa | sort > bb` to sort eveything


src/config/dark-sites.config
```diff
@@ -1,5 +1,3 @@
-*.iaas.store
```
alexanderby, line 1: It seems you added a single site to the list, but many records moved around. Please check.

alexanderby (changes requested):

BjarneVerschorre:
I've commited the change mr darkreader person

PR #12801, opened 2024-06-04
Update subdomains of docusign.com

felixonmars:
Wildcard subdomain was removed when merging https://github.com/darkreader/darkreader/pull/12706 which broke the inner pages. Let's add them back.

I'm not sure if you prefer a wildcard again or specific ones, so I chose the latter this time.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -7978,6 +7978,8 @@ img[src="assets/Uploads/dlp/images/logo-libreoffice.png"]
 ================================
 
 docusign.com
+apps.docusign.com
+account.docusign.com
 
 INVERT
 a[data-context="nav-main-logo"]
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -7978,6 +7978,8 @@ img[src="assets/Uploads/dlp/images/logo-libreoffice.png"]
 ================================
 
 docusign.com
+apps.docusign.com
+account.docusign.com
```
Myshor, line 7982: Maybe use `*.docusign.com` instead of adding all found subpage?

Myshor (commented):
See my proposal.

src/config/dynamic-theme-fixes.config
felixonmars: I did that in #12706 but it was removed ._.

I'll add it back.

PR #12802, opened 2024-06-04
Fix for dynadot.com

felixonmars:
Invert logo in signin form

Before:
Screenshot from 2024-06-04 18-45-11

After:
Screenshot from 2024-06-04 18-45-07

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -8342,6 +8342,7 @@ dynadot.com
 
 INVERT
 .navbar-logo
+.signin-form-content-new img
 
 ================================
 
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -8342,6 +8342,7 @@ dynadot.com
 
 INVERT
 .navbar-logo
+.signin-form-content-new img
```
Myshor, line 8345: after some analysis I propose to remove both classes from INVERT section and add:
`img[data-src="/tr/mainsite2023/navbar-logo-dark-2023.png"]`

As I found all logos uses the same image.

And as a bonus I would add:
`.icons-line img`
which would inver some icons in footer

felixonmars: Updated as suggested :)

Myshor (approved):

PR #12803, opened 2024-06-04
Fix for account.booking.com

felixonmars:
Invert logo

Before:
Screenshot from 2024-06-04 18-48-36

After:
Screenshot from 2024-06-04 18-49-17

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -689,6 +689,13 @@ INVERT
 
 ================================
 
+account.booking.com
+
+INVERT
+span[data-testid="header-logo"]
+
+================================
+
 account.live.com
 
 INVERT


src/config/dynamic-theme-fixes.config
```diff
@@ -689,6 +689,13 @@ INVERT
 
 ================================
 
+account.booking.com
```
Myshor, line 692: The whole page of booking has this black logo problem as I checked.

I propose remove this part with account.booking com then:
- go to booking.com part of fixes
- add `*.booking.com` line under booking.com to include all subdomains (account will be included)
- add span[data-testid="header-logo"] into the INVERT section there.

I don't see need of separate subpage fix if it's needed for whole booking.com 😉

Myshor (changes requested):

src/config/dynamic-theme-fixes.config
felixonmars: Understood. Merged into the other :)

PR #12823, opened 2024-06-08
dynamic-theme-fixes.config :: copilot.microsoft.com

Kaens:
Inverted the buttons that pop on hover in the history panel.


Myshor:
Can you share the screenshots before fix and after? (because I presume it's needed to be logged in or something)

Kaens:
Sorry! I don't know how GitHub hadn't notified me about the screenshot question, but I missed it... I suppose it's not needed anymore but I'll be checking my commits separately moving forward.

PR #12828, opened 2024-06-09
Fix math expressions for fandom.com subdomains

greentea-cup:
Fixes black math expressions on dark background.
Before:
Снимок экрана_2024-06-10_01-11-22
After:
Снимок экрана_2024-06-10_01-12-41


src/config/dynamic-theme-fixes.config
```diff
@@ -134,6 +134,13 @@ CSS
 
 ================================
 
+*.fandom.com
```
Myshor, line 137: As you start search fandom.com in sitefixes there are many different subpages for it and all have different fixes.
Adding *.fandom.com can break other subpages fixes so I propose to switch that to subpage you found with problem... If you have more than one then you can add other as aliases just under the first address.

Myshor (changes requested):

src/config/dynamic-theme-fixes.config
greentea-cup: I did a little research:
1. According to [Fandom documentaion on editing math formulas](https://community.fandom.com/wiki/Help:Math) their wiki engine generates `span.mwe-math-element` with various contents which are all considered a `<math>` tag (one that shows formulas). Such convertion (wiki -> browser) should be expected within all fandom wiki pages.
2. Searching through the up-to-date config shows that current fandom.com and <x>.fandom.com styles are only applied to other content, math formulas (`.mwe-math-element`) are untouched. **So existing fixes should not be broken with that change**. There are only one case that should be fixed (could be done within this pr) — that's `science.fandom.com` which inverts only inner parts inside `.mwe-math-element` which I suppose is partially correct or at least an obscure way to achieve the same behaviour.
3. There are countless other fandom wiki subdomains with math formulas aside `science`. To name a few: `math`, `latex-programming`, `lcme`.
4. Besides fandom, any MediaWiki powered page (such as Wikipedia) also uses `.mwe-math-element`.
5. Wikipedia fix also states two above-mentioned mwe parts [here](https://github.com/darkreader/darkreader/blob/390a1b7acd8a7b526054f373c898d6acafeb2c03/src/config/dynamic-theme-fixes.config#L28214) and this config is global to all wikipedia pages. (Also, I don't know in which cases outer element will work while inner won't, just stating one outer seems clearer to me).

From that I can state that the pr-proposed change doesn't require to be changes, besides, maybe, checking whether to leave outer `.mwe-math-element` or to copy two `.mwe-...` from existing fixes, or to change them instead to use a single outer tag.

Myshor: At least I would merge this *.fandom.com with fandom.com as it would be pain in the (...) to search if something is already included in fixes or not.

So move alias `*.fandom.com` under `fandom.com` then add 4th selector `.mwe-math-element` to INVERT section. 😉

greentea-cup:
Updated pr (message me if I did it wrong).
Removed `science.fandom.com` since it does literally the same but for a single subdomain.

Myshor (approved):

PR #12832, opened 2024-06-10
Update dynamic-theme-fixes.config for talishar.net

n21lv:
Restoring the backgrounds of images that shouldn't be inverted since IGNORE IMAGE ANALYSIS does not seem to work


n21lv:
Before fixes
image

After fixes
image

Extension disabled for the website
image

PR #12844, opened 2024-06-13
fix for messenger.com

Myshor:
Fix for strange color around bubbles.


dimostheenis:
Still an issue for me on messenger.com
librewolf_bnetPDrkLB

Myshor:
As there is possibility to use dark mode in messenger it was hard to see this by me because this white parts shows only when dark mode is disabled. :) I will push some fix to that ia a sec. 😉

Myshor:
Fix push in #13079

PR #12870, opened 2024-06-19
Update dark-sites.config

merajmasuk:


src/config/dark-sites.config
```diff
@@ -419,6 +421,7 @@ forum.atlasos.net
 forum.cswarzone.com
 forum.lastos.org
 forum.tribler.org
+forum.warthunder.com
```
Myshor, line 424: I recommend to remove this one as it has possibility to open in light mode too so not always dark site.

Myshor (changes requested):
One url to remove.

PR #12875, opened 2024-06-21
Added chatgpt.com to dark-sites.config

Gitoffthelawn:


Myshor:
On default settings it uses autodetection of browser/system theme - so it can end with light mode enabled. I can not add it to darksites because of that...

And when you have option to detect dark theme then it detects it nicely on this site.

Rejecting and closing.

PR #12884, opened 2024-06-22
fix smallcultfollowing.com

flippette:
fixes smallcultfollowing.com/babysteps

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -23670,6 +23670,7 @@ pre code
 pre
 .summary
 .markdown
+.markdown ul
 .footnotes
 
 CSS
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -23659,6 +23659,25 @@ img.section-bg-absolute
 
 ================================
 
+smallcultfollowing.com
```
Myshor, line 23662: I have nicer and smaller solution which won't invert whole page and picture.

You can use it here for your PR:
```
smallcultfollowing.com

CSS
*:not(a) {
}

```

Myshor (changes requested):

src/config/dynamic-theme-fixes.config
flippette: Great solution, I will update the PR. Thank you!

Myshor (approved):

PR #12886, opened 2024-06-23
Fix for mubi.com

McBruno712:
Fixes the 'Now showing' movie banner, mubi logo, notification icon and partially fixes movie critics's star rating (partially filled stars still show as empty).

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -17645,6 +17645,29 @@ INVERT
 
 ================================
 
+mubi.com
+
+INVERT
+.css-jd0r3a.e1kjxayp2
+.css-k7pew6.ef3qs2t4 img
+
+CSS
+.css-ztue73 {
+    filter: brightness(1) !important;
+}
+.css-1fdxz4i.e1f1mjpr0 path,
+.css-1fdxz4i.e1f1mjpr0 rect,
+.css-1fdxz4i.e1f1mjpr0 circle {
+    fill: #FFFFFF !important;
+}
+.css-ker29f.ed6irnx8 {
+    filter: invert(1) hue-rotate(180deg) !important;
+}
+
+IGNORE INLINE STYLE
+.star
+
+================================
 mullvad.net
 
 CSS


src/config/dynamic-theme-fixes.config
```diff
@@ -17645,6 +17645,29 @@ INVERT
 
 ================================
 
+mubi.com
```
flippette, line 17648: These class names look randomly generated, are they resilient to future site changes?

McBruno712: Let's hold the PR for a month or so, I'll let you know if it's still working by then, they change the banner every week or so.

src/config/dynamic-theme-fixes.config
```diff
@@ -17645,6 +17645,29 @@ INVERT
...
+.css-1fdxz4i.e1f1mjpr0 rect,
+.css-1fdxz4i.e1f1mjpr0 circle {
+    fill: #FFFFFF !important;
+}
+.css-ker29f.ed6irnx8 {
+    filter: invert(1) hue-rotate(180deg) !important;
+}
+
+IGNORE INLINE STYLE
+.star
```
alexanderby, line 17668: Maybe try specifying something like `svg.some-class *` here.

McBruno712: That breaks it, I could do .star-ratings svg * though.

src/config/dynamic-theme-fixes.config
McBruno712: > These class names look randomly generated, are they resilient to future site changes?

I sent an email about this to Mubi yesterday, let's see if they reply.

McBruno712:
@alexanderby  @flippette 

Since we're are holding this one for a while I took the time to fix some other things and I found something that I think could be considered a bug or an oversight in Dark Reader itself.

image
image

Dark Reader creates a new style, overriding the previous one, but it should include the url tag from the previous style so the image isn't removed. This behaviour is also present without any custom CSS added to Dark Reader. I am using the Dynamic mode, other modes don't have this issue.

Should I create a new Issue?

McBruno712:
If I understand the documentation correctly, adding that class below IGNORE IMAGE ANALYSIS should fix it, but is doesn't, could it be that IGNORE IMAGE ANALYSIS only work for images added in the HTML itself?

src/config/dynamic-theme-fixes.config
McBruno712: @flippette it's been 3 weeks and they've changed the banner multiple times, the fix is still working, I think it's ok to merge as the CSS classes don't seem to change.

flippette: Yup, looks good to me.

Besides, if things change we can always make another PR.

flippette (approved):
Seems like it works, I don't see why not to merge.

src/config/dynamic-theme-fixes.config
McBruno712: @alexanderby I added another commit that fixes the stars completely. Please have a look.

McBruno712:
What is missing for this to be merged?

flippette:
@Myshor Can you take a look at this?

src/config/dynamic-theme-fixes.config
```diff
@@ -17645,6 +17645,34 @@ INVERT
...
+}
+.css-ker29f.ed6irnx8 {
+    filter: invert(1) hue-rotate(180deg) !important;
+}
+
+IGNORE INLINE STYLE
+.star-ratings svg *
+
+================================
 mullvad.net
```
Myshor, line 17676: here we must have empty line before next url or unit tests won't pass

src/config/dynamic-theme-fixes.config
```diff
@@ -17645,6 +17645,34 @@ INVERT
...
+.css-ztue73 {
+    filter: brightness(1) !important;
+}
+.css-1fdxz4i.e1f1mjpr0 svg *,
+.css-1gez225.e1t0dj1j2 svg *,
+.css-1gez225.e13p4uxl2 svg * {
+    fill: #FFFFFF !important;
+}
+.css-ker29f.ed6irnx8 {
+    filter: invert(1) hue-rotate(180deg) !important;
```
Myshor, line 17669: why forcing invert this way and not include .css-ker29f.ed6irnx8 in INVERT section?

src/config/dynamic-theme-fixes.config
```diff
@@ -17645,6 +17645,34 @@ INVERT
...
+.css-1dev2nn.e18wowia1 .star
+
+CSS
+.css-ztue73 {
+    filter: brightness(1) !important;
+}
+.css-1fdxz4i.e1f1mjpr0 svg *,
+.css-1gez225.e1t0dj1j2 svg *,
+.css-1gez225.e13p4uxl2 svg * {
+    fill: #FFFFFF !important;
```
Myshor, line 17666: Which SVH is this that we force it to fill always white?

Myshor (changes requested):
Some changes and describing fixes needed.

Myshor:
One more important thing... I checked how the page will look like when Light theme of Dark Reader will be used with this fixes and ended with something like that:
obraz

So check if some of forced colors doesn't need to be used as inverted with ${} for example if you use #FFFFFF then to use black in light mode and white in dark you use ${#000000} instead of #FFFFFF

One last thing:
https://mubi.com/en/awards-and-festivals

All logos are black here... would be nice if there will be added `img[src*="logo-small_black.png"]` into INVERT section too.

Summarizing.
I want to have nicely looking page in dark and light theme working without any problems before merging. 😉

src/config/dynamic-theme-fixes.config
McBruno712: It's the bottom logo.

McBruno712 (commented):
Fixed all you mentioned @Myshor

McBruno712:
Are there any plans to address [this bug ](https://github.com/darkreader/darkreader/pull/12886#issuecomment-2198436506)?

Myshor:
Checked how it looks using light/dark theme in Dark Reader + compared how it looks with Dark Reader disabled.

I will push one change to this PR which won't be even visible for dark theme. 😝

Myshor (approved):

Myshor:
Merged... If you find something you will want to fix and I didn't catch with my fast walk through the site then just create new PR. 🙂

PR #12900, opened 2024-06-27
Fix code blocks on gnu.org

alexalekseyenko:
gnu.org makes pretty heavy use of `<pre>` tags (see the [elisp manual](https://www.gnu.org/software/emacs/manual/html_mono/elisp.html) for instance) and this fixes the inversion of those.

Before:
Screenshot 2024-06-26 at 23 34 10

After:
Screenshot 2024-06-26 at 23 34 23


Myshor:
I can not find the issue you try to fix using Firefor or MS Edge with Dark Reader in Filter mode:
obraz

What version of Dark reader and what browser you are using?

alexalekseyenko:
You're right, I think it must have been something I forgot to delete from my inversion fix editor. Thanks for checking and sorry for the bother!

PR #12913, opened 2024-06-28
Change restricted pages option text

Tiagoquix:
Change text to let users know about the README.

Is there any way to add a hyperlink to this text? Would be nice.

Also, shouldn't this option be hidden for non-Firefox users? As far as I know, Chromium-based browsers do not allow any extension on restricted pages, and there's no easy way to disable this protection.


Tiagoquix:
@alexanderby friendly ping

PR #12916, opened 2024-06-28
add proton.me to dark sites

jcassette:


alexanderby:
I'm afraid these websites can have both dark and light mode.

PR #12925, opened 2024-07-03
fix for wanikani.com

eric-r-n:
wanikani.com:
Removed text-shadow on certain selectors to fix difficult to read text.


Myshor:
I can not check it. I presume it's needed to be signed in. Can you at least share some before/after fix screens?

eric-r-n:
Here are some before and after screenshots. While not included, it's particularly difficult to read on mobile as the shadows make the text look "fuzzy" on the smaller screen. Please let me know if you need anything else.

full_after
full_before
zoom_after
zoom_before

PR #12926, opened 2024-07-03
added fix for elektronik-kompendium.de

GrohJann:
The latex formulas weren't inverted, so that these formulas were completely unreadable

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -8749,7 +8749,8 @@ img[alt="logo"]
 elektronik-kompendium.de
 
 INVERT
-img[alt="Formel"]
+img[src*="formel/"]
+img[src*="schalt/"]
 
 ================================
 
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -8746,6 +8746,13 @@ img[alt="logo"]
 
 ================================
 
+elektronik-kompendium.de
+
+INVERT
+img[alt="Formel"]
```
Myshor, line 8752: I checked some problematic graphs and formulas and I did not found that alt="Formel" in them so I suggest to change it into:
`img[src*="formel/"]`
as all problematic pictures are stored in this folder on the server.

Myshor (changes requested):
Check proposition.

src/config/dynamic-theme-fixes.config
GrohJann: You are are right it should have been `img[src*="formel/"]`.
I also noticed that circuit diagrams weren't looking correctly.

Now I just have to figure out how to add the changes to the pull request

PR #12941, opened 2024-07-06
Fix for wmata.com

pnaf:
- Fixes logo on most pages.
- Improves legibility of language widget.

Before:

After:


Myshor:
I can not reach this site. Can you at least share some screens before and after the fix?

pnaf:
Good to know. I added another fix and attached some screenshots.

PR #12946, opened 2024-07-09
Update dark-sites.config: Add lyricify.app

tsavpyn:


Myshor:
Page detects system/browser settings and even can be forced to be in light mode so it's not always dark.
obraz

Rejectring Pull Request.

PR #12971, opened 2024-07-15
Add fix for campus.datacamp.com

hrnaltun:
When dark theme is enabled, the code tag is not readable. I fixed this for campus.datacamp.com 
Default
Default
Fixed
Fixed

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -6773,8 +6773,14 @@ video#vjs_video_3_html5_api.vjs-tech {
     color: var(--darkreader-neutral-text) !important;
     mix-blend-mode: normal !important;
 }
-.css-alxior code {
-    color: #f0f0f0 !important;
+.css-alxior code,
+pre {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -6773,6 +6773,10 @@ video#vjs_video_3_html5_api.vjs-tech {
     color: var(--darkreader-neutral-text) !important;
     mix-blend-mode: normal !important;
 }
+.css-alxior code {
+    color: #f0f0f0 !important;
```
Myshor, line 6777: As Dark Reader can be used with Dark Theme (by default) it can be switched to Light theme.
That's why we try to avoid forcing colors that are good only in dark. ;)

I propose to use:
`color: ${#0f0f0f} !important;`

this will invert the color to #f0f0f0 in dark theme and use #0f0f0f in light.

Myshor (changes requested):

Myshor (approved):

PR #12992, opened 2024-07-20
Fix for fairvote.org

pnaf:
Fixes navigation menu.

Before:

After:


Myshor:
Checked how this work with light scheme in Dark Reader and Approve. 😉

Issue #10922, opened 2023-03-16
[Bug] Dynamic Filter edits Google Sheets Cell Contents


Obsidianninja11:
### Prerequisites

- [X] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this bug to avoid opening a duplicate.
- [X] I can reproduce this bug in a new, unmodified web browser profile with Dark Reader installed as the only extension.
- [X] I understand I need to use the [Broken Website Report template](https://github.com/darkreader/darkreader/issues/new?assignees=&labels=Broken+Website&template=broken-website-report.yml&title=%5BBroken+Website%5D+Replace+with+title) if this bug I am reporting occurs on a single website.

### Bug Description

When editing a cell in google sheets with the dynamic filter on, the text color is reset to black (or white if light mode is enabled).
This does not happen in the other filters (Filter, Filter+, Static).

Why this is a problem:
Dynamic filter provides the best dark mode for google sheets, but editing a cell that utilizes colored text is a hassle with the dynamic filter on (you'd have to manually change the text color back to what it was before you edited it).

### Website Address

https://docs.google.com/spreadsheets

### Steps To Reproduce

(See screenshots for additional info)

1. Turn dark reader and the dynamic filter on.
2. Edit a cell with colored text.
3. Press enter to confirm editing the cell.

4. Additional info: The bug does not appear if you finish editing the cell with dark reader toggled off.

### Expected Behavior

Editing a cell with the dynamic filter on shouldn't change cell contents, and the cell should remain the same color after editing it with the filter enabled.

### Actual Behavior

Editing a cell with the dynamic filter on changes the cell's text color to black.

### Screenshots

1. A red colored cell with dynamic filter on: 
![image](https://user-images.githubusercontent.com/108832807/225727364-9a031f49-273d-4411-bf5d-0d2db519da48.png)
The same cell with dark reader toggled off:
![image](https://user-images.githubusercontent.com/108832807/225727393-b746c368-d54f-4beb-a393-20731518f3e7.png)

2. The red colored cell after starting to edit it (with filter on):
![image](https://user-images.githubusercontent.com/108832807/225727422-cb4aa473-eea4-4ebb-834f-b2de07054102.png)
The same cell with dark reader toggled off:
![image](https://user-images.githubusercontent.com/108832807/225727435-0d8a4a91-0273-4153-933c-6f1c4a5c9afd.png)

[...]

PR #13009, opened 2024-07-23
Fix dark reader *permanently* modifying text color in google sheets

Obsidianninja11:
Resolves: https://github.com/darkreader/darkreader/issues/10922

Dark Reader was permanently modifying text color of edited cells in google sheets (as in when editing a cell with Dark Reader enabled, it would change the color and stay changed even when refreshing or revisiting the sheet)
This pr fixes that behaviour.

For some reason, if I were to add a `docs.google.com/spreadsheets` url to dynamic theme editor it would break the invert filter (outlined in my issue), so I was forced to put the fix under `docs.google.com` which isn't ideal, but I don't think it will be a problem? If it is, a more specific selector could be used like `body:has(.classSpecificToSheets) .cell-input > span` or something, but that seems a bit janky to me.

If there's a better way to do this or any other way it can be improved, please let me know.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -8011,6 +8011,9 @@ input.docs-title-input {
     fill: var(--darkreader-selection-background) !important;
 }
 
+IGNORE INLINE STYLE
+.cell-input > span
+
 ================================
 
 docs.google.com/picker


Myshor:
@Obsidianninja11 Can you share link to some document the fix is needed? I went through different sheets I have available in my google docs and I don't see any changes with and without the fix.

PR #13013, opened 2024-07-24

Kaens:
Fixes a rather unreadable white-on-thick-yellow highlighter.


Myshor:
@Kaens Can you share exact URL where it is used that we can check how it looks before and after fixes? I just want to be sure the fix will work fine when dark and light theme of Dark Reader will be used or it will need some cosmetical changes. 😉

Kaens:
Sure, for example this https://www.baitoru.com/contents/serial/0037.html

PR #13018, opened 2024-07-25
Adding my personal website that automatically activates dark theme if system preferred.

tonideleo:


Myshor:
As this site can be dark/light it can not be added into darksites - this list is for permament dark ones only.

Closing this PR but...

You can add entry to https://github.com/darkreader/darkreader/blob/main/src/config/detector-hints.config using below schema (as I already checked how it works on your site). 😉

```
TARGET
html

MATCH
[data-theme="dark"]

```

PR #13038, opened 2024-07-29
Fix home.gamer.com.tw user-styled pages

lumynou5:
This affects, for instances:
- https://home.gamer.com.tw/homeindex.php?owner=vc
- https://home.gamer.com.tw/homeindex.php?owner=s861102

### DIFF
--- a/src/config/detector-hints.config
+++ b/src/config/detector-hints.config
@@ -263,6 +263,12 @@ NO DARK THEME
 
 ================================
 
+home.gamer.com.tw/profile/index.php
+
+NO DARK THEME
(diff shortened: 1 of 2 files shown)


lumynou5:
The text colors of elements with ignored background images are still inverted by DarkReader. It results white-on-white or black-on-black text colors.
Is there a way to indicate DarkReader not to set `color`?

Or should we just remove all background images and keep images without text on them (i.e. banners) only? This breaks page appearance a lot, but it seems hard to keep user-generated styles and dark theme at same time.

PR #13058, opened 2024-08-01
Fix cnrtl.fr syntagma highlighting

nicolasmeric:
The original color, rgb(192, 255, 192), for syntagma highlighting makes the text illegible

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -140,6 +140,15 @@ INVERT
 
 ================================
 
+*.cnrtl.fr
+
+CSS
+.tlf_csyntagme {
+    background-color: rgb(22, 89, 0) !important;
+}
+
+================================
+
 *.crm*.dynamics.com
 
 CSS
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -143,8 +143,14 @@ INVERT
 *.cnrtl.fr
 
 CSS
+.tlf_cdefinition {
+    background-color: ${rgb(255, 255, 0)} !important;
+}
+.tlf_cdomaine {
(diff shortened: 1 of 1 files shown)


Myshor:
@nicolasmeric Can you share URL where this fix applies?

nicolasmeric:
Yes, for example:
https://www.cnrtl.fr/definition/r%C3%A9petition

Way to reproduce the issue:

- Disable dark reader
- Go to the url: the color set for .tlf_csyntagme is rgb(192, 255, 192) from the script https://www.cnrtl.fr/js/utilitiesv30.js if I'm not mistaken
- Enable dark reader (in dynamic mode): the color is reset to rgb(22, 89, 0) by default which is fine and allows legible text
- Reload the page: the color is reset to its original illegible value rgb(192, 255, 192)

Incidentally, I found out that .tlf_cdefinition and .tlf_cdomaine have the same issue using other colors.

Should I submit another pull request?

src/config/dynamic-theme-fixes.config
```diff
@@ -140,6 +140,15 @@ INVERT
 
 ================================
 
+*.cnrtl.fr
+
+CSS
+.tlf_csyntagme {
+    background-color: rgb(22, 89, 0) !important;
```
Myshor, line 147: change to:
`background-color: ${rgb(233, 166, 255)} !important;`

This will use (22, 89, 0) when dark theme is used or (233,166, 255) if light theme with dark text.

Myshor (changes requested):
After life check using light/dark theme some polishing change is needed.

Myshor:
> Should I submit another pull request?

You have some change request so you can add the second class into fix too with changes. 😉

Myshor (approved):

PR #13094, opened 2024-08-07
Add kasmedia to dark list

imalfect:
KasMedia has an available built-in dark mode, this pull request is also influenced by the fact that dark reader is breaking framer-motion animated images on the website, causing a black box instead of the image (see featured article, which is the biggest image on top of the website, it breaks with dark reader being enabled.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -600,6 +600,7 @@ kadantiscam.netlify.app
 kaioken.dev
 kalence2.github.io
 kaoskrew.org
+kasmedia.com
 kazimagazine.com
 kbh.games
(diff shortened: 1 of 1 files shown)


Myshor:
As this page can be set into dark and light mode then it can not be added to Dark Site list.

It can be added here: https://github.com/darkreader/darkreader/blob/main/src/config/detector-hints.config

using:
```
TARGET
html

MATCH
.dark

```

And for the missing picture I will push sitefix in a minute.

Myshor:
So if you want to put this site into detector-hints just go ahead but this PR i must reject and close.

PR #13096, opened 2024-08-07

ForeverFloating:
Added West Virginia University custom CSS

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -301,6 +301,110 @@ CSS
 
 ================================
 
+*.wvu.edu
+
+CSS
+.bg-wvu-blue {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -301,6 +301,110 @@ CSS
 
 ================================
 
+*.wvu.edu
```
Myshor, line 304: The whole fix should be under *.webex.com

Now test for alphabetical order fails.

Myshor (changes requested):

Myshor:
Only URL was needed to be fixed as the sites must be in aphabetical order... as I see you even sorted selectors. For the future - they are not needed to be sorted alphabetically. 😉

Myshor (approved):

PR #13097, opened 2024-08-07
Add Valve-related sites to global dark list

Tiagoquix:
- `half-life.com` looks fine with Dark Reader, but the special subpage `half-life.com/*/halflife25` has some problems with Dark Reader ON, and it can be disabled specifically for that page.
  - `*` here is for language (`/en/`, `/fr/` etc.).
- `l4d.com` is already dark.
- `thinkwithportals.com` is already dark, except for the background, but enabling Dark Reader there defeats the site purpose's of resembling the in-game levels.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -515,6 +515,7 @@ hackthebox.com
 hackthebox.eu
 hackthissite.org
 hacktoberfest-projects.vercel.app
+half-life.com/*/halflife25
 halowaypoint.com
 hardforum.com
 hardstuck.gg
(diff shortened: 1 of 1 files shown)


Myshor:
Except l4d.com I have doubts the rest should be added to dark list as they are not dark, and Dark Reader makes some cosmetic changes there.

portal2 page could be accepted as it looks strange but halflife makes some texts back to black.

I won't merge it. Let @alexanderby decides.

Myshor:
I am strict about pages that are not fully dark. That's why I called Alexander here. 😉

Tiagoquix:
friendly ping @alexanderby

PR #13100, opened 2024-08-08
Add dumpspace.spuckwaffel.com to global dark list

Wuemeli:
Add dumpspace.spuckwaffel.com to global dark list


Myshor:
As this page can be set into dark and light mode then it can not be added to Dark Site list.

It can be added here: https://github.com/darkreader/darkreader/blob/main/src/config/detector-hints.config

using:
```
TARGET
html

MATCH
.dark

```

Rejecting PR and closing.

Issue #12762, opened 2024-05-26
[Broken Website] Google Sheets Loads White


Michael13435:
### Prerequisites

- [X] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this website issue to avoid opening a duplicate.
- [X] I can reproduce this website issue in a new, unmodified web browser profile with Dark Reader installed as the only extension.

### Website Issue Description

In both Dynamic and filter+ mode the page briefly flashes white while loading

### Website Address

https://docs.google.com/spreadsheets/u/0/

### Steps To Reproduce

1. Open a new spreadsheet

### Operating System

Windows Version	10.0.22631 Build 22631

### Web Browser name and version

Firefox 126.0

### Dark Reader version

4.9.85

Michael13435:
Note that the current fix for docs.google.com only works some of the time, and only when reloading the page, not when going to a new spreadsheet

rajkishan16:
This happens to me as well in Microsoft Edge, Dark Reader version: 4.9.86. This is occurring in Google Docs as well as Google Sheets. Reloading the page doesn't help.

@Michael13435 what is the current fix that you use?

Kwispy222:
I was able to recreate this in both a fresh version of Firefox and Google Chrome. I will be working on it and hope to have a PR linked to this by Thursday night (8/8) at the latest.

Myshor:
I merged the mentioned fixes but not closing the bug as main spreadsheet still flashes on Dynamic mode and Filter mode was not touched at all. 😉

Myshor:
Reverted last changes in #13119 as there was some issue mentioned in PR after merge.

PR #13101, opened 2024-08-08
Made some CSS changes in order to smooth transition on the website

Kwispy222:
I was able to fix a large amount of what was mentioned in this issue (https://github.com/darkreader/darkreader/issues/12762).
Specifically, I was able to eliminate the lag time of swtitching on and off for the top bar and option sections of a new spreadsheet as described in the issue.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -8092,6 +8092,75 @@ input.docs-title-input {
 .sketchy-text-background + g g rect {
     fill: var(--darkreader-selection-background) !important;
 }
+.docs-sheet-active-tab {
+    background-color: var(--darkreader-neutral-background) !important;
+    border-color: rgb(64, 69, 72) !important;
+    color: var(--darkreader-neutral-text) !important;
+}
+.goog-toolbar-separator {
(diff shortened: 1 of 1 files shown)


Myshor:
Tested in on Firefox and I do not see any differences with flashing when this fix is applied in sitefixes or is not appled.

Still I see white flash before everything is set to dark.

As I do not see any changes after fix I leave it not merged.

Kwispy222:
Interesting. @Myshor thanks for taking a look at this PR! Are you seeing any difference between current main branch and this branch? I installed my build from scratch on a fresh firefox install and I see that there is still white flash with the "t-formula-bar-input" but the rest of the top section (i.e. spreadsheet title, share button, bold/itallic/underline buttons, etc) switch immediately which is not the case when I was testing on the current main branch.

Thanks for any input you can give!

Myshor:
As you mentioned the top section now I see this part is not flashing that much. OK. Starting unit tests if they pass I will merge. 😉

Obsidianninja11:
This messes up the look of the header, and doesn't even help with flashing very much, since the main flash of the waffle grid can still happen.
Before: 
After:

Myshor:
ok... I'm reverting as it won't help that much and causes problems.

Issue #13044, opened 2024-07-30
[Feature Request] Revise Labels for Improved Usability in the Site List Tab


zrob1122:
### Prerequisites

- [X] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this feature request to avoid opening a duplicate.

### Is this feature request related to a problem?

Yes

### Feature Request Description

The current labels for the tabs within the "Site List" tab are "Invert Listed Only" and "Not Invert Listed." I feel these are not immediately intuitive for certain users, or at least not as immediately intuitive as they can be. These labels are somewhat "technical" and may not clearly convey their functionality to users unfamiliar with the terminology. This can lead to confusion and may result in users skipping over these features entirely, not realizing the options available to them. I feel that changing these labels to something along the lines of (if not this exactly) "Sites to Keep in Light Mode" and "Sites to Keep in Dark Mode" could go a long way in alleviating this issue.

### Screenshots

![issue pic](https://github.com/user-attachments/assets/6544bc1f-77d2-4dca-8ed5-bf99f4bc4ff3)

### Additional Context

Again, I feel the names of the tabs within the Site list tab should be changed to something more immediately "recognizable." Even more simplistic names like "Permanent Light Mode" and "Permanent Dark Mode" could be good options. I realize this feature request may sound somewhat sophomoric, but I do think this would make the Site list tab more user-friendly and intuitive for all users.

PR #13103, opened 2024-08-09
Update index.tsx, in order to update Site List tab labels for better usability

zrob1122:
Updated Site List tab labels to be more intuitive, as discussed in issue #13044. This PR changes the current labels in the Site List tab from "Invert Listed Only" and "Not Invert Listed" to "Sites to Keep in Dark Mode" and "Sites to Keep in Light Mode" to improve clarity for all users.


Myshor:
This is BIG NO! You change variable names here that are used in translation configurations.

Closing without merge.

PR #13104, opened 2024-08-09
Update en-US.config, in order to update Site List tab labels for better usability

zrob1122:
Updated Site List tab labels to be more intuitive, as discussed in issue #13044. This PR changes the current labels in the Site List tab from "Invert Listed Only" and "Not Invert Listed" to "Sites to Keep in Dark Mode" and "Sites to Keep in Light Mode" to improve clarity for all users.

### DIFF
--- a/src/_locales/en-US.config
+++ b/src/_locales/en-US.config
@@ -95,10 +95,10 @@ Apply settings to current website only
 Site list
 
 @invert_listed_only
-Invert listed only
+Sites to Keep in Dark Mode
 
 @not_invert_listed
-Not invert listed
+Sites to Keep in Light Mode
 
 @add_site_to_list
 Add site to list


Myshor:
Not sure it works like you proposed to change... As I understand Invert Listed only enables dark mode only on sites that are on invert list and do not touch other.

When you switch to Not Invert ListedDark Reader enables dark mode on all except listed on this list.
Changing the description on EN-US can be more confusing... especially when you change it only on this language config and all the rest stays in older version.

For me it's NO for merge... but I will leave it here so @alexanderby can give more feedback.

alexanderby:
Thank you for your pull request. We will have 2 separate lists after the redesign. This PR will break the layout, because there's not enough space.

Issue #13110, opened 2024-08-10
[Broken Website] Contrast of highlight of Culter, Maine website is not strong enough


byrnelo:
### Prerequisites

- [X] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this website issue to avoid opening a duplicate.
- [X] I can reproduce this website issue in a new, unmodified web browser profile with Dark Reader installed as the only extension.

### Website Issue Description

Hi -- I went on the town website for Cutler, Maine and the contrast on some of the blocks is not strong enough to see so the user cannot tell if they are hovering on it.

### Website Address

https://www.cutlermaine.net/

### Steps To Reproduce

1. Go to website
2. Scroll down to Renew boat registration 

### Screenshots

![image](https://github.com/user-attachments/assets/0ab7e28f-195e-4065-a3d1-d5c4cf069b00)
![image](https://github.com/user-attachments/assets/83852f30-d73b-48bc-b1f3-8536aa11d5ee)

### Operating System

Windows 64

### Web Browser name and version

Chrome Version 127.0.6533.72

### Dark Reader version

Version 4.9.88

byrnelo:
Should  be a lighter grey and not completley black -- cannot see in the first screenshot.

Myshor:
The gray on disabled Dark Reader comes from opacity set to 0.8 and white background. When Dark Reader makes background dark then opacity from black to dark is almost invisible. Adding gray background on hover in sitefix.

Myshor:
Sitefix added.

- If you want to have this fix faster you can set **Synchronizee sites fixes** option to check in Advanced section of Settings.
![obraz](https://github.com/user-attachments/assets/9dea8b66-b77a-46ab-8aa6-e50d995718c3)
- With that enabled newest sitefixes will be updated every browser start. 😉
- To open all settings you must click on:
(Old/classic design) More -> All Settings
![obraz](https://github.com/user-attachments/assets/8c17b1f1-e85e-4672-8a88-85930e178e02)
(New/Preview design) Settings in popup
![obraz](https://github.com/user-attachments/assets/e4f586f8-8ce4-47bc-90d7-7451eb030439)

PR #13111, opened 2024-08-10
Update to the Cutler Maine Site

byrnelo:


Myshor:
If this was made as a fix for #13110 then I do not see what should it fix.
2nd thing - as I first check bugs then PRs then I already made fix for the bug in #13115 

As I do not see what your proposal whuld do I reject this PR and close.

PR #13123, opened 2024-08-12
Fix for  eeworld.com.cn

felixonmars:
Invert logo


src/config/dynamic-theme-fixes.config
```diff
@@ -9002,6 +9002,13 @@ CSS
 
 ================================
 
+eeworld.com.cn
```
Myshor, line 9005: I would add alias *.eeworld.com.cn under
then add needed invert from #13124

Myshor (changes requested):

src/config/dynamic-theme-fixes.config
felixonmars: Updated as requested.

Myshor (approved):

PR #13124, opened 2024-08-12
Fix for m.eeworld.com.cn

felixonmars:
Invert logo


src/config/dynamic-theme-fixes.config
```diff
@@ -16147,6 +16147,13 @@ INVERT
 
 ================================
 
+m.eeworld.com.cn
+
+INVERT
+.logo
```
Myshor, line 16153: I would use .header-logo-search-login

as there are two black icons on the right too.

And merge everything and 1 PR - just check review in #13123

Myshor (changes requested):

src/config/dynamic-theme-fixes.config
felixonmars: Updated as requested.

PR #12466, opened 2024-03-18
Invert potential article images in transparent background for common blogs including silent blogs

fritx:
Style in silent blogs gets overriden and takes no effect:
[fix transparent image background in darkmode
](https://github.com/fritx/silent/commit/22ae7e04b42691f0a8c4888ce01ec81a01a12e16)
Preview Link: https://blog.fritx.me/?2015/01/fs-auth

Before fix:

After fix:

PR #13137, opened 2024-08-14
Revert patch for fritx.me to fix unexpected image filter

fritx:
Reverts previous PR: https://github.com/darkreader/darkreader/pull/12466 , which I have fixed on my own side.

And fixes all the unexpected image filter.

image

-- https://blog.fritx.me/?aboutme

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -10908,16 +10908,6 @@ body {
 
 ================================
 
-fritx.me
-*.fritx.me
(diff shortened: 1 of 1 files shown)


Myshor:
BTW if you want to exclude some pic like this you can add :not() to the selector

so if you want to have previous fixes excluding your photo you can create PR with code below: 😉
```
fritx.me
*.fritx.me
*.*.fritx.me
fritx.github.io

INVERT
.main-page img:not(img[src="p/avatar.jpg"])

```

PR #13158, opened 2024-08-20
Update dynamic-theme-fixes.config to fix logo on *.abc.net.au

colin-tso:
Update dynamic-theme-fixes.config to fix logo on *.abc.net.au, by removing existing svg path fill to make darkreader inversion work.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -102,10 +102,11 @@ IGNORE INLINE STYLE
 
 *.abc.net.au
 
-CSS
-[data-component="ABCLogo"].path {
-    fill: "";
-}
+INVERT
+[data-component="ABCNewsLogo"] *
+
+IGNORE INLINE STYLE
+[data-component="ABCLogo"] *
 
 ================================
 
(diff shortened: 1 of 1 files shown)


Myshor:
Can you share some example URL? Because I think it could be fixed with IGNORE INLINE STYLE not that CSS

and as I went through some pages on abc.net.au I can not find anything with `data-component="ABCLogo"` in the html code.

colin-tso:
image
This is on the main page ([www.abc.net.au](url)) and is on the "Masthead" for most pages.
I think instead I can use` INVERT` - for some reason I did not think of this
I had tried `IGNORE INLINE STYLE` which did not work.

colin-tso:
Below seems to address the intent of the fix better.

```
*.abc.net.au

INVERT
[data-component="ABCLogo"] > path
[data-component="ABCNewsLogo"] > path
*.abc.net.au/emergency
```

Will put in additional commits if happy to proceed.

Myshor:
As you started this PR I will give you the code what I should prefer. 😉
```
*.abc.net.au

IGNORE INLINE STYLE
[data-component="ABCLogo"] *

```

Tested it and it works fine.

PR #13166, opened 2024-08-23
Fix for GitHub.com

oberrich:
Applies `display:inline-flex` to `<a href"...">` sequences inside markdown bodies, this globally fixes an issue with annoying whitespaces between badges on GitHub.

See also:
https://stackoverflow.com/a/65645230
https://github.com/orhun/git-cliff/pull/809


Myshor:
Maybe stupid question but where I can see badges on github to see live what this fix would do?

PR #13168, opened 2024-08-23
 Fixed fritz.box navigation Menu colors in dynamic-theme-fixes.config

nodscher:
Fixed white Gradients on the navigation Menu on the fritz.box website (local dns to 192.168.178.1 for controlling the AVM fritz box; only accessible with a fritz box router).


Myshor:
As it is only available through router then I will need some screens:

1. Before fix.
2. After fix in dark theme set in Dark Reader.
3. After fix in light theme set in Dark Reader.

I see you force some color but I am afraid it will break the menu when someone would use the light theme in Dark Reader.

PR #13170, opened 2024-08-24
fix(reg): apply font to inline styled elements setting `font-family`

oberrich:
Fixes a regression introduced in my previous commit to the file.


oberrich:
Actually we should also replicate the `:not(...)` parts on the `[style*="font-family"]` selector to handle `pre`, `code` tags or the presence of a common icon font class.
Might be worth to just drop the `font-` wildcard entirely to avoid this headache?

Myshor:
Not sure if it won't break the icon fonts. @alexanderby there was some code changes for not including icon fonts in Dark Reader font forcing. Maybe you take a look here if it can be added like proposed.

oberrich:
Closing this for now as we have no good way of dealing with this as of today.

PR #13171, opened 2024-08-24
Add *.furaffinity.net to dark sites

pnaf:


Myshor:
@pnaf Are there some more furaffinity.net sites?
Maybe it would be easier to merge www.furaffinity.net and this one into *.furaffinity.net and have it as one entry?

pnaf:
@Myshor Good question. I thought this would be the only other subdomain, but I've discovered `shop.furaffinity.net`, which also uses a dark theme.

Given that all their subdomains I've encountered so far seem to be using a dark theme, switching to `*.furaffinity.net` sounds like a good idea.

PR #13187, opened 2024-08-27
Added acumatica.com to dynamic-theme-fixes.config

bengineered1:
Making checkbox background dark for `*.acumatica.com`. Without this, checkmarks would invert and their background would not, turning it white-on-white.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -110,6 +110,18 @@ IGNORE INLINE STYLE
 
 ================================
 
+*.acumatica.com
+
+CSS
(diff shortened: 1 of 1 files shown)


bengineered1:
Putting back to draft - was using it more today and found an error is being thrown in a different place in the site with these rules added. Going to investigate further and resolve before marking this ready to merge.

`undefined: Failed to execute 'insertRule' on 'CSSStyleSheet': The index provided (4294967295) is larger than the maximum index (0).`

Myshor:
And if you can share screens what exactly you fix change... I mean before/after fix.

PR #13210, opened 2024-09-02
fix icloud.com

sorpdev:
### Homepage

Before
image

After
image

### Subpages (e.g. icloud.com/mail)

Before
image

After
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -13253,13 +13253,22 @@ INVERT
 #cloudos-loading-spinner
 .editor-container
 .light > svg
-.apple-icloud-logo > g > path:first-child
-.apple-icloud-logo.dark > g > path:last-child
+#.apple-icloud-logo > g > path:first-child
+#.apple-icloud-logo.dark > g > path:last-child
 
 CSS
 .cw-pane-container,
 .reminders {
     background-image: none !important;
+}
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -13253,14 +13253,23 @@ INVERT
 #cloudos-loading-spinner
 .editor-container
 .light > svg
-.apple-icloud-logo > g > path:first-child
-.apple-icloud-logo.dark > g > path:last-child
+#.apple-icloud-logo > g > path:first-child
```
alexanderby, line 13256: Hi, could you please tell about the `#` symbol? If you wanted to disable these lines, you had to simply remove them. The `#abc` is an `id="abc"` selector in CSS, but together with a dot it produces an invalid selector.

sorpdev: oh sorry my bad 😓, you're right I should've just removed the lines. I didn't think about the id selector at the time, thought I would comment it out in case it may be needed in the future, but I don't think it will.

sorpdev: but it seems to support comments. just tried with the dev tools.
just had to use // instead of # 

(I was at the same time working on configuration stuff and there is # used as a comment, so I mixed it up 😅)

image

PR #13218, opened 2024-09-04
Add some sites to dark-sites.config

halilim:
primevideo.com, codeium.com, hachyderm.io, and orbstack.dev


alexanderby:
I decided to remove the regular expression. If there's no simple solution, let's not make things too complex. We have a dark theme detection feature for this.

PR #13227, opened 2024-09-07
Fix for www.iyingdi.com

lumynou5:


lumynou5:
I forgot it actually begins with `i`.

alexanderby:
Thank you. I'm surprised that our formatting tests didn't break for the multi-line gradient.

PR #13249, opened 2024-09-15
`envs.net` isn't entirely a `dark-site`

Rudxain:
Some URLs may point to arbitrary HTML+CSS, which may not support dark-theme


Myshor:
Do you have some examples where the Dark mode is not supported? I went through some pages and did not find any issues.

Rudxain:
https://envs.net/~binarycat/pokestat

PR #13251, opened 2024-09-15
Fix for eksisozluk.com

hrnaltun:


Myshor:
Can you tell me what exactly is fixing this PR?

I went to eksisozluk.com and can not see any difference with and without fix.

hrnaltun:
When I first wrote the code, the site looked like this
GXlfFdPWQAAazLR
Now it seems to have fixed itself without any code changes, so I guess we can close the pull request.

PR #13270, opened 2024-09-21
dark-sites: Added epilogue sites 

0PandaDEV:


corentinderuyck:
Hello, according to [https://github.com/darkreader/darkreader/blob/main/CONTRIBUTING.md#adding-a-website-that-is-already-dark](the contribution rules) we can only add completed websites and not websites that are under construction.

PR #13273, opened 2024-09-21
Aliexpress RU checkbox fix

TheMiner203:


Myshor:
Can you show some screens before and after fix to see which checkbox exactly is fixed.

TheMiner203:
изображение
without fix

изображение
with fix

PR #13275, opened 2024-09-22
Fix Codeforces.com Navbar

HussainElemam:
fix the color of the nav bar elements. It was not visible before because it was black on a dark background.

Navbar before fix:
before

Navbar after fix:
after


HussainElemam:
Navbar before fix:
before

Navbar after fix:
after

PR #13283, opened 2024-09-24
Fix for typst.app/universe

Andrew15-5:
The problem I faced is that `#results li`/`#banner` style inverts the style for all children, which is why I had to _additionally_ invert all of their key children.

Note that the default `filter: invert(100%) hue-rotate(180deg) contrast(90%) !important;` inversion style doesn't look good here, so I used `filter: invert(95%) hue-rotate(180deg) contrast(90%) !important;` instead.

I made a separate `/universe` entry because Typst Universe is a different/isolated website, so it insures that some other `#banner` etc. styles on other pages won't be affected. And the previous SVG fix is more of a global fix (applies to Universe and docs websites).

## Before

https://typst.app/universe/search:

image
---

https://typst.app/universe/package/alchemist:

image
---

https://typst.app/universe/package/a2c-nums:

image

## After

https://typst.app/universe/search:

image
---

https://typst.app/universe/package/alchemist:

image
---

https://typst.app/universe/package/a2c-nums:

image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -28017,6 +28017,7 @@ img[src$=".svg"]:not([alt="Typst"])
 .carousel .top-text
 .carousel .title
 .carousel .bottom-text
+#categories ul li a
 
 CSS
 #results li a,
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -27999,6 +27999,25 @@ img[src$=".svg"]:not([alt="Typst"])
 
 ================================
 
+typst.app/universe
+
+INVERT
+#results li .title .name
+#results li .description
+#results li .template-thumbnail
+#banner
+#banner .title
+#banner .description
+#banner .featured-label
+
+CSS
+#results li a,
+#banner {
+    filter: invert(95%) hue-rotate(180deg) contrast(90%) !important;
+}
+
+================================
+
 ubank.com.au
 
 CSS
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -28002,18 +28002,23 @@ img[src$=".svg"]:not([alt="Typst"])
 typst.app/universe
 
 INVERT
-#results li .title .name
-#results li .description
-#results li .template-thumbnail
+img[alt*="icon"]
+#results li
+#results .title
+#results .description
+#results .template-thumbnail
+#template-thumbnail
 #banner
 #banner .title
 #banner .description
 #banner .featured-label
+#banner .icon
 
 CSS
 #results li a,
 #banner {
-    filter: invert(95%) hue-rotate(180deg) contrast(90%) !important;
+    border: 1px solid !important;
+    border-color: lightgray !important;
 }
 
 ================================
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -28012,7 +28012,7 @@ img[alt*="icon"]
 #banner .title
 #banner .description
 #banner .featured-label
-#banner .icon
+#banner .featured-label .icon
 
 CSS
 #results li a,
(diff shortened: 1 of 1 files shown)


Myshor (changes requested):
It works fine with dark scheme but fixes should work fine with light scheme too and search ends with inverted pictures:
obraz

I propose:
```
typst.app/universe

INVERT
img[alt*="icon"]
#results li
#results img
#results .title
#results .description
#banner
#banner .title
#banner .description
#banner .featured-label
#banner .icon

CSS
#banner {
    border: 1px solid !important;
    border-color: lightgray !important;
}
#results li {
    border: 1px solid !important;
    border-color: lightgray !important;
    border-radius: 12px !important;
}

```

It ends with almost the same background as page background for "a2c-nums" and others using the same color but at least doesn't break loading page in light scheme and dark scheme.

In CSS I would add borders that are almost invisible but at least goes around that very dark ones.

Result can be seen below:
obraz
obraz

P.S.
`img[alt*="icon"]` inverts all black icons I saw on the mentioned pages.

Andrew15-5:
> fixes should work fine with light scheme

There is a light scheme? D: Does it make all websites bright? I did a lot of patches before but never tested that, I didn't even know about this until just now.

> `img[alt*="icon"]` inverts all black icons I saw on the mentioned pages.

Yeah, that's a good catch, didn't notice that.

> It ends with almost the same background as page background for "a2c-nums" and others using the same color but at least doesn't break loading page in light scheme and dark scheme.

I specifically used the `invert(95%)` because uncolored packages are just plain black and one with the background, even though for default theme this is kinda also true. I want them to have a slightly different background so that they are easily distinguished from background/each other. With your proposal, they again go black.

> In CSS I would add borders that are almost invisible but at least goes around that very dark ones.

As a substitution for distinguished background? I didn't think of this, but I guess I can do that, if background change is out of the question. The thin border does add a little bit of accumulative vertical height.

Andrew15-5:
image
image
image

Andrew15-5:
> Just be careful... I made some edits with more fixes to my proposal.

Wait, there is more??? /hj

OK, let's read that comment again.

Andrew15-5:
I'm not sure about `#categories ul li a`:

image

Andrew15-5:
I noticed that after switching to light and then back to dark scheme, the styling breaks and the page needs a reload:

image

Maybe a bug or limitation.

Myshor:
> I'm not sure about `#categories ul li a`:
> 
> image

For that I am not sure what browser you use but in my Firefox they look fine like screen below:
obraz

That's why I did not touch them.

Even with dark text they looks fine to me.



> I noticed that after switching to light and then back to dark scheme, the styling breaks and the page needs a reload:
> 
> image
> 
> Maybe a bug or limitation.

I noticed that too... May be some issue when we invert too much 😂 but most of the people if use light/dark scheme configured for dark reader they do not switch them that often as people doing and checking fixes. (I hope) So minor bug.

The light/dark scheme is mostlu used for this option:
obraz
obraz

As it sometime can change dark sites into light ones user can set to change scheme of Dark Reader automaticaly depending on system theme.

Not many people use it but there was some bugs about broken light theme in the past that's why I am trying to check if everything works fine with my PRs and with PRs I review. 😉

Myshor:
OK... Finish this PR calmly.

I need to switch off. So I will come back tomorrow to that when I will be back to my PC. 😉

Andrew15-5:
> That's why I did not touch them.
> 
> Even with dark text they looks fine to me.

Someone linked https://webaim.org/resources/contrastchecker/ from where it is said that black text would have more contrast than the white, so I think I will make it black after all.

> OK... Finish this PR calmly.

This will be the calmest PR in human history!

Andrew15-5:
Since I already tested it a few times, I think it should be ready.

Myshor (approved):

Myshor:
Done... Thanks for your cooperation and for showing me that I didn't catch I could border something deeper to not use border-radius manually. 😁

Andrew15-5:
> for showing me that I didn't catch I could border something deeper to not use border-radius manually.

Yeah, I just thought that it looks weird that an explicit border radius is set when there is already a hover outline that has this style set. That's how I found that the additional line can be removed and CSS blocks merged. I also had `#results li a` at first, but you proposed `#results li`, so this also was a small pointer.

PR #13288, opened 2024-09-25
Wikipedia: add fix for Creative Assembçy

Tiagoquix:
Fixes black logo in infobox https://en.wikipedia.org/wiki/Creative_Assembly


Myshor:
Not sure... it's only on this page... I will add forcing transparent background to this image in a sec.

PR #13289, opened 2024-09-26
add vudu to dark-sites

gordcodes:


Myshor (changes requested):
For https://www.vudu.com/region-error I end with this region error so I can not check.

For X I can confirm even by this video: https://www.youtube.com/watch?v=kEx5WbFwHqY that it's not always dark and light mode can be selected and used by default!

So X must be removed for sure.

gordcodes:
I removed X, I had no idea there was a darkness setting that persisted across devices

Myshor (approved):

Myshor:
OK... Installed some VPN for a while and didn't find any switching to light mode on vudu so I will merge. 😉

PR #13290, opened 2024-09-26
Remove ns.nl

Joristdh:
The input box is already a correct darkened version of the original yellow. There's no need to force it to be the same color as the background.


Myshor:
I think someone wanted to have the page fully dark... As it was pushed here #5570 
Ok. We will revert. 🙂

Issue #10441, opened 2022-12-05
[Feature Request] Update Dark Reader style based on Firefox theme


MorsMortium:
### Prerequisites

- [X] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this feature request to avoid opening a duplicate.

### Is this feature request related to a problem?

No

### Feature Request Description

Firefox has an api [theme.getCurrent](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/theme/getCurrent). This allows for addons to get the currently used browser themes colors, or an empty result, when no theme is set. It would be nice, if this could be used, and the addons own popups would use these colors. This could be optional too.

MorsMortium:
I actually wrote a much simpler extension, that uses this API, and found every color (background, text, hover), with trial and error. I also found out what Firefox uses for hover color, when the theme does not specify it.
It is here: https://codeberg.org/MorsMortium/Synclair/src/branch/main/Script/Theme.js

Or here:
```

async function GetTheme() {
    SetStyle(await browser.theme.getCurrent());
}

function SetStyle(Theme) {
    if (Theme.colors) {
        var Parser = new DOMParser();

        PopupText = Theme.colors.popup_text;
        Color = "rgba" + PopupText.slice(3, PopupText.length - 1) + ", .17)"

        if (Theme.colors.popup_highlight) {
            Color = Theme.colors.popup_highlight
        }

        var Style = Parser.parseFromString(`
        <style>
        * {
            color: ${Theme.colors.popup_text};
            background-color: ${Theme.colors.popup};
        }
        button:hover, input:hover {
            background-color: ${Color};
        }
        </style>`, "text/html");
        document.head.appendChild(Style.head);
    }
}

GetTheme();

```

One thing with this approach is that even Firefox doesn't always uses the hover color, it specifies in the theme. It has a different hover color for buttons and for menu items in a popup, spawned from these buttons. It might use the fallback color then, but i haven't investigated that yet

bershanskiy:
@MorsMortium I like the idea of using Firefox (or any other browser's) APIs for style data to sync browser theme and Dark Reader settings. However, I'm still not certain how to do this via the Theme API: I simply don't understand which values we should pick. For example, [this theme](https://addons.mozilla.org/uk/firefox/addon/nicothin-space) populates  almost all values:
[...]

PR #13309, opened 2024-09-29
Option to use Firefox theme for Dark Reader GUI

MorsMortium:
Fixes https://github.com/darkreader/darkreader/issues/10441

image
image
image
image
Disabled:
image

### DIFF
--- a/src/defaults.ts
+++ b/src/defaults.ts
@@ -101,4 +101,5 @@ export const DEFAULT_SETTINGS: UserSettings = {
     enableForProtectedPages: false,
     enableContextMenus: false,
     detectDarkTheme: true,
+    themeWithFirefox: false,
 };
--- a/src/ui/utils.ts
+++ b/src/ui/utils.ts
@@ -1,6 +1,130 @@
-import {isMobile} from '../utils/platform';
+import {isMobile, isFirefox} from '../utils/platform';
+import UserStorage from '../background/user-storage';
 
 declare const __CHROMIUM_MV3__: boolean;
+
+declare const browser: {
+    theme: {
+        getCurrent: (() => Promise<void>);
+    };
+};
+
+export async function getTheme() {
+    // Check if theme setting should be used
+    await UserStorage.loadSettings();
+    if (UserStorage.settings.themeWithFirefox && isFirefox && typeof browser !== 'undefined' && browser.theme && browser.theme.getCurrent) {
+        setStyle(await browser.theme.getCurrent());
+    }
+}
+
+function parseColor(color: any) {
+    // Create a div, set a color, get it back in unified rgb format
+    var div = document.createElement('div'), m;
+    document.body.appendChild(div);
+    div.style.color = color;
+    m = getComputedStyle(div).color.match(/^rgb\s*\(\s*(\d+)\s*,\s*(\d+)\s*,\s*(\d+)\s*\)$/i);
+    div.remove();
+    if(m) return [Number(m[1]),Number(m[2]),Number(m[3])];
+    return [0,0,0];
+}
+
+function setStyle(Theme: any) {
+    if (Theme.colors) {
+        // Get color or use fallback white
+        var Color = Theme.colors.popup_text;
+        if (Color.toString() == "hsl(0, 0%, 100%)") {
+            Color = "rgb(255, 255, 255)";
+        }
+
+        // Get 2 different strength highlights
+        const BackgroundColor = "rgba" + Color.slice(3, Color.length - 1) + ", .17)"
+        const ActiveBackgroundColor = "rgba" + Color.slice(3, Color.length - 1) + ", .34)"
+
+        // Check whether theme is light or dark
+        var CRGB = parseColor(Color);
+        var BRGB = parseColor(Theme.colors.popup);
+        var Light = 1;
+        if(BRGB[0] + BRGB[1] + BRGB[2] <= CRGB[0] + CRGB[1] + CRGB[2]) {
+            Light = 0;
+        }
+
+        // Create and apply css
+        var Parser = new DOMParser();
+        var Style = Parser.parseFromString(`
+        <style>
+        * {
+            color: ${Color} !important;
+            border-color: ${BackgroundColor} !important;
+            scrollbar-color: ${Color} ${BackgroundColor} !important;
+        }
+
+        .color-picker--focused .color-picker__wrapper {
+            border: unset !important;
+            outline: 2px solid ${BackgroundColor};
+        }
+
+        .check-button .checkbox__input ~ .checkbox__checkmark {
+            background-color: ${ActiveBackgroundColor} !important;
+        }
+
+        .preview .settings-button-icon, .preview .m-help-button__text::before, .reset-button__icon, .dropdown__selected::after, .preview .theme-group__more-button::after, .preview .collapsible-panel__group__expand-button::before, .preview .page__back-button::before, .nav-button::after, .preview .m-help-button::after, .check-button .checkbox__input ~ .checkbox__checkmark {
+            filter: invert(${Light});
+        }
+
+        .color-picker__input, .color-picker__wrapper {
+            background-color: unset !important;
+        }
+
+        .site-list__item__remove:hover {
+            color: ${Theme.colors.popup} !important;
+        }
+
+        .preview .theme-group__presets-wrapper::before {
+            border-color: ${BackgroundColor} !important;
+        }
+
+        .loader__message {
+            display: none;
+        }
+
+        .textbox::placeholder {
+            color: ${Color};
+        }
+
+        .header__more-settings__shortcut-wrapper--set svg {
+            --icon-color: ${Color};
+        }
+
+        .tab-panel__buttons::before, .tab-panel__buttons::after {
+            border-bottom-color: ${BackgroundColor};
+        }
+
+        .footer-help-link:hover, .donate-link:hover, .site-list__textbox:focus, .automation__behavior .dropdown, .preview .theme-preset-picker .dropdown__list {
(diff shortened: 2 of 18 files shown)


Myshor:
I won't touch it as it changes too many things in the code but...

1. Where will be the option to select theming with Firefox available because I do not see anything in the screens to show that?
2. You shared example screens with old design of settings popup. What about new design which can be enabled in Dark Reader's Developer Tools?
3. Last but not least... Have you tried to compile Dark Reader with this change and check how it will work on Chrome based browsers? Will there the option visible too or not? If yes will it break something if enabled or not? 😝

And after looking into the feature request. I (and I bet Anton) thought we want to get colors from Firefox theme and apply them into Dark Reader color schemes:
obraz

But here you only change the colors of Dark Reader popup.

Is it really worth that changes in the code?

P.S. Assigning @alexanderby to review it.

MorsMortium:
image
image
image
image
image
image

There should be no option, you can see that here, it only appears on firefox: https://github.com/darkreader/darkreader/blob/3f4cfccaf502636471d9f07b589aa16fb9705ca9/src/ui/options/general/general-tab.tsx
But even if enabled, there is a runtime test for the browser being firefox, and that the functions are present, here: https://github.com/darkreader/darkreader/blob/fa7784cba7ffcbcb42728e6a3de1c05ae706e888/src/ui/utils.ts in getTheme
From a test all I can see that the setting is not there and as both the default is off and the browser is not firefox, I can't trigger it to run.

Not sure what you wanted back then, i opened the feature request two years ago for what I implemented. But, there could be another separate one, for checking if the theme is available, and putting it in that list, after some conversion. And yes, it's worth it, like themed apps are worth it.

Myshor:
> Not sure what you wanted back then, i opened the feature request two years ago for what I implemented. But, there could be another separate one, for checking if the theme is available, and putting it in that list, after some conversion. And yes, it's worth it, like themed apps are worth it.

Just when I went through that Feature Request my thoughts what was requested was totally different that's why I mentioned my thoughts.

Thanks for the rest of the screens and more info how it will work.

Still I am only cleaner here to check and make CSS fixes and other not much complicated things in Config.
That's why we will wait until @alexanderby will look through the code. 😉

src/ui/utils.ts
```diff
@@ -1,7 +1,134 @@
...
+    m = getComputedStyle(div).color.match(/^rgb\s*\(\s*(\d+)\s*,\s*(\d+)\s*,\s*(\d+)\s*\)$/i);
+    div.remove();
+    if(m) return [Number(m[1]),Number(m[2]),Number(m[3])];
+    return [0,0,0];
+}
+
+function setStyle(Theme: any) {
+    if (Theme.colors) {
+        // Get color or use fallback white
+        var Color = Theme.colors.popup_text;
```
alexanderby, line 34: Using `const` or `let` is preferred over `var`. In JavaScript naming convention Capitalized variable names are used for classes and Enums.

src/ui/utils.ts
```diff
@@ -1,7 +1,134 @@
...
+        if (Color.toString() == "hsl(0, 0%, 100%)") {
```
alexanderby, line 35: We use single quotes `'`.

src/ui/utils.ts
```diff
@@ -1,7 +1,134 @@
...
+            Color = "rgb(255, 255, 255)";
+        }
+        if (Color.toString() == "hsl(0, 0%, 0%)") {
+            Color = "rgb(0, 0, 0)";
+        }
+
+        // Get 2 different strength highlights
+        const BackgroundColor = "rgba" + Color.slice(3, Color.length - 1) + ", .17)"
```
alexanderby, line 43: Using template strings (`\`\``) is preferred.

src/ui/utils.ts
```diff
@@ -1,7 +1,134 @@
...
+        const ActiveBackgroundColor = "rgba" + Color.slice(3, Color.length - 1) + ", .34)"
+
+        // Check whether theme is light or dark
+        var CRGB = parseColor(Color);
```
alexanderby, line 47: We have `getSRGBLightness()` function for determining color lightness.

src/ui/utils.ts
```diff
@@ -1,7 +1,134 @@
...
+        var BRGB = parseColor(Theme.colors.popup);
+        var Light = 1;
+        if(BRGB[0] + BRGB[1] + BRGB[2] <= CRGB[0] + CRGB[1] + CRGB[2]) {
+            Light = 0;
+        }
+
+        // Create and apply css
+        var Parser = new DOMParser();
+        var Style = Parser.parseFromString(`
```
alexanderby, line 56: Parsing HTML strings is open to XSS attacks or other unwanted content. There's `new CSSStyleSheet()` and `adoptedStyleSheets` for creating CSS in JS, if for some reason CSS files cannot be used. Or `document.createElement('style')` and `element.textContent = '...'` could be used as an alternative.

src/ui/utils.ts
```diff
@@ -1,7 +1,134 @@
...
+        <style>
+        * {
+            color: ${Color} !important;
```
alexanderby, line 59: `*` and `!important` are discouraged for theming.

src/ui/utils.ts
```diff
@@ -1,7 +1,134 @@
...
+
+        .checkbox__checkmark::before, .checkbox__checkmark::after, .updown__icon::before, .updown__icon::after, .checkbox__input:checked + .checkbox__checkmark::before, .checkbox__input:checked + .checkbox__checkmark::after, .select__expand__icon::before, .select__expand__icon::after, .check-button .checkbox__checkmark, .slider__thumb, .site-list__item__remove:hover {
+            background-color: ${Color};
+        }
+
+        .header__more-settings::after, .news::before, .news__header {
+            background-image: unset;
+        }
+
+        html, button, .tab-panel__tab, .toggle__btn, .updown__icon, .track--clickable, .toggle::before, .header__more-settings, .multi-switch, .checkbox, .header__more-settings__shortcut, .news__list, .select__option, .select__list, .news__header, .check-button .checkbox__checkmark, .dropdown__list__item, .message-box, .dropdown__list, .editor, .preview .page, .preview .news-section, .preview .news-section__popover, .color-picker__hsb-line, .site-list__textbox, .time-range-picker__input, .automation__location-control__latitude, .automation__location-control__longitude, .text-list__textbox, .select__textbox, .dynamic-per-site__search-input {
```
alexanderby, line 117: Listing selectors within the app is absolutely not scaleable and unmaintainable. CSS variables (custom properties) should be used and redeclared like `:root { --color-bg: #000; ... }`.

alexanderby (changes requested):
Dear @MorsMortium,

Thank you for your interest in Dark Reader. Please understand Dark Reader is used by millions. What solves your problem can make the app harder to use others and can make the code harder to maintain.

I cannot remember any other request to change the colors of our user interface. The result doesn't look visually appealing to me. The app's design will change completely soon.

However I appreciate your effort and reviewed your Pull Request. I hope it will help you learn more about JavaScript best practices, naming conventions and available browser APIs.

Regards,
Alexander.

src/defaults.ts
```diff
@@ -101,4 +101,5 @@ export const DEFAULT_SETTINGS: UserSettings = {
     enableForProtectedPages: false,
     enableContextMenus: false,
     detectDarkTheme: true,
+    themeWithFirefox: false,
```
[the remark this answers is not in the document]
MorsMortium: What would be correct for it?

src/ui/utils.ts
MorsMortium: Do you mean the entire css of the entire extension should be rewritten for this?

src/defaults.ts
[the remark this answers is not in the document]
alexanderby: Something that when you read it, you understand what it does. But as short as possible.

src/ui/utils.ts
alexanderby: We have `theme.less` file with LESS variables. In theory you should be able to have something like:
```less
:root {
    --color-back: #141e24;
}

@color-back: var(--color-back);
```

And then you can insert CSS like
```less
:root {
    --color-back: #000000;
}
```
and the colors within the entire UI should change. However there could be issues with the colors of icons (they should be moved from LESS to HTML/JSX markup (`currentcolor` can be used for SVG to make it use CSS `color` value). Another issue could be LESS color mixing functions (like `fade` or `mix`), they won't process the CSS variables.

Please understand that even if you do all those changes, the PR most likely won't be merged due to reasons I described earlier. However it could be useful to replace LESS color variables with CSS variables in our code. If you would like to have some practice, you can submit a separate PR addressing these color variables. Then another for moving icons from CSS to SVG (we have  `src/ui/icons` folder, see how they are used in the code).

MorsMortium: @alexanderby I did clean it up as much as I could (while still using css selectors/classes). I'd like an answer to my question. If it wouldn't be merged even with using css variables, I'll just close the pr/issue and keep it for my own usage

src/ui/utils.ts
```diff
@@ -1,7 +1,161 @@
...
+        const style = document.createElement('style');
+        style.textContent = `
+        * {
+            color: ${text} !important;
+            border-color: ${highlightWeak} !important;
+            scrollbar-color: ${text} ${highlightWeak} !important;
+        }
+
+        /* Icon inversions */
+        .settings-button-icon, .m-help-button__text::before, .nav-button::after,
```
alexanderby, line 68: Sorry, but I cannot merge this implementation. I have some ideas about changing the UI theme dynamically, and when I get to refactoring the use of color variables it will be easier for you to implement your option.

PR #13312, opened 2024-09-30
fix(text-style): use case-insensitive CSS selectors

oberrich:
See also
- https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors
- https://caniuse.com/css-case-insensitive

### DIFF
--- a/src/generators/text-style.ts
+++ b/src/generators/text-style.ts
@@ -12,7 +12,7 @@ const excludedSelectors = [
     // Generic matches for icon/symbol fonts
     '.icofont',
     '[style*="font-"]',
-    '[class*="icon"i]', 
+    '[class*="icon"i]',
     '[class*="symbol"i]',
 
     // Glyph Icons
--- a/src/generators/text-style.ts
+++ b/src/generators/text-style.ts
@@ -10,9 +10,10 @@ const excludedSelectors = [
     '.fa', '.fab', '.fad', '.fal', '.far', '.fas', '.fass', '.fasr', '.fat',
 
     // Generic matches for icon/symbol fonts
-    '.icofont', '[style*="font-"]',
-    '[class*="icon"]', '[class*="Icon"]',
-    '[class*="symbol"]', '[class*="Symbol"]',
+    '.icofont',
+    '[style*="font-"]',
+    '[class*="icon"i]', 
+    '[class*="symbol"i]',
 
     // Glyph Icons
     '.glyphicon',


Gitoffthelawn:
Has anyone performance tested this change?

I think there are only 3 cases that likely need to be covered: all lowercase, initial cap, all uppercase. Is it faster/slower to check for those 3 cases or to check for every possible case as this change implements?

oberrich:
> Has anyone performance tested this change?
> 
> I think there are only 3 cases that likely need to be covered: all lowercase, initial cap, all uppercase. Is it faster/slower to check for those 3 cases or to check for every possible case as this change implements?

I haven't done any benchmarks on this but I'd imagine a single case-insensitive selector being more performant than three case-sensitive ones, especially since we're "fuzzy matching".

Considering the sheer amount of CSS rules being injected, this is just a drop in the bucket.

Gitoffthelawn:
> > Has anyone performance tested this change?
> > I think there are only 3 cases that likely need to be covered: all lowercase, initial cap, all uppercase. Is it faster/slower to check for those 3 cases or to check for every possible case as this change implements?
> 
> I haven't done any benchmarks on this but I'd imagine a single case-insensitive selector being more performant than three case-sensitive ones, especially since we're "fuzzy matching".

Best to benchmark instead of imagine!  Especially since the code is "fuzzy matching", the difference will be magnified.

> Considering the sheer amount of CSS rules being injected, this is just a drop in the bucket; let's not optimize prematurely.

All the oceans of our world are filled with nothing but drops.  Yes, it's a real concern that too many inefficient generic CSS rulesets are now being injected into every page.  Whenever possible, host-specific CSS rulesets need to be used.  You indicated you feel it is premature to optimize.  When do perceive it is time to optimize?  Is that scheduled?

Gitoffthelawn:
> Feel free to benchmark this change and come back with the numbers but as far as I’m concerned this is not worth discussing further.

You made the change to the existing code.  Why?  What bug were you trying to fix?  There are no links to any issue reports attached to your code change.

Why didn't you benchmark your code to see if it performed better or worse than the existing code before submitting it to be merged?

It appears that you are brand new to this project and your questionable coding practice based on non-scientific *"I imagine"* combined with your *"I'm not interesting in discussing further"* attitude isn't helpful.

oberrich:
> You made the change to the existing code. Why? What bug were you trying to fix? There are no links to any issue reports attached to your code change.

This PR reduces the overall amount of CSS selectors and is semantically in line with what we actually want to do here.

> Why didn't you benchmark your code to see if it performed better or worse than the existing code before submitting it to be merged?

There are close to 8000 PRs, 12 of which mention the [keyword "benchmark"](https://github.com/darkreader/darkreader/pulls?q=benchmark). I don't think it's fair to assume I have to provide a nanosecond-scale benchmark because I'm changing CSS selectors in a code path only a low percentage of users are even going to hit in the first place.

> It appears that you are brand new to this project and your questionable coding practice based on a non-scientific _"I imagine"_ hypothesis combined with your _"I'm not interesting in discussing further"_ attitude isn't helpful.

You haven't brought up any argument as to why you would assume it's going to be slower in the first place, other than me not having submitted a benchmark, which, again, has never been expected before.

I think it's sound to assume three fuzzy matches are more expensive than a single fuzzy match, so this PR is most likely a (negligible) performance win.

Calling in question my "non-scientific" and "questionable coding practices" all because I'm making an *informed* guess about the runtime performance of my code is uncalled for.

> @Gitoffthelawn I can revert this change if you think it would be better. If yes I'll come back here to give feedback it's done and give @oberrich possibility to make PR again.

I don't think me, or anyone else for that matter, should be beholden to a [different standard](https://github.com/darkreader/darkreader/pulls?q=benchmark) than any other contributor.

Myshor:
As I said... For me it is good PR that's why I merged it.

Just I don't want to have quarrel here.

As I remember this part is responsible to exclude icons from font change which can be applied from Dark Reader options so sometimes "slower" (still not confirmed if it is😝) but better coverage should be better IMHO.

Sorry @oberrich - I do not have any problems with your code. I just not merge anything outside sitefixes/darksites normally and I feel a little gulty that I did and there is someone who doesn't like it. 😉

oberrich:
> Just I don't want to have quarrel here.
> Sorry @oberrich - I do not have any problems with your code. I just not merge anything outside sitefixes/darksites normally and I feel a little guilty that I did and there is someone who doesn't like it. 😉

I'm reverting this patch and will do another PR for Alex.

oberrich:
> Revert merged. @oberrich you can make new PR and assign Alex to review.
> 
> And one more time sorry for the problem which appeared here.

I've resubmitted as #13351 however I cannot assign a reviewer myself.

Myshor:
I assigned Alex there. 😉

Gitoffthelawn:
@Myshor No problem.  I was surprised you merged it, but I had a hunch you thought it was a PR that applied to something else. ;)  Don't worry about a quarrel, I won't be responding to their comments.

oberrich:
> @Myshor No problem. I was surprised you merged it, but I had a hunch you thought it was a PR that applied to something else. ;) Don't worry about a quarrel, I won't be responding to their comments.

You’ve already created a quarrel, they don’t want to hurt ur feelings in case you still haven’t managed to read the room princess.

He merged it because it’s a good change.

PR #13315, opened 2024-09-30
Fix for newreleases.io

rakleed:


src/config/dynamic-theme-fixes.config
```diff
@@ -19486,6 +19486,15 @@ CSS
 
 ================================
 
+newreleases.io
+
+CSS
+.bg-white > .bg-cover {
+    background-image: none !important;
```
Myshor, line 19493: maybe instead of removing whole background image we will mix it with background that will be dark

Check how it can look with this part of code:
```
    background-blend-mode: multiply !important;
    background-color: ${rgba(255,255,255,0.95)} !important;
```

Myshor (changes requested):
Take a look on my comment and make a state if you still want to remove background or do what I proposed. 😉

Myshor (approved):

PR #13316, opened 2024-10-01
Dynamic theme fixes for Washington Post comments

StevenSchifter:
Fix for comments initially loading with a light theme. Also fixes the comment counter and sort menu.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -29218,6 +29218,17 @@ div[data-qa="desktop-copyright"] {
     border-left-color: transparent !important;
     border-right-color: transparent !important;
 }
+.App-root,
+select {
+    background: var(--darkreader-neutral-background) !important;
+    color: var(--darkreader-neutral-text) !important;
+}
+span[class^="Counter-root"] {
+    background: var(--darkreader-neutral-background);
+}
+i[class^="Icon-"] {
+    color: var(--darkreader-neutral-text);
+}
 
 IGNORE INLINE STYLE
 .center > svg > path[fill="white"]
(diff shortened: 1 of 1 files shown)


Myshor:
I added your fix to my local Dark Reader then opened this page.

https://www.washingtonpost.com/advice/2024/10/02/carolyn-hax-boyfriend-obsessed-ex/

It seems it's still not working good. What I found it makes dark background without this fix when I disable/enable Dark Reader or reload sitefixes. (I pressed Reset button on sitefixes and got dark background for comments).

So If we want to have some permament fix for comments it needs some changes in this PR and checking how they look like after page refresh.

StevenSchifter:
@Myshor You are correct, my fix is not working as expected. What kind of change could I make so the styling would work after a page refresh?

Myshor:
Not sure if it's possible to do in sitefixes. It seems the comments are dynamicly loaded after clicking show button with their own CSS which has some things set to important. The only one possible fix would need some Dark Reader code changes I think.

Summarizing:
Nothing is applied to comments as they are not loaded on the page. After showing them they load with their own CSS.
Making Dark Reader off/on in this moment applies Dark Reader changes to backgrounds which are already loaded on the page.

So to fix this I think there should be some changes in the code that "reaaplies" Dark Reader if something is added dynamically to the page.

For this I can only propose to create a bug for this issue and mention `@alexanderby` in it.

Myshor:
You can mention this one too however comments are different part and are always with white background so it's not the same issue. 😉

StevenSchifter:
Done (#13324)!

Myshor:
Added comment there as my URL doesn't ask for email or login. 😉

BTW I think we can close this PR as we do not have any possible solution in sitefixes and it won't be merged.

StevenSchifter:
OK, I will close this PR.

PR #13341, opened 2024-10-05
Fix for staging.typst.app{,/universe}

Andrew15-5:
During the testing phase (before each new release) a lot of people in the Typst community are using staging.typst.app instead of typst.app to find bugs etc.

I don't know if there is a regex syntax to do `(staging.)?typst.app`, so I created 2 secondary entries.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -28000,7 +28000,43 @@ img[src$=".svg"]:not([alt="Typst"])
 
 ================================
 
+staging.typst.app
+
+INVERT
+img[src$=".svg"]:not([alt="Typst"])
+
+================================
+
 typst.app/universe
+
+INVERT
+img[src$=".svg"]:not([alt="Typst"])
+#results li
+#results .title
+#results .description
+#results .template-thumbnail
+#template-thumbnail
+#banner
+#banner .title
+#banner .description
+#banner .featured-label
+#banner .featured-label .icon
+.carousel
+.carousel .top-text
+.carousel .title
+.carousel .bottom-text
+#categories ul li a
+
+CSS
+#results li a,
+#banner {
+    border: 1px solid !important;
+    border-color: lightgray !important;
+}
+
+================================
+
+staging.typst.app/universe
 
 INVERT
 img[src$=".svg"]:not([alt="Typst"])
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -28000,6 +28000,13 @@ img[src$=".svg"]:not([alt="Typst"])
 
 ================================
 
+staging.typst.app
```
Myshor, line 28003: Adding this would need to have it alphabetically sorted not under typst.app

src/config/dynamic-theme-fixes.config
```diff
@@ -28029,6 +28036,35 @@ CSS
 
 ================================
 
+staging.typst.app/universe
```
Myshor, line 28039: Adding this would need to have it alphabetically sorted not under typst.app/universe

src/config/dynamic-theme-fixes.config
Myshor: However it would be easier to add alias staging.typst.app as a next line under typst.app as it will be the same fixes used and remove this additional copy of fixes.

src/config/dynamic-theme-fixes.config
Myshor: However it would be easier to add alias staging.typst.app/universe as a next line under typst.app/universe as it will be the same fixes used and remove this additional copy of fixes.

src/config/dynamic-theme-fixes.config
Myshor: Example how aliases work:
obraz

Myshor (changes requested):

Andrew15-5:
Not really sure why the CI failed again. :/

Myshor:
It's failing because first URL starts with S (staging) and it's placed in the sitefixes between URLs starting with T.
I will make a qucik edit to maki it good. (I mean all staging URLS will be second into the URLs.

Myshor (approved):

Andrew15-5:
But are the aliases also must be sorted between themselves?

Myshor:
>But are the aliases also must be sorted between themselves?

No... It's nice to have it sorted under the main but it's not needed. Specially it's nicely visible on wikimedia screen - All `*.` subpages are added under the their main URL as Alias but if there would be forcing all aliases sorted then all `*.` should be first and it should be moved to the top of aliases in the file.

Andrew15-5:
Thanks for that.

PR #13348, opened 2024-10-07
fix for seedr.cc

ramavats:
fixed some text were dark in the dark mode, and an icon which was black in the dark mode.


Myshor:
Found everything in fix - merging.

PR #13350, opened 2024-10-07
Revert "fix(text-style): use case-insensitive CSS selectors (#13312)"

oberrich:
This reverts commit a605d61b2a6965f385d351aa65bef42cbdeba5dd.


Myshor:
Accepting as it was already mentioned in #13312

PR #13351, opened 2024-10-07
fix(text-style): use case-insensitive CSS selectors

oberrich:
See also
- https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors
- https://caniuse.com/css-case-insensitive

This PR was previously submitted as #13312 and landed by @Myshor however @Gitoffthelawn has raised concerns about performance so we're asking for review by @alexanderby.

### DIFF
--- a/src/generators/text-style.ts
+++ b/src/generators/text-style.ts
@@ -10,9 +10,10 @@ const excludedSelectors = [
     '.fa', '.fab', '.fad', '.fal', '.far', '.fas', '.fass', '.fasr', '.fat',
 
     // Generic matches for icon/symbol fonts
-    '.icofont', '[style*="font-"]',
-    '[class*="icon"]', '[class*="Icon"]',
-    '[class*="symbol"]', '[class*="Symbol"]',
+    '.icofont',
+    '[style*="font-"]',
+    '[class*="icon"i]',
+    '[class*="symbol"i]',
 
     // Glyph Icons
     '.glyphicon',


alexanderby (commented):
I really don't know the performance implications of case insensitive selectors. It could be optimal in Chrome, but slow in Firefox or Safari. It's definitely slower, but telling how much slower is hard and requires time to build the test cases and measure the performance. These `*=` selectors are already slow, but there is no easier way to fix the issue. So let's stick to the previous case-sensitive version if it already works. It's fine to have 2 extra code lines if it's faster.

oberrich:
This is micro optimization on n=2 CSS selectors, at the cost of semantical correctness and, as @Myshor pointed out earlier, coverage.

I'm not going to try to please everybody, so I'm fine whether or not this gets merged, but do feel obligate to at least reiterate this point.

Gitoffthelawn:
@alexanderby I agree that we should prioritize performance over 2 short lines of code, especially when a PR is presented that is not fixing any documented issue.

oberrich:
Clearly the maintainers feel it’s best to make the Senior Contributor happy and not listen to any technical arguments instead.
Closing this PR to keep my sanity.

alexanderby:
Please guys stop this argument. It's not worth it. The impact of this change is +0.0001% positive or -0.0001% negative. I mean, maybe there is a website that has `<div class="ICON">` or similar, and maybe someone with Dark Reader's custom font will ever meet it. Or maybe there is a website with a lot of elements like `<div class="abdcefgh_ICO">` and this string search will repeat over and over again for people with custom font settings. Once we have an issue reported we can return to this change.

Everyone has good points and important problems were raised:
1. The impact of Dark Reader's global font overriding the font icons and an enormous amount of different class name variations and the necessity to maintain the list of exceptions.
2. The performance of attribute selectors with a matching string.
3. Premature optimisations.
4. Premature problem solving.
5. The impact of arguments on the internet on mental wellbeing.

I hope everyone learned something new from this argument and will consider it when making future changes and comments.

A few useful articles on CSS selector performance and measurement:
https://blogs.windows.com/msedgedev/2023/01/17/the-truth-about-css-selector-performance/
https://developer.chrome.com/docs/devtools/performance/selector-stats

PR #13363, opened 2024-10-11
Fix for Clint Eastwood Wikipedia page

Tiagoquix:
Fixes signature color.
image


Myshor:
I checked how it is used on en version and pl version and I will edit this so all signatures will work fine (again) and not adding single ones to invert. 😉

PR #13369, opened 2024-10-13
Fix the company logo icon for https://www.elastic.co/

papadakis-k:
Fixes the company logo icon for https://www.elastic.co/
The selector used might be catching other SVG graphics also on the same domain with fill="#000" but that should work just fine for us.

### DIFF
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -702,6 +702,13 @@ INVERT
 
 ================================
 
+elastic.co
+
+INVERT
+path[fill="#000"]
+
+================================
+
 electronics-tutorials.ws
 
 REMOVE BG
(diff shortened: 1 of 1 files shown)


Myshor:
Can you give some example where it will apply?

I found one place where Logo is displaying with dark text:
obraz

The main page scrolled down.

But it still uses fill="#fff" there so no fix applied.

papadakis-k:
It's all over the docs https://www.elastic.co/docs (which is I guess the most frequently visited part of the site), but also other links like here https://www.elastic.co/community here https://www.elastic.co/elasticsearch etc. It seems the site is 50% light mode 50% dark mode.
I am not sure how you get that result in the front page, mine appears like this 
image

PR #13386, opened 2024-10-16
Fix for Google Maps

oberrich:
Fixes wheelchair symbol in result list.

_before_
image

_after_
image


Myshor:
What browser do you use?
Do you force font changing? If yes then to what kind exactly.

I am asking because I checked your example and could not reproduce the issue you try to fix:
obraz

oberrich:
Font: Gadugi
Chromium Version 128.0.6613.114
Dark Reader Version 5 Preview (4.9.85)

image

Myshor:
> Dark Reader Version 5 Preview (4.9.85)

Why this old version when newer is available with more fixes in code?

I tried with 4.9.95 and your font and can not reproduce problem you try to fix either:
obraz

Myshor:
Ok now I can confirm it fixes it. 🙂

PR #13401, opened 2024-10-20
Add grammatikfejl.dk to global dark list

jesperbakhandskemager:


Myshor:
Checked with light system and browser theme.
Site is theme based so can not be added to global dark list:
obraz

PR #13402, opened 2024-10-21
Fix docs.deno.com code previews

ChristopherHaws:
Before:
image

After:
image


Myshor (changes requested):
With our without my proposed changes it is ugly in Firefox:
obraz

Maybe use `border-color: ${rgb(194, 196, 201)} !important;`

It will make darker color in darks scheme and lighter in light.

PR #13414, opened 2024-10-23
fix for wiki

Myshor:
fixes img link with ( and )


Myshor:
All `*=` search in CSS costs some time. Making 2 separate runs to check is painful butl when I must use it then I use it - 3 is not acceptable by me at all. 😉

PR #13417, opened 2024-10-23
Adjust Quad9 sites

Tiagoquix:
Remove `*.quad9.net` from the dark sites list because https://status.quad9.net/smap/ is not dark.
Add other sites to the coverage.

See also: https://github.com/darkreader/darkreader/issues/13418.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -10,7 +10,6 @@
 *.newgrounds.com
 *.paydaythegame.com
 *.pleasefix.gg
-*.quad9.net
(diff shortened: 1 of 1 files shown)


Myshor (commented):
Maybe add `quad9.net` without `*.` in the beginning as I see it could be used without Dark Reader too.

Myshor:
Or you don't want to as It will work fine with or without Dark Reader applied then I will merge.

Myshor:
I did not expand the code to check. 😅

PR #13421, opened 2024-10-26
Dynamic theme fixes for AP News

StevenSchifter:
These fixes are for the breaking news banner.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1973,6 +1973,8 @@ INVERT
 .hamburger-box
 .sticky-part > div svg
 .LeftRail > div svg
+.Banner
+.Banner-link
 
 ================================
 


Myshor:
Can yoiu share some URL where I can look on that banner?

StevenSchifter:
Unfortunately, the breaking news banner is only visible when there is breaking news. I apologize about that. I last saw one [here](https://apnews.com/hub/israel-hamas-war) after Israel struck Iranian military sites. It's not there now though. If you want, I could post "before" and "after" screenshots next time such a banner appears.

Myshor:
That could be nice. I just want to be sure we don't invert some photos or similar things that will look strange INVERTed. 😉

Myshor:
Ok I think I will merge. If there will be some problems with that banner we will fix that ad-hoc. 😉

PR #13433, opened 2024-10-28
Fix for wiki fix

Tiagoquix:
(Continuation of https://github.com/darkreader/darkreader/pull/13426)

Fixes the fix of invert for Xbox logo (parentheses don't work properly).

@Myshor You said we should avoid using 3 look-ups ("[src]"), but in this case not using three may result in some other random image being inverted I believe. It's fine to use 3 this time then?

https://en.wikipedia.org/wiki/Xbox
https://commons.wikimedia.org/wiki/File:Xbox_logo_(2019).svg

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -17556,7 +17556,7 @@ img[src*="DragonFly_BSD_Logo.svg"]
 img[src*="Fallout_logo.svg"]
 img[src*="PlayStation_4_logo_and_wordmark.svg"]
 img[src*="PlayStation_logo_and_wordmark.svg"]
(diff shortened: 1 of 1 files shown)


Myshor:
I won't merge. 😝 We have already many selectors here in wikipedia.
Check localy using Dark Reader Developer Tools available in Settings > Advanced if `img[src*=Xbox_logo_%282019%29.svg"]` works. If yes then I would prefer to use this than 3 look-ups.

If you won't have much time I will take a look if my proposal works a little bit later 'cause I am a little bit busy right now. 😉

Myshor:
As you mentioned it I will change it in a sec. 😉

PR #13439, opened 2024-10-29
wprost.pl contrast fonts

FadeMind:
https://www.wprost.pl/


Myshor (changes requested):
First - try to fix this black text... 
(you can try using `:not(selector to exclude)` in already used selectors)

Second (more important)
Add alias `*.wprost.pl` because many added things will be needed for opened articles. For example:
https://biznes.wprost.pl/gospodarka/11846397/laptop-dla-nauczycieli-ministerstwo-cyfryzacji-zwrocilo-sie-do-nauczycieli.html

Just check if we really want to have black text where it was already white.

Myshor:
> @Myshor proszę sprecyzuj, jak mają wyglądać fixy z `:not` dla `wprost.pl` - nie mam aż takich umiejętności w kodowaniu DR. Masz możliwość edytowania PR.

I will add some example in review under the code. And I will try to use english as some other people can look here too. 😋

Myshor:
Check comments in review about how to use `:not`

I still find two places that are needed to be examined if they can be excluded from inverts somehow:
1. INVERTED as part of `strong span:not(ul li.bli-tp-default a strong span)` so one more not to find (or we leave it as it is 😋)
obraz
2. INVERTED as part of `nav ul li a` (or we leave it as it is 😋)
obraz

Myshor:
I could try to make all fixes by myself but today I have some task to do in the evening and looking into Github for short time is possible. Doing fixes must wait for tomorrow. 😉

Myshor (approved):

Myshor:
I made that fixes but next time I will just reject. 😋

PR #13442, opened 2024-10-30
insomniac.com fixes

FadeMind:
https://www.insomniac.com/music/artists/paul-kalkbrenner/

@Myshor  I learned something about `:not` . 
Weird bug: cannot handle invert twitch social icon without specify selector.


Myshor:
There is difference in rendering this page with Dark Reader in Chrome and Firefox I see.

In Chrome Dark Reader makes some blob with already inverted icon. In Firefox it's not touched.

Again I will update your PR to make it work the same in both browsers then I will add comment that will give you more info what I did. 😉

PR #13457, opened 2024-11-03
Invert Audible header navigation logo

Byteflux:
Inverts the header logo (dark grey text over a transparent background) on Audible.com.


Myshor:
The same class is used for logo on main page which is already white.

I will make some fix in PR to exclude it.

PR #13480, opened 2024-11-08
Fix button display on wishlist view on amazon.*

cj-wong:
Before

firefox_Kjot48P5QM

After

firefox_XkIKQooJYB

Note: I'm not quite knowledgeable enough to fix the odd color for the share button.


Myshor:
I found solution for this share button. :)

Will push to your PR and merge.

PR #13489, opened 2024-11-09
feat(darkreader#13250): Allow matchMedia function to be undefined

jugglingcats:
Enables Jest tests to run without errors when using the darkreader API.


jugglingcats:
Thank you 🎉

PR #13495, opened 2024-11-11
Update dark-sites.config to reflect domain name change (slither.io to slither.com)

Gitoffthelawn:


Myshor:
maybe we will need to have them both... when I try to open .com it redirects me to .io

so I can not merge removing (changing from) .io to .com

Myshor:
If it always ends with 

> > maybe we will need to have them both... when I try to open .com it redirects me to .io
> > so I can not merge removing (changing from) .io to .com
> 
> Are you sure you're seeing that right? When I go to the .io it redirects to http://slither.com/io and when I go to the .com it also goes to http://slither.com/io

Ok... When I open  the .com version it goes to http://slither.com/io but .io still goes to .io
obraz

For today I still recommend to have both entries.

Gitoffthelawn:
Yes, if it's still going to .io for you, let's do that.  I wonder if it's browser-dependent or dependent on which server of theirs you connect to.  Either way, I agree, let's do both.

Gitoffthelawn (commented):
Will add the original .io too

Gitoffthelawn (commented):
Done!

PR #13501, opened 2024-11-13
fix wikipedia vector theme icon inversion

GalaxySH:
For DR dynamic theme this change fixes inversion for various icons in the 2022 Wikipedia vector theme. These changes have been tested in Light/Dark mode in Firefox and Chrome.

One line adds an inversion fix for a third-party plugin icon.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -17662,9 +17662,14 @@ img[src*="uc_lc"][src*=".svg"]
 img[src*="Greek_letter"][src*="sans.svg"]
 img[src*="Psi2.svg"]
 li.menu-tooltip:not(.lang_btn)
-td.icon
-#vector-main-menu-dropdown
+#vector-main-menu-dropdown .vector-icon
 #vector-page-tools-dropdown-label::after
+.vector-user-links-main .vector-icon
+#vector-user-links-dropdown .vector-icon
+#vector-user-links-dropdown .vector-icon:after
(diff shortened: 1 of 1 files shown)


Myshor:
Can you share some example URLs to compare how it looks with and without this fix?

Myshor:
> I believe this PR added something that isn't present without DR:
> 
> image
> 
> This border highlight doesn't exist with DR off. https://en.wikipedia.org/wiki/Systemd

This border shows up only when hover to me and it's old feature fix to add white background to pictures that have transparent background. I can not see that "border" without any hover.

Myshor:
> I believe this PR added something that isn't present without DR:
> 
> image
> 
> This border highlight doesn't exist with DR off. https://en.wikipedia.org/wiki/Systemd

@Tiagoquix This border shows up only when hover to me and it's old feature fix to add white background to pictures that have transparent background. I can not see that "border" without any hover.

And it's this part of sitefixes:
```
.mw-mmv-image .svg,
.fullImageLink [src*=".svg"],
a[href$=".svg"]:hover > img,
a[href*=".gif"]:hover > img {
    background-blend-mode: color;
    background-color: rgba(255, 255, 255, 0.75) !important;
}
```

Myshor:
It was added before Dark Reader went to mobile version of Firefox. So if there was some picture that was black and was not readable then until we think if we want to add it to invert then you see it nice with white background when you hover the mouse. 😉

PR #13507, opened 2024-11-14
Invert Stockbit's site logos and stock charts

partingscientist:
This pull request inverts company logos and stock charts (requires an account) shown on Stockbit's site.
This pull request readdresses the work previously underlined in https://github.com/darkreader/darkreader/pull/11245 and prior related pull requests.


Myshor:
As you only added/removed some inverts I won't ask for screenshots with examples. (I hope you use the site so you won't break anything else there.) 😉

Issue #888, opened 2018-12-06
Extension suddenly ignoring whitelist and causing lag(?)


SharkWipf:
- Using Chrome 70.0.3538.110 on Gentoo Linux

(Presumably) the latest update suddenly causes the plugin to ignore the whitelist, running dark themes on every site even after a clean reinstall.  
Not just that, it also seems to suddenly cause massive slowdowns in website load times, even crashing some tabs.  
Had to disable the extension to keep it workable.

SharkWipf:
Oops, nvm, duplicate of already closed #885

PR #13509, opened 2024-11-14

erkingoksuu:
-i dont know why but although youtube default logo red side hex color #ff0033, when dark reader in dark mode it converts into #ff1e68 ,and when dark reader in light mode it converts into #ff5880. I fixed it as default value. SS are below.

Before:
logo

After:
logoFixed

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -33101,6 +33101,9 @@ tp-yt-paper-menu-button {
 #paid-comment-background.ytd-comment-view-model {
     background-color: transparent !important;
 }
+svg[id^="yt-ringo2"] :not([id^="youtube-paths"]) path:nth-child(1) {
+    fill: #ff0033 !important;
+}
 
 IGNORE INLINE STYLE
 yt-live-chat-ticker-paid-message-item-renderer *


erkingoksuu:
-additionally, fixed description section border top size and color for dark reader light and dark mode. SS are below.

Before:
description

After:
descriptionFixed

src/config/dynamic-theme-fixes.config
```diff
@@ -33101,6 +33101,16 @@ tp-yt-paper-menu-button {
 #paid-comment-background.ytd-comment-view-model {
     background-color: transparent !important;
 }
+svg[id^="yt-ringo2"] :not([id^="youtube-paths"]) path:nth-child(1) {
```
Myshor, line 33104: Instead of forcing color in CSS move this selector to IGNORE INLINE STYLE then it will be using original color from youtube. 😉

src/config/dynamic-theme-fixes.config
```diff
@@ -33101,6 +33101,16 @@ tp-yt-paper-menu-button {
...
+    fill: #ff0033 !important;
+}
+ytd-horizontal-card-list-renderer.ytd-structured-description-content-renderer:not(:first-child),
+ytd-metadata-row-container-renderer.ytd-structured-description-content-renderer:not(:first-child),
+ytd-structured-description-content-renderer[inline-structured-description] ytd-horizontal-card-list-renderer.ytd-structured-description-content-renderer,
+ytd-structured-description-content-renderer[inline-structured-description] ytd-error-corrections-section-renderer.ytd-structured-description-content-renderer,
+ytd-structured-description-content-renderer[inline-structured-description] ytd-video-description-infocards-section-renderer.ytd-structured-description-content-renderer {
+    border-top: 2px solid #222 !important;
```
Myshor, line 33112: Have you checked how this would look like using light scheme in Dark Reader?
Maybe we should use `${#DDD}` instead to have it light in light scheme and dark in dark?

Myshor (changes requested):

erkingoksuu:
ı try to put "svg[id^="yt-ringo2"] :not([id^="youtube-paths"]) path:nth-child(1)" line to IGNORE section but does not work then try with "svg[id^="yt-ringo2"] *" does not work also then try with "svg *" does not work also lastly try with "svg" again does not work just put it same.

additionally change border-top color with #888.

you can see that border like any video. example video "https://www.youtube.com/watch?v=v0dRWqk7qzE" just expand the description section

Myshor (approved):

Myshor:
OK... Let's merge it.

PR #13527, opened 2024-11-18
Update dynamic-theme-fixes.config for medium.com fix

TobyDS:
This fixes the medium.com logo in the nav bar being unreadable by switching it to a contrasting color.

`a[href="/"]>svg` - Fixes the logo on the homepage
`a[data-testid="headerMediumLogo"]` - Fixes the logo on all other pages


Myshor:
Many "articles" opens with totally different site in URL and it's not that easy to include it there - the only possibility is to move `a[data-testid="headerMediumLogo"]` selector to `*` section.

I will make this change into the PR and merge. 😉

TobyDS:
That's a great point, thanks for correcting that oversight on my part

PR #13528, opened 2024-11-18
Fix for wiki

Tiagoquix:
Fixes logo in infobox https://en.wikipedia.org/wiki/Persona_(series)

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -17715,6 +17715,7 @@ img[src*="Systemd-logo.svg"]
 img[src*="PulseAudio-logo.png"]
 img[src*="Pipewire_logo.svg"]
 img[src*="Metal_Gear_franchise_logo.svg"]
+img[src*="Persona_PSP_logo.svg"]
 
 CSS
 :root {


GalaxySH:
Any idea on how much of a performance cost all these custom image selectors have? What about after a hundred more have been added? There's a lot of infoboxes on Wikipedia.

GalaxySH:
What about adding a blanket white background:

```css
.infobox-image img[src*=".svg"] {
    background-color: white;
}
```

I know this undermines the dark mode, but SVG logos are guaranteed to work on a white background, and this would only affect infobox logos. This seems better to me than gradually adding "fixes" for 100,000 logos. I just ran into about 20 more broken logos when testing this solution that would need individual patches otherwise.

It also seems better than inverting all SVG logos, which would break the coloring in many cases, but that's a matter of opinion.

Myshor:
I made some feature for wiki to all transparent pictures - on hover I add white background. It was enough until Dark Rider went to mobile and on mobile you can't do hover. 😕

But thanks @GalaxySh to your comment I found that native Dark Mode of Wiki sometimes inverts logos (like this in Persona) by themselves so I will change this PR to include their way to invert some things - maybe there will be less logos to invert after that. 😉

PR #13534, opened 2024-11-20
Make code multi language for new design

JNylson:


Myshor:
The idea is nice as the new design is mostly not translated in many places but making review of it will take much time I think.

First look is it's ok. At least we have added all necessary variables into tranlation files. All of them are in English here but that means it will work the same as it was before without variables.

What is needed to be checked - all places where static text was changed to variable to be sure we don't have some typo in code.

Again today I will call @alexanderby as I do not know how the future plans for Dark Reader look like and if the new design will be default in the future or there will be total change of design so this PR would not be needed.

alexanderby:
You did a big chunk of work. But the problem now is this list of translations is hardly maintainable. And the design planned is totally different. Let me think how I can improve the structure of the translations files, and I will get back to you.

Currently, please move changes to existing Portuguese translations for old design into a separate PR.

PR #13535, opened 2024-11-20
change sepia filter to night mode

sangjun19:
I've changed sepia filter to night mode filter.
The filter protect our eyes from the blue light.


Myshor:
For me it's not acceptable as it totally changed variable for localization and only localization in korean language is added.
For the configuration for new setting I won't comment.

Requesting review from @alexanderby - maybe he will give some comments what to do with this PR to accept (or he will reject it).

alexanderby:
It would be great if you could explain what are you trying to achieve.

sangjun19:
Thank you for checking my code.
I'm student now, we are practicing contribute to open source.
Actually I tried to add new filter, night mode. I want to develop protecting blue light mode. But, it's too hard to understand whole code and there are so many code to fix. I thought sepia filter looks similar to night mode. So I changed sepia filter to night mode.
Thank you.

PR #13537, opened 2024-11-20
streamable.com fixes

FadeMind:
- remove orphans rules
- fix logo top header

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -26314,15 +26314,8 @@ IGNORE IMAGE ANALYSIS
 streamable.com
 
 INVERT
-.customer-logos
-.landing .features-pane .feature > span > svg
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -26314,15 +26314,8 @@ IGNORE IMAGE ANALYSIS
 streamable.com
 
 INVERT
-.customer-logos
```
Myshor, line 26317: This was used in https://github.com/darkreader/darkreader/pull/9866

And now I see it changed to .customer-logos-v2

Please update.

Myshor (changes requested):
Add customer logos new selector.

For the rest it seems it's doesn't be included anymore as I can see the player buttons and progress is not needed to be touched. 😉

Myshor (approved):

PR #13540, opened 2024-11-20
update wiki user talk highlighting

GalaxySH:
#### What
Fixes an issue with message highlighting in user talks on Wikipedia. Users receive notifications about messages on their user talk, clicking the notification brings them to the user talk where the message is highlighted.

#### Why
DR does not properly highlight the message. Disabling the color would prevent all message highlighting, and the default coloring on Wikipedia is a harsh white background. Recommend using one of the special DR background color variables.

#### Examples
Link with an example of it working: https://en.wikipedia.org/wiki/User_talk:PolyversialMind/Archive_1#c-Lajmmoore-20240523070600-Women_in_Red_June_2024
(highlighting disappears when you click elsewhere on page)

Before:
https://i.imgur.com/MlerCFV.png

After:
https://i.imgur.com/vrnyXYK.png


GalaxySH:
I would welcome other suggestions for coloring or no coloring

Myshor:
I tried to make the color not related to variable and I think the closest one to original is to use

```
.ext-discussiontools-init-targetcomment {
    background-color: rgba(192,192,255, 0.25);
}
```

It looks nice in Dark scheme and light scheme of Dark Reader.

Check if you confirm and if you like this then push this code to PR and I will merge.

If not then try to find different color. 😉

Myshor (changes requested):
The result with this code is not very nice. Check my comment.

GalaxySH:
> The result with this code is not very nice. Check my comment.

|||
|--------|--------|
| after image | after image | 

I think it looks much better

Myshor (approved):

PR #13559, opened 2024-11-24
Add react.dev and reactnative.dev to dark-sites.config

Abbondanzo:
Both https://react.dev and https://reactnative.dev are dark mode by default. Adding them to the `dark-sites.config` per [contribution guidelines](https://github.com/darkreader/darkreader/blob/main/CONTRIBUTING.md#adding-a-website-that-is-already-dark)

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -983,6 +983,8 @@ ray.so
 razeenf.ca
 razer.com
 razorsecure.com
+react.dev
+reactnative.dev
 redacted.ch
 redasm.io
 redbox.com


Myshor:
Rejecting as both sites checks preferences of system/browser.

In Light Mode set in Windows 11 they look like below:
obraz

obraz

Myshor:
They could be added into [detector-hints.config](https://github.com/darkreader/darkreader/blob/main/src/config/detector-hints.config)

However both of them are detected in my Dark Reader when option to detect dark theme on site is enabled.

PR #13560, opened 2024-11-25
Invert 2700chess' live ratings table

partingscientist:
This pull request inverts the table that displays chess live ratings on the main page for better readability.
This pull request also simplifies existing selectors.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -612,11 +612,8 @@ body {
 2700chess.com
 
 INVERT
-.cg-wrap piece.queen.black
-.cg-wrap piece.bishop.black
-.cg-wrap piece.rook.black
-.cg-wrap piece.knight.black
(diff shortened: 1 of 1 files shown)


Myshor:
Did you check the inverts on the chessboard?
I did in Firefox and Chromium based browser. 😝

Now all black pieces must be inverted to look OK.

Pushing fix to PR and merge. 😉

Myshor:
I looked on on some game preview using Firefox and Edge and in both browsers it was white when :not(.king) was added.

When exactly the king would be white without this not so I will try to confirm and find other selector to exclude.or think of other fix for it?

Myshor:
Seem that sometimes Dark Reader does the inverted blob picture for black king sometimes not. I still can not confirm on my side.

But I will make new fix which remove invert for black pieces totally and will IGNORE IMAGE ANALYSIS for black pieces. 😉

PR #13586, opened 2024-12-05
Fix for my.canngo.express

oberrich:
NB: This website is a german telemedicine provider and requires login.

_before_
image

_after_
image


Myshor:
As I do not see anythis suspicious that would break light scheming in Dark Reader so merging. 😉

PR #13595, opened 2024-12-07
Update dark-sites.config

stastix:


Myshor:
2 things that make this Pull Request unacceptable:

1. Sites in dark-sites.config must be entered alphabetically. You just added site in one before last line.
2. Sites must be always dark. See => [contribution guidelines](https://github.com/darkreader/darkreader/blob/main/CONTRIBUTING.md#adding-a-website-that-is-already-dark)
https://nextjs.org/ is white when system/browser is set to light theme:
obraz

Dark Reader nicely detects it's Dark Mode when Detect Dark Theme option is enabled so there is nothing to do here. I close this PR.

PR #13610, opened 2024-12-12
Fixed menu button on IKEA homepage.

SidRoberts:
Before:
image

After:
image

It also applies to the close button when the menu is open.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -13765,6 +13765,9 @@ body,
 
 ikea.*
 
+INVERT
+.menu-btn
+
 CSS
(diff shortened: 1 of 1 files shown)


Myshor:
Can you share some URL? I managed to find when hamburger menu icon is showing but I can not find used class in fix and for me the menu and close button is not dark.

obraz

SidRoberts:
It's only on https://www.ikea.com/ - the other webpages follow a different style where the menu button already works. The homepage has `html.startpage` so perhaps it might be safer to change the selector to `html.startpage .menu-btn` to avoid affecting other elements on other pages?

image

Myshor:
Thanks for the info how I can check it.... Merging it. 😉

PR #13611, opened 2024-12-13
Update dynamic-theme-fixes.config For Overleaf.com

Qrible:
Website: overleaf.com
config: Dynamic

Addition:
I added ".canvasWrapper>canvas" in INVERT so the pdf is more readable but still gets the luminosity and contrast effect because the black version of pdf is usually less readable (for math articles at least). 
I propose it more as an alternative than a permanent change but I can't create a pdf exception for a website.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -21243,6 +21243,7 @@ overleaf.com
 
 INVERT
 .canvasWrapper
+.canvasWrapper>canvas 
 
 ================================
 


Qrible:
No problem. Here is a document I have on overleaf and the change it does. However, I hope to make it more like a second possibility than a permanent fix if possible, but inverted pdf are not always easily readable.
Here It keeps the luminosity (90%) and contrast (120%) changes I have put in.
Before :
Darkreader - before

After:
Darkreader - after

oskarkupke:
Is there a way this can be reverted or, as @Qrible suggested, modified somehow to be just an option? I have bad eyesight and find looking at light mode pages pretty straining on the eyes, so this change has made overleaf a lot less usable for me :/

Or is there any workaround for this already that I'm not aware of?

Myshor:
@Qrible What about you? It was your PR so maybe you can make some comment. 😉

Of course I can do the revert however I am the guy who prefers PDFs not inverted and having white background that's why I am not doing it ad-hoc. 🙄

Qrible:
I've tried tweaking the html and other things, but the only way I found was to do this.
I understand that some prefer black PDFs with white writing, but in many cases, this results in much more eye-straining reading. Disabling darkreader on Overleaf is sad as I find it very useful. I would love to have an option like a check box to enable it, but I'm not good enough to do it in the current add-on. It is my primary goal, but I'm not competent enough for that.
I would find it wonderful if any good coder wants to do it. Otherwise, having to add the code line in the advanced setting every time I open my browser would probably make me give up on darkreader even tho I find it very nice

oskarkupke:
Very interesting to see what's straining for different people's eyes ^^

> Otherwise, having to add the code line in the advanced setting every time I open my browser would probably make me give up on darkreader even tho I find it very nice

Can you elaborate on this? I haven't really used the advanced settings so far, I tried editing a bit there now but no luck. Is there something I could add to get the old behavior back? Even if I had to reapply it every time I restart the browser, I think it would work for me, at least as a short term solution.

Qrible:
> Can you elaborate on this? I haven't really used the advanced settings so far, I tried editing a bit there now but no luck. Is there something I could add to get the old behavior back? Even if I had to reapply it every time I restart the browser, I think it would work for me, at least as a short term solution.

If you go to the darkreader setting > advanced > development tools > "per site editor" (up right) > type :"overleaf" > remove the line ".canvasWrapper > canvas" > Apply
Then you have the old version.
I added manually this line for a few weeks then asked for the change in the github. But I don't know if adding a button could be easily done for the HTML

Myshor:
I won't revert still the PR but I will have some question. Is there a possibility to share some URL I could see that "live". It's not easy to do anything with canvas but sometimes it could be possible to change the background only to some shade of gray. But like I wrote already without look how the whole page is created I can not take a part in discussion.

Qrible:
> I have proposal that may satisfy both sides. 😁
> 
> Instead of having `.canvasWrapper > canvas` in INVERT section I would use this code:
> 
> ```
> CSS
> html[data-darkreader-scheme="dark"] .canvasWrapper > canvas {
>     filter: invert(100%) hue-rotate(180deg) contrast(60%) !important;
> }
> ```
> 
> Which would INVERT that part of canvas too but contrast set to 60% would make the white background less painfull to the eyes. 

With the '.canvasWrapper > canvas' you don't touch the contrast and I've did it so my "contrast" and "luminosity" change affed also my pdf and I have a 80-90% luminosity with 120% contrast so if we could make it interactive with a slider or something It would be nice to everyone otherwise I think it'll never be perfect for everyone.
My problem is that I don't know how to have button dedicated for pdf rendering or canvas rendering.

Myshor:
> My problem is that I don't know how to have button dedicated for pdf rendering or canvas rendering.

You can't... There is some option to include invert PDF or not but it works in Chrome only as I remember it right... and it works for pure PDFs not PDFs embedded into site.

For the PDF embedded on site there is only possible to do some sitefix like was done here.

Let's see what @oskarkupke will reply for my offer.

oskarkupke:
> If you go to the darkreader setting > advanced > development tools > "per site editor" (up right) > type :"overleaf" > remove the line ".canvasWrapper > canvas" > Apply

This works, thanks 👍 

> I have proposal that may satisfy both sides. 😁

Hmm, definitely better than before, although I think I still prefer the dark-mode PDFs 🤔 

Ideal would probably be an option in the settings to "Invert PDFs" or something. I dont know how easy or hard it would be to add something like this in darkreader, I havent worked on any browser extensions yet.

Also a thing I noticed, the INVERT rules `.canvasWrapper` and `.canvasWrapper > canvas` seem to work against each other, if I edit the rules to use either one, the PDF is dark, but if i use both OR neither, the PDF is normal. It seems to me someone added the rule to invert the PDF to dark-mode some time ago, and now this second rule was added to re-invert it back to normal? ^^

oskarkupke:
> You can't... There is some option to include invert PDF or not but it works in Chrome only as I remember it right... and it works for pure PDFs not PDFs embedded on site.

Ah, that's a bummer.

Myshor:
> Also a thing I noticed, the INVERT rules `.canvasWrapper` and `.canvasWrapper > canvas` seem to work against each other, if I edit the rules to use either one, the PDF is dark, but if i use both OR neither, the PDF is normal. It seems to me someone added the rule to invert the PDF to dark-mode some time ago, and now this second rule was added to re-invert it back to normal? ^^

I checked what you wrote and... I confirm. The first one inverts only the PDF. The second is inside so it inverts it back. So to have white background PDF we could have no INVERT at all.

That makes my proposad even smaller - we would set filter with contrast only 😝 and remove all INVERTs 😁

Qrible:
> > Also a thing I noticed, the INVERT rules `.canvasWrapper` and `.canvasWrapper > canvas` seem to work against each other, if I edit the rules to use either one, the PDF is dark, but if i use both OR neither, the PDF is normal. It seems to me someone added the rule to invert the PDF to dark-mode some time ago, and now this second rule was added to re-invert it back to normal? ^^
> 
> I checked what you wrote and... I confirm. The first one inverts only the PDF. The second is inside so it inverts it back. So to have white background PDF we could have no INVERT at all.
> 
> That makes my proposad even smaller - we would set filter with contrast only 😝 and remove all INVERTs 😁

It make a difference: The luminosity and contrast  modifier doesn't apply to the PDF then so it's pure white

Myshor:
> Also a thing I noticed, the INVERT rules `.canvasWrapper` and `.canvasWrapper > canvas` seem to work against each other, if I edit the rules to use either one, the PDF is dark, but if i use both OR neither, the PDF is normal. It seems to me someone added the rule to invert the PDF to dark-mode some time ago, and now this second rule was added to re-invert it back to normal? ^^

I checked what you wrote and... I confirm. The first one inverts only the PDF. The second is inside so it inverts it back. So to have white background PDF we could have no INVERT at all.

That makes my proposad even smaller - we would set filter with contrast only 😝 and remove all INVERTs 😁

Summarizing: We need to make some decision what to do with this fix as now the whole overleaf.com sitefix could be removed to have the same result as it is now. 🤔

Qrible:
If something is done, either remove my fix or let it be like this but don't remove everything please

Myshor:
> It make a difference: The luminosity and contrast modifier doesn't apply to the PDF then so it's pure white

That's why I am asking what to do... We have 2 options as I understand:

1. Remove this invert and leave only 1 so the PDF is back with dark background.
2. Remove both inverts but add filter contrast to have it not full white.

And because we have 3 people here we can "vote".
I am for light background but more contrast to make it more grayed and less eye-painfull.

So guys choose dark or light background and we will have this brainstorm finished. 😁

Qrible:
> > It make a difference: The luminosity and contrast modifier doesn't apply to the PDF then so it's pure white
> 
> That's why I am asking what to do... We have 2 options as I understand:
> 
>     1. Remove this invert and leave only 1 so the PDF is back with dark background.
> 
>     2. Remove both inverts but add filter contrast to have it not full white.
> 
> 
> And because we have 3 people here we can "vote". I am for light background but more contrast to make it more grayed and less eye-painfull.
> 
> So guys choose dark or light background and we will have this brainstorm finished. 😁

Then as I said, remove my line and put the dark background and I'll manually add it back myself

Myshor:
OK... so just revert this.

oskarkupke:
If it's not too inconvenient for you @Qrible then sure, revert to the darkmode PDF.

Thank you both for looking into this!

PR #13673, opened 2024-12-26
Fix for app.element.io

Andrew15-5:
I tried using `IGNORE INLINE STYLE` and `INVERT` at the start, but it doesn't seem to work.

After:

screenshot_2024-12-26_18-30-16

screenshot_2024-12-26_18-34-59

image

Before:

screenshot_2024-12-26_18-39-23

image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -2133,63 +2133,63 @@ svg text.time_cursor {
 app.element.io
 
 CSS
-.mx_NotificationBadge,
-.mx_ReplyPreview_header_cancel,
-.mx_RoomTile_menuButton::before,
-.mx_IconizedContextMenu_icon::before,
-.mx_RoomTile_notificationsButton::before,
-.mx_RoomTile_notificationsButton_show::before {
+.cpd-theme-dark .mx_NotificationBadge,
+.cpd-theme-dark .mx_ReplyPreview_header_cancel,
+.cpd-theme-dark .mx_RoomTile_menuButton::before,
+.cpd-theme-dark .mx_IconizedContextMenu_icon::before,
+.cpd-theme-dark .mx_RoomTile_notificationsButton::before,
+.cpd-theme-dark .mx_RoomTile_notificationsButton_show::before {
     background: var(--cpd-color-icon-primary) !important;
 }
-.mx_MessageContextMenu_iconRedact::before {
+.cpd-theme-dark .mx_MessageContextMenu_iconRedact::before {
     background: var(--cpd-color-icon-critical-primary) !important;
 }
-.mx_AccessibleButton.mx_NotificationBadge {
+.cpd-theme-dark .mx_AccessibleButton.mx_NotificationBadge {
     background-clip: padding-box !important;
     border: 3px solid rgba(38, 39, 43, 0.82) !important;
 }
-.mx_EventTile[data-layout="group"].mx_EventTile_selected > .mx_EventTile_line {
+.cpd-theme-dark .mx_EventTile[data-layout="group"].mx_EventTile_selected > .mx_EventTile_line {
     background-color: var(--cpd-color-bg-subtle-secondary);
     box-shadow: rgb(235, 238, 242) 54px 0px 0px -50px inset;
 }
-.mx_ReplyChain {
+.cpd-theme-dark .mx_ReplyChain {
     border-left: 2px solid var(--username-color);
 }
-.mx_UserMenu,
-.mx_RoomHeader,
-.mx_TimelineSeparator > hr,
-.mx_LeftPanel_filterContainer {
+.cpd-theme-dark .mx_UserMenu,
+.cpd-theme-dark .mx_RoomHeader,
+.cpd-theme-dark .mx_TimelineSeparator > hr,
+.cpd-theme-dark .mx_LeftPanel_filterContainer {
     border-bottom: 1px solid var(--cpd-color-gray-400) !important;
 }
-.mx_RoomSublist_resizeBox:hover .mx_RoomSublist_resizerHandle,
-.mx_SpaceButton_icon::before {
+.cpd-theme-dark .mx_RoomSublist_resizeBox:hover .mx_RoomSublist_resizerHandle,
+.cpd-theme-dark .mx_SpaceButton_icon::before {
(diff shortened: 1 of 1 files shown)


Myshor:
Looked into it and... it fixes colors when element is using Dark Mode already on it's side.
When I changed manualy the theme to white this fix is not doing any fixes at all.

Maybe we need just add detection-hint for this page to just not use Dark Reader when page opens in dark mode already.
obraz

For me this fix is not acceptable to merge. I'll be back here some days later to see how it goes in the conversation. 😉

Andrew15-5:
> Looked into it and... it fixes colors when element is using Dark Mode already on it's side.

Yes. That is indeed what it does.

> When I changed manualy the theme to white this fix is not doing any fixes at all.

Why would you change manually to the white theme? There is probably nothing to fix there anyway. If you see any issue, I can update the patch.

> Maybe we need just add detection-hint for this page to just not use Dark Reader when page opens in dark mode already.

IIUC, this will add the issue of white background during the page load. This is literally part of why I use dark reader in the first place. Unless it's a different kind of disabling, this is not a good solution.

Myshor:
> Why would you change manually to the white theme? There is probably nothing to fix there anyway. If you see any issue, I can update the patch.

Some people can use light theme for system, have some theme applied to browser and use Dark Reader.

With Dark Reader enabled in this setup I ended with:
obraz

So colors for icons are broken even with sitefix. When we do sitefix we must be sure it will work fine with Light/Dark theme (if site uses different ones). We need to be sure it will it will work fine with light/dark scheme set in Dark Reader.

> this will add the issue of white background during the page load.

For me it adds white background on page load (I mean login page) even with the sitefixes when I use Firefox.



> Can you clarify on the reason?

All above mentioned.

And I see that on original Dark Mode all works fine after loggin in that's why I think about adding detection-hint for this page instead of using this CSS fixes.

Andrew15-5:
> For me it adds white background on page load (I mean login page) even with the sitefixes when I use Firefox.

Uhh, whaa? What do you mean "it adds white background"? Who is "it"? With dark reader on, I now never ever see a bright loading page in Firefox (and probably when I was using Chrome too). Either I don't know something, or sounds like something is broken then.

Does using `@media (prefers-color-scheme: dark)` fix the issue (wrapping the patch code in it)? Since Element is using system theme and `@media` will detect it the same way, with auto-theme mode on, it will only apply the patch for the dark theme.

Other than this, I don't know what else can I do. There could be some hacks with `:has([class="cpd-theme-dark"])` or something (plus maybe `:not()`), but it's way too hacky and the feature is somewhat recent, at least in Firefox.

Myshor:
I meant this white background:
obraz

And it is always white for me even when I have all set to dark in system and browser + Dark Reader is enabled.

And I still prefer detection hints if we want to "fix" already dark mode than making CSS fixes that is not universal to all page possible schemas.

But ok... Let's call @alexanderby to as he is the man dev for Dark Reader. 😉

Andrew15-5:
Oh! This one? Yeah, now I understand, I have the same thing, thanks for clarifying that. I meant the screen/moment when pretty much any website is loading or being reloaded, then there is either a short or a very long (5+ s) period of time when it's just blasting unsolicited almost fullscreen pure white rectangle into my eyeballs (between being super dark). Not a fan of that (when dark reader is disabled).

Myshor:
I tried to check and find some workaround that will work on both light and dark theme and ended with:
```
app.element.io

INVERT
.mx_NotificationBadge
.mx_ReplyPreview_header_cancel
.mx_RoomTile_menuButton::before
.mx_IconizedContextMenu_icon::before
.mx_RoomTile_notificationsButton::before
.mx_RoomTile_notificationsButton_show::before

```

This and the detection-hints add to detect dark theme should work fine with the site.

Andrew15-5:
Is the above patch supposed to... fix the dark theme? because if the dark theme detection is to be added, I don't understand if a patch is even necessary.

The simplified fix as expected cut some corners. 3 of the most noticeable:

The plus button is practically gone:

image

The selected message is very hard to easily distinguish from other ones:
image

And the currently selected chat is also not highlighted on the left side.

Everything else is _mooostly_ the same, but still under direct comparison there are many more discrepancies.

Myshor:
The main thing we should summarize is... Are we adding that dark `detection-hints` so we do not need to make fixes for dark mode at all.

If yes then the INVERT
```
app.element.io

INVERT
.mx_NotificationBadge
.mx_ReplyPreview_header_cancel
.mx_RoomTile_menuButton::before
.mx_IconizedContextMenu_icon::before
.mx_RoomTile_notificationsButton::before
.mx_RoomTile_notificationsButton_show::before

```

would be needed if page will load in light mode.

If we still consider to not add detection hints then as your fix works only with dark mode then it should have in every selecton `.cpd-theme-dark` as pre-class that would tell it should be touched only when dark theme is used on the site.

And then we can do invert but with pre-class `cpd-theme-light`.

As this is your PR then you must make some decission. 😉

I can make the fix I think it's best - and for me it's detection + inverts only in dynamic theme fixes. But this would end with closing this PR or just changing it to INVERTs only. 😈

Andrew15-5:
> I can make the fix I think it's best - and for me it's detection + inverts only in dynamic theme fixes. But this would end with closing this PR or just changing it to INVERTs only. 😈

Bruh...

OK, so the explanation clears all the questions I have left. I'm surprised I didn't realize that the theme class can be just prepended to the fixes, and it will not touch the light theme. As I/we don't know when or if the hinting will be added, I'll use your suggestion and combine everything together.

Myshor (approved):
OK... That is acceptable to me. 😉

Andrew15-5:
Right, I forgot to mention that since Element is a FOSS web app that anyone can host, there are other websites that someone can add in the future. But I suppose this isn't an issue since multiple websites can be stacked like I did with Typst:

```
typst.app
staging.typst.app
```

So, for example:

```
app.element.io
element.4d2.org
element.sibnsk.net
```

Do you think I should add those now, since I happen to know them?

Myshor:
typst has already included sitefixes based on invert as I checked.

the element ones you can add as aliases in the new PR... aliases are not that much restricted witch checking 😁

Andrew15-5:
Great, then here is the freshly baked PR: https://github.com/darkreader/darkreader/pull/13740.

Issue #11778, opened 2023-10-04
[Bug] DR sometimes no longer correctly detects dark theme on websites


rakleed:
### Prerequisites

- [X] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this bug to avoid opening a duplicate.
- [X] I can reproduce this bug in a new, unmodified web browser profile with Dark Reader installed as the only extension.
- [X] I understand I need to use the [Broken Website Report template](https://github.com/darkreader/darkreader/issues/new?assignees=&labels=Broken+Website&template=broken-website-report.yml&title=%5BBroken+Website%5D+Replace+with+title) if this bug I am reporting occurs on a single website.

### Bug Description

DR after recent updates sometimes does not correctly detect dark theme on websites.

### Website Address

https://github.com/darkreader/darkreader

### Steps To Reproduce

1. Enable `Detect Dark Theme` option in Dark Reader Settings
2. Choose to enable Dark Reader automatically according to the theme system
3. Enable automatic theme change on GitHub in accordance with the system theme
4. Try opening GitHub in a new tab every few minutes

### Expected Behavior

DR is automatically disabled every time you open GitHub with a dark theme.

### Actual Behavior

DR is sometimes enabled when opening GitHub with a dark theme.

### Screenshots

![image](https://github.com/darkreader/darkreader/assets/19418601/e392307f-4c19-442a-afac-a28f66e6c9e8)

### Operating System

macOS Sonoma 14.0 (23A344)

### Web Browser name and version

Chrome 117.0.5938.149

### Dark Reader version

4.9.67

### Additional Context

In version 4.9.65, incorrect operation of the Detect dark theme option happened once every few days, but now I encounter this problem several times or even dozens of times a day.

bloodgain:
This seems to be a broader issue with dark theme detection since the last update on 10/2. There are several reports for individual websites being detected incorrectly as using a dark theme when they are not. I have seen several others, which I came here to start a report on -- e.g. arstechnica.com is detected as dark even when the light theme is applied. While this is the more noticeable situation, DR's false negatives (no dark theme detected when there is one) also seems to be spotty. Some websites, such as logitechg.com, appear to have confusing theming, and this may be hard for Dark Reader to detect correctly.

[...]

PR #13674, opened 2024-12-26
Add detector hints for Wikipedia

rakleed:
Add detector hints for new Wikipedia theme [`Vector (2022)`](https://en.wikipedia.org/wiki/Special:Preferences#mw-prefsection-rendering). Also, add the second hint for cases when the user chooses the `Automatic` option, the same as for GitHub.

### DIFF
--- a/src/config/detector-hints.config
+++ b/src/config/detector-hints.config
@@ -694,6 +694,17 @@ NO DARK THEME
 
 ================================
 
+wikipedia.org
+
+TARGET
+html
+
+MATCH
+.skin-theme-clientpref-night
+.skin-theme-clientpref-os
+
+================================
+
 x.com
 
 TARGET
--- a/src/config/detector-hints.config
+++ b/src/config/detector-hints.config
@@ -701,7 +701,6 @@ html
 
 MATCH
 .skin-theme-clientpref-night
-.skin-theme-clientpref-os
 
 ================================
 


src/config/detector-hints.config
```diff
@@ -694,6 +694,17 @@ NO DARK THEME
...
 ================================
 
+wikipedia.org
+
+TARGET
+html
+
+MATCH
+.skin-theme-clientpref-night
+.skin-theme-clientpref-os
```
Myshor, line 704: I can not accept with this class as you can end with white system theme and dark reader installed and used... then this will leave wikipedia on light mode.

I can accept the selector for permament night only.

Myshor (changes requested):

src/config/detector-hints.config
rakleed: Please take a look at https://github.com/darkreader/darkreader/issues/11778 and https://github.com/darkreader/darkreader/pull/11995

Myshor: You want me to revert the #11995 for the auto selector? 🙄
In 2023 I wasn't experienced how the Dark Reader can work with Light Theme and that auto can end with light one and we should not disable Dark Reader for white one.

@alexanderby Can you tell how you think of adding detectors here for system-preffered?

alexanderby: I would only leave the first line, as there is a high chance of people getting a white theme. Unfortunately I wasn't able to find a reliable solution for auto-detection at this time, but will try improving it in the next version.

Myshor (approved):

Issue #13678, opened 2024-12-28
[Broken Website] oed.com Broken search dropdown


Obsidianninja11:
### Prerequisites

- [x] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this website issue to avoid opening a duplicate.
- [X] I can reproduce this website issue in a new, unmodified web browser profile with Dark Reader installed as the only extension.

### Website Issue Description

Due to improper svg image inverting in darkreader sync, the search dropdown at oed.com is bugged.

### Website Address

https://oed.com

### Steps To Reproduce

1. Go to oed.com
2. Look at the search bar dropdown

### Screenshots

Without darkreader:
![image](https://github.com/user-attachments/assets/1e8cd525-7275-44d6-a03d-dc4bbd652f80)

With darkreader and the cause:
![image](https://github.com/user-attachments/assets/19612c6c-c989-4c0b-96b0-5d72ad64d116)
![image](https://github.com/user-attachments/assets/e645e3c5-a288-4a18-b305-1b4ffc87c2a5)

### Operating System

Windows 11 Pro

### Web Browser name and version

Google Chrome 131.0.6778.205

### Dark Reader version

Version 4.9.96

## Fix:
A possible fix is to hardcode the properly inverted svg, which obviously isn't ideal, but using a transparent background works pretty well even for different color schemes.
```css
select {
    background-image: url("data:image/svg+xml,%3Csvg id='Layer_1' data-name='Layer 1' xmlns='http://www.w3.org/2000/svg' viewBox='0 0 4.95 10'%3E%3Cdefs%3E%3Cstyle%3E.cls-1%7Bfill:transparent;%7D.cls-2%7Bfill:%23bbb;%7D%3C/style%3E%3C/defs%3E%3Ctitle%3Earrows%3C/title%3E%3Crect class='cls-1' width='4.95' height='10'/%3E%3Cpolygon class='cls-2' points='1.41 4.67 2.48 3.18 3.54 4.67 1.41 4.67'/%3E%3Cpolygon class='cls-2' points='3.54 5.33 2.48 6.82 1.41 5.33 3.54 5.33'/%3E%3C/svg%3E");
}
```
The fix at default brightness
![image](https://github.com/user-attachments/assets/38e5a7dc-be26-4610-9cd7-cd02d54d3907)

* **Pull request implementing this fix: https://github.com/darkreader/darkreader/pull/13679**

Myshor:
[...]

PR #13679, opened 2024-12-28
Fix for oed.com

Obsidianninja11:
* Resolves https://github.com/darkreader/darkreader/issues/13678

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -20774,6 +20774,9 @@ CSS
 #maincontainer {
     color: var(--darkreader-neutral-text) !important;
 }
+select {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -20774,6 +20774,9 @@ CSS
 #maincontainer {
     color: var(--darkreader-neutral-text) !important;
 }
+select {
```
Myshor, line 20777: I know you tried to make nice fix but we try not use `url("data:image"` in sitefixes.
Second thing... As Dark Reader can use light or dark scheme for applying it's changes to the site then `fill:%23bbb` should be changed to something more neutral like `fill:%23888`

Myshor (changes requested):
Do the changes for more neutral color... I will push the fix to exclude the invert for this background image in a minute. Because it's better to have it always with white background than broken.

And I will call @alexanderby here to make his review if we would accept your fix or not in.

src/config/dynamic-theme-fixes.config
Obsidianninja11: > we try not use url("data:image" in sitefixes
Yeah that makes sense.
I didn't realize `IGNORE IMAGE ANALYSIS` existed.

Myshor (approved):
For me it's ok now. Just need to wait for Alexander statement about `data:image` usage.

PR #13684, opened 2024-12-31
Fix CW PSA icons for all instances

jerko-lenex:
#12526 already added NA to the config, this PR would enable white icons for all CW PSA instances.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -261,6 +261,15 @@ CSS
 
 ================================
 
+*.myconnectwise.net
+
+CSS
+.cwsvg {
+    fill: white;
+}
+
+================================
+
 *.myworkday.com
 
 INVERT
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -19666,7 +19666,7 @@ span[class*="nk-icon_id_business-feedback-task"] > svg > g > path
 
 ================================
 
-na.myconnectwise.net
+*.myconnectwise.net
```
Myshor, line 19669: Sitefixes must be sorted alphabetically.
When you remove main URL and change it to another then the fix must be moved - in this case before `*.myworkday.com`

Myshor (changes requested):

src/config/dynamic-theme-fixes.config
jerko-lenex: Thank you for the feedback, I've moved the entry to before `*.myworkday.com`.

Myshor (approved):

Myshor (approved):

PR #13693, opened 2025-01-03

Sooperfish:
genius.com 
- Header: Color, logo and signup link fix nytimes.com
- Games hub: NYT Games logo, hamburger menu and progress icons fix
- Letter Boxed: Game icon fix
- The Crossword: Replaced some variable colors for cells with hex colors to match the color scheme to the android app in dark mode

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -11845,6 +11845,20 @@ genius.com
 
 INVERT
 .texmath
+.krOZTj
+
+CSS
+.header-primary,
+.dhlnLG {
+    background-color: rgb(255, 255, 100);
+}
+.header-action,
+.iiLMJr {
+    color: rgb(121, 115, 103);
+}
+
+IGNORE IMAGE ANALYSIS
+.header-primary .logo_link
 
 ================================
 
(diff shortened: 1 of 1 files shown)


Myshor:
First thing: Please share URLs we can check the differences before and after fixes.
Second thing: As I found you changed some `${color}` parts of fixes into static one colors then it's needed to be tested a little bit more before merged as this can (I am not saying it will for sure but it can) break the colors when light scheming in Dark Reader is used.
And last but not least: For the next time it's better to push 2 different Pull Requests for each site. Having 2 sites here in the fix will take more time to check everything + can block one part of the fix if other will need some changes.

Summarizing:
We need URLs to check changes, as screenshots for bigger changes can be not enough.

Sooperfish:
Sorry for the late reply. I have edited my original comment to add links, and I'll use separate pull requests next time. Thanks for the feedback.
About the usage of static colors instead of `${color}`, I did this mainly because I wanted to have an accurate 1 on 1 color match to the official dark mode the android app uses. I'm not that proficient with CSS, so if there was a better way to achieve that, please let me know.
One issue: I noticed the progress icons issue on `https://www.nytimes.com/crosswords` persists (some of them should be inverted). I'm still a little confused what selector to use for those, so I've reverted that one for now.

src/config/dynamic-theme-fixes.config
```diff
@@ -20671,19 +20690,32 @@ a[data-testid] > svg {
...
+    stroke: #171717 !important;
+}
+#xwd-board path {
+    stroke: #171717 !important;
+}
+#js-logo-nav > svg > rect {
+    fill: #000 !important;
 }
 
 IGNORE INLINE STYLE
```
Myshor, line 20721: Add ti this section:
`#js-logo-nav > svg > rect` and the logo will work better in light/dark scheme.

src/config/dynamic-theme-fixes.config
```diff
@@ -20671,19 +20690,32 @@ a[data-testid] > svg {
     fill: ${goldenrod} !important;
 }
 #xwd-board rect[class^="xwd__cell--cell"] {
-    fill: ${darkgrey} !important;
+    fill: #595963 !important;
```
Myshor, line 20693: If you want to have this color in dark scheme then use `${#ada59b}` - this will do lighter cell in light scheme.

Myshor (changes requested):
Still I found some places to fix. But it's less than last time. 😉

PR #13698, opened 2025-01-04
Fix for wunderground.com

pnaf:
Fixes map widget on "Weather Conditions" pages.

Before:

After:

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -31361,6 +31361,10 @@ select:focus {
 
 wunderground.com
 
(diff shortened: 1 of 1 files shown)


Myshor:
As you can change the map to satellite and it's not possible to catch it when it's map and satellite in canvas. That's why we do not INVERT that kind of maps even in fixes because we do not want to break satellite view.

Not merging and closing.

pnaf:
Ah, you're right, I somehow missed checking the satellite view, and it doesn't seem possible to distinguish between the two views at the moment. Good call! :smile:

PR #13707, opened 2025-01-06
Fix techdocs.akamai.com

zmousm:


Myshor (changes requested):
First of all... it would be nice to have some screens and/or URLs to see how the site looks before fixes and with fix.
2nd thing... I checked some pages in Firefox and found your fix not working so good... here is my URL:
https://techdocs.akamai.com/ion/docs/set-up-adaptive-acceleration so you will need to make some changes to it.
obraz


Last but not least: I checked the page https://techdocs.akamai.com/ion/docs/set-up-adaptive-acceleration on Chromium based browser without your fix and it works fine:
obraz

So even adding some fix here will need to check both kind of browsers if it's working fine in both.

Myshor:
So... best thing here would be sharing URLs you were taking to fix so doing the new review we can check them in both kind of browsers and give a tip how to make fix working for both or even push the fix that will work fine in both. Now I do not even want to make any changes in this PR as I can make more damage than fix. 😁

PR #13708, opened 2025-01-06
Fix www.in.gr

zmousm:


Myshor:
Thanks for the PR...

Sorry for what I will do with it but I found more places to fix and different approach to do that.
Because you started the PR I will do the changes on your PR and merge. 😉

PR #13712, opened 2025-01-07
Navigation bar fixes for apple.com

StevenSchifter:
CSS rules for navigation bars are now just under apple.com.


src/config/dynamic-theme-fixes.config
```diff
@@ -2399,21 +2403,6 @@ INVERT
...
 
-CSS
-#globalnav {
-    --globalnav-background: var(--darkreader-neutral-background) !important;
-}
-
-================================
-
-apps.apple.com
-support.apple.com
```
Myshor, line 2410: You miss this alias after your changes which breaks the navbar!
obraz

I think the easiest way to fix this is to add alias under `apple.com` and if we want to be sure we won't miss anything with prefix on the apple.com then I would add `*.apple.com`

I will add the other comment in the place I mention.

src/config/dynamic-theme-fixes.config
```diff
@@ -2369,6 +2369,10 @@ INVERT
 .rf-ccard-content-info
```
Myshor, line 2369: I can't add comment earlier but adding the `*.apple.com` alias should be included as 2365 line and empty line should be under it.
obraz

To know what I am writing about check next comment. 😉

Myshor (changes requested):
Need changes to include missing support.apple.com fix.

Myshor (approved):

src/config/dynamic-theme-fixes.config
StevenSchifter: I added the *.apple.com alias on line 2365 with an empty line underneath.

PR #13716, opened 2025-01-11
fix for *.docs.kubernetes.io

dudeofawesome:
I'm not sure if this is the right way to fix the random underscores that I'm seeing.

image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -252,6 +252,15 @@ CSS
 
 ================================
 
+*.docs.kubernetes.io
+
+CSS
+.highlight pre code > span {
+    text-decoration: none !important;
+}
+
+================================
+
 *.maricopa.edu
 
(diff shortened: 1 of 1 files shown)


Myshor:
Can you share some example URL that can be checked and analyzed if there could be some other solution?

dudeofawesome:
Here's an example of a [page with the underline issue](https://kubectl.docs.kubernetes.io/references/kubectl/kustomize/).

Myshor:
OK... but as I can see the underlines are there even when dark reader is not appled (but they are white so almost invisible).
I am little afraid that removing textdecoration for all spans can break something else.

I will push more detailed selector into your PR and merge. 😉

Myshor:
Something breaks the tests on merge. I will try to push it again with changes given as a tip in tests.

dudeofawesome:
Huh. I didn't notice the underlines in light mode.

PR #13721, opened 2025-01-12
Fix for spaceship.com

felixonmars:
- Invert bright logo
- Invert a bright background image for search bar (after login).

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -25964,9 +25964,8 @@ CSS
 
 spaceship.com
 
-INVERT
-.lf-logo.lf-color-scheme--light
-.lp-background-image
+IGNORE INLINE STYLE
+svg *
 
 ================================
 
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -25962,6 +25962,14 @@ CSS
 
 ================================
 
+spaceship.com
+
+INVERT
+.lf-logo.lf-color-scheme--light
+.lp-background-image
+
+================================
+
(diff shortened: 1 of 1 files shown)


Myshor:
The whole page is making my Firefox hang up.

So I tried different approach:
```
spaceship.com

IGNORE INLINE STYLE
svg *
```

That gave me the first logo fixed without INVERT.

And fixed treedots too... So maybe check if my fix would do the trick with "Invert a bright background image for search bar (after login)."

If not then change the first invert to that IGNORE INLINE STYLE section if yes then just change the fix to the code above.

src/config/dynamic-theme-fixes.config
```diff
@@ -25962,6 +25962,14 @@ CSS
 
 ================================
 
+spaceship.com
+
+INVERT
+.lf-logo.lf-color-scheme--light
```
Myshor, line 25968: This will be removed and changed to IGNORE INLINE STYLE fix.

src/config/dynamic-theme-fixes.config
```diff
@@ -25962,6 +25962,14 @@ CSS
...
+.lp-background-image
+
+================================
```
Myshor, line 25971: If INVERT will be still needed then IGNORE INLINE STYLE will be as the next section under it to add.

Myshor (changes requested):

src/config/dynamic-theme-fixes.config
felixonmars: It appears this has been solved as well. Changed accordingly and thanks!

Myshor (approved):

PR #13731, opened 2025-01-15
Fix www.mozilla.org (dynamic)

Gitoffthelawn:
Logo in upper-left corner  
Example page: https://www.mozilla.org

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -32495,6 +32495,7 @@ www.mozilla.org
 INVERT
 img.sidebar-icon
 h1.c-logo
+.c-navigation-logo-image
 .m24-c-navigation-logo-image
 
 CSS
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -32495,6 +32495,7 @@ www.mozilla.org
 INVERT
 img.sidebar-icon
 h1.c-logo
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -32495,6 +32495,7 @@ www.mozilla.org
 INVERT
 img.sidebar-icon
 h1.c-logo
+.m24-c-navigation-logo-image
```
Myshor, line 32498: I checked and for me it seems the selector is `.c-navigation-logo-image`

Can you check again?

Myshor (changes requested):

src/config/dynamic-theme-fixes.config
Gitoffthelawn: Yes, thanks.  It's different for me.  Which browser did you use?  I think we will need to add both classes.

Myshor: obraz
Width: 2560px as my monitor is 2560x1080 and I use Firefox in Full Screen.

But for me the main page opens as /pl/ when I changed to En-US I see your class.

So yeah... I think it will be better to add both classes. 😉

Gitoffthelawn: Done! Thank you!

Gitoffthelawn:
Additional class added. Thank you for the review.

Myshor (approved):

PR #13753, opened 2025-01-20
Add gethomepage.dev to dark sites list

APT37:


Myshor:
It has some switch to make dark or light but as it only makes it dark or darker I will merge this PR. 😉

PR #13754, opened 2025-01-20
Fix for www.bilibili.com

IcedDog:
Fixed TeX Images not inverted.


Myshor:
Can you share some example URL where these TeX Images are included on site?

There are not on main page and as I am not good in chinese language it would be hard to find it fast. 😅

PR #13766, opened 2025-01-21
fix for docs.redhat.com

fmigneault:
Code blocks background colors are inverted.

See for example: 
https://docs.redhat.com/en/documentation/red_hat_openstack_platform/8/html/command-line_interface_reference_guide/openstackclient_commands

Prior to fix:

{410E769F-7C6B-4348-9348-29D051307BD0}

After fix:

{804181CE-4748-4527-95CD-CEA2BEEC73F7}

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -8927,6 +8927,15 @@ body {
 
 ================================
 
+docs.redhat.com
+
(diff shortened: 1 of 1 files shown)


Myshor:
Rejecting as it was reseting colors after page reload... I will push some other fix for docs in a minute. 😉

PR #13780, opened 2025-01-23
Fix for www.meridianproject.tw

lumynou5:


lumynou5:
Ideally, it would look better if we manually adjust the gradient backgrounds because of the limitation of `hue-rotate()`, but it costs too much to maintain.

PR #13785, opened 2025-01-24
Update dark-sites.config Adding iTrade.gg

whoops1e:
Including iTrade.gg to the list.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -1,3 +1,4 @@
+*.itrade.gg
 *.cases.gg
 *.clash.gg
 *.cncnet.org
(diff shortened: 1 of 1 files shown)


src/config/dark-sites.config
```diff
@@ -1,3 +1,4 @@
+*.itrade.gg
```
Myshor, line 1: Why you add it with `*.` in the beginning?
Is this site has some version with prefixes? If not then it should be added without that prefix!

Myshor (changes requested):

Myshor (approved):

PR #13789, opened 2025-01-25
Fix for theguardian.com

pnaf:
- Improves visibility of header and footer by using site-specific colors.
- Fixes logo on some article pages.

Before:

After:


Myshor:
Ok... Needed to check how it looks in Dark Reader light scheme but it's ok. 😉

PR #13802, opened 2025-01-26
Fix for typst.app (sidebar, dashboard, settings)

Andrew15-5:
Before:

image image

image

After:

image image

image

I wasn't able to fix the blue gradient background for the theme switcher because some `null` "inline:1" style was being added no matter what. But if I add it manually inline, then it starts working.


Myshor:
OK... specially created account to log in and check how it works when Dark Reader scheme will be switched to light/dark. :)

Andrew15-5:
Ha! I tricked you into becoming a Typst user! My plan worked flawlessly! Now enjoy better document quality! :D

PR #13815, opened 2025-01-28
Fix for Wikipedia

Tiagoquix:
Fixes 

The "standalone" version of the WW II Luftwaffe KG 54 wing's dead's head unit insignia

in

https://en.wikipedia.org/wiki/Totenkopf


Myshor:
For me it was "visible" but as you wish. 😉

PR #13833, opened 2025-02-01
Added fix for bright.uvic.ca to dynamic-theme-fixes.config

anOgden:
Darkens white navigation containers & makes top navigation bar match color scheme.


anOgden:
I've looked again and this was just another instance of dynamic mode not applying properly on loading the page.

PR #13848, opened 2025-02-05
Fix for Marquette.edu

CarterFike:
Dark reader on Marquette made it difficult to read some text due to the gradient. Made a change to gradient color to match font style.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -18065,6 +18065,18 @@ INVERT
 .svg-icon-block
 
 ================================
+marquette.edu
+
+CSS
+.h1 {
+    background-image: linear-gradient(
+        to right,
+        var(--darkreader-background-0098ca, #8fcdff) 20%,
+        var(--darkreader-background-003366, #8fcdff) 40%,
+        var(--darkreader-background-003366, #8fcdff) 100%
+    ) !important;
+}
+================================
 
 mastarti.com
 streamguard.cc


Leaf45:
peer programming moment

src/config/dynamic-theme-fixes.config
```diff
@@ -18064,6 +18064,18 @@ INVERT
 .mp-svg-profile
 .svg-icon-block
 
+================================
```
Myshor, line 18067: Missing empty line after '===' line

src/config/dynamic-theme-fixes.config
```diff
@@ -18064,6 +18064,18 @@ INVERT
...
+CSS
+.h1 {
+    background-image: linear-gradient(
+        to right,
+        var(--darkreader-background-0098ca, #8fcdff) 20%,
+        var(--darkreader-background-003366, #8fcdff) 40%,
+        var(--darkreader-background-003366, #8fcdff) 100%
+    ) !important;
+}
 ================================
```
Myshor, line 18079: Missing empty line before '===' line

Myshor:
The whole fix is not working as it should I think:
obraz

And when I checked that on light mode it looks even worse:
obraz

Myshor (changes requested):
I checked the CSS for not gradient parts that breaks with your fix and I propose to use this code:
```

marquette.edu

CSS
.h1 {
    -webkit-text-fill-color: unset !important;
    background-color: unset !important;
    background-image: unset !important;
    color: var(--darkreader-text--primary_blue, var(--darkreader-text-000000)) !important;
}

```

This will force non-gradient text for all places but will not create some problematic backgrounds.

And this will work nice in light scheme too:
obraz

PR #13849, opened 2025-02-05
add `nitter.net` to `dark-sites.config` (again)

Rudxain:
- Reverts #13215
- [Info](https://github.com/zedeus/nitter/pull/1210)


Myshor:
Themes can be set through settings and there is Twitter theme which is full white!

obraz

That disqualify this page to the dark-sites!

Rudxain:
oh ok! It seems like a new (perhaps it's not new?) feature I missed

PR #13880, opened 2025-02-09
add riflesniper.art + subdomain to dark sites

riflesniper:
i find that even with the dark lock put in the head, the sites still show up as white and unformatted for a split second before the css kicks in when i have darkreader turned on. hoping that adding my sites to the global list will help people avoid any possible trouble by turning it off automatically, since i try to be mindful of people who are photophobic/epileptic.


riflesniper:
thank you!!

PR #13882, opened 2025-02-10
Add codebuff.com to dark-sites.config

bherila:
# Codebuff.com: 
Site already supports dark mode, and currently Dark Reader is causing certain information like Pricing to fail to display

## Before 

Screenshot 2025-02-09 at 4 03 53 PM

## After

Screenshot 2025-02-09 at 4 03 59 PM

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -228,6 +228,7 @@ cnc-comm.com
 cncnet.org
 co.wukko.me
 cobalt.tools
+codebuff.com
 codeium.com
 codepen.io
 coder.com


Myshor:
Site can't be added as it has switch to make site light or dark:
obraz

But you can add the detection-hint here => https://github.com/darkreader/darkreader/blob/main/src/config/detector-hints.config
Just use this code in correct place (it must be sorted alphabetically)
```
codebuff.com

TARGET
html

MATCH
.dark

```

bherila:
OK, sent PR in https://github.com/darkreader/darkreader/pull/13885

PR #13883, opened 2025-02-10
fix: Improve cursor visibility for console.cloud.google.com

jinwoo1225:
# Description
Dear Dark Reader Maintainers, This PR fixes the cursor visibility issue in `console.cloud.google.com` when Dark Reader is enabled. Previously, the text cursor was difficult to see due to its dark color against dark backgrounds, especially in query editors.

# Changes
Added site-specific CSS rules for console.cloud.google.com to make the Monaco Editor cursor white:
```css
.monaco-editor .cursor,
.monaco-editor .cursor-block,
.monaco-editor .cursor-line {
    background-color: #ffffff !important;
    border-color: #ffffff !important;
}
```

# Before/After
I attached a screen capture for ease of checking out.

https://github.com/user-attachments/assets/ba9827f8-b5ae-42af-b6b3-b1acf2016b0b

Before: Text cursor is dark and barely visible in dark mode
After: Text cursor is white and clearly visible against dark backgrounds
This change specifically improves the user experience when editing queries in Google Cloud Console's query editor.

# Testing
Tested in Google Cloud Console (console.cloud.google.com)
Verified that the change only affects console.cloud.google.com

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -6866,8 +6866,8 @@ input[type=text] {
 .monaco-editor .cursor,
 .monaco-editor .cursor-block,
 .monaco-editor .cursor-line {
-    background-color: #ffffff !important;
-    border-color: #ffffff !important;
+    background-color: ${#000000} !important;
+    border-color: ${#000000} !important;
 }
 
 ================================
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -6863,6 +6863,12 @@ input[type=text] {
 .bq-results-table-header-cell {
     background: var(--darkreader-neutral-background) !important;
 }
+.monaco-editor .cursor,
+.monaco-editor .cursor-block,
+.monaco-editor .cursor-line {
+    background-color: #ffffff !important;
+    border-color: #ffffff !important;
+}
 
 ================================
 


src/config/dynamic-theme-fixes.config
```diff
@@ -6863,6 +6863,12 @@ input[type=text] {
 .bq-results-table-header-cell {
     background: var(--darkreader-neutral-background) !important;
 }
+.monaco-editor .cursor,
+.monaco-editor .cursor-block,
+.monaco-editor .cursor-line {
+    background-color: #ffffff !important;
```
Myshor, line 6869: We can use Dark or Light scheme in Dark Reader.
This color looks nice in Dark scheme but I think it will be not visible if we change the scheming to light.

To fix this we need to use `${#000000}` instead of `#ffffff`.

This will give `#000000` black in light scheming and inverted `#ffffff` in dark.

src/config/dynamic-theme-fixes.config
```diff
@@ -6863,6 +6863,12 @@ input[type=text] {
...
+    border-color: #ffffff !important;
```
Myshor, line 6870: We can use Dark or Light scheme in Dark Reader.
This color looks nice in Dark scheme but I think it will be not visible if we change the scheming to light.

To fix this we need to use `${#000000}` instead of `#ffffff`.

This will give `#000000` black in light scheming and inverted `#ffffff` in dark.

Myshor (changes requested):

src/config/dynamic-theme-fixes.config
jinwoo1225: thanks for review! just applied and tested on my local setup!

> Video Attachment

https://github.com/user-attachments/assets/ac3bd248-fb0b-4562-8871-31130d49a0a3

commit : [review: use ${#000000} as color of cursor](https://github.com/darkreader/darkreader/pull/13883/commits/c8fea9036b83e39a9fbf82e5c730d47b79149592)

PR #13887, opened 2025-02-10
Fixes for Wiktionary and Wikipedia

adamhotep:
This fixes the Wiktionary logo and interproject box as well as returning the open-book background for Wikipedia's Monobook skin (which used to be the default and is still quite common).


Myshor:
Can you share the example URLs?

As you make some more changes than one small then I would like to check how the fixes work live and how they will work when light scheme will be selected in Dark Reader.

src/config/dynamic-theme-fixes.config
```diff
@@ -18420,6 +18420,7 @@ INVERT
 .mw-kartographer-map
 .mw-kartographer-mapDialog-map
 .mw-logo-tagline
+.mw-logo-icon
```
Myshor, line 18423: that will need to be changed

maybe for en.wiktionary it will work but on pl.wiktionary I have this logo and it should not be inverted:
obraz

So as a resolve I propose `img[src="/static/images/icons/enwiktionary.svg"]`

src/config/dynamic-theme-fixes.config
```diff
@@ -18445,7 +18446,7 @@ img[alt="audio speaker icon"]
 img[alt="logo Wikisource"]
 img[src*="Wiktionary-logo"]
 img[src="images/black.png"]
-img[src*="Wikipedia-logo-textonly.svg"]
+img[src*="Wikipedia-logo"]
```
Myshor, line 18449: Why you want to invert this logo:
obraz

As it looks ok as it is right now?

Is there some other logo you wanted to invert so you removed `-textonly.svg` part?

Myshor (changes requested):
More review when you share examlpe URLs.

However two changes already found problematic.

adamhotep:
> Can you share the example URLs?
> 
> As you make some more changes than one small then I would like to check how the fixes work live and how they will work when light scheme will be selected in Dark Reader.

As noted in my commit log, https://en.wiktionary.org/wiki/satchel is one example. The monobook changes are for literally any article page as viewed with the Monobook theme. IIRC, there's a way to call it by GET parameter. I'll get that for you next week (I'm away right now).

The other points above are quite valid and I'll take another look next week or so.

Myshor:
Does this monobook skin is appliet through some settings / properties after loggin in? I still can not see it anywhere.

src/config/dynamic-theme-fixes.config
adamhotep: I was matching the inverted logo at the top left of the MonoBook view. I take it that was *not* intended, which is fine—but means the MonoBook skin has a bug in this regard.

Here is how it looks for me with this merge request (which brings back the book pages in the background):
image

With Dark Reader disabled, it looks like this:
image

src/config/dynamic-theme-fixes.config
adamhotep: that works for me

adamhotep:
> Does this monobook skin is appliet through some settings / properties after loggin in? I still can not see it anywhere.

The MonoBook skin can be found at https://www.mediawiki.org/wiki/Skin:MonoBook and can be previewed at https://en.wikipedia.org/w/index.php?useskin=monobook or https://www.mediawiki.org/w/index.php?title=Skin:MonoBook&useskin=monobook (change the `title` parameter to preview a different page).

I believe I have addressed your other concerns. Note that the MonoBook-related bug referenced above is likely nontrivial since it's pulled in as a `background-image` to an `<a>` element and the text below the logo is black, so preventing the inversion is also problematic:

image

Myshor:
Ok... You gave some examples which doesn't move me forward.

I am still not sure how I can switch and use than skin... which makes that much problems. 🙄

adamhotep:
I provided links which you can use to preview any skin. If you actually want to set up your account to use the MonoBook skin, the instructions at [Wikipedia:Skin](https://en.wikipedia.org/wiki/Wikipedia:Skin) say:

> Once you have an [created an account](https://en.wikipedia.org/wiki/Wikipedia:WCAA) and logged in, go to [Preferences](https://en.wikipedia.org/wiki/Special:Preferences) → [Appearance](https://en.wikipedia.org/wiki/Special:Preferences#mw-prefsection-rendering) → [Skin](https://en.wikipedia.org/wiki/Special:Preferences#mw-prefsection-rendering-skin). In the default Vector 2022 skin, the Preferences page can be accessed at any time from the 👤 🮦 Personal settings dropdown in the top right corner and selecting ⚙️ Preferences. Choose your skin and click Save. Then, all pages will be loaded with the new selected skin.

Myshor:
Ok... now I see the two examples uses skin already.

src/config/dynamic-theme-fixes.config
```diff
@@ -18671,6 +18675,14 @@ html[data-darkreader-scheme="dark"] img[src*="Eagle_with_fasces.svg"] {
     background-color: ${#BBBBBB} !important;
     mix-blend-mode: screen !important;
 }
+body.mediawiki.skin-monobook:has(> #globalWrapper) {
+    background-color: #f9f9f9 !important;
+    background-image: linear-gradient(#0000, #0000), url(https://upload.wikimedia.org/wikipedia/commons/f/fb/Monobook_skin_background.svg) !important;
+    filter: invert();
```
Myshor, line 18681: As Dark Reader can be used in dark or light scheme filtering like this will force dark in both.

For better handling light and dark scheme I would remove invert from here.\
For solution check my next comment.

src/config/dynamic-theme-fixes.config
```diff
@@ -18671,6 +18675,14 @@ html[data-darkreader-scheme="dark"] img[src*="Eagle_with_fasces.svg"] {
...
+}
+body.mediawiki.skin-monobook > * {
```
Myshor, line 18683: As we remove the invert from previous selector we need to make some changes here so
We change this selector to:
```
html[data-darkreader-scheme="dark"] body.mediawiki.skin-monobook > *,
html[data-darkreader-scheme="dark"] body.mediawiki.skin-monobook:has(> #globalWrapper) {
```

This will apply invert only when dark scheme is applied.

src/config/dynamic-theme-fixes.config
```diff
@@ -18655,6 +18656,9 @@ div.flaggedrevs_short,
 table.layouttemplate {
     border: var(--border-color-subtle) 2px solid !important;
 }
+.mw-parser-output .interproject-box {
+    background-color: ${Canvas} !important;
```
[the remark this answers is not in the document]
Myshor: For me this is to remove.

Myshor (changes requested):
Checked again the pages with monobook skin and requested some changes.

P.S. I could push them by myself and merge but I leave it so you can take a look and expend some CSS experience in Dark Reader sitefixes. 😉
P.P.S. If you just want to have fixes merged as fast as possible you can ask me to push requested changes. I know sometimes people can take a look on the comment but doesn't have much time to sit and make code changes. 😅

PR #13912, opened 2025-02-16
Fix for wiki

Tiagoquix:
Fixes infobox in https://en.wikipedia.org/wiki/President_of_India

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -18555,6 +18555,7 @@ img[src*="Sampad.svg"]
 .mw-mmv-restriction-insignia
 img[src*="Yocto_Project_logo.svg"]
 img[src*="The_Pirate_Bay_logo.svg"]
+img[src*="President_of_India_Logo.png"]
 
 CSS
 :root {


imsys:
This kind of fix can be done in Wikipedia itself. just use "class=skin-invert-image"
like:
 [[File:President of India Logo.png|class=skin-invert-image|300px]]

If you don't want to fix this yourself, you can report it in the Appearance bar, on "Report an issue with dark mode", that brings you to: https://www.mediawiki.org/wiki/Reading/Web/Accessibility_for_reading/Reporting/en.wikipedia.org

Anyway, I fixed this for you.

Myshor:
I wanted to say what is needed to be fixed as it shows inverted on the page but that `.skin-invert-image` fix on wiki side made that.

@Tiagoquix 
Ok I will merge it as it will invert pic on other related sites too. However maybe it would be better to start reporting images to wiki like imsys wrote.
That would exclude adding new svg files to the fixes. 🙄

Myshor:
For now I try to make the selectors not breaking others... As there was already some problem with `.skin-invert-image` double invert... now it's fixed so even if there is selector for this it will select only image which can be included in `img[src*"..."]` selector. And when Dark Reader's INVERT part catches exactly the same element from more than one selector it just aplies one INVERT. 😁

Still I was already writing that I add this INVERTs but if there would be no mobile Firefox with Dark Reader then I would not as on desktop you can always do the hover to add white background. At least it was white background to transparent pictures because now we have it white if it's not inverted or black if INVERT applies.

Until Alexander Shutau is not complaininmg about wikipedia fixes size we can still do the INVERTs. 😉
It's still not that much lines compared to the rest wikipedia CSS fixes. 😅

PR #13914, opened 2025-02-17
Add remaining Tracker Network domains

lauriys:
Many of our sites are already on the list (e.g. `tracker.gg`, `destinytracker.com`, `fortnitetracker.com`, etc.), but some of them are still missing (e.g. `apex.tracker.gg`, `r6.tracker.network`, and others).

`tracker.network` does not function on its own, it's just a redirect to `tracker.gg`, which is why only the wildcard is added as a catch-all.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -15,6 +15,8 @@
 *.starbreeze.com
 *.suyu.dev
 *.tickettool.xyz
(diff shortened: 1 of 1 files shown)


lauriys:
yeah that's what i meant, just didn't phrase it very well

`rocketleague.tracker.network` is also another example

PR #13925, opened 2025-02-17
fixed deepseek.com dark mode

Leaf45:
issue with background image and changed corresponding elements on the page to match new colors

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -7804,6 +7804,19 @@ body {
 
 ================================
 
+deepseek.com
+
+CSS
+.bg-cover,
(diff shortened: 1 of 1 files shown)


Leaf45:
if someone could, please lmk how to specify the difference between changes for dark mode and light mode

Myshor:
> if someone could, please lmk how to specify the difference between changes for dark mode and light mode

I will push the fix for that so you will be able to see how we use colors to light/dark scheming. 😉

Leaf45:
@Myshor would you be able to provide me a discord account so we can chat? I would like to ask some more in-depth questions.

Myshor:
Just search for myshor 😅
obraz

PR #13939, opened 2025-02-20
Fix for theuselessweb.com

papajojoes:
Fixed the colors/some bottom info text to correctly fit between light and dark mode, images provided
image
image


Myshor:
- There is already other marquette.edu fix pushed.
- I do not see any point to force all colors on the theuselessweb.com to white and dark and removing the backgrounds. The page looks fine to me without this fixes to me. Sitefixes should be used only when necessary when something is not readable and/or broken.

Summarizing. I am rejecting this PR.

PR #13940, opened 2025-02-21
Fix embedded image FDIC-Insured text visibility on AmericanExpress.com

sergio-sketch:
-Ensure the FDIC-Insured text remains visible in Dark Mode
-Targets embedded image while keeping other elements unchanged and matching color with the rest of the page.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -201,6 +201,10 @@ select[aria-labelledby="bankAccount-label"] > option {
     background-color: var(--darkreader-neutral-background) !important;
 }
 
+img[alt*="FDIC-Insured"] {
+    filter: invert(1) hue-rotate(180deg) brightness(0.75) contrast(1.7) !important;
+}
+
 ================================
 
 *.archlinux.org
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -201,6 +201,10 @@ select[aria-labelledby="bankAccount-label"] > option {
     background-color: var(--darkreader-neutral-background) !important;
 }
 
+img[alt*="FDIC-Insured"] {
```
Myshor, line 204: Dark Reader can be used with dark (default) scheme or light. Forcing invert like that will break light scheming.
obraz

To fix this remove all `img[alt*="FDIC-Insured"]` related CSS part and add INVERT Section before CSS so the whole fix will look like this:
```
*.americanexpress.com

INVERT
img[alt*="FDIC-Insured"]

CSS
select[aria-labelledby="bankAccount-label"] > option {
}

```

That will INVERT image only in dark scheme and leave it untouched in light. 😉

Myshor (changes requested):

src/config/dynamic-theme-fixes.config
sergio-sketch: Hello,

Thank you so much for taking the time to reply it was very helpful to learn! I have just corrected this with the INVERT. Thanks again! :)

Myshor (approved):

PR #13941, opened 2025-02-21
fix for baike.baidu.com

Yersmagit:
fix for baike.baidu.com

|Before|After|
|-|-|
|image|image|


Myshor:
Can you share some example URLs to check the fixes?

I checked https://baike.baidu.com/ without any login in as I do not have any account there and I saw this:
- without your fix:
obraz

- with your fix:
obraz

And looking into your screenshot:
obraz

I do not understand why invert for this part (as I do not see anything more fixes except already catched logo) needs 52 lines of sitefixes.

PR #13959, opened 2025-02-25
Fix for exercism.org

Adrien-LUDWIG:
Fix unsupported rgba of hex color var causing box-shadow to disappear

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -10400,6 +10400,12 @@ header.c-header-with-bg {
 .theme-light {
     --darkreader-bg--page-dashboard-background-image: none;
 }
+.c-ed.--available:not(.disabled) {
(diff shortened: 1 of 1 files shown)


Myshor:
Can you share some URL to test it before merge? I just want to be sure the fix is working fine in both kind of browsers - Chromium based and Firefox.

Adrien-LUDWIG:
Thank you very much for the tests and fix!
I only checked [caniuse.com](https://caniuse.com/mdn-css_types_color_color-mix) because I don't usually a Chromium based browser. I'll take this into next time. :wink:

Adrien-LUDWIG:
I discovered some more things to fix so I made this a draft PR. I'll resume it once I'm finished.

Adrien-LUDWIG:
Oops, looks like I forgot to mark the PR as ready!

Myshor:
Added one more selector to INVERTs and merging. 😉

Adrien-LUDWIG:
Thanks again for your help @Myshor! :wink:

PR #13978, opened 2025-03-01
fix for bilibili.com

rintrint:
Fix bili-text-button element in dark mode

Previously fixed in #12937 by me, but the issue has resurfaced after recent site updates. 
This PR:
- Removes the previous fix that no longer works
- Adds a new solution using CSS custom properties
- Sets `--darkreader-bg--_container-color: rgba(0, 0, 0, 0)` to make button text visible
- Updates other related variables to maintain consistent dark theme appearance

Tested on bilibili.com with latest site version as of 20250301.

# Before：
Snipaste_2025-03-01_20-35-18

# After：
Snipaste_2025-03-01_20-36-05

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -32952,11 +32952,13 @@ CSS
 .qrcode {
     background: white !important;
 }
-.button::before {
-    background-color: var(--_container-color) !important;
-}
 body {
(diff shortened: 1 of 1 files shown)


C-Zking:
Very useful, thank you

Zhou-Jianhao:
Nice! I have tried many changes on the previous `.button::before`, but neither of them worked......
Thank you for fixing this!

PR #13997, opened 2025-03-04

ForeverFloating:
Added fixes for origene.com

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -22116,6 +22116,24 @@ INVERT
 
 ================================
 
+origene.com
+
+CSS
+.block .title *,
+.items > li,
+.nav-item *:not(:hover),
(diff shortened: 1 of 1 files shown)


Myshor:
Can you share some specific URL with examples? Because when I saw main page then you just made some visible text to more light. As it was not some critical fix needed then I am considering if this fixes should be merged into the code.

ForeverFloating:
No, the fixes are purely for the text. The text was too dim for me and so I changed the text color to respect the user's settings for the font color (in my case, white). Here's the [page](https://www.origene.com/catalog/cdna-clones/expression-plasmids/sc320242-grp75-hspa9-nm-004134-human-untagged-clone) I was referencing when creating the CSS rules.

Myshor:
OK... It's not breaking anything and tht text is more "distinct". I will merge. 😉

PR #14014, opened 2025-03-07
Add itch.io/embed-upload/ to dark sites list

Tiagoquix:
`itch.io/embed-upload/` is a common subpage used by games that can be played in the browser via the website. Adding it to the list so Dark Reader doesn't interfere with those games and potentially reduce performance by trying to inject dark themes.

It's games, so Dark Reader should not touch them.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -594,6 +594,7 @@ intothetrenches.1917.movie
 iskdeusingqt6.org
 isthereanydeal.com
 isxander.dev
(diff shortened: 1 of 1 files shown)


Myshor:
Hard to confirm it should/could be added to Dark Sites

When I try to open the link I end with:
obraz

And as the whole page is normally white then I can not merge any entry based on itch.io before checking if it's really dark and not changing to light when light mode is set into system/browser.

PR #14018, opened 2025-03-07
added fix for not updating background image on *.pycheung.com/checker/

The-Doggo:


src/config/dynamic-theme-fixes.config
```diff
@@ -23954,6 +23954,16 @@ INVERT
 
 ================================
 
+*.pycheung.com/checker/
```
Myshor, line 23957: remove `*.` as it won't pass the tests for alphabetical order in fixes.

Do you have some subpages that uses prefix before pycheung.com so it needs to have this asterisk here?

src/config/dynamic-theme-fixes.config
```diff
@@ -23954,6 +23954,16 @@ INVERT
...
+
+IGNORE INLINE STYLE
```
Myshor, line 23959: Remove whole IGNORE INLINE STYLE

This is only used for not using Dark Reader INLINE FILL for element. As you fix the invert for background image the next section is needed only.

Myshor (changes requested):

src/config/dynamic-theme-fixes.config
The-Doggo: no, i just copie pasted the line selector from another existing selector

src/config/dynamic-theme-fixes.config
The-Doggo: It appers to not work if I remove the line.

With `IGNORE INLINE STYLE`:
image
Without `IGNORE INLINE STYLE`:
image

Myshor: OK

Myshor (approved):

PR #14021, opened 2025-03-07
Update dark-sites.config

eki-github:
i want too add my site do global blacklist because its already black


Myshor:
Not so nice with that background picture... I mean can be eye painfull but still when Dark Reader is force applied that background is shown too so...

You will have your site in dark sites. 🙂

PR #14049, opened 2025-03-11
fix for jsr.io

MAKS11060:
- fix code block
---
before:
image
after:
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -15716,6 +15716,27 @@ INVERT
 
 ================================
 
+jsr.io
+
+CSS
+.pl-smi {
+    color: white !important;
+}
+.pl-s {
+    color: #a5d6ff !important;
+}
+.pl-c1 {
+    color: #2f7bd0 !important;
+}
+.pl-en {
+    color: #a668ff !important;
+}
+.ddoc table td code {
+    background-color: #252b33 !important;
+}
+
+================================
+
 jsware.net
 
 INVERT
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -15720,19 +15720,19 @@ jsr.io
 
 CSS
 .pl-smi {
-    color: white !important;
+    color: ${black} !important;
 }
 .pl-s {
-    color: #a5d6ff !important;
+    color: ${#00295a} !important;
 }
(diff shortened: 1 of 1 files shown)


Myshor:
This is nice if we work with dark scheme only:
obraz

But when we switch to light scheme in Dark Reader we will end with:
obraz

so we need to use ${primary_color} instead of already inverted ones in sitefix.

Will try to fix them fast for you.

Myshor:
After my fix the dark scheming is like before and light:
obraz

Myshor:
Give me the URL with the code I could check all selectors.... As I do not want to push the fixes with checking everything in light/dark.

Myshor:
I changed static color to dynamic related to light/dark scheme. I was able to check how they work for every selector except `.ddoc table td code`. So if it will break anything with my change come back here with URL where it is used and I will try to fix it but I hope it will be ok.

Merging. 😉

MAKS11060:
I do not know how to write a convenient selector for this case.

https://jsr.io/@dldc/chemin#api and https://jsr.io/@maks11060/crypto#key-features
The text is inside the `<code>` element
image


This fix only works for tables. https://jsr.io/@maks11060/crypto#algorithms
The text inside the `<code>` block but inside the table, this is corrected using
```css
.ddoc table td code {
    background-color: ${#ccd4da} !important;
}
```
image

Myshor:
OK... Now I see what it is and what can be done. :) I will push one more fix (the one I used is nice and can stay for table but I will add selectors for the code pills in text too.

MAKS11060:
@Myshor 
- I found the most universal selector for `code` (fixes https://jsr.io/@maks11060/discord-interactions).
- Improved the color of the link inside the `code`.
```css
.ddoc :not(.highlight) > code {
    background-color: ${#ccd4da} !important;
}
.ddoc a > code {
    color: ${#205faa} !important;
}
#usageSelector > summary {
}
```
The last noticed error remained in the 'docs' tab https://jsr.io/@dldc/chemin@13.0.0/doc

Myshor:
In the last mentioned palce the CSS used originally for site includes `!important;` that's why we can not override the colors but I was managed to make similar effect adding additional effects into that elements:
```
.markdown_summary p code {
    color: ${#222222} !important;
    opacity: 0.85 !important;
}
```

I will push your changes and mine in a minute. 😉

Issue #14052, opened 2025-03-12
[Broken Website] Open-Meteo.com 1) has non-default dark mode 2) DarkReader breaks graph


insipidlight:
### Prerequisites

- [x] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this website issue to avoid opening a duplicate.
- [x] I can reproduce this website issue in a new, unmodified web browser profile with Dark Reader installed as the only extension.

### Website Issue Description

Included URL is a sample request for London

Dark Reader does not detect site theme; using Dark Reader with website-configured light/dark themes breaks graph

Graph is only usable with Dark Reader disabled

### Website Address

https://open-meteo.com/en/docs?latitude=51.5085&longitude=-0.1257&hourly=temperature_2m,relative_humidity_2m,dew_point_2m,apparent_temperature,precipitation_probability,precipitation

### Steps To Reproduce

1. Dark reader on default, fresh installation
2. Go to URL / select check boxes to create API request
3. scroll to graph
4. disable Dark Reader, website observes browser/system setting or by button on navbar, graph works correctly

### Screenshots

![Image](https://github.com/user-attachments/assets/71ddc460-eea0-46b6-a09a-2c60d9ee38f5)

### Operating System

Windows 11

### Web Browser name and version

Firefox 136.0

### Dark Reader version

Version 4.9.103

Myshor:
I made some fix to your PR and will merge it in a while but I will make one more PR to detector-hints.config as the page nicely adds `.dark` class when applies own dark mode.

Myshor:
Sitefix added.

- If you want to have this fix faster you can set **Synchronizee sites fixes** option to check in Advanced section of Settings.
![obraz](https://github.com/user-attachments/assets/9dea8b66-b77a-46ab-8aa6-e50d995718c3)
- With that enabled newest sitefixes will be updated every browser start. 😉
- To open all settings you must click on:
(Old/classic design) More -> All Settings
![obraz](https://github.com/user-attachments/assets/8c17b1f1-e85e-4672-8a88-85930e178e02)
(New/Preview design) Settings in popup
![obraz](https://github.com/user-attachments/assets/e4f586f8-8ce4-47bc-90d7-7451eb030439)

Myshor:
Additionally detection hint is added too... We can say we fixed everything what was needed to be fixed. 😁

PR #14053, opened 2025-03-12
Added open-meteo.com

insipidlight:
https://github.com/darkreader/darkreader/issues/14052

largely fixed entire site in dark/light mode regardless of website setting, several inputs/buttons not completely fixed, and header background image remains removed by DarkReader


Myshor:
Added two selectors to fix buttons color:
obraz
obraz

PR #14065, opened 2025-03-15
fix for jsr.io

Myshor:
Additional fixes to https://github.com/darkreader/darkreader/pull/14049

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -15732,11 +15732,19 @@ CSS
 .pl-en {
     color: ${#009759} !important;
 }
-.ddoc table td code,
-p code,
-ul code,
-h2 code {
+.ddoc :not(.highlight) > code,
+ {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -15732,12 +15732,20 @@ CSS
 .pl-en {
     color: ${#009759} !important;
 }
-.ddoc table td code,
-p code,
-ul code,
-h2 code {
+.ddoc :not(.highlight) > code,
```
MAKS11060, line 15735: Bug, you need to remove the comma.

Myshor:
Yea... and this makes the code for `.markdown_summary p code` need to fix too.

One more fix in a minute... However the page https://jsr.io/@dldc/chemin@13.0.0/doc will have a little bit more gray text in the code 'cause the opacity must be changed to 0.5 to have similar "pills" background color.

PR #14067, opened 2025-03-15
Fix for theguardian.com

pnaf:
- Fixes graphics on "Video" section on home page by using site-specific colors (supercedes fixes in #12786).
- Improves visibility of star ratings on "Reviews" page and article pages by using site-specific colors.
- Clarify several rules to allow for easier identification and sorting.

Before:

After:


Myshor:
I merged however I see that it uses many generated class names. Would be nice to take a look on that sitefix and try to use some selectors less related to classes with generated names.

Now as you even mentioned there is a need to make updates to the fixes.

pnaf:
Thanks for pointing that out. Sometimes it can be a bit tricky to rely less on generated class names, but I can definitely look into improving those fixes in subsequent updates. :slightly_smiling_face:

PR #14070, opened 2025-03-17
fixes for amys.com

StevenSchifter:


Myshor:
Checked your fix and it works fine when dark scheme is used in Dark Reader but leaves dark backgrounds when switched to light ones.

I saw you forced background images which I will remove too.

And last but not least... We do not use `data:image/` in sitefixes.

I will make some edit to your PR and merge. 😉

StevenSchifter:
Thanks for fixing my fixes. 🙂 Before today I was not familiar with `background-blend-mode`. Next time I come across background images that don't darken as expected I will try using that instead of forcing background images or using data:image/.

PR #14071, opened 2025-03-17
Update dark-sites.config

EyBmr:
Added "hadesblack.com" and its subdomains.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -536,6 +536,8 @@ hackthebox.com
 hackthebox.eu
 hackthissite.org
 hacktoberfest-projects.vercel.app
+hadesblack.com
+*.hadesblack.com
 halotracker.com
 halowaypoint.com
 hang.fm


src/config/dark-sites.config
```diff
@@ -536,6 +536,8 @@ hackthebox.com
 hackthebox.eu
 hackthissite.org
 hacktoberfest-projects.vercel.app
+hadesblack.com
```
Myshor, line 539: This entry is Ok.

src/config/dark-sites.config
```diff
@@ -536,6 +536,8 @@ hackthebox.com
...
+*.hadesblack.com
```
Myshor, line 540: This entry is not ok as entries must be in alphabetical order.

With this order tests will not pass.

2nd thing. Why to use asterisk (*) when I think there are only 3 pages to add:

1. about.hadesblack.com
2. privacy.hadesblack.com
3. contact.hadesblack.com

Just remeber when adding this three subdomains - they are needed to put in the list alphabetically too!

P.S. I could make edit in your PR but sometimes it's better to leave this to author so you will learn something new how the list is working in project. 😉

Myshor (changes requested):

src/config/dark-sites.config
EyBmr: As you wish, Milord.

I just thought that since the website has a general dark theme it would be easier to just wildcard any potential future subdomains to it to prevent Dark Reader interfering with subdomains that may exist in the future but haven't been added to this list.

I was just trying to keep things simple here, but if you insist, I will bow.

EyBmr:
Closed to satisfy @Myshor's requests.

src/config/dark-sites.config
EyBmr: As you wish, Milord.

I thought that since the website has a general dark theme it would be easier to just wildcard any potential future subdomain to it to prevent Dark Reader interfering with future subdomains that haven't been added to this list. Not sure if there are more subdomains.

For some reason I thought alphabetical order meant the domains and not the subdomains. But since all wildcards are at the top of the .config file, it should've made sense to me.

PR #14084, opened 2025-03-20
Fixes for zerohedge.com

wigbam:
Inverts images in the header.

Before:
zerohedge_before

After:
zerohedge_after

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -35274,7 +35274,7 @@ zerohedge.com
 *.zerohedge.com
 
 INVERT
-* header > * img
+header img
 
 ================================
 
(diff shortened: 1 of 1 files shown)


oberrich:
Can reproduce locally, lgtm!

@Myshor *kindly pinging.*

Myshor:
Not sure why to use `* header > * img`
If we want to use `header` anywhere in the html code and invert img inside this `header` (again it can be deep inside this header) then it's easier to use and easier to read `header img`.

I will edit PR and merge. 😉

wigbam:
Thanks @Myshor for the merge and for putting my primitive CSS skills on blast for the whole world to see! 🤣

Myshor:
Everybody was doing something first time. The feedback is not to blast something to whole world but to give the tip how you can make it shorter and easier to check in the future. 😉

Better to push some PR that will be reviewed and corrected than leave some almost invisible things on site.

Issue #14095, opened 2025-03-22
[Broken Website] klibs.io: Hard to read text


Bartuzen:
### Prerequisites

- [x] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this website issue to avoid opening a duplicate.
- [x] I can reproduce this website issue in a new, unmodified web browser profile with Dark Reader installed as the only extension.

### Website Issue Description

Some texts are hard to read, and some are even barely visible.

### Website Address

https://klibs.io

### Steps To Reproduce

Open the website to see cards with texts on them that are hard to read.
Select a library to see barely visible title, description, tabs, and metadata labels.

### Screenshots

![Image](https://github.com/user-attachments/assets/73924792-3fff-42c2-aa8c-f56f49895655)
![Image](https://github.com/user-attachments/assets/27257277-331f-4354-8612-ea72b27dfafd)

### Operating System

Windows 11

### Web Browser name and version

Firefox 136.0.2

### Dark Reader version

4.9.103

Bartuzen:
Thanks for your work! However, the big library title is still the wrong color.

![Image](https://github.com/user-attachments/assets/98e7f360-961a-4eb4-b73a-3770902ed02d)

PR #14099, opened 2025-03-24
fix for klibs.io

Myshor:


Myshor:
This was fix for #14095

PR #14110, opened 2025-03-27
update my site URLs

tycrek:
remove old domains (no longer resolve), add active domain


Myshor:
I will merge it but there is one place which will not be dark and stay white when system/browser is set to light theme.

https://sylviefox.ca/info/compose/
obraz

Just check if you can force Dark Theme there somehow. I know it's subpage but still. 😉

PR #14111, opened 2025-03-28
Fix for www.gamer.com.tw

lumynou5:
This fixes the new home page of the website.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -33662,6 +33662,21 @@ CSS
 
 ================================
 
+www.gamer.com.tw
+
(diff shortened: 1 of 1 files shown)


Myshor:
As we have already fixes for `*.gamer.com.tw` and `home.gamer.com.tw` so I will merge all fixes including yours into one `*.gamer.com.tw`

lumynou5:
@Myshor Will this affect the performance because it tries to insert useless CSS?

Myshor:
Still will be better than going through 3 different fixes to check if something should be injected or not because some pages could be used with different fixes. Even your proposed `www.` would be shared with `*.`

PR #14117, opened 2025-03-29
Uninvert unsplash logo

barraIhsan:
Ok, follow up to my old pr #12188 for some reason it's now looking like this?
image

And removing the invert line will fixes this
image

I don't even know svg can revert on its own without the `prefers-color-scheme` media query.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -30647,9 +30647,6 @@ textarea {
 
 unsplash.com
 
-INVERT
-svg[aria-labelledby="unsplash-home"]
-
 CSS
 li > a[aria-current="page"] {
     box-shadow: ${rgb(25,25,25)} 0px -2px inset;


Myshor:
I checked and now Dark Reader uses text color to fill this logo that's why it's light in Dark and dark in Light scheme. 😉

Merging uninvert.

barraIhsan:
> I checked and now Dark Reader uses text color to fill this logo that's why it's light in Dark and dark in Light scheme. 😉
> 
> Merging uninvert.

text color? so it will only work with the `fill="currentColor"` one?

Myshor:
I mean some style was changed on site and now this logo uses this color to fill:
(Example screen without Dark Reader applied):
obraz

And when Dark Reader is applied it uses it's own variable which is light for dark mode and dark for light:
obraz

That's why the invert was not needed anymore in the sitefixes. 😉

barraIhsan:
ahh ok

PR #14119, opened 2025-03-29
Added dignityofwar.com to darksites

Maelstromeous:
https://dignityofwar.com
https://wiki.dignityofwar.com

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -319,6 +319,8 @@ dicebear.com
 diep.io
 digi77.com
 digikam.org
+dignityofwar.com
+wiki.dignityofwar.com
 disasm.pro
 disboard.org
 discadia.com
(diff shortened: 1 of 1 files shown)


src/config/dark-sites.config
```diff
@@ -319,6 +319,7 @@ dicebear.com
 diep.io
 digi77.com
 digikam.org
+dignityofwar.com
```
Myshor, line 322: This part is definitely not DARK.
obraz

I can accept only wiki.

Myshor (changes requested):

src/config/dark-sites.config
Maelstromeous: Whoops, you're entirely right sorry!

Myshor (approved):

PR #14130, opened 2025-04-01
Update for Aliexpress RU

TheMiner203:
The website has been updated, now the class is called differently


Myshor:
Can you share some URL where I could find that?

On main page I can not see new one nor old one.

Myshor:
OK... Nevermind... I think when I look on the name it is some page with already payer order so it even won't be possible to share.

PR #14137, opened 2025-04-02
fix for some buttons in arch wiki

nodscher:
mainly affects hamburger menus, expand/ collapse buttons and the button bar in the top right

(this fix also applies to other media wiki sites (like community.kde.org), but I have only applied it to the arch wiki for now)


Myshor:
Need to make a small change as this fix breaks icon on mobile:
obraz

After my fix it will be:
obraz

PR #14140, opened 2025-04-04
fix openstreetmap background

GalaxySH:
- [x] tested in firefox
- [x] tested in chrome


GalaxySH:
Failed testing in Chrome. Apparently this breaks the background color, but I can't make sense of it. The color is computed correctly, but shows up inverted? Not what happens in Firefox.

|Chrome|Firefox|
|--------|--------|
| image | image |

PR #14155, opened 2025-04-08
Added fix for https://cursoreffects.com/

Leaf45:
changed some color values

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -7684,6 +7684,19 @@ IGNORE IMAGE ANALYSIS
 
 ================================
 
+cursoreffects.com
+
+CSS
+.box {
+    background-color: #808080;
+    border-bottom-color: var(--darkreader-border-000000, #FF1694);
+    border-left-color: var(--darkreader-border-000000, #728bd4);
+    border-right-color: var(--darkreader-border-000000, #728bd4);
+    border-top-color: var(--darkreader-border-000000, #FF1694);
+}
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -7684,6 +7684,19 @@ IGNORE IMAGE ANALYSIS
 
 ================================
 
+cursoreffects.com
+
+CSS
+.box {
+    background-color: #808080;
```
alexanderby, line 7691: Unfortunately this fix may be unstable. Could you please use the following pattern:
```css
.box {
    background-color: ${#FFFFFF} !important;
}
```
where `#FFFFFF` should be a color value without any changes (or close to it).

src/config/dynamic-theme-fixes.config
```diff
@@ -7684,6 +7684,19 @@ IGNORE IMAGE ANALYSIS
...
+    border-bottom-color: var(--darkreader-border-000000, #FF1694);
```
alexanderby, line 7692: If the value should be strictly `#FF1694`, please use just `#FF1694 !important` without referencing Dark Reader internal variables. `--darkreader-border-000000` refers to a color, that used to be black initially.

alexanderby (changes requested):

Myshor:
Checked the site what it does and what the fixes do... In the end I pushed the fixes related to background and removed all entries related to borders.

Myshor (approved):

PR #14168, opened 2025-04-12
Fix for ecco website

ForeverFloating:
Fixes for the colors in the "color" filter when browsing products. I couldn't get `IGNORE INLINE STYLES` to work with the selector `button[role="menuitem"] div:empty`, so I manually replaced the classes with the colors originally on the website without Dark Reader.


Myshor:
As there is error in the PR that will not accept tests then:
Can you share some URL and screens what exactly you are changing? It's not easy to find that manually.

PR #14191, opened 2025-04-17
Fixes for bahai.org

StevenSchifter:
These fixes make the header background dark while keeping text and icons visible. I also added a fix for footnotes, which appear when numbers in brackets are clicked on a page [such as this one](https://www.bahai.org/documents/nsa-usa/two-wings-bird).


Myshor:
I am adding one more selector into CSS fixes for:
obraz

Then I will merge.

PR #14195, opened 2025-04-19
dynamic fix: shapecatcher.com  INVERT img

effctcodes:
compare @ https://shapecatcher.com/unicode/block/Basic_Latin

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -26761,6 +26761,13 @@ INVERT
 
 ================================
 
+shapecatcher.com
+
+INVERT
+img
+
+================================
+
 share.dmhy.org
 
 CSS


rich-ayr:
**before** 
image

**after**
image

Logo green now.

```css
shapecatcher.com

INVERT
img[src^="https://shapecatcher.com/svg/"]

CSS
img[src*="header.png"] {
    background-color: ${#aaaaaa} !important;
}

```

maybe something like this? not sure we can perfectly fix it.

image

rich-ayr:
Your CSS turns the duck green. Maybe something like this could work as a compromise?

```css
shapecatcher.com

INVERT
img[src^="https://shapecatcher.com/svg/"]

CSS
img[src*="header.png"] {
    background-color: ${#aaaaaa} !important;
}

```

image

Myshor:
@rich-ayr 
I will push your fix but with one more fix to not include any background for the logo when light scheme is used in Dark Reader.

@effctcodes The `${...}` just invert(100%) and hue-rotate(180deg) the color inside the brackets. So the original `...` is used when Dark Reader is used to create light scheme and INVERTED to create dark.

However I will use one more html tag to set dark color to the dark mode only. 🙂

PR #14214, opened 2025-04-26
Fix for internationalrightsadvocates.org

pnaf:
- Fixes logo on most pages.
- Fixes bright background on home page.

Before:

After:


Myshor:
Always check how the page will look like when switched to light scheming after doing the fixes.
I needed to add text color fix because some of text was not readable or even not visible at all.

PR #14236, opened 2025-05-01
Add `project2025.observer` to dark-sites.config

cclypeatus:


Myshor:
Can not add it into Dark sites as it can be light and it has button to switch styles:
obraz

But you can make Detection hint here: https://github.com/darkreader/darkreader/blob/main/src/config/detector-hints.config
using this code:
```
project2025.observer

TARGET
html

MATCH
.dark

```

PR #14237, opened 2025-05-02
Update dark-sites.config

WeiChongDevelops:
✅ Dark mode implemented via button in navbar
✅ Placed in alphabetically ordered location in dark-sites.config

Thank you :)


Myshor:
> Dark mode implemented via button in navbar

That's why I can not add it into Dark Sites. 

But you can make Detection hint here: https://github.com/darkreader/darkreader/blob/main/src/config/detector-hints.config
using this code:
```
getdiverge.com

TARGET
html

MATCH
.dark

```

PR #14240, opened 2025-05-03
fix for objshadow.is-a.dev (and mirrors)

ShadowObj:
This is my personal blog website: objshadow.pages.dev
(`objshadow.pages.dev`, `objshadow.tun.webredirect.org`, `shadow.tun.webredirect.org`, `shadowobj.eu.org` are mirrors for this website)
I've already designed dark theme for my website,
and Dark Reader **CANNOT** detect the dark theme,
Even if I added `color-scheme="dark"` to the tag.
Besides, Dark Reader makes the dark theme of my website (especially the colors) looks strange,
so I hope that Dark Reader can be disabled on my website.

Best wishes,
Object Shadow


src/config/detector-hints.config
```diff
@@ -573,6 +585,17 @@ NO DARK THEME
 
 ================================
 
+shadow.tun.webredirect.org
+shadowobj.eu.org
```
Myshor, line 589: I could merge it as it is but as this is the same page but with different URL address then I would add this two URLS under 
`objshadow.tun.webredirect.org` alias and then remove this section as it is duplicate to the earlier one.

Myshor (changes requested):
Make changes or give me the nice reason why to not merge two sections into one.

Myshor (approved):

PR #14249, opened 2025-05-05
Adding my dark themed site under dark-sites.config

TheSigmaArts:
Added my own site due to it already being dark themed.


Myshor:
Can not add it as it doesn't meet 3rd requirement:
obraz

I see the site is still not finished so I can not add it.

PR #14262, opened 2025-05-08
Fix weird blue color on openstreetmap.org

TesseractPi:
Blue color for icons on OpenStreetMap looks off


Myshor:
I use detection of Dark Theme option and I didn't catch that. 😁

Checked how it will look like with light and dark scheming using in Dark Reader and I think it's ok to make that fix.
Merging.

PR #14268, opened 2025-05-10
Fix bgcolor of <button> in cart on instacart.com

ron-wolf:
The text of `<button>` elements in the cart, mainly cart items and <kbd>✏ Add instructions</kbd>, has gray text on a white background and is not legible. This fix uses CSS to override the style.

The reason I don't just add an INVERT section is that the text color is correct, it's just the background that needs resetting to a darker color.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -15175,8 +15175,8 @@ svg text {
 instacart.com
 
 CSS
-.e-im51v0 {
-    background-color: unset !important;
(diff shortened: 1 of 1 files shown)


Myshor:
This won't fix anything as the class name is generated. At least today nothing was fixed with this code so...

Instead of forcing background to some buttons I will push fix which will make darker text on the buttons when Dark scheme in Dark Reader is used.

This made all buttons on main page readable so it should make buttons in cart more visible too. 😉

ron-wolf:
Ok! Will test when I get around to it to see if it looks alright.

PR #14272, opened 2025-05-10
Bug in modify-css.ts

dcog989:
duplicate thumb - should be track?

### DIFF
--- a/src/inject/dynamic-theme/modify-css.ts
+++ b/src/inject/dynamic-theme/modify-css.ts
@@ -661,7 +661,7 @@ export function getScrollbarColorModifier(value: string): string | CSSValueModif
         return null;
     }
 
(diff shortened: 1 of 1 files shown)


Myshor:
I saw a bug earlier.

Can you show some screens what this will fix? Or else it will need to wait for @alexanderby to review.

dcog989:
It requires review by someone who understands the *intent* of the code. 

New track color is derived from original thumb color, not original track color. This seems wrong - hence the PR.

alexanderby:
Thanks for noticing. Did you use some AI for this?

dcog989:
Hi. I used AI to confirm / explain it - but was just looking at the code to try and understand how it all worked. Seeing FG and BG set to same color seemed suspicious! Happy to help a little.

PR #14285, opened 2025-05-15
Fix for zi.tools

lumynou5:


Myshor:
I went through the whole page and I will make some changes in your sitefix to make all needed dark parts visible. 😉

PR #14303, opened 2025-05-16
Fixes for www.vjw.digital.go.jp

felixonmars:
- Invert logo
- Set text color to make them readable

Before:
Screenshot_20250517_011010

After:
Screenshot_20250517_011002

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -34964,6 +34964,18 @@ object {
 
 ================================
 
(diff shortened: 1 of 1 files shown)


felixonmars:
Hi, it seems this fix was gone from current latest version, but the problem persists. I failed to find where and why it got removed. Any ideas?

Myshor:
I won't search when and why it was removed. I remeber there was some PRs removing not working URLs. Maybe it was down for a while. I will push back this fixes in a sec. 😉

PR #14304, opened 2025-05-16
Invert input tools icon on mail.google.com

StevenSchifter:


Myshor:
Can you share some screenshot with before/after so we would know what exactly you are trying to invert?

I tried to find that selector in my mail.google.com and nothing was found.
Additionally I did not see any icon that is needed to be INVERTED.

StevenSchifter:
Gmail_InputToolsIcon_BeforeAndAfter
In the upper right corner of Gmail, there may be an icon for Input Tools, depending on whether or not [the setting](https://www.groovypost.com/howto/enable-gmail-input-tools/) is enabled. The icon is not inverted by default (left side of image).

Myshor:
OK... So I knopw you use chromium based browser as it was not needed in Firefox and adding your fix as it is will break Firefox presentation.

But do not worry. I spent some time and I will push additional fix to your code that will will work in both browser types. 😉

PR #14307, opened 2025-05-18
Fix for *.tripadvisor.com

felixonmars:
Invert logo.

Before:
image

After:
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -569,6 +569,13 @@ CSS
 
 ================================
 
+*.tripadvisor.com
+
+INVERT
+img[alt="Tripadvisor"]
+
+================================
+
 *.webex.com
 
 CSS


felixonmars:
Same as [#12982](https://github.com/darkreader/darkreader/pull/12982), no idea why it got removed 🤔

Myshor (changes requested):
For me it's working fine without your fix.
obraz

It's working fine because of this fix:
```
tripadvisor.com
tripadvisor.com.tw
*.tripadvisor.com

INVERT
img[alt="Tripadvisor"]

```

obraz

So... Can you check if you are using newest sitefixes? Because adding your additional and separate `*.tripadvisor.com` can do more mess than fix. 😋

P.S. I do not request changes but I need to block merging until we confirm it's not needed or I have something older on my side. 😉

felixonmars:
Oops, you are right. Not sure why I have "Synchronize sites fixes" enabled but it didn't sync :(

Myshor:
I can tip why synchronization can not work.

When you apply some fixes in Dark Reader's developer tools that fixes are used in Dev Tools until "Reset Changes" button is pressed. (at least it was working like that when I wasn't one of main contributors to the Dark Reader.

Now I do not even notice how it works as I always "Reset Changes" after pushing and merging my PRs to the main branch. 😅

However... To be sure that newest sitefixes are applied no other changes should be applied in the Dark Reader. 😉

felixonmars:
I see. Thanks for the info!

PR #14352, opened 2025-05-29
Fix detexify.kirelabs.org in dynamic-theme-fixes.config

LyroyTheToad:
Symbols in https://detexify.kirelabs.org were not inverted correctly.

I have no idea why `backgound-image` in `.symbols > img` is overwritten with:
`url("blob:https://detexify.kirelabs.org/875e9562-3745-41cd-b978-58c3b4f07d1d")`
only when dark reader is active.

The only solution I found was to hard-code back the correct image:
`url(https://detexify.kirelabs.org/images/symbols-a95e7763.png)`.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -8432,8 +8432,12 @@ IGNORE INLINE STYLE
 
 detexify.kirelabs.org
 
-INVERT
-.symbol
+CSS
+.symbol > img {
+    background-color: rgba(0, 0, 0, 0) !important;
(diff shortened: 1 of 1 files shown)


Myshor:
It used blob as Dark Reader was trying to make invert for this pic by himself. However something is not working right for that and it ended without anything but...

You nicely catched what to use to fix it back but I can not accept this PR as it breaks light scheming.

Forcing invert works fine with Dark Scheme:
obraz

but it break light scheme:
obraz

That's why we use separate section with INVERT (the one you removed completely). It is used when Dark Scheme of Dark Reader is used only.

I will make quick fix to your PR and merge it as it's not big thing to fix. 😉

PR #14355, opened 2025-05-31
Fix for instagram.com

pnaf:
Uses grayed-out text on posts with limited comments.

Before:

After:


Myshor:
Why to make it more grayed out as the original Dark Mode on instagram (without Dark Reader enabled) doesn't do that?
obraz

pnaf:
Ah, thanks, I didn't check the original Dark Mode on Instagram and just went by the color scheme as seen on the site's original Light Mode.

I'll close this pull request for now.

PR #14367, opened 2025-06-02
Adding my now completed site to the list

TheSigmaArts:
Now that my site is complete, I'm asking to add it to the list.
(continuation of #14249)

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -15,7 +15,6 @@
 *.sapph.xyz
 *.starbreeze.com
 *.suyu.dev
-*.thesigmaarts.com
 *.tickettool.xyz
 *.tracker.gg
 *.tracker.network
@@ -1131,6 +1130,7 @@ thelinuxcast.org
 theme-park.dev
 themes.vscode.one
 themetphilly.com
+thesigmaarts.com
 thespacedevs.com
 thetatoken.org
 thetrackernetwork.com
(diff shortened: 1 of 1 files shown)


src/config/dark-sites.config
```diff
@@ -15,6 +15,7 @@
 *.sapph.xyz
 *.starbreeze.com
 *.suyu.dev
+*.thesigmaarts.com
```
Myshor, line 18: Do you use prefixes for the site address for other sub-pages? I could not find any like that using links on your site.

As I do not saw anything with prefix I need to asc to add this page without `*.` in the correct order.

Myshor (changes requested):
Additionally - shop and Collaborations have white boxes:

obraz

I can tolerate it as a part of Dark Scheme but I just want to mention it was painfull to my eyes. 😋

src/config/dark-sites.config
Myshor: `furaffinity.net` is bad example - it has sites with prefixes like `https://shop.furaffinity.net/`

I went through all links of `thesigmaarts.com` and all other sites are open as suffix `https://thesigmaarts.com/shop`

and for that `*.` prefix is not needed in dark-sites

I just want to have clear entry. Just don't want everyone adding *.mysite.isdark with asterisks because they are lazy.

You change `*.thesigmaarts.com` to `thesigmaarts.com` and I will merge.

Myshor (changes requested):
Still I request change.

TheSigmaArts:
Changes have been made accordingly~

kekkodance:
Not happy about this but feel free to merge.

Myshor (approved):

PR #14380, opened 2025-06-08
Fix for thenewslens.com

felixonmars:
Invert bright header.


Myshor:
Can you share your broken site?

I checked in my Firefox and in my MS Edge and I have same results in both:
obraz

So no invert needed or it will be broken. 🤪

felixonmars:
Hrm, now here it's not needed either. That's weird :/

PR #14381, opened 2025-06-08
Broken website: Public Interest Research Group

erkinalp:


Myshor:
When I tested this I found one more place broken.
Adding one more selector and merging. 😉

PR #14393, opened 2025-06-15
Add `abhiyandhakal.com.np` to dark-sites.config

abhiyandhakal:
The site [abhiyandhakal.com.np](https://abhiyandhakal.com.np/) is dark by default. So, I'm adding the site to the dark-sites.config

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -35,6 +35,7 @@
 a-maze.jothin.tech
 aagaming.me
 abhiyan.me
+abhiyandhakal.com.np
 absolucy.moe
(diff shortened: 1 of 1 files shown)


Myshor:
I can not add site which remeber my change to light mode using switch:
obraz

Dark-sites are reserved for pages that are always opened as dark.

For the pages that have light/dark mode we use https://github.com/darkreader/darkreader/blob/main/src/config/detector-hints.config

For your site this code should be enough:
```
abhiyandhakal.com.np

TARGET
html

MATCH
.dark

```

PR #14395, opened 2025-06-15
invert logo for CVF open acess

pedrormjunior:
invert logo for CVF open acess


Myshor:
First I wanted to reject as img selector is too general to use but as I went through whole page it seem the all docs are in PDF or on other sites in the links and IMGs are only logos. 😅

Merging.

PR #14400, opened 2025-06-17
Fix card backgrounds on myndisportal.ndis.gov.au

non-bin:


non-bin:
nevermind, it works when I press apply or change from light to dark mode, but breaks when I reload

PR #14404, opened 2025-06-18
Fix archive.org collection pages

McBruno712:
Various fixes for archive.org collection pages (example https://archive.org/details/alexacrawls?tab=collection) both for Collection and About tabs. This also includes fixes for the page logo, search bar and some other icons as well.


Myshor:
Works fine in Chromium based browsers:
obraz

But in Firefox I end with:
obraz

Not merging yet as I will try to do some fixes to make it work in both browsers.

Myshor:
Fixes added. Merging. 😁

PR #14409, opened 2025-06-19
Fix spaces in code blocks on palletsprojects.com

wheelercj:
Some of the spaces in code blocks on *.palletsprojects.com appear as white underscores in dark mode, which makes the code more difficult to read. You can see this in the first code block on https://flask.palletsprojects.com/en/stable/async-await/. There should be a space, not an underscore, between `def` and `get`.

Screenshot from 2025-06-19 01-45-03

The bug appears to be present for every instance of `def` in a code block throughout the site.

This PR changes those underscores to the same color as the background.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -23278,6 +23278,15 @@ img[src="images/vpsserver.png"]
 
 ================================
 
+*.palletsprojects.com
+
+CSS
+span.w {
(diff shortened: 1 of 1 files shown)


Myshor:
Ok... it seems the class is `.w` in both cases on this page. I hope there won't be other classes with spaces on other sites and or different entry in class `.w` than empty spaces.

For now I am merging.

PR #14412, opened 2025-06-19
Improve Serbian locale

rammba:
Hey @alexanderby, I'm happy to contribute to your project.
I'm currently working on translating the [official React documentation](https://github.com/reactjs/sr.react.dev) to Serbian, but I also contributed with translations to few other open-source projects.

My change here was mainly in using cyrillic instead of latin version of Serbian language. Cyrillic is formal and it's used in all legal papers, so I thought it should be used here as well.
If I understood [Chrome docs](https://developer.chrome.com/docs/extensions/reference/api/i18n#locales) well, they don't differentiate two Serbian versions, right? If I'm missing something, please tell me how to name latin file to have both versions available.

Hey @kizule @nebocoder, since I'm chaning your previous work, please feel free to give your suggestions.

P.S. Is there a way to add a comment into a `.config` file mentioning who worked on the translations?

### DIFF
--- a/src/_locales/store/store.sr.config
+++ b/src/_locales/store/store.sr.config
@@ -1,10 +1,10 @@
 @store_listing
-Ova ekstenzija brine o vašim očima tako što omogućava tamnu temu za sajtove. Dark Reader invertuje svetle boje i pojačava kontrast, što omogućava lakše čitanje noću.
+Ово проширење брине о вашим очима тако што омогућава тамну тему за сајтове. Dark Reader инвертује светле боје и појачава контраст, што омогућава лакше читање ноћу.
 
-Možete podesiti osvetljenje, kontrast, sepija filter, tamni mod, font i listu ignorisanih sajtova.
+Можете подесити осветљеност, контраст, сепија филтер, фонт преференце, бирати између тамног и shader режима, и омогућити или онемогућити проширење за специфичне сајтове.
 
-Dark Reader Plus otključava premium korisnički interfejs sa prilagodljivim temama boja uz naknadu.
+Dark Reader Plus уз накнаду откључава premium кориснички интерфејс са прилагођеним темама.
 
-Molimo vas da pročitate uslove korišćenja https://darkreader.org/terms/
+Молимо вас прочитајте услове коришћења https://darkreader.org/terms/
 
-Pre instalacije, isključite druge slične ekstenzije. Uživajte!
+Пре инсталације, молимо вас да искључите остала слична проширења да бисте избегли конфликте. Уживајте!


kizule:
Hi @rammba, I'm very happy to see that there are more of our people that are enjoying and using DarkReader.

I left some suggestions, but I like this idea and what you did so far.

> If I understood [Chrome docs](https://developer.chrome.com/docs/extensions/reference/api/i18n#locales) well, they don't differentiate two Serbian versions, right? If I'm missing something, please tell me how to name latin file to have both versions available.

That's unfortunately true.

alexanderby:
Hi @rammba, thank you for the contribution. This Latin translation was initially made by @nebocoder 4 years ago. We haven't had any complaints since then. Some Cyrillic sentences may be caused by me using Google Translate API for multiple languages at once. We also have a Help page translation made by @bzzrak 3 years ago in Latin alphabet.

Changing the alphabet may cause a heatwave of email from users. But let's hear what the initial translation authors will say. I hope it won't spark any heated political debate.

rammba:
Hey guys, thanks for the quick response!

> I left some suggestions, but I like this idea and what you did so far.

Am I missing something, since I don't see your suggestions?

> Changing the alphabet may cause a heatwave of email from users.

I don't think users will complain a lot, but statement makes sense. Maybe bosnian and montenegrin people also uses serbian version because their languages are not supported by Chrome, unfortunatelly. Still, I think all serbian speakers knows both cyrillic an latin. But, of course, we should see what other contributors think.

> We also have a Help page translation made by @bzzrak 3 years ago in Latin alphabet.

I'm planning to check help page translation as well, and cover both cyrillic and latin versions, because it's not strictly binded to Chrome supported locales.

kizule:
> Am I missing something, since I don't see your suggestions?

These should be visible as reviews.

rammba:
> These should be visible as reviews.

Do you mean this? Maybe you didn't publish comments?
image

kizule:
> > These should be visible as reviews.
> 
> Do you mean this? Maybe you didn't publish comments? image

Ah, I forgot that there's a button where I have to publish it. Sorry, done now.

rammba:
Thanks for the suggestions, now they are fixed.

Myshor:
For the Cyryllic changes I think @alexanderby would be better person to merge. 😅
But I think it's very nice work done here. 😉

rammba:
Hello @alexanderby, is it ok to use cyrillic version, or to switch back to latin?

kizule:
Cyrillic is fine by me as well. :)

alexanderby:
Well, since initial contributors disappeared, we can consider this.

PR #14432, opened 2025-06-30
Fix for start.gg

miss-manners:
Fixes white gradients that make text difficult to read

Before
image

After
image


Myshor:
OK... found the places for that and checked using both dark and light scheming.

PR #14441, opened 2025-07-04
Add *.uncoverit.org to dark-sites.config

WarFiN123:


Myshor:
I can not merge this as:
`uncoverit.org` can be switched to light mode and then opens in light.
obraz

Same with `status.uncoverit.org`
obraz

I can only recommend adding dark theme detection into => https://github.com/darkreader/darkreader/blob/main/src/config/detector-hints.config

```
*.uncoverit.org

TARGET
html

MATCH
.dark

```

Above code would do the trick.

PR #14443, opened 2025-07-04
Fix colors for the connection game on nytimes.com

zUnixorn:
This makes the colors used in the nytimes.com game "connections" more accurate to the intended colors (especially the blue category)

# Without darkreader (original)
image
image

# With darkreader (before)
image
image

# With darkreader (after)
image
image


Myshor:
Seems ok if Dark Scheme is used.

For light it needs fixes. 😋
obraz

I will work a while to set correct colors that will be inverting where needed.

PR #14469, opened 2025-07-13

rahulharpal1603:

This site already has a dark mode implemented, so it will be good to have this site added into the Global Dark List.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -753,6 +753,7 @@ mixxx.org
 mizik.eu
 mmorpg.com
 mnsr.win
+modelcontextprotocol.io
 modworkshop.net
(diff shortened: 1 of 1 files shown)


Myshor:
The site has possibility to switch into light mode. So it's not always dark and can not be added into Dark Sites list:
obraz

Alternatively you can add detection hint here: https://github.com/darkreader/darkreader/blob/main/src/config/detector-hints.config

This code should work:
```
modelcontextprotocol.io

TARGET
html

MATCH
.dark

```

rahulharpal1603:
Thank you for clarifying this, I will open the PR with your suggested changes. 

Here is the new PR: https://github.com/darkreader/darkreader/pull/14474

PR #14470, opened 2025-07-14
added geeksforgeeks.org to global dark list

dragon-slayer27:
https://www.geeksforgeeks.org

This site already has a dark mode implementation in place, so it would be nice if it were included in the global dark list.


Myshor:
- Opened default in light mode:
obraz
- Has a switch to switch modes.

Two point that disqualify it to Dark Sites list.

What's worst - I can not see any changes in html code when theme is changed so I do not see possibility to add some detection hint for this site either.

PR #14471, opened 2025-07-14
Added_neetcode.io_to_listof_dark_sites

Ksh1tij777:
This PR adds https://neetcode.io/roadmap to the dark sites list. This page is actively used by many and already has a native dark theme, and Dark Reader's override causes visibility and readability issues. Adding it to the dark sites list resolves this by preventing unnecessary styling.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -788,6 +788,7 @@ nationsglory.es
 nationsglory.fr
 neal.fun/size-of-space
 nee.lv
+neetcode.io/roadmap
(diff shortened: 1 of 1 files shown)


Myshor:
As the page have switch for the theme it can not be added into list of Dark Sites:
obraz

But you can add detector hint here => https://github.com/darkreader/darkreader/blob/main/src/config/detector-hints.config

This code should work for the site:
```
neetcode.io

TARGET
html

MATCH
.dark-theme

```

Issue #14483, opened 2025-07-17
[Broken Website] guideautoweb.com


Gitoffthelawn:
### Prerequisites

- [x] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this website issue to avoid opening a duplicate.
- [x] I can reproduce this website issue in a new, unmodified web browser profile with Dark Reader installed as the only extension.

### Website Issue Description

In dynamic dark mode, much of the text on this site is dark on dark.

### Website Address

https://www.guideautoweb.com

### Steps To Reproduce

1. Enable dynamic dark mode
2. Go to https://www.guideautoweb.com
3. Notice dark on dark text

### Operating System

Linux

### Web Browser name and version

Firefox 140.0.4

### Dark Reader version

4.9.108

Myshor:
There was some big fix proposed already but I will do that with small sitefix which can be updated faster. 😉

Myshor:
Sitefix added.

- If you want to have this fix faster you can set **Synchronizee sites fixes** option to check in Advanced section of Settings.
![obraz](https://github.com/user-attachments/assets/9dea8b66-b77a-46ab-8aa6-e50d995718c3)
- With that enabled newest sitefixes will be updated every browser start. 😉
- To open all settings you must click on:
(Old/classic design) More -> All Settings
![obraz](https://github.com/user-attachments/assets/8c17b1f1-e85e-4672-8a88-85930e178e02)
(New/Preview design) Settings in popup
![obraz](https://github.com/user-attachments/assets/e4f586f8-8ce4-47bc-90d7-7451eb030439)

Gitoffthelawn:
Thank you! ❤️

PR #14495, opened 2025-07-18
Fix/guideautoweb.com Issue of the text on this site is dark on dark.

arnim2006:
What does this PR do ?

This PR  fixes the issue of the text on this site is dark on dark of the site https://www.guideautoweb.com/ .

Fixes: #14483

Screenshot :

image


Myshor:
Why so much changes into CSS style and additional changes into Dark Reader's code? I will reject as the fix is "one liner" (adding fix to one selector) into sitefixes => https://github.com/darkreader/darkreader/pull/14497

PR #14496, opened 2025-07-18
Store translation

jenny-sunshine:


Myshor:
What kind of translation you use... Why there is some mention about paid Dark Reader Plus option?

For me this is reject. @alexanderby maybe you know something more about some Dark Reader Plus? 🤔

alexanderby:
Hi, that's alright, @jenny-sunshine Is our tech support. She's doing the translations for the Edge store. We are testing an alternative paid UI on Edge to see if it promoting it will be more successful than begging to pay.

PR #14500, opened 2025-07-20
Add dark theme support for swiggy.com

AmrendraTheCoder:
- Fix search container background to use neutral background color

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -29590,6 +29590,21 @@ CSS
 
 ================================
 
+swiggy.com
+
+INVERT
+.logo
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -29590,6 +29590,37 @@ CSS
...
+.logo
+.brand-logo
+.site-logo
+.main-logo
+.header-logo
+header .logo img
+nav .logo img
+.navbar .logo img
+
+NO INVERT
```
alexanderby, line 29605: `NO INVERT` is used for Filter mode, a different config file. Did you test this fix?

PR #14503, opened 2025-07-20
Add dark theme support for phonepe.com

AmrendraTheCoder:
- Fix header background to use neutral background color

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -24039,6 +24039,21 @@ CSS
 
 ================================
 
+phonepe.com
+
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -24039,6 +24039,37 @@ CSS
...
+.logo
+.brand-logo
+.site-logo
+.main-logo
+.header-logo
+header .logo img
+nav .logo img
+.navbar .logo img
+
+NO INVERT
```
alexanderby, line 24054: `NO INVERT` is not relevant here.

PR #14505, opened 2025-07-20
Add dark theme support for zomato.com

AmrendraTheCoder:
- Fix search input background to use neutral background color


Myshor:
Something broke up here.... First some additional files has been added into the PR.

Second thing.
NO INVERT won't work for Dynamic fixes. This section is reserved for other modes.

status_summary.sh
```diff
@@ -0,0 +1,39 @@
...
+echo "🔄 NEXT STEPS:"
+echo "   1. Test locally using: ./test_local_changes.sh"
+echo "   2. Add comment to PR #14509 explaining the fix"
+echo "   3. Wait for maintainer reviews"
+echo "   4. Close original PR #14485 with explanation"
+echo ""
+echo "💡 BENEFITS ACHIEVED:"
+echo "   • Smaller PRs = faster reviews"
+echo "   • Specific selectors = no content image issues"
+echo "   • Proactive fixes = impressed maintainer"
```
alexanderby, line 36: What can I say. The purpose of Dark Reader's dynamic mode is to automatically determine what to invert and what to not without impacting the performance too much.

That's why the list of fixes is needed. Particularly the image sprites with mixed dark and light icons can be tricky. That's why we need a person to solve non-trivial problems.

Please submit an Issue before opening any more PRs.

verify_all_prs.sh
```diff
@@ -0,0 +1,95 @@
...
+        pr_num=14510
+    elif [ "${websites[$i]}" = "zomato" ]; then
+        pr_num=14505
+    fi
+    echo "  ${websites[$i]}.com: https://github.com/darkreader/darkreader/pull/${pr_num}"
+done
+
+echo ""
+echo "🎯 VERIFICATION COMPLETE!"
+echo "If all checks show ✓, your PRs are ready for review!"
```
alexanderby, line 95: Did you use ChatGPT to generate these scripts?

PR #14513, opened 2025-07-20
Fix: Inverted Country Flags while using Darkreader 

### LINKED ISSUE #14512: [Broken Website] mpb.com, country chooser inverting country flags
### Prerequisites

- [x] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this website issue to avoid opening a duplicate.
- [x] I can reproduce this website issue in a new, unmodified web browser profile with Dark Reader installed as the only extension.

### Website Issue Description

When going into the country selector/region selector it is inverting the country flags colours.

### Website Address

https://www.mpb.com/en-eu

### Steps To Reproduce

1. Go to site
2. Scroll all the way down and click `Change location`

### Screenshots

Image

### Operating System

MacOS

Zen 1.14.5b

### Dark Reader version

4.9.110

### Additional Context

_No response_

arnim2006:
What does this PR do ?
This PR improves the functioning of Darkreader at the website https://www.mpb.com/de-at where the country flag colors  were inverted while using darkreader . 

Fixes #14512

what did i do ?
1. Reproduced the issue 
2. saw the code problem at console 
3. solved it 

SCREENSHOT:

BEFORE:

image

AFTER:

image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -20893,7 +20893,7 @@ mpb.com
 IGNORE INLINE STYLE
 svg[class*="flag"]
 
-========
+================================
 
 msi.com
 
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -20888,6 +20888,13 @@ INVERT
 
 ================================
 
+mpb.com
+
+IGNORE INLINE STYLE
+svg[class*="flag"]
+
+========
+
 msi.com
 
 INVERT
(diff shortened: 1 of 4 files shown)


arnim2006:
hii @Myshor , i tried to solve the issue and from my end it is working . It would be better if you see this and i want to say that there it is showing that i have changed 4 files but the main change in only on the  one file which is dynamic-theme-fixes.config . For the other files , i basically did changed some line of code which were not relevent to this issue so i just converted back them to normal . I hope you will understand

Myshor:
OK. But for merge I accept all changes or nothing. And as you mentioned. We want to see only dynamic-theme-fixes.config changed to merge.

Other files should not be even listed here as I won't allow to change anything in manifests.

I can try to "Remove" this files from PR but I can do more mess so I think you should do that by yourself.

Myshor:
Fixed last file.

And fixed amount of "=" to separate fixes in dynamic-theme-fixes.config.

PR #14517, opened 2025-07-21

ArjavJ-code:

This site already has dark mode implemented, so it will be good to have this site added into the Global Dark List


alexanderby:
This website can have both dark and light themes, thus doesn't qualify.

PR #14520, opened 2025-07-21

Mardavweb:

This site already has a dark mode implemented, so it will be good to have this site added into the Global Dark List.


alexanderby:
This site also has light theme.

PR #14522, opened 2025-07-22
Fix/add inoreader site to dark list

Mardavweb:
https://www.inoreader.com/dashboard

This site has a dark mode implemented, so this site should be added into the Global Dark List


alexanderby:
Only sites that are always dark should be added to this list. For other cases we have automatic dark theme detection.

PR #14534, opened 2025-07-26
Site added

techvoyager-varun:
midjourney.com add this site to dark-site.config as this site has default dark mode no light mode but the dark reader extension interpret wrong and chnage colour


alexanderby:
Both websites have light themes available.

PR #14536, opened 2025-07-27
Add export site list test to 'Site List' settings

kelson8:
This is a work in progress and early version of this, I have the basic functions working for an 'Export site list' button.
I have added an `Export site list` button into the `Site List` settings, which gets saved to a `list.txt` file in the users browser.

This should make it easier to save websites to a list, I would need to figure out how to add an import button but this seems to work well for me.

Can be useful if a user switches browsers or has a lot of sites to keep excluded from Dark Reader.
Feel free to adapt this and change it up if you decide to use this.

### DIFF
--- a/src/ui/options/site-list/export-site-list.tsx
+++ b/src/ui/options/site-list/export-site-list.tsx
@@ -1,0 +1,84 @@
+import {m} from 'malevic';
+import {getContext} from 'malevic/dom';
+
+import type {ViewProps} from '../../../definitions';
+import {Button, ControlGroup, MessageBox} from '../../controls';
+import {DeleteIcon} from '../../icons';
+
+// Mostly copied out of clear-site-list.tsx
+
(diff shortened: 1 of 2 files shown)


alexanderby:
Hi @kelson8. We have Import/Export ability in the Settings. Maybe this will solve your use case?

kelson8:
Yeah that works too, better then what I was trying to do which was just save the list instead of all the settings. I don't know how I missed that option myself.
That's my fault really, should've looked in the other options menu because I never did see those options before under advanced settings.

skellitor301:
So, I'm not sure if I'm reading things wrong or if there's a general confusion on this, but I'm looking through the issues for solutions to export the site list itself and people keep saying try exporting the settings and that somehow works? I've tried this myself and tried importing the resulting .json into another browser and it doesn't carry the site list over. Am I missing something or is there no way to export just the site list so one can import it into another browser currently?

PR #14542, opened 2025-07-29
Added primevideo.com to dark-sites.config

rudrakshtank:


alexanderby:
This website's other pages are white, so the PR is rejected. I suspect some coach told you guys how great it is to contribute to open source. But I don't appreciate these low quality pull requests that come not from genuine use of Dark Reader while trying to solve a real problem, but are made just to be there.

Issue #14548, opened 2025-07-31
[Broken Website] Text appears black on sankei.com articles


diontools:
### Prerequisites

- [x] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this website issue to avoid opening a duplicate.
- [x] I can reproduce this website issue in a new, unmodified web browser profile with Dark Reader installed as the only extension.

### Website Issue Description

Article text appears in black color when Dark Reader is enabled, making it difficult or impossible to read on dark backgrounds.

### Website Address

https://www.sankei.com/article/20250730-UDZ7DUQVIBILPI2PPM7CKL322Q/

### Steps To Reproduce

1. Install Dark Reader extension
2. Navigate to https://www.sankei.com/article/20250730-UDZ7DUQVIBILPI2PPM7CKL322Q/
3. Observe that article text appears in black color

### Screenshots

<img width="1237" height="898" alt="Image" src="https://github.com/user-attachments/assets/ac4e7636-d6d3-43ad-a137-0f82e7bd762a" />

### Operating System

Windows 11, Android

### Web Browser name and version

Firefox 141

### Dark Reader version

4.9.110

PR #14550, opened 2025-07-31
Fix: Improving the text quality and some more enhanced changes added to sankei.com

arnim2006:
What does this PR do ?
This PR changes the text color which was earlier dark while using dark reader due to which it was not easier to read . When i fixed this issue i noticed one more problem which was that when we write something on the writing section of the site then the text we wrote was also appearing to be dark so i also fixes that .

What did i do ?
1. Reproduced the issue 
2. Inspected the problem text 
3. Solved the Problem 

Fixes: #14548

SCREENSHOT:

BEFORE:

Screenshot 2025-07-31 213454

After:
When i solved the issue which was mentioned , i noted one more thing as you can see in the arrows i have tried to show the text which i have written in the writtes section is appearing dark .

Screenshot 2025-07-31 222258

After solving this problem:
Now you can see the text is clearly shown 

Screenshot 2025-07-31 221151

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -27125,6 +27125,38 @@ CSS
 
 ================================
 
+sankei.com
+
+INVERT
+# (none)
+
+CSS
+article,
+.article-body,
+.caption,
+figcaption,
+footer,
+.article-footer,
+.sps-article *,
+#consent-wrapper,
+#consent-wrapper .text,
+#consent-wrapper a,
+input,
+textarea,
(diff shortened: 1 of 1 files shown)


alexanderby:
The problem with this website is `* { color: inherit; }`. Every element inherits the text color from its parent. But the root HTML element has no color assigned, so Dark Reader has no color to override. So the issue is fixed by assigning a color to HTML element https://github.com/darkreader/darkreader/commit/956d88b9a2a1f843e155283361b19b264d65f9de

PR #14556, opened 2025-08-03
Added a dark website which got disturbed when the Darkreader applied.

Dibbu-cell:
In this  "https://element.io "  website darkreader should not be applied because it is already a dark website  but darkreader is 

applicable for this website and icon's of website got disappeared when we apllied the darkreader.


alexanderby:
This website has light theme pages, like Pricing. But we should improve dark theme detection there.

PR #14559, opened 2025-08-06
Add leetcode to detector-hints.config

anikett718:


Myshor:
Can you share the steps how this page can be switched to dark mode? I tried to find the switch. Even signed in and went through setting and this page is always in light mode for me. Can not confirm it can be switched into dark mode.

Edit..
Nevermind I found this when clicked "Try new features"

Issue #14199, opened 2025-04-21
[Feature Request] Transparency as an Option


aliyss:
### Prerequisites

- [x] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this feature request to avoid opening a duplicate.

### Is this feature request related to a problem?

No

### Feature Request Description

Using the ``dynamic theme`` and adding following rules just above IGNORE INLINE STYLE:

```css
:root {
    --darkreader-background-25292e: transparent !important;
    --darkreader-background-f1f3f4: transparent !important;
    --darkreader-background-f2f2f2: transparent !important;
    --darkreader-background-f6f8fa: transparent !important;
    --darkreader-background-ffffff: transparent !important;
}
```

I can get following output:

![Image](https://github.com/user-attachments/assets/23b07265-337b-4cbd-9440-5c07a82402ce)

I am now wondering if it would make sense to add this feature directly to darkreader. This would be especially useful for popups and stuff like that where one would have to add a backdrop-filter: blur(10px) or similar to actually have it working as intended.

I am very willing to help out. So if I could get a hint on which files to edit that affect the logic on how the background is calculated that would be great.

Myshor:
I am sorry but I can not catch what exactly you want to achieve,

Where this trasparency option should be included. If in the developer tools in sitefixes I think you can use the example for specific site and it will work.

I somewhere in the options I only think of background and text color settings. Why we should set them to transparent?

ckissane:
@Myshor I believe @aliyss is referring to making a flag to enable behavior like the Zen Internet firefox extension https://addons.mozilla.org/en-US/firefox/addon/zen-internet/

In particular setting webpage background transparent as well as potentially preserving opaque or blurring the backgrounds of popup elements or dropdown menus for legbility.

the code they posted is a shim to emulate such behavior.

Issue #13017, opened 2024-07-25
[Feature Request] Transparency in colors


pallaswept:
### Prerequisites

- [X] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this feature request to avoid opening a duplicate.

### Is this feature request related to a problem?

Yes

### Feature Request Description

Allow users to specify transparency/opacity in their colours.

### Additional Context

With firefox 128 this week, we now officially are able to set transparency in our windows and their content. Unfortunately, I found that enabling dark reader replaced my transparent background with the fully-opaque background colour I had selected in DR's UI. I attempted to enter a transparent value, but DR continued using the colour without its opacity value. So, DR kinda breaks the new transparency feature.

Related to this, it has been noted that while transparency can be nice, [enabling it for specific sites can be fiddly](https://github.com/cyp3rr/firefox-lab/blob/main/no-background/userContent.css). DR takes care of much of this by theming our background colours already, so, all that users would need, to get a very good coverage, is to use DR - if it allowed transparent colours.

So this would not only fix a problem related to DR, but also use DR to solve a problem inherent to theming. I hope you'll consider this as a new feature :) Thanks!

pallaswept:
FYI for maintainers: I was not required to tick a box to submit this issue. The "I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this feature request to avoid opening a duplicate." was not present until after I submitted. I thought you might like to know.

colrobloxkid69420:
I need this

pallaswept:
> I would have expected this to work since hex values allow alpha values, but when you change the alpha values it simply doesn't change anything.

I know that under the hood, DR does some clever magic to blend in between colours. Those functions will probably need a little tweak to make it work. 

At the moment, it appears to not even store the values for the alpha channel, or perhaps they are stored but removed when displayed, I can only guess at this point... But if you enter them, they won't stay. I wonder if there will need to be wide-reaching changes to data structures to hold the extra bytes. I sure hope not.

Anyway, I can see that this has the potential to be quite far away :( but, maybe not....we'll see.

PR #14560, opened 2025-08-06
Support for transparent backgrounds

PsychedelicPalimpsest:
Added an optional feature to add transparency to the background color. (As requested in #14199, #13017, and a few others)

Still has some kinks to work out (Most sites look awful), and transparency (To my knowledge) only seems to be supported in the Zen browser.  Because of that, I hid the feature behind an "Enable background transparency" setting in the advanced options. As of now, it is functional, however has some graphical issue. 

To enable transparency in the zen browser, you need to set the following in your `about:config`:

* browser.tabs.allow_transparent_browser = true
* On **windows** set browser.tabs.allow_transparent_browser, widget.transparent-windows and widget.windows.mica to true
* On **linux** set browser.tabs.allow_transparent_browser and zen.widget.linux.transparency to true. (If you are on arch like myself, you also will need a compositor)

I opened this PR to gather feedback, what do you all think of this, does this deserve up in the upstream package or should I keep it as a fork? 

Ps: This is my first time writing typescript, or code for a browser extension, so feedback is greatly appreciated! 
Pps: This image below also has some userChrome.css modifications and the Transparent Zen mod.
gscreenshot_2025-08-06-162226


PsychedelicPalimpsest:
Now is in a usable state
gscreenshot_2025-08-07-142511

PsychedelicPalimpsest:
Now it is in a usable state
gscreenshot_2025-08-07-142850

Myshor:
First of all... If you add new variable into localization it should be added into all languages as it can break the functionality of Dark Reader when used with other language than English.

The rest I can not comment as I am not that much involved in other places of Dark Reader's code.
This must be reviewed by @alexanderby

alexanderby:
Let me think what's the best way to implement in.

PsychedelicPalimpsest:
> If you add new variable into localization it should be added into all languages as it can break the functionality of Dark Reader when used with other language than English.

I added some placeholder text so it won't break in other languages, but definitely do need somebody else to do the actual translation (Unless you want me to use ai or something)

PsychedelicPalimpsest:
Currently my biggest issue is some sites just seem to ignore the transparency. I suspect that the issue is sites setting their own theme. 

But other than that, I feel this is functional.

Aqa-Ib:
This is really cool. Any updates? Thanks!

PR #14584, opened 2025-08-18
Fix for battlelog BF3 in dynamic-theme-fixes.config

JustCryen:
Fixed Assignments Overview section colors, Assignments List progress bar and Assignments Details progress bars.
I confirm that the page still looks correct in an original / light mode.

Before the fix:  
Screenshot From 2025-08-18 13-36-35

After the fix:  
Screenshot From 2025-08-18 13-36-24


Myshor:
Can you share the same page with light scheme enabled in Dark Reader after your fix applied?

Just want to be sure nothing is broken with that setting.

JustCryen:
@Myshor 
Sure thing!

Before the fix:
image

After the fix:
image

Looks exactly the same to me.

PR #14592, opened 2025-08-22

AzoSBear:
Updated the configuration to include the CSS selector targeting the bright master images on microsoft.com within the invert section of the microsoft.com entry, excluding the blue master image to prevent inversion from enhancing its brightness.


AzoSBear:
The webpage affected was this one
https://www.microsoft.com/en-us/microsoft-365/outlook/email-and-calendar-software-microsoft-outlook/

Without Fix
image

With Fix
image

> Remove this line as double empty lines will break tests.

Done

PR #14593, opened 2025-08-23

byMr712:
Changing the display of one style on the Avito.ru website so that the content is displayed correctly and does not burn the eyes.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -227,16 +227,6 @@ body {
 
 ================================
 
-*.avito.ru
-
-CSS
-.price-yellowHighlight-l8ZLI {
-    background-color: initial !important;
-    background-image: none !important;
-}
-
-================================
-
 *.babbel.com
 
 IGNORE INLINE STYLE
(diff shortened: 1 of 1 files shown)


byMr712:
Sorry for the comment in css, it seems I misunderstood you when reading what needs to be explained how it works.

Here is a link where you can check
https://www.avito.ru/all/tovary_dlya_kompyutera/komplektuyuschie/videokarty-ASgBAgICAkTGB~pm7gmmZw?cd=1&q=rtx+5090
(4th ad at the moment, it may move!)
(you need to pay attention to the yellow highlighting, it looks bad when dynamically adjusted, mostly because the text in the yellow price highlighting becomes white, the color combination is bad, and other prices normally have white color, they do not have a "colored background")

I will attach a screenshot of what problem I am trying to say

image

byMr712:
Of course you can tell me why I didn't just change the text color to black so it would look normal with the yellow background. The problem is that I couldn't do it, every time I tried it didn't work, and also the dark reader extension tried to "rebel and change" my fix.

At the moment, the ad has become number 1 in searches

byMr712:
Sorry, I'm new to GitHub and accidentally "completed" your change request. I removed the comment from the css and saved the change, hopefully it showed up for you.

Myshor:
I will totally refactor this sitefix and put everything under avito.ru fixes as we have already section for that.

Myshor:
Now that yellow background is darkened with invert.

And additionally I added fix for different colors on the "tags" under price.

byMr712:
Thank you very much, you are the best!

PR #14626, opened 2025-09-10

star4570:
Fixed the posts on the Activity Feed of the Simcoe County District Schoolboard version of D2L to appear dark.

Added the following code:

scdsb.elearningontario.ca

INVERT
.my-activities


alexanderby:
INVERT should be used for elements like images. Is it your case?

star4570:
> INVERT should be used for elements like images. Is it your case?

No, sorry. I'll submit a new pull request with the proper format. Thank you!!

PR #14636, opened 2025-09-13
Fix drop down menu items on d2l.lonestar.edu

SAMC-Coffee:
fixed the background color of all drop down menu items. previously only some menu items had a dark background while other were white

Before:
image

After:
image


SAMC-Coffee:
Also fixed a color variable that is used on the site not using the dark reader version of that color

PR #14659, opened 2025-09-25
Add ostra.gg to dark-sites.config

NickantX:
Adding Ostra's website to the dark sites list


GalaxyCatD3v:
I agree this would be good idea. vroom vroom uwu

PR #14677, opened 2025-09-30
Add detector hint for mermaid.live/edit

cclypeatus:


alexanderby:
I feel like using the class name `dark` should be more reliable. The `color-scheme` property should normally be in a CSS file, but not as inline property (it defines the look for native controls).

PR #14679, opened 2025-10-01

AzoSBear:
added 2 selectors that fix the top bar for google docs and sheets

Without fix

image
image

With fix
image
image


alexanderby:
Inverting the header looks like a risky change. Could you try using CSS like `background-color: var(--darkreader-neutral-background) !important;` for those selectors instead?

AzoSBear:
with css background color change
image
image

alexanderby:
Looks beautiful, thank you! There're a missing color variable, which refers another missing color variable, and Dark Reader can't handle this at the moment. By default we use Filter mode specifically for Google Docs.

Issue #10757, opened 2023-02-09
[Broken Website] openingtree.com


lica98:
### Prerequisites

- [X] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this website issue to avoid opening a duplicate.
- [X] I can reproduce this website issue in a new, unmodified web browser profile with Dark Reader installed as the only extension.

### Website Issue Description

dark reader breaks the look of the chess board 

### Website Address

https://www.openingtree.com/

### Steps To Reproduce

1

### Screenshots

![image](https://user-images.githubusercontent.com/74957561/217780925-60d92c33-7098-4ee6-8aaa-76427d7959ad.png)

### Operating System

Windows 10

### Web Browser name and version

Chrome

### Dark Reader version

4.9.60

PR #14710, opened 2025-10-14
Update dark-sites.config: add openingtree.com

insipidlight:
https://github.com/darkreader/darkreader/issues/10757

Site is dark by default, though the chessboard is still broken with darkreader turned on, with all combinations of darkreader light/dark / site light/dark. Apologies for the mis-wording of previous commit, it's not "complete" because it's still broken


Myshor:
Can not add this to dark sites as it has mode switch.

I will push sitefix for broken black pieces in a minute.

PR #14716, opened 2025-10-17
Allow building outside of the git tree

kewisch:
This will make it possible to build from sources when you zip up the files for the Mozilla review in timezones with an offset. In Germany it will pass -7200 as the time and the yazl expects to write an uint32.


jobisoft:
I can confirm, that the Thunderbird add-on review team is able to build the add-on from source with this patch applied.

@kewisch : Thank you very much for digging that up!

alexanderby:
Hi @kewisch, thank you for looking into this!

PR #14725, opened 2025-10-20
Fix for classroom.google.com

es1228:
Fixed ribbon color for classroom.google.com.


alexanderby:
Could you please share a screenshot?

es1228:
Default (disabled)
image
Hard to see in this image but there is a round border part of the ribbon here

Dark Reader (default)
image
Miscolored border

Dark Reader (with patch)
image
Correctly colored border

Light Reader (default)
image
Miscolored border

Light Reader (with patch)
image
Correctly colored border

The issue is that #14479 used the page color rather than the ribbon color when filling in the round border

PR #14728, opened 2025-10-22
Fixed unreadable text and broken white element for pons.com

mihailkondov:
original look:
image

look before fix
image

look after fix
image


alexanderby:
Let me look into the issue, some problem with how CSS layers are handled.

alexanderby:
Just found a cause, should work well in the next release.

PR #14736, opened 2025-10-25
Fix for zohranfornyc.com

pnaf:
Fixes graphics on "Featured Press" section on home page.

Before:

After:

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -37691,6 +37691,18 @@ INVERT
 
 ================================
 
+zohranfornyc.com
+
+INVERT
+.bg-background > .mb-4
+
+CSS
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -37691,6 +37691,18 @@ INVERT
 
 ================================
 
+zohranfornyc.com
+
+INVERT
+.bg-background > .mb-4
+
+CSS
+.bg-\[url\(\/static\/images\/clipping\.svg\)\] {
```
alexanderby: Is it a real class name?

pnaf: It's defined in https://www.zohranfornyc.com/_next/static/css/c8ae659f49548637.css as of this writing. This should also work if you wanted to go with a more readable name:

```suggestion
.p-px > .h-2 {
```

PR #14760, opened 2025-11-02
fix for docs.google.com

es1228:
Fixed title ribbon color in docs.google.com homepage

Dark Reader Off
Screenshot 2025-11-02 142422

Dark Reader On
Screenshot 2025-11-02 142352

Dark Reader With Patch
Screenshot 2025-11-02 142604

Does not seem to break the light mode version of dark reader as well (similar to #14725)


alexanderby:
That's strange, it should have been fixed last week, but is broken again. Let me find a stable solution...

alexanderby:
Seems like another change on their website, color variables, that depend on other variables, that are not defined.

Issue #14690, opened 2025-10-06
[Feature Request] Can you provide support for Reactnative Expo?


Super1Windcloud:
### Prerequisites

- [x] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this feature request to avoid opening a duplicate.

### Is this feature request related to a problem?

Yes

### Feature Request Description

[Feature Request] Can you provide support for Reactnative Expo?  Great thank for you

Issue #12599, opened 2024-04-14
[Feature Request] System Theme Support


ShalokShalom:
### Prerequisites

- [X] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this feature request to avoid opening a duplicate.

### Is this feature request related to a problem?

Yes

### Feature Request Description

I dont know, if this possible. 

But Startpage.com comes by default with a dark theme, if your system has set a dark theme. So currently, we apply a less pretty dark theme onto it, despite that they come with a custom, pretty theme themselves anyway. 

Can we detect, if a page does use the system theme - and especially, if that uses the dark theme?

https://www.startpage.com

### Screenshots

With Darkreader:
![Screenshot_20240414_123933-1](https://github.com/darkreader/darkreader/assets/6344099/6737d65b-87c9-4893-b0e9-599951dd092e)

Without Darkreader:
![Screenshot_20240414_123921-1](https://github.com/darkreader/darkreader/assets/6344099/7a2ae5ac-894c-4d18-8693-e1192e475f60)

scrlkx:
Have you tried the "Detect dark theme" setting? It solves this problem on many pages that I use. If no, try enabling it on settings.

Issue #11644, opened 2023-08-28
[Broken Website] https://www.startpage.com/


Lungenpest:
### Prerequisites

- [X] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this website issue to avoid opening a duplicate.
- [X] I can reproduce this website issue in a new, unmodified web browser profile with Dark Reader installed as the only extension.

### Website Issue Description

white background, black input, dynamic theme

### Website Address

https://www.startpage.com/

### Steps To Reproduce

visit https://www.startpage.com/

### Screenshots

![grafik](https://github.com/darkreader/darkreader/assets/64774660/9f12b572-d858-40b7-a29c-b92df578fcbb)

### Operating System

manjaro stable

### Web Browser name and version

firefox 116

### Dark Reader version

4.9.65

alexanderby:
Looks like fixed now.

ghost:
It still does not detect the dark mode for me.

PR #14783, opened 2025-11-11
Fix for various

06kellyjac:
- Detector hint for `startpage.com`
- Fix detector for `developer.mozilla.org`
- Detector hint for `opentitan.org`
- Detector hint for `supabase.com`
- Fix for `ebpf.io`
- Add `maxroll.gg` to dark sites
- Add full detector for `android-x86.org`
- Detector hint for `docs.renovatebot.com`
- Detector hint for `*.zulipchat.com`
- Detector hint for `expo.dev`
- Also detect `gist.github.com`
- Detector hint for `developer.hashicorp.com`
- Detector hint for `zen-browser.app`

I've tested all but `*.zulipchat.com` in a fresh zen-browser. That one I've tested by hand with `document.querySelector()`
I'll log in to a chat at some point tomorrow.

The auto detection for `opentitan.org` mostly works but it can get stuck as "detected dark mode" on light mode until refreshed. Lightmode to darkmode is identified properly without refreshing.
Auto detection for startpage is a bit odd since it can set 2 different default colours when using `device`. I couldn't find something fixed to use.

There aren't any other css selectors using `:has` or `:not` so I'll keep this draft to discuss in-case that's not allowed

I can also split this out into separate PRs if necessary but I wanted to get a discussion going in one place first :)

### DIFF
--- a/src/config/detector-hints.config
+++ b/src/config/detector-hints.config
@@ -864,6 +864,19 @@ MATCH
 
 ================================
 
+startpage.com
+
+TARGET
+html
+
+MATCH
+:has(> head > meta[data-site-theme="dark"])
+:has(> head > meta[data-site-theme="night"])
+:has(> head > meta[content="#0c0d0f"] ~ meta[data-site-theme="device"])
+:has(> head > meta[content="#0a0b0b"] ~ meta[data-site-theme="device"])
+
+================================
+
 steamgifts.com
 
 NO DARK THEME


06kellyjac:
Addresses #12599 (at least for startpage itself)
Might address #11644 for startpage, depends if that background/theme still exists
Probably addresses #14690 if it means expo.dev

alexanderby:
Please separate the PRs, and only supply a fix where it's really necessary. I couldn't reproduce the issues with the start page, it's probably fixed now. The react native issue doesn't seem to be related to HTML, but rather Android and iOS UI.

06kellyjac:
Thanks.
Yeah I can split it.

Just to clarify slightly more complex css selectors like `:has` `:not` etc are ok?

---

On startpage with "System default" it fails to recognize dark mode on 4.9.112. I can check again for latest `main`

image

image

image

---

image

image

image


It looks like explicit dark and night work fine. Shall I remove those hints and just keep what's needed for System Default to work?

PR #14804, opened 2025-11-21
Extend IGNORE IMAGE ANALYSIS to BBC subdomains

Tenzian:
#14019 added IGNORE IMAGE ANALYSIS for bbc.co.uk

This extends the ignore to images on BBC subdomains.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -242,14 +242,6 @@ CSS
 
 ================================
 
-*.bbc.co.uk
-*.bbc.com
-
-IGNORE IMAGE ANALYSIS
-*
-
-================================
-
 *.bitrix24.ru
 
 INVERT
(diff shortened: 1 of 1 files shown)


Tenzian:
I've consolidated the existing rules in to one: The behaviour shouldn't change from what already exists.

The bitesize and CBBC sites have some (existing) issues that'll take more targetted use of selectors: I'll address those when I have time in another PR.

I've also extended the list of sites that are dark by default (mostly parts of iPlayer) in `dark-sites.config`.
(The `bbc.co.uk/tv/` prefix _could_ be used for sites that aren't dark by default so I've been specific)

src/config/dynamic-theme-fixes.config
```diff
@@ -3938,12 +3938,14 @@ div[class^='scrollableList'] {
 ================================
 
 bbc.co.uk
-bbc.co.uk/weather
-bbc.com/weather
+bbc.com
+*.bbc.co.uk
```
alexanderby: Looks good, thank you!

alexanderby (approved):

PR #14829, opened 2025-12-03

ForeverFloating:
Added corrections to GlassDoor


alexanderby:
Nice catch with the class name. However I replaced the inversion with removing the fill for the case, if we ever fix respecting the precedence of the SVG attributes and CSS values, and this inversion becomes irrelevant.

PR #14865, opened 2025-12-19
fix for mozillafoundation.org

Myshor:
Fix selector (removed additional "a" which accidently appeared after needed one)


Gitoffthelawn (commented):
LOL :)

PR #14890, opened 2025-12-29
Add hacks.guide to the global dark list

FarmYard-Gaming:


Myshor:
Can not add site as it has possibility to switch modes manually.
I will add it to detection hints instead.

Myshor:
As other solution is pushed I close this PR.

PR #14895, opened 2025-12-31

S1n7h:
Youtube's Pole Borders were barely visible. Changed the borders of the poll and the different choices to a different colour.

Before:
before

After:
after

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -37334,6 +37334,13 @@ CSS
 #search-icon-legacy.ytd-searchbox {
     border: 1px solid var(--ytd-searchbox-legacy-border-color) !important;
 }
+ytd-backstage-post-renderer[uses-full-lockup] {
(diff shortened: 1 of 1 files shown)


alexanderby:
Could you please share a link to see this?

S1n7h:
Any poll will do, but here are some.
https://www.youtube.com/post/Ugkx796Tr4jbjaa2A2OOxN5wcfVkvm2X3wYk
https://www.youtube.com/post/UgkxYHEETzePzC40Xb7yuMPrX12515D1dqzV

I have noticed that this is especially a case when the youtube theme is set to light and then the dark mode of dark reader is turned on. Another thing is, this problem is apparent in firefox regardless of youtube's theme.

alexanderby:
Screenshot 2026-01-07 at 10 54 39

Looks ok to me. Probably some redesign is underway. I would wait and see what is the actual problem, why Dark Reader fails to recognise the correct colors.

PR #14896, opened 2025-12-31
Fix to D2L Content screen to remove background

ptyork:
Remove background image / gradient from the left-side content navigation panel. Replaced with subtle (but still high-contrast) coloring using standard d2l color variables.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -176,7 +176,7 @@ d2l-empty-state-simple {
 .d2l-loading-spinner-wrapper svg circle {
     fill: var(--darkreader-neutral-background) !important;
 }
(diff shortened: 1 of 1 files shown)


Myshor:
Can you share some URLs to check this sitefix?
If not then I would like to see screens:
- before fixes
- after fixes
- after fixes but switched to light scheme in Dark Reader

Specially when the fix is added into `*` section it would be nice to see what exactly we want to fix.

ptyork:
Glad you had me do this, as I found and fixed an issue. Now I know you can add a second commit to a PR without reverting to the command line. :) Anyway, these are definitely all behind university logins, so here are screen caps. I checked this "edit" mode (instructor) and student view mode to ensure it didn't introduce anything unexpected.

DR OFF
image

DR ON - NO CHANGES
image

DR ON - WITH CHANGES DARK
image

DR ON - WITH CHANGES LIGHT
image

PR #14898, opened 2025-12-31
Invert and fix pdfViewer in hypothes.is

ptyork:
Fixes need to apply to multiiple hosts, so switched to wildcard format. I suppose not everyone wants the PDF Viewer inverted, but for me it's a severe accessibility issue. I also had to fix some really bad CSS issues (yellow on white with an opacity of 0.4 and alpha of 0.3 = TOTALLY INVISIBLE to me, even inverted). If it is possible to apply these only if the user has opted to invert PDF's, that may be desirable.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -407,6 +407,27 @@ IGNORE IMAGE ANALYSIS
 
 ================================
 
+*.hypothes.is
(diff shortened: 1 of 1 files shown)


ptyork:
Forgot to fix the broken toolbar buttons, so I closed this in favor of a single, complete fix.

PR #14902, opened 2025-12-31
Invert and fix pdfViewer in hypothes.is (revised)

ptyork:
Fixes need to apply to multiiple hosts, so switched to wildcard format. Fixed toolbar icons that were not displayed. I suppose not everyone wants the PDF Viewer inverted, but for me it's a severe accessibility issue. I also had to fix some really bad CSS issues (yellow on white with an opacity of 0.4 and alpha of 0.3 = TOTALLY INVISIBLE to me, even inverted). If it is possible to apply these only if the user has opted to invert PDF's, that may be desirable.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -410,6 +410,33 @@ IGNORE IMAGE ANALYSIS
 
 ================================
 
+*.hypothes.is
+
+INVERT
+a[href="https://hypothes.is"] > svg[aria-labelledby="logo-title"] > g > path[fill-rule="nonzero"]
+img[class="fusion-standard-logo"]
+.pdfViewer .canvasWrapper
+
+CSS
+.pdfViewer .textLayer {
+    opacity: 1 !important;
+}
+.pdfViewer .hypothesis-highlight {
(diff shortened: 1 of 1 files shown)


Myshor:
Can you share some URLs to check this sitefix?
If not then I would like to see screens:
- before fixes
- after fixes
- after fixes but switched to light scheme in Dark Reader

I could not see any changes after entering on hypothes.is (which loads as web.hypothes.is) and applying this fix.

ptyork:
Sure! It's hard to find a public annotation that uses the pdfviewer. That seems mostly relegated to it's academic role. So here are some screen caps. Note that I added a couple of annotations to show the changes to the highlights. I did also verify that these changes didn't break anything that I could find on non-PDF annotations (just improves visibility of the highlights in both light and dark mode). I can't be 100% sure everything works, but I think this is fairly benign and fixes what almost certainly is a bug (an overlay with both alpha and opacity applied making one type of annotation nearly invisible).

DR OFF
image

DR ON WITH CHANGES
image

DR ON WITH CHANGES LIGHT (I could probably get rid of that rogue light line in the middle, but it's not really hurting anything)
image

DR ON WITHOUT CHANGES (note the "hidden" toolbar buttons)
image

PR #14917, opened 2026-01-08
Fix `select` widget in Drupal Claro

smokris:
The [Drupal content management system](https://drupal.org)'s [default admin theme](https://www.drupal.org/docs/core-modules-and-themes/core-themes/claro-theme) styles `<select>` elements, adding a custom arrow image.  Dark Reader's adjustments cause the arrow image to be rendered too large.  This PR adds a CSS override to improve the styling.

| Default (Dark Reader disabled) | Dark Reader before | Dark Reader after |
| - | - | - |
| drupal-claro-light | drupal-claro-dark-before | drupal-claro-dark-after |

In the "Dark Reader before" screenshot above, I added pink arrows pointing to the misrendered `<select>` arrow.

### DIFF
--- a/src/inject/dynamic-theme/image.ts
+++ b/src/inject/dynamic-theme/image.ts
@@ -39,20 +39,24 @@ export async function getImageDetails(url: string): Promise<ImageDetails> {
                     if (encoding === 'base64') {
                         svgText = atob(svgText);
                     }
-                    if (svgText.startsWith('<svg ')) {
-                        const closingIndex = svgText.indexOf('>');
-                        const svgOpening = svgText.slice(0, closingIndex + 1).toLocaleLowerCase();
-                        if (svgOpening.includes('viewbox') && !svgOpening.includes('width') && !svgOpening.includes('height')) {
+                    else {
+                        svgText = decodeURIComponent(svgText);
+                    }
+                    const svgOpeningMatch = svgText.match(/^<svg( .*?)?>/);
+                    if (svgOpeningMatch) {
+                        const svgOpening = svgOpeningMatch[0];
+                        if (svgOpening.toLowerCase().includes('viewbox') && !svgOpening.includes('width') && !svgOpening.includes('height')) {
                             useViewBox = true;
 
                             // Explicitly set size due to unexpected drawImage() behavior
-                            const viewboxIndex = svgOpening.indexOf('viewbox="');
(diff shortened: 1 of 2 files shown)


alexanderby:
Could you please screenshot the actual CSS background value of the icon? If it's SVG, it may be potentially fixed in the next release.

smokris:
Sure, here are screenshots that include the Inspector view of the CSS values:

| Dark Reader disabled | Dark Reader enabled |
| - | - |
| drupal-claro-light-css | drupal-claro-dark-css |

alexanderby:
Thanks, the issue was those SVGs with viewBox but no width and height were behaving differently to those the Dark Reader was making. Should be fixed in the next release.

Myshor:
So... there is new version of Dark Reader released. @smokris  can you check if your page works fine with it? If yes then this PR should be closed without merge.

smokris:
Thanks for the info.  I updated to Dark Reader 4.9.119 just now, but unfortunately the problem remains.

Here's a simplified test case that reproduces the problem: [test.html](https://github.com/user-attachments/files/24697977/test.html)

I think there are 2 separate problems here:

1. The recent SVG parsing changes (8093d8031c6154945d4cebfc711978cc26984cd3 and 54a14b9355254ec7b05e543b7e7d19e981c49d25) don't handle the way this particular SVG is encoded.  I just pushed changes to handle that encoding, including an automated test.

2. The original CSS sets `[background](background: var(--input-bg-color))`, and a higher-priority rule specifies `background-image`, `background-repeat`, `background-position`, and `background-size`.  Then Dark Reader injects a rule that sets `background: var(--darkreader-bg--input-bg-color);`.  Since this rule is for `background` (rather than specifically `background-color`), it resets the `background-repeat`, `background-position`, and `background-size` values.  Would it be possible for Dark Reader to override only the `background-color`?  I didn't immediately see how to implement that, so for now I left in the change to `dynamic-theme-fixes.config`.

Myshor:
@alexanderby I am not that much experienced into code outside config folder. Maybe you can take a look here again. 😉

alexanderby:
@smokris Thank you for providing a test case and for pointing out the single quotes and the URL encoding. Brilliant!

This is fixed now, as well as overriding of the `background-size` property, a long-lasting issue.

`indexOf` has a slight performance and readability advantage over regexps here, so I preferred to use those. But I will happily merge your test case.

alexanderby (changes requested):

PR #14928, opened 2026-01-12
Allow specifying `file:///` URIs inside Dev tools > Dynamic Theme Editor > Per Site Editor

SuperCowProducts:
TL;DR: I commented what I want to fix in the source which is causing [this](https://postimg.cc/SnfJc5n3)

Here is an example:

```
mkdir -p ~/baf76686-ef9c-11f0-b6e5-8e390d695a9c && cd $_
wget "https://download.oracle.com/otn_software/java/jdk/25.0.1+8/2fbf10d8c78e40bd87641c434705079d/jdk-25.0.1_doc-all.zip"
unzip *
firefox docs/index.html
```

Dark mode for this site only works well with 'Static' mode, yet I want all other site to be in 'Dynamic' mode. Only way to do this seemed to be ` Dev tools > Dynamic Theme Editor > Per Site Editor`: actually you can get it done via GUI [like this](https://postimg.cc/ns1cQQP9), but that's hacky and hard to come up with (that special menu started appearing once I had opened `Dev tools` in a new window).

funnily enough, if in the same prompt we run

```
python -m http.server && firefox http://0.0.0.0:8000/
```

we can get around the `file://` problem as well.

The point is: if this is possible and could help some users, why don't we make the Dark Reader tool more flexible and allow `file://` URIs (or potentially others if that would be even better), instead of being in my opinion overly restrictive?

Not that I didn't get it to work in the end, this is mainly to help future users have to do less tinkering and access the feature more easily :)

### DIFF
--- a/src/ui/stylesheet-editor/components/body.tsx
+++ b/src/ui/stylesheet-editor/components/body.tsx
@@ -86,6 +86,7 @@ export default function Body({data, actions}: ExtWrapper) {
                     Reset changes
                     {dialog}
                 </Button>
+                <!--one of the functions triggered by this button caused `TypeError: can't access property "url", fix is undefined`-->
                 <Button onclick={apply}>Apply</Button>
             </div>
             <Overlay />


Myshor:
It's not allowed because sitefixes in dev tools are done for as the name says sites. `file://` is not the address just the protocol similar to `http://` or `https://`

In sitefixes we have fixes for specific sites not protocols.
What you can do if you want to have some fix for local file page is to include it in your local Dark Reader under `*` section.
This section is applied for all the pages opened in browser.

That's the solution I use when sometimes I had to debug some page sent to me zipped. 😉

So for me it's a big NO!

Brent-Sanchez:
An outstsnding Idea, Ive had to workaround this local rules only stuff with the persistsnce of an 8 year old who was promissed a super nintendo with the full on duck hunt package, controllers, and mario cart, but in modern times, so it cant be plugged in to the tv without bridging into an hdmi input, when his mom's not bird dogging.  

Of course this is actually worse now, with the recent changes to local network calls, its the rainbow boss of the interzones watching to make sure we dont sneak our fun, dymanic css into the file:/// view . 

So, one way to do this, might be extinct now, and thats to just use a real, but generic CA cert, and just treat your local lan like a web server. And yes, I hate this idea too, and yes it does suck a great deal more on windohz.  

Thats why I just took the dynamic theme I liked and slapped a static version right into windohz, in the kegacy internet options (IE 7) settings.  This worked pretty well, because i had dark reader in edge too, and had to make sure edge and (brave) other chromium clients never match as far as my standard dynamic theme anyway.  Otherwise, edge forgets who it is, and will cause some epic cascading style collisions.  Plus, god knows we can't let edge just start handling all the local stuff, like.file:///  or it'll be that directory plus the weather, and allegations of illegal chrome-urls flags and one-drive appears misconfigured!  Except, thats exsctly what i did, I just opened file:// up in edge, and paased it over to the system browsee with open-in Brave Browser, which used my native node, and dark reader could just go ahead and treat it dynamkcally.  

This is the only thing that worked for me, I also had an extensive blocklist set up for that ad-block extension that needs manifest v2 to run, i forget the name, but it is a must, edge likes to think it's in charge!  And it is, it's in charge of the local dark reader, and keepingn the file:/// access locked up, phoning Home://  and reminding me that I cant use IE, and every single local adress "doesnt exist".

This all replaces my old workaround btw.  The system as pdf's.  :)

mindcloud69:
@SuperCowProducts Maybe create a pr for a tips page explaining your workarounds.

Edit: I think it is actually.
https://github.com/darkreader/darkreader.org/blob/main/src/help/topics/en.md

https://github.com/darkreader/darkreader.github.io/tree/master/tips

PR #14946, opened 2026-01-16
Swordofsecrets

henrygab:
Multiple readability issues:
* Scrolling matrix-inspired background text interferes with text legibility
* Title is ... all wrong

Example URL:
* https://swordofsecrets.com/mini-rv32ima.html

The fix sets the matrix-inspired background to not be visible, and overrides some properties on the title text so it appears clear.

<hr/>
<details><summary>Original non-dark site</summary>
<p>

image

Note: load live to appreciate.

</p>
</details> <hr/>
<details><summary>Without the fix</summary>
<p>

image

</p>
</details> <hr/>
<details><summary>With the fix</summary>
<p>

image

</p>
</details>

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -30792,12 +30792,12 @@ swordofsecrets.com
 
 CSS
 .matrix-bg {
-    content-visibility: hidden !important;
+    display: none !important;
 }
 .title {
+    -webkit-background-clip: text;
     animation: none !important;
-    background: none !important;
-    color: var(--darkreader-selection-text) !important;
+    background-clip: text;
     text-shadow: none !important;
 }
 


alexanderby:
Dark Reader successfully detects dark theme here. Is there a need in this fix?

henrygab:
Without this, the title is not dark and difficult to read.  The legibility of the title is greatly increased by removing the rotating-color glow effect and the scrolling text background effects.

Therefore, I recommend the changes for usability reasons.  If accepted, and  then someone doesn't like the modifications, it's easy to disable DarkReader for the site.  On the flip side, it's non-trivial for folks to figure out how improve legibility of a site.  Since it's easy to disable and hard to create improvements, I offer this PR for your consideration as (imho) it improves the situation.

Myshor:
I found what is missing after Dark Reader changes fix.

To have the same view after Dark Reader applied we just need:
```
swordofsecrets.com

CSS
.title {
    -webkit-background-clip: text;
    background-clip: text;
}

```

I do not understand why to remove whole background.

I am not touching this PR. Let author decide if my proposal is good enough (I allow to just change PR to this code). Or we close this one and add another with my proposal.

Right now it detects dark theme on this site for me like it does for Alexander that's why I do not see the big urgent to apply sitefix for it.

Myshor (changes requested):
Added my proposal already. Current code is not acceptable for me.

henrygab:
> To have the same view after Dark Reader applied we just need:
> ```
> .title {
>     -webkit-background-clip: text;
>     background-clip: text;
> }
> ```

Thanks for figuring out that `-webkit-background-clip` fix.

I've applied that instead of removing the background.

I have retained the removal of the text-shadow and animation, both of which impact contrast and legibility.

I addition, I applied the automated GitHub Copilot recommendation to use `display: none` instead of `content-visibility: hidden`.

I tested the resulting file, and it works.

I believe I have now addressed all your concerns.  If not, could you help me understand what remains?

Thank you!

Myshor:
For me removal of animated background for .matrix-bg when it's dark is not a fix but integrating totally in content displayed by site. I won't accept it now, Let's see what @alexanderby would say.

I just extend my thought just to show why I do not like this:
Sitefixes are done to correct display because it can be broken after Dark Reader changes. In most cases (don't want to say it's always) we are doing sitefixes to fix broken colors etc. We do it for inverts dark elements, fix color of text or remove background if it's light and makes reading of the text difficult.

On this site - the site is dark by default. In most cases when user will have "Detect Dark Theme" option enabled in setting it won't be touched by Dark Reader at all. In that case adding sitefixes which removes features of that site like animated background and animatet effect for title text is not a fix but total change of site features.

I can accept only the code:
```
swordofsecrets.com

CSS
.title {
    -webkit-background-clip: text;
    background-clip: text;
}

```

but as @alexanderby was adding some comment here too I will give him the final word.

alexanderby:
@Myshor nicely noted that `background-clip` is the problem here (-webkit is a fallback for older browsers and can be ommited). The reason it failed was the combination of `background` property with variable followed by that `background-clip`. When the `background` is overriden, the `clip` part is lost. I made a fix to handle the shorthand background value with a variable before the other properties, and added the clip value copying. Hopefully it will fix similar issues with other `background-*` values.

So, I'm closing the PR as the fix becomes unnecessary after the next update.

henrygab:
Thank you for taking the time to explain the reasoning and bar.  I'm glad this issue highlighted an issue, and that you found a way to improve the handling more generically as a result.

PR #14953, opened 2026-01-18
Add urbandictionary.com to global dark list

FarmYard-Gaming:


Myshor:
Can not add it into global dark list as it sets dark or light mode - depending how system/browser preference is set.
Below when I switched browser preferences from Dark to Light:
obraz

Issue #10668, opened 2023-01-21
[Broken Website] The Verge: Link underline is invisible


ajsirota:
### Prerequisites

- [X] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this website issue to avoid opening a duplicate.
- [X] I can reproduce this website issue in a new, unmodified web browser profile with Dark Reader installed as the only extension.

### Website Issue Description

On theverge.com articles with the site's new-ish design, the underline on links is implemented with CSS shadow and is unreadable in Dynamic mode.

### Website Address

https://www.theverge.com/2023/1/19/23562966/cnet-ai-written-stories-red-ventures-seo-marketing

### Steps To Reproduce

1. Find a link
2. Notice there is no underline when not hovering

### Screenshots

In Dynamic mode:
<img width="590" alt="image" src="https://user-images.githubusercontent.com/67526318/213842930-70908347-e5f8-469e-9a5b-ded823763d54.png">

Filter+:
<img width="603" alt="image" src="https://user-images.githubusercontent.com/67526318/213842964-0ed427f6-faf4-48bd-a1c9-764869350eb2.png">

Disabled:
<img width="589" alt="image" src="https://user-images.githubusercontent.com/67526318/213842976-33488612-b826-481e-8826-c2deb2678fa8.png">

### Operating System

macOS 13.1

### Web Browser name and version

Edge 109.0.1518.49 (Official build) (arm64) 

### Dark Reader version

Version 5 preview (4.9.62)

PR #14957, opened 2026-01-19
Use website-themed link colors for The Verge

futurejohn:
Currently, links are not highlighted when viewing articles on The Verge. The Verge uses a custom box-shadow, which Dark Reader seems to have an issue inverting. This CSS uses the box-shadow color and applies it directly to the font color, making links easily visible while maintaining the color scheme of the website.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -31766,10 +31766,10 @@ INVERT
 
 CSS
 .duet--article--article-body-component p a {
-    color: #3cffd0 !important;
+    box-shadow: inset 0 -1px 0 0 ${#131313} !important;
 }
-.duet--content-cards--content-card p a {
-    color: #631aff !important;
+.duet--article--article-body-component p a:hover {
+    box-shadow: inset 0 -0.5em 0 0 ${#3cffd0} !important;
 }
 
 ================================


alexanderby:
Could you please share a sample link with a screenshot to take a look?

futurejohn:
> Could you please share a sample link with a screenshot to take a look?

Yes, here's a link: https://www.theverge.com/news/864229/openai-focus-practical-adoption-sarah-friar

The Verge is semi-paywalled now so you may need to open in an incognito/private window. This example is for a standard article and uses a greenish color for links. The Verge also has "quick posts" that use a purple color scheme, which is why there are 2 classes with different colors for each one. 

Screenshots:

Original 
Original

Default Dark Reader
Default Dark Reader

Dark Reader with proposed fix implemented 
Dark Reader with proposed fix implemented

alexanderby:
I've used `box-shadow` property as it was designed. Unfortunately I couldn't find an element for the second selector. If you could add a similar fix in another PR that would be great.

I will try improving box-shadow support for CSS variables.

futurejohn:
Ok, I can do that. The other selector was for "quick posts" on The Verge which use different CSS classes and a purple color scheme. I'll note that in the next PR.

matteocontrini:
This should close #10668, probably.

PR #14963, opened 2026-01-21

Hamham6:
invert dark image


Hamham6:
Example of page affected: https://www.overleaf.com/learn/latex/Learn_LaTeX_in_30_minutes#Writing_your_first_piece_of_LaTeX
The part under "This example produces the following output: "
Without fix
Screenshot 2026-01-21 at 14 03 36
With fix
Screenshot 2026-01-21 at 14 03 19

PR #14976, opened 2026-01-26
Removed references to gsuite from other Google hosts

Gitoffthelawn:
Likely safe to remove all these given that gsuite is defunct.  You may want to verify, as I only performed a cursory verification, but it's likely safe.


alexanderby:
Thank you, indeed we should remove old mess.

Gitoffthelawn:
> Thank you, indeed we should remove old mess.

You're welcome.  Glad I could help.

PR #14977, opened 2026-01-26
Provide a setIgnoredCSSURLs option

baptistejamin:
Adds a new ignoreStylesheetURLs: string[] option to DarkReader's DynamicThemeFix configuration that allows excluding stylesheets from dark mode processing based on their URL.


alexanderby:
Hi @baptistejamin, do you have an example where it could be useful?

baptistejamin:
Sure!

At Crisp (app.crisp.chat), we are using darkreader as a library. Since stylesheets are converted on the fly there is no direct way to ignore transformations.

For instance, we are having issues with our widget that is getting auto-transformed, while it supports darkmode already. 

ignoreStylesheetURLs is a simple way, to just say darkreader "don't process this please"

alexanderby:
Wonderful, thank you.

PR #14989, opened 2026-02-04
Fixes for BrickLink's Color Guide (v2)

StevenSchifter:
These fixes are intended to prevent Dark Reader from inadvertently styling color swatches in [BrickLink's Color Guide](https://v2.bricklink.com/en-us/catalog/color-guide).


Myshor:
Small fix will be added by me before merge. I will just leave the banner color but make it darker in dark mode. 
obraz

PR #15000, opened 2026-02-08
Fix links and support card on upcloud.com

okuuva:
Gradients on various links caused them to completely disappear unless hovered on. "Scroll to top" link has a shadow that stayed white when in dark mode making it look hideous. And the support card gradient was unnecessarily stripped, looks good as is on both light and dark modes.


okuuva:
https://upcloud.com/docs/guides/tooling/

Before:
image
image
After:
image
image

PR #15003, opened 2026-02-10
Improve dynamic-theme-fixes.config

Gitoffthelawn:
I wasn't feeling well, but I'm not a TV watcher... so I improved `dynamic-theme-fixes.config` for everyone instead.

The goals:  
1. Fix bugs (e.g. syntax errors)  
2. Reduce memory footprint  
3. Improve code consistency  
4. Update to modern CSS syntax  
5. Slightly improve performance

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -367,6 +367,7 @@ discourse.cataclysmdda.org
 discuss.logseq.com
 disneyplus.com
 distro.tube
+distrobox.it
 djflame.tech
 dlive.tv
 docs.adonisjs.com


Gitoffthelawn (commented):
Code checker didn't like spacing on line 4842, so copied spaces from another line to hopefully satisfy it.

alexanderby:
Hi @Gitoffthelawn. I wish you get well soon.

The incorrect syntax fixes are indeed important. Some changes like `transparent -> #ffff` or `50% -> .5`, `0px -> 0`, make it less readable. It might have 0.001% performance benefit, but it also removes some Git functionality, like when you click on a line you can see why it appeared there without digging through diffs. The differences in syntax are due to fixes being usually based on the website's CSS. I would prefer to not enforce a common format for the properties values (but missing spaces are indeed annoying).

The error is due to CSS properties not being sorted alphabetically.

Please take a rest.

PR #15047, opened 2026-02-22
Fix for nbc.com

pnaf:
Fixes slideshow background on home page.

Before:

After:


alexanderby:
Dark Reader mistakenly decided this is text color, need to investigate why...

pnaf:
That commit fixed the issue for me. Thank you!

PR #15048, opened 2026-02-22
Add sub-domain support for baobun.dev in dark-sites.config

tookender:
This is added as baobun.dev has multiple sub-domains (dash.baobun.dev and docs.baobun.dev) which are all dark mode by default. Previously, only baobun.dev was in this list, which only worked on the main site.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -1,4 +1,3 @@
-*.baobun.dev
 *.cases.gg
 *.clash.gg
 *.cncnet.org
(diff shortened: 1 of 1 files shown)


alexanderby:
I'd prefer that the subdomains are explicitly listed, since there are only a few of them. The wildcards make the list harder to maintain.

tookender:
Thanks for the feedback. I have updated the dark-sites.config to list the subdomains (and main site) instead of using wildcards. It now features baobun.dev, dash.baobun.dev and docs.baobun.dev

PR #15071, opened 2026-03-01
Add fixes for map on fuel sites

rakleed:
- https://auto.lukoil.ru/ru/ProductsAndServices/PetrolStations?
- https://azs.tatneft.ru/locator
- https://azs.teboil.ru/map/
- https://www.azsgazprom.ru/azs-map/
- https://kirishiavtoservis.ru/stations/
- https://lk.teboil.ru/
- https://neftm.ru/#gasstations
- https://www.novnp.ru/azs/?f=azs
- https://opti-24.ru/azs/
- https://rosneft-azs.ru/stations
- https://sngbonus.ru/#kartaAZS

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3514,6 +3514,13 @@ button[data-handler="authorise-github"] > img.icon.icon-default
 
 ================================
 
+auto.lukoil.ru
+
+INVERT
+ymaps[class$="ground-pane"]
+
+================================
+
 autodesk.com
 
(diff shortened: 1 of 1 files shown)


Myshor:
Can you add some URLs to this maps? It's hard to find some if you do not understand Cyrillic at all. 😅

Or at least share the screens before and after adding the fix.
As you use INVERT then after in Dark mode should be enough. 😉

Edit: Sorry you shared it already I was looking into URLs into sitefixes file. 😅
Merging. 😉

PR #15085, opened 2026-03-05
invert review stars and fix review text color for archive.org

TipsyTheCat:
what it says on the tin :3
i should mention that i have VERY little html and (even less) css knowledge, so it's rather likely that this isn't perfect. like maybe i should have picked a child class or something. what im saying is please make edits where you see fit.
anyways, it works and i'm not seeing it break anything.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3022,6 +3022,7 @@ p + fieldset a svg
 .activity-icon
 #branding
 .status-text
+.review-star
 
 CSS
 #search-input {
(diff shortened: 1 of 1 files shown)


alexanderby:
Hi @TipsyTheCat, if it worked in the devtools, then it's fine.

TipsyTheCat:
just noticed that i missed the "there are X reviews for this item" text that shows before you show the reviews. it's in a div with the class `message` which feels too generic, but i couldn't find it used anywhere else so i think it's fine. i'll make a new PR for it.
also while trying to see if that class was used elsewhere i found out that the BG isn't changed on alot of pages and on some elements. item pages are fine, this is when browsing. i don't even know where to get started trying to fix these so i'll just make an issue if there isn't one already.

PR #15120, opened 2026-03-16
Fix for ulta.com

insipidlight:
fixes the review stars, nav icons, and qr code

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -33829,10 +33829,16 @@ IGNORE INLINE STYLE
 ulta.com
 
 CSS
-.pal-c-Icon,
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -33826,6 +33826,16 @@ IGNORE INLINE STYLE
 
 ================================
 
+ulta.com
+
+CSS
+.pal-c-Icon,
+.QrCode > svg:nth-child(1) > path:nth-child(2) {
```
alexanderby: Hi, does the fix work? Maybe try using `IGNORE INLINE STYLE` for the QR code selector? I'm worried that black will become white unexpectedly.

Myshor:
Ow... It's doesn't have any change request. I will apply final fixes to the PR instead of pasting solution. 😉

PR #15121, opened 2026-03-17
Fixed orteil.dashnet.org homepage

johnnycharlesw:
I just added in a rule to fix the homepage of `orteil.dashnet.org`, I tested it, it does not break Cookie Clicker it seems.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -37746,3 +37746,20 @@ CSS
 .zyimage {
     background-color: var(--darkreader-neutral-text) !important;
 }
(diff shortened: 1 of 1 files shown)


Myshor:
I tried to apply this "fix" after adding  `!important` to every line but it did not fix anything.

I bet someone is trying to apply background like original:
obraz

I am with big no... but as @alexanderby you started to review this then consider if we want to merge or just reject.

Now the page without any fix looks like below:
obraz

johnnycharlesw:
Well, I tried...I'll just keep it in my fork until it is ready, I guess...

Issue #15090, opened 2026-03-07
[Feature Request] Disable text-shadow effect in Filter mode


Gravemind:
### Prerequisites

- [x] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this feature request to avoid opening a duplicate.

### Is this feature request related to a problem?

Yes

### Feature Request Description

In Filter and Filter+ modes, text appears bolder with a thicker stroke compared to Dynamic mode. This kind of limits how "soft" dark I can go using Filter. I even find bold text can look a bit gritty and with poor kerning.

**Root Cause:** Dark Reader Filter mode adds `html { text-shadow: 0 0 0 !important; }` (see #18 [here](https://github.com/darkreader/darkreader/blob/174cf8fad5bfb77f036c749c3671371a76e63d45/src/generators/css-filter.ts#L84)).

But, I found that a text stroke of 0.1 produces imperceptibly similar results than `text-shadow`.

**Proposition:** So, is it possible to remove the hard-coded `text-shadow` declaration and let users control text appearance via the existing text stroke feature instead ?

### Screenshots

Here are screenshot of github.com (soft dark theme). This is very subtle.

Dark Reader OFF:

<img width="896" height="138" alt="Image" src="https://github.com/user-attachments/assets/aab49a9d-3bad-4b42-8ac3-ab7ae162eb70" />

Dark Reader Filter, no effects:

<img width="896" height="138" alt="Image" src="https://github.com/user-attachments/assets/98ace208-ac7a-456f-8624-094a64e99fea" />

Dark Reader Filter, no effects, text-shadow manually removed - **Looks the same as Dark Reader OFF**:

<img width="896" height="138" alt="Image" src="https://github.com/user-attachments/assets/af8e2fc4-afbe-4043-9f49-c1b2105c884f" />

Dark Reader Filter, stroke +0.1, text-shadow manually removed - **Imperceptibly same as Dark Reader ON with text-shadow**:

<img width="896" height="138" alt="Image" src="https://github.com/user-attachments/assets/074de06b-612b-482a-9704-48d3acf1e204" />

To me, text-shadow looks imperceptibly the same as stroke +0.1.

For comparison, here is Dark Reader Dynamic, brightness +5 - **Stroke looks more like OFF than Filter**:

<img width="896" height="138" alt="Image" src="https://github.com/user-attachments/assets/68cae594-af85-49dc-b19c-a885eacea9ae" />

### Additional Context

[...]

Issue #18, opened 2016-03-05
Text rendering issues


Hexodus:
The text isn't rendered properly IMHO - somehow it's not looking clean and is difficult to read. I even see it on your screenshots. There is a noticeable difference between the font on pages that are designed to be dark and the view created by this extension. It feels like the "text stroke" option is never really disabled even when it tells that it's switched off.

Anyways - this is a very cool and useful extension I've been waiting for. A big thank your for sharing it!

UPDATE:
I just found an easy work around:

	html {
	    text-shadow: 0 0 0 rgba(0,0,0,0.5);
	}

This is also being inverted so the result is higher text contrast - the bitten out parts are being smoothed ;)

alexanderby:
Nice solution, thanks, will be released soon.

Hexodus:
Many thanks ;)

PR #15123, opened 2026-03-18
Remove hard-coded text-shadow in Filter mode

Gravemind:
Remove the hard-coded text-shadow that slightly embolden fonts in Filter and Filter+ modes.

This text-shadow was intentionally added by #18, but didn't leave you the choice to remove it. 

You can now increase **More > Text stroke** to get back the (imperceptibly) same result as the removed text-shadow.

Fix darkreader/darkreader#15090

### DIFF
--- a/src/generators/css-filter.ts
+++ b/src/generators/css-filter.ts
@@ -77,13 +77,6 @@ export function cssFilterStyleSheetTemplate(filterRoot: string, filterValue: str
         lines.push(createTextStyle(config));
     }
 
-    // Fix bad font hinting after inversion
-    lines.push('');
-    lines.push('/* Text contrast */');
-    lines.push('html {');
-    lines.push('  text-shadow: 0 0 0 !important;');
-    lines.push('}');
-
     // Full screen fix
     lines.push('');
     lines.push('/* Full screen */');


alexanderby:
Hi @Gravemind, do you use the Light mode or Dark mode? What's your operating system?

The issue with Filter mode is, it operates pixel by pixel, and when the text gets inverted and the color's hue is restored, the sub-pixel rendering appears inverted:
```
Normal pixels
RGB|RGB|RGB
110|000|011
Yellow-black-cyan makes it look a bolder black line on white background

Inverted pixels
RGB|RGB|RGB
001|111|100
Blue-white-red makes it look a bolder white line on black background

Inverted + Hue restored
RGB|RGB|RGB
100|111|011
A weird red-white-cyan glow appears around the text
```
 
So this forced text shadow with 0 values disables the clear-type (sub-pixel rendering). If you are using the Filter mode in Light mode, we could try removing this forced effect in light mode only.

Gravemind:
I don't have subpixel antialiasing/hinting, so I was not affected. But it makes sense.

I enabled subpixel hinting to check, but I could not reproduce the issue: subpixel hinting seems to already be automatically disabled on Firefox 148 and Chrome 146 as soon as there is a `filter: invert(X)` or `filter: hue-rotate(X)` (maybe the same way it was with `text-shadow`...). I could not find a reference on when this behavior started yet.

@alexanderby, If I find the proper versions, does a version check before adding the text-shadow sound good?

> If you are using the Filter mode in Light mode, we could try removing this forced effect in light mode only.

I do use Dark mode too. I usually use Light mode to adjust existing dark themes.

alexanderby (approved):
I think we are safe to remove that 10 year old fix.

PR #15136, opened 2026-03-22
add play.hbomax.com to global dark list

gordcodes:
The dark list already has hbomax.com but looks like it doesn't work when logged in. When you're logged in it's using play.hbomax.com which does not work on global dark list without wild card.


pnaf:
Have you seen [help.hbomax.com](https://help.hbomax.com/) yet? That site doesn't use a dark theme.

PR #15166, opened 2026-04-06
Fix plaintext files display in Firefox

LogicDaemon:
To me, in Firefox with default settings and Dark Reader using "Filter" mode, all plain text files (ex: <https://raw.githubusercontent.com/darkreader/darkreader/refs/heads/main/src/config/inversion-fixes.config>) display with dark letters on a dark background (black on black basically).
This is a fix for that specific behavior: 1st non-class &lt;pre&gt; element on a plain html page (which is created by Firefox to display plaintext files) is inverted. Background stays dark but the text is readable now.

### DIFF
--- a/src/config/inversion-fixes.config
+++ b/src/config/inversion-fixes.config
@@ -16,6 +16,7 @@ twitterwidget
 .sr-backdrop
 iframe:fullscreen
 [class="opinary-iframe"]
+html > body > pre:nth-child(1):not([class])
 
 NO INVERT
 [style*="background:url"] *


alexanderby:
Could you please share a link for an incorrectly looking page?

LogicDaemon:
> Could you please share a link for an incorrectly looking page?

It's in the original message: 
> (ex: https://raw.githubusercontent.com/darkreader/darkreader/refs/heads/main/src/config/inversion-fixes.config)

Here are screenshots
image
image

alexanderby:
Looks like dark theme detection didn't work for some reason. Let me think what's broken in there.

LogicDaemon:
> Looks like dark theme detection didn't work for some reason. Let me think what's broken in there.

It's like that for all plaintext files for many years, and I've been using the fix all that time. It's just after every update I have to re-apply it so I decided to finally create the PR :)

YagUber:
Would the correct fix not be do an early return if the content type is `text/plain`?

```diff
--- a/src/inject/detector.ts
+++ b/src/inject/detector.ts
@@ -137,6 +137,12 @@ function canCheckForStyle() {

 export function runDarkThemeDetector(callback: (hasDarkTheme: boolean) => void, hints: DetectorHint[]): void {
     stopDarkThemeDetector();
+
+    if (document.contentType === 'text/plain') {
+        callback(isSystemDarkModeEnabled());
+        return;
+    }
+
```

LogicDaemon:
> Would the correct fix not be do an early return if the content type is `text/plain`?
> 
> ```diff
> --- a/src/inject/detector.ts
> +++ b/src/inject/detector.ts
> @@ -137,6 +137,12 @@ function canCheckForStyle() {
> 
>  export function runDarkThemeDetector(callback: (hasDarkTheme: boolean) => void, hints: DetectorHint[]): void {
>      stopDarkThemeDetector();
> +
> +    if (document.contentType === 'text/plain') {
> +        callback(isSystemDarkModeEnabled());
> +        return;
> +    }
> +
> ```

maybe, I did not read the code at all. I just shared the workaround I've been using for years.

alexanderby:
Firefox reports the root background to be transparent. Fixed now (after an update, probably next week).

PR #15180, opened 2026-04-11
Remove ecosia.org from dark sites

pnaf:


Myshor:
Before I merge I need to add this site to detection hints as it seems to use dark mode when Dark is set in system/browser. :)

PR #15187, opened 2026-04-12
feat(detector-hints): system-aware theme detection and config hygiene

YagUber:
## Summary

This PR adds system theme aware detector hint rules, cleans up a significant number of broken config entries caught by new automated validation, and rewrites the contributing guide from scratch. 

This was done as a response to #15155 and as a way of apologizing to @Myshor for the extremely negligent github detector hint I added that caused the regression he had to fix. Sorry!

### `feat(detector-hints)`: System theme aware detection

Adds two new detector hint commands:

- `MATCH SYSTEM DARK`
  - applies selectors only when `prefers-color-scheme: dark`
- `MATCH SYSTEM LIGHT`
  - applies selectors only when `prefers-color-scheme: light`

These are intended for sites that follow the system theme but store separate values for each state, for example, a site that keeps `data-color-mode="auto"` fixed while separate `data-dark-theme` / `data-light-theme` attributes determine the active palette. A plain `MATCH` on those attributes would fire regardless of system theme, causing false positives. `MATCH SYSTEM DARK` and `MATCH SYSTEM LIGHT` handle this correctly.

The GitHub detector hint has been [updated](https://github.com/YagUber/darkreader/blob/724e6ed578e31ddfbad70d8d951d85349343d3b0/src/config/detector-hints.config#L450-L463) to use these new rules, replacing the previous `[data-color-mode="dark"]` selector which only caught the explicitly dark case and missed the system aware auto mode entirely.

### `fix(dynamic-theme-fixes, inversion-fixes)`: Remove invalid and no-op CSS rules

The new config validation (see below) caught a number of entries that were silently broken:

- **Invalid `${...}` color templates**
  - Several entries used malformed syntax like `${fff}` (missing `#`), `$var(...)`, `$(...)`, or nested `${${...}}`. These were never applied.
- **Invalid plain CSS**
  - Entries with syntax errors such as `; !important` (semicolon before `!important`), invalid selector syntax like `div[dir="ltr", id="map"]`, and malformed property values like `--darkreader-bg--charcoal-gray-20:` used as a color. These were also no-ops.
- **Dynamic-theme-only variables in `inversion-fixes.config`**
  - `var(--darkreader-*)` custom properties and `${COLOR}` templates are only meaningful in dynamic theme mode. Several entries in `inversion-fixes.config` used them, where they had no effect.

All removed entries were verified to be genuinely broken rather than intentional.

### `chore(tests)`: Config test rewrite

The existing config tests were a flat collection of loosely related assertions that were difficult to read, hard to extend, and provided limited signal when they failed.

This rewrites them using `describe`/`it` blocks with shared `beforeAll` setup, a custom `toBeFormattedAs` matcher, and uses the existing `multiline` helper that eliminates the `.join('\n')` boilerplate that made inline test configs hard to read.

The rewrite also adds new validation that wasn't present before:

- **CSS validation**
  - For both `dynamic-theme-fixes.config` and `inversion-fixes.config`, using `css-tree` to parse each `CSS` block and validate that any `${...}` color templates contain a valid CSS color value.
- **Dynamic-theme-only pattern detection**
  - For `inversion-fixes.config`, rejecting `var(--darkreader-*)` and `${COLOR}` templates which are meaningless in filter mode.
- **Stricter `detector-hints.config` validation**
  - Including checks that `TARGET` is always paired with at least one `MATCH` variant, that `NO DARK THEME` and `SYSTEM THEME` are never combined with match rules, and that the new `matchSystemDark` / `matchSystemLight` fields are included in the comma selector checks.
- **Unit tests for all four parsers** (`parseDetectorHints`, `parseDynamicThemeFixes`, `parseInversionFixes`, `parseStaticThemes`)
  - Covering empty input, multi-URL blocks, multiple entries, unsupported commands, and edge cases.

> **Note:** while writing these tests I noticed that optional fields on `DetectorHint` and `DynamicThemeFix` are not typed as nullable, which caused some friction. I didn't fix this because doing it properly requires touching internal null handling code and I didn't want to risk affecting production behavior in a larger way than I already had, potentially worth correcting.

### `docs(CONTRIBUTING)`: Full rewrite

The existing CONTRIBUTING.md covered the basics but left a lot of gaps that made it harder than it needed to be to make a first contribution. I spent considerable time on this rewrite and think it meaningfully improves the onboarding experience, I wish it had existed when I started.

Highlights:
- A navigation section at the top that routes contributors to the right section for their use case.
- Expanded and corrected documentation for disabling Dark Reader on a site, including framework specific examples (React, Next.js App Router, Next.js Pages Router, Vue, Nuxt, Svelte), a detailed explanation of how the `MutationObserver` works and what can go wrong, and a recommendation to prefer the standard `color-scheme` meta tag over the proprietary `darkreader-lock` where possible.
- Full documentation of the new `MATCH SYSTEM DARK` / `MATCH SYSTEM LIGHT` rules and when to use them, including a worked example.
- Improved rule reference tables for `dynamic-theme-fixes.config`, `inversion-fixes.config`, and `detector-hints.config`.
- Clearer documentation of `${COLOR}` template behavior, including how Dark Reader decides which transformation to apply based on luminance.
- Updated development setup instructions covering the MV2 and MV3 Chromium build targets and Thunderbird.

> **Note:** I included a note as a comment in the file flagging an open question about whether `darkreader-lock` and `runCheck` automatic detection is genuinely intended to be dynamic-theme only. I couldn't come up with a reason as to why it is.
> Additionally, I wasn't sure how deep to get into some things, such as the plethora of npm scripts, since I wanted it to be somewhat beginner friendly, I kept it pretty light.

You can preview the rendered version here: [CONTRIBUTING.md](https://github.com/darkreader/darkreader/blob/2de7918c4b7fab828ba7823bb6e9b72869dcd48a/CONTRIBUTING.md)

If the code changes need more review time, I'm happy to split this into a separate PR so the docs can be merged independently.


Myshor:
Not looking deep into the code because I already said that when I focus on CSS fixes here I can do reviews for config files only.

I have 2 questions:

1. Why do we need MATCH SYSTEM LIGHT ?? In most cases if user won't set manually dark mode then if site uses option to autodetect it will be light not dark mode.
2. I saw you have differences in it but was you able to see data-light-theme="dark" in browser developer tools? Because I was not.

Summarizing. I think only MATCH SYSTEM DARK would be enough. Because you would check for `html[data-color-mode="auto]` if system is dark.

Last thing. Is this MATCH SYSTEM DARK is the same bool used for detecting system theme used here?
obraz

If yes then that really could be useful because not only github has system preferenced setting by default and that could be the possibility to add more pages in detection hints.

P.S. Please consider if we really need to have 2 new matches.

YagUber:
To answer your first question:

<details>
<summary>Justification for two new matchers</summary>
<ul>

The strongest justification is GitHub itself, and the reason I am arguing both rules are needed is not purely for symmetry, but GitHub’s actual state model.

The important distinction is that `data-light-theme` / `data-dark-theme` are not “the page is light” / “the page is dark”. They are the stored theme selections for the two `data-color-mode="auto"` branches.

So when `data-color-mode="auto"`:
- If the OS theme is dark
  - GitHub applies `data-dark-theme`
- If the OS theme is light
  - GitHub applies `data-light-theme`
<br>
And those values are not restricted by branch name. The selected theme in either branch can itself be a dark or light variant.

So these are both valid dark-theme cases that need different detection:
- `prefers-color-scheme: dark` + `data-color-mode="auto"` + `data-dark-theme^="dark"`
- `prefers-color-scheme: light` + `data-color-mode="auto"` + `data-light-theme^="dark"` 
<br>
That is why I added both <code>MATCH SYSTEM DARK</code> and <code>MATCH SYSTEM LIGHT</code>.
The system theme decides which stored branch is active, but the branch value decides whether the active GitHub theme is dark.
<br><br>
I explain this in slightly more detail in the answer below.

</ul>
</details>

To answer your second question (using GitHub's [Appearance](https://github.com/settings/appearance) page):

<details>
<summary>GitHub in practice</summary>
<ul>

(skip to the bottom for the screenshots of the browser developer tools)

In practice GitHub has the following internal list of themes:

```
light
light_colorblind
light_tritanopia
light_colorblind_high_contrast
light_tritanopia_high_contrast
dark
dark_colorblind
dark_tritanopia
dark_dimmed
dark_colorblind_high_contrast
dark_tritanopia_high_contrast
dark_dimmed_high_contrast
```

In "Single Theme" mode any dark theme you choose will set `html[data-color-mode="dark"]`, and any light theme you choose will set `html[data-color-mode="light"]`, making detection trivial with the current matching system.

However, when in "Sync with system" mode, you are given the ability to set an individual theme per system theme. Such that 
```css
html[data-color-mode="auto"]
html[data-light-theme="[any_from_above]"]
html[data-dark-theme="[any_from_above]"]
```

This already breaks the [current rule](https://github.com/darkreader/darkreader/blob/main/src/config/detector-hints.config#L460-L466) in the production Dark Reader, which only matches `html[data-color-mode="dark"]`, just because we have the "Sync with system" mode enabled.

The issue with any potential fixes arises when you want to actually determine whether the current mode is either light, or dark. Now, we could match against styles selectively applied depending on each theme, but that would just be sidestepping the functionality that currently does not exist in Dark Reader with a more fragile alternative. The functionality that does not exist is the ability to detect this pattern:
```scss
html[data-color-mode="auto"] AND (prefers-color-scheme: dark) AND html[data-dark-theme^="dark"]
OR
html[data-color-mode="auto"] AND (prefers-color-scheme: light) AND html[data-light-theme^="dark"]
```

Dark Reader's detector-hints (and the other configs matchers by extent) are doing a simple `Element.matches` on each list item under `MATCHES`, meaning we don't have access to the prefers-color-scheme media feature, since there is no `:prefers-dark` pseudo class or selector equivalent.

<details><summary>Images of example in the browser developer tools</summary>
<p>
<ul>
image

image

Why would anyone do this? I really don't know. But I am sure this pattern can be extended to many other sites.

</ul>
</p>
</details> 

</ul>
</details>

To answer your final question about what MATCH SYSTEM DARK uses to detect the theme:

Both the automation detection (the one in your screenshot) and the new matching system introduced, are determined by `matchMedia('(prefers-color-scheme: dark)').matches`.

Besides that, hopefully this clears things up, let me know if any further clarification is needed.

YagUber:
Quick answer:

`html[data-color-mode="auto]` cannot be the only match because you can select a non-dark theme that GitHub will selectively apply when the system theme is set to dark.

No clue why anyone would do this, but its a possibility.

And yes, both systems use the same bool for detecting the system theme.

Myshor:
Ok... Checked github settings and confirm there is possibility to set light mode for dark mode detected. 😕
Am I the only one who sees that UX experience is going wrong with every update of every app/site? 🙄

alexanderby:
Oh my days, a new wave of AI slop is coming. @Myshor That looks like a bot to me. Please be aware if you see any changes to `package.json` or `package-lock.json` files. There were recent attempts to hijack dev accounts, like [that one](https://socket.dev/blog/axios-maintainer-confirms-social-engineering-behind-npm-compromise).

PR #15189, opened 2026-04-12
Add new rules for dynamic theme fixes for https://www.nytimes.com/puz…

colin-tso:
…zles/spelling-bee*

Before

image

After

image


Myshor:
Idea was good but not working at all for me and when I removed * from the URL it started to work but broke the rest of fixes applied to nytimes.com.

Moving this invert to nytimes.com and then I will merge.

PR #15222, opened 2026-04-18
Add CSS fixes for dynamic themes for *.autotask.net

redistributor:
Added primary and hover background colors to fix issue with menus being displayed with transparent background.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -266,10 +266,10 @@ body {
 
 CSS
 .bg-background-primary {
-    background-color: #1a1d1e !important;
+    background-color: ${rgb(255 255 255)} !important;
 }
 .bg-background-hover {
-    background-color: #1f2b34 !important;
+    background-color: ${rgb(207, 217, 226)} !important;
 }
 
 ================================
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -262,6 +262,18 @@ body {
 
 ================================
 
+*.autotask.net
+
+CSS
+.bg-background-primary {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -262,6 +262,18 @@ body {
 
 ================================
 
+*.autotask.net
```
alexanderby, line 265: Is there a way to use a fixed domain name (without "*" wildcard)?

src/config/dynamic-theme-fixes.config
```diff
@@ -262,6 +262,18 @@ body {
...
+
+CSS
+.bg-background-primary {
+    background-color: #1a1d1e !important;
```
alexanderby: These fixed colors might break in Light mode and are not adjustable. I'd suggest to use something like
```
background-color: ${rgb(10 10 10)} !important;
```

Where `rgb(10 10 10)` is a source color value, that's going to be modified.

alexanderby (changes requested):

src/config/dynamic-theme-fixes.config
redistributor: Autotask uses 18 different subdomains, depending on hosting region, so the wildcard is useful here. [See here](https://ww5.autotask.net/help/Content/4_Admin/TechnicalStuff/Allowlist.htm?highlight=Zone) for listing of subdomains.

src/config/dynamic-theme-fixes.config
redistributor: I've changed to rgb values and confirmed that it works in light mode as well.

alexanderby (approved):

PR #15227, opened 2026-04-20
Fix for star-history.com chart images

joganubaid:
Fix star-history.com embedded chart images in dark mode.

Star-history.com provides an API that generates SVG charts. When embedded on other sites (like GitHub README), these charts appear as light images on dark backgrounds, making them hard to read.

I was browsing trending GitHub repos at 2 AM (as every good developer does) and noticed star-history charts showing up as bright white blobs. After some investigation, I realized these charts come from api.star-history.com with a data-canonical-src attribute.

This fix:
- Inverts images from api.star-history.com that don't already have theme=dark
- Uses CSS :not() selector to avoid double-inversion for images that already use the dark theme parameter

For developers who somehow missed that theme=dark exists in the API (seriously, it's right there in the README) - your charts will now be readable in dark mode too.


alexanderby:
Could you please share an example? Is it worth having this fix for GitHub only, to avoid applying it for every website?

PR #15229, opened 2026-04-21
Add developers.openai.com to dark sites config

xiangyann:
Default is dark on light sys theme.


alexanderby:
Thank you for the proposal. This website offers both dark and light themes, and the dark theme detection works as expected. So it doesn't fit for this list.

PR #15242, opened 2026-04-27
Add puuilo.fi

okuuva:
puuilo.fi sets `background-color: #F8F6F4 !important` inline on the search input, which Dark Reader cannot override via standard `background-color` rules. As a result the input keeps the light background while the rest of the site is dark, leaving the typed text and placeholder near-invisible.

Fix: paint the background via `background-image` linear gradient (not locked by inline style) using `--darkreader-neutral-background`, and set text/icon colors via `--darkreader-neutral-text`. Works in both dark and light DR schemes.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -27142,6 +27142,19 @@ CSS
 
 ================================
 
+puuilo.fi
+
+CSS
+#customSearchInput {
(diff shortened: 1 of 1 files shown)


okuuva:
# Before:
## Light:
image
image

## Dark:
image
image

# After:
## Light:
image
image

## Dark:
image
image

PR #15261, opened 2026-05-02
Fix Petdex sprite rendering

zachvivier:
When Dark Reader is enabled, sprites disappear on Petdex. This proposes a dynamic-theme fix.

Preserves `.pet-sprite` inline styles and image analysis so WebP spritesheets remain visible in Dark Reader dark mode.

Before
image

After
image


zachvier:
https://petdex.crafter.run/#gallery

PR #15283, opened 2026-05-07
Fixes for admin.shopify.com

TheCreamGravy:
Wow, baby's first pull request. Apologies if anything is not as it should be!

Fixes for buttons and sections within the Admin page for Shopify stores. Customer information has been removed for screenshot purposes.

Dark Reader, Before Patch: 

image

image

Dark Reader, After Patch:

image

image

Also checked in light mode, and it works just fine. :)

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -1653,6 +1653,21 @@ CSS
 
 ================================
 
+admin.shopify.com
+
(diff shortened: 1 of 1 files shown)


Myshor:
Not sure how setting background should help with darker text but this part of fix at least removes some white backgrounds.

If more fixes will be needed then new PR can be added later.

PR #15298, opened 2026-05-09
Fix for nytimes.com

pnaf:
Fixes logo on most Wirecutter pages.


Myshor:
Can you share some URL to check or at least screens before/after? Because "fix" with remove of something that was in the sitefixes is not what I would like to merge just like that.

pnaf:
Sorry about that! Example URL is https://www.nytimes.com/wirecutter/. Most linked articles within that page should also work.

Deleting the sitefix "fixes" the site for me on Firefox. Let me know if you still need screenshots.

PR #15315, opened 2026-05-12
Fix dynamic theme styles in app.hibob.com

louis6321:


Myshor (changes requested):
Thanks for the PR. However, I noticed a lot of redundant selectors. Using triples like .b-body, b-body, [b-body] is a 'shotgun approach' that adds unnecessary bloat to the config and looks like it was blindly generated.
Please clean up the selectors to be more specific and efficient.

Also, I cannot access the login area of this site, so please provide 'before' and 'after' screenshots to verify the changes. I'll review it again once these are provided.

louis6321:
@Myshor Unfortunately I can't provide screenshots, I'll get in big trouble as the pages contain private company data.

Myshor:
Since this platform is restricted to corporate environments and we cannot access it to verify the changes visually, we are unable to review or maintain this configuration long-term. In this case, it's best to use Dark Reader's Dev Tools to add these specific fixes locally to your own extension. Thank you for your effort, but I'm closing this PR to avoid adding unmaintainable code to the project.

PR #15323, opened 2026-05-15
Update dark-sites.config with domain aerography.eu

SeDi343:


Myshor:
Thank you for the PR, but I'm closing this.

The website already natively supports both light and dark modes using prefers-color-scheme and the color-scheme meta tag. As you can see here:

obraz
(The site is perfectly light when the browser is in light mode)

obraz
(Meta tags and CSS logic are handled natively)

Because of this implementation, Dark Reader handles the site automatically as long as the 'Detect Dark Theme' option is enabled:

obraz

Since the automation already covers this behavior, adding the site manually to the list is redundant. Closing.

P.S. If this is your own website and you have the ability to modify the code, I suggest adding a `.dark` or `.light` class to the `<html>` tag. If you do that, we can consider adding the site to `detection-hints`. This would allow the site to trigger native Dark Mode regardless of whether the 'Detect Dark Theme' option is enabled in the user's settings.

SeDi343:
Hey @Myshor yes I own the website. I added a script that dynamically loads `class="dark"` or `class="light"` into the `<html>` tag.

SeDi343:
Should the `detection-hints` file contain following content additionally?

```

aerography.eu

TARGET
html

MATCH
.dark
```

SeDi343:
Created https://github.com/darkreader/darkreader/pull/15337

PR #15329, opened 2026-05-15
Add reset site settings button to popup

02H20:
This PR introduces a "Reset" button in the popup interface to allow users to quickly clear custom settings for the current website and revert to global theme settings.

### DIFF
--- a/src/_locales/en.config
+++ b/src/_locales/en.config
@@ -91,6 +91,9 @@ Only for
 @only_for_description
 Apply settings to current website only
 
+@reset_site_settings
+Reset site settings
+
 @site_list
 Site list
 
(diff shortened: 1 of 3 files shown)


Myshor:
I won't review the code as for that part of project @alexanderby is main maintainer and main developer.

However I would like to see the screens of this change. I use new prototype design and it has reset settings button already 😋 so I think you are trying to add similar (or even the same) button somewhere to old (default for most of the users) design.

Can you share screens where this button is located after your changes and how it looks like?

alexanderby:
The "Only for" button acts like a toggle button. Clicking it again resets the theme for the website.

Myshor:
> The "Only for" button acts like a toggle button. Clicking it again resets the theme for the website.

Still it leaves the values untouched.

Compared to "Reset to defaults" in preview design which resets evert value to default.
obraz
After click "Reset to defaults:
obraz

That's why I asked for presentation how and where this button is placed into the popup. 🙂 Without visualisation I do not even look deeper into the code. 😏

PR #15335, opened 2026-05-18
Dynamic theme fixes for freecommander.eu (previously freecommander.com)

Gitoffthelawn:
The previous fixes (which were mostly written by me) were for the FreeCommander English forums.  Those forums recently moved to freecommander.eu, so I updated the TLD.

Also, some of the fixes don't seem to be needed any longer, so I removed them.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -13592,17 +13592,13 @@ CSS
 
 ================================
 
-freecommander.com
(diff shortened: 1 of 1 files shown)


Myshor:
Good you mentioned forums as I was almost rejecting why you remove .com site when it still works. 😉

Gitoffthelawn:
@Myshor Yes, I figured this one needed a good explanation! 😄

PR #15337, opened 2026-05-18
Update detector-hints.config with aerography.eu

SeDi343:


Myshor:
I had no time to check it earlier but now everything is fine. Merging. 😉

Issue #15346, opened 2026-05-19
[Broken Website] CSS attribute selectors supercede IGNORE INLINE STYLE on *.dreamwidth.org


pinions:
### Prerequisites

- [x] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this website issue to avoid opening a duplicate.
- [x] I can reproduce this website issue in a new, unmodified web browser profile with Dark Reader installed as the only extension.

### Website Issue Description

Global wildcard selectors like "*[style*="background:#fff"]" and "*[style*="color:#000"]" override all code on the page, even if it's explicitly told not to with IGNORE INLINE STYLE. This frequently creates unreadable text.

I have encountered this on other sites, namely news, but I can't remember which and didn't realize what was causing it at the time.

### Website Address

https://iyashikeimemes.dreamwidth.org/4549.html?view=top-only

### Steps To Reproduce

Load the page with Dark Reader in dynamic mode.

### Screenshots

<img width="816" height="589" alt="Image" src="https://github.com/user-attachments/assets/827f9ba2-0138-4087-b1cd-7bb01f3c263d" />

### Operating System

Windows 11

### Web Browser name and version

Firefox 150.0.3, Edge 148.0.3967.70

### Dark Reader version

4.9.125

### Additional Context

Another example: https://buttrinthum.dreamwidth.org/34838.html?view=top-only

<img width="845" height="312" alt="Image" src="https://github.com/user-attachments/assets/ed846b23-a5a7-4193-9889-ad6de0ea33fa" />

Myshor:
As some people force colors it's not that easy for Dark Reader to change them. Sometimes we can fix them with forcing our colors in sitefixes sometimes not.

At least for two examples you shared here I was able to force text colors and backgrounds that was acting ugly because color forcing made by the users.

I will put fixes in a minute.

Myshor:
Sitefix added.

- If you want to have this fix faster you can set **Synchronizee sites fixes** option to check in Advanced section of Settings.
![obraz](https://github.com/user-attachments/assets/9dea8b66-b77a-46ab-8aa6-e50d995718c3)
- With that enabled newest sitefixes will be updated every browser start. 😉
- To open all settings you must click on:
(Old/classic design) More -> All Settings
![obraz](https://github.com/user-attachments/assets/8c17b1f1-e85e-4672-8a88-85930e178e02)
(New/Preview design) Settings in popup
[...]

PR #15359, opened 2026-05-21
fix for dreamwidth.org

pinions:
Selects problematic colors and sets a unified background and text color (current fix is general enough it breaks some backgrounds / images). Also updates some class selectors to avoid odd behavior in other site skins.


Myshor:
Checked again the sites you mentioned in #15346 

Merged, thanks for clean up!

PR #15360, opened 2026-05-21

alfema-esp:
The goal is for the banner containing the advertisement to fit the available space on the left and not overlap with the right side of the page. Also, if there's no banner ad, the div shouldn't expand.

1_without_DR
2_with_DR
3_with_filter

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -21692,6 +21692,17 @@ CSS
 
 ================================
 
+meneame.net
+
+CSS
+.ads-interlinks {
+    display: flex !important;
+    height: auto !important;
+    width: auto !important;
+}
+
+================================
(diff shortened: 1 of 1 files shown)


Myshor:
I know we were talking about it in discussions. But now as I understand this banner overlap the right side even when Dark Reader is not applied? If yes then I am not sure we should interfere in site code.

2nd thing. I disabled uBlock Origin for a second a tried to reproduce your problem you try to fix and wasn't able to reproduce in on desktop. And when I switched to mobile view I see it's how the page is generated by default.
obraz


We use sitefixes to make fixes for elements that are not visible because Dark Reader's changes applied - not making fixes to site bugs. 😉

So it's nice you wanted to fix this issue but I am sorry to say we won't include this fix as it's not our bug to be fixed.

PR #15396, opened 2026-05-24
Fix system theme on syzygy-tables.info

kirelagin:
The website follows the system theme correctly, however dark theme detection does not work for it for some reason.

### DIFF
--- a/src/config/detector-hints.config
+++ b/src/config/detector-hints.config
@@ -1019,6 +1019,12 @@ NO DARK THEME
 
 ================================
 
+syzygy-tables.info
+
+SYSTEM THEME
+
+================================
+
 t-j.ru
 
 TARGET


src/config/detector-hints.config
```diff
@@ -1019,6 +1019,12 @@ NO DARK THEME
 
 ================================
 
+syzygy-tables.info
```
Myshor, line 1022: I cannot add an unreachable site to the configuration. syzygy-tables.info is currently returning a 500 Internal Server Error (Server got itself in trouble). We don't merge sitefixes or hints for sites that are down, as they cannot be verified. Please update this PR once the server is back online and testable.
obraz

Myshor (changes requested):
Not merging without possibility to test.

src/config/detector-hints.config
kirelagin: It is back up, check again please. (It looks like they had some issue with certificate renewal on the backend.)

Myshor (approved):

PR #15449, opened 2026-05-26
Add new websites to global dark list

arturstat:
Add the following websites to the global dark list:

https://homicide.igarape.org.br/
https://ncased.com/

And the following websites from streaming services:

https://music.apple.com/
https://globoplay.globo.com/
https://www.rakuten.tv/

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -792,7 +792,6 @@ mstdn.social
 mtv.com
 musedash.moe
 music.com
-music.apple.com
 music.youtube.com
 muttwizard.com
 mutualaiddisasterrelief.org
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -534,6 +534,7 @@ git.gnous.eu
 github1s.com
 gitkraken.com
 gitmoji.kaki87.net
+globoplay.globo.com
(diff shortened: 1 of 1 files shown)


src/config/dark-sites.config
```diff
@@ -790,6 +792,7 @@ mstdn.social
 mtv.com
 musedash.moe
 music.com
+music.apple.com
```
Myshor: Apple music is light/dark depends on system theme.

Will be needed to be removed.

Myshor (commented):
music.apple.com must be removed I will add it to detection-hints just after merging this.

src/config/dark-sites.config
arturstat: I did not know a SYSTEM THEME existed in the hints.
That will solve one other issue that i have.
Thank you.

Myshor: I found this TAG today during discussion with some other user. And after PR which he created. May be helpful for some sites which do not have any manual theme change possible and no information about change in html code. 😉

arturstat: @Myshor 

This website might be of interest:
https://deepwiki.com/darkreader/darkreader/4.5-dark-theme-detection

PR #15423, opened 2026-05-25
Update adguard-dns.io dark theme detection

arturstat:
Fix #15395

PR #15452, opened 2026-05-26
fix theme detector for adguard sites

Myshor:


Myshor:
Created because I had some issue in [##15423](https://github.com/darkreader/darkreader/pull/15423)

But it seems that github has issues.
obraz

PR #15478, opened 2026-05-28
Add dark theme detection for archdaily.com

arturstat:
Website uses the system theme:

https://www.archdaily.com/

### DIFF
--- a/src/config/detector-hints.config
+++ b/src/config/detector-hints.config
@@ -163,6 +163,12 @@ MATCH
 
 ================================
 
+archdaily.com
+
+SYSTEM THEME
+
+================================
+
 arstechnica.com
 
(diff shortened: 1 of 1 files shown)


arturstat:
Dark reader is detecting dark theme inconsistently on this website.
This will fit it.

src/config/detector-hints.config
```diff
@@ -163,6 +163,12 @@ MATCH
 
 ================================
 
+archdaily.com
+
+SYSTEM THEME
```
Myshor: I know it's nice tag to use but still if site has possibility to check in site code we preffer code check.

Myshor (commented):
I will make change to the PR and merge. 😉

PR #15494, opened 2026-05-29
Add rog-forum.asus.com to global dark sites list

arturstat:


Myshor:
It's almost full dark... I found difference without Dark Reader enabled. Dropdown menus:
obraz

Dark Reader's dynamic engine handles this better than the native site currently does.
obraz

 Closing this for now – global list is for 100% dark sites only.

arturstat:
Actually, this one looks better with dark reader enabled!
Sorry for that.

Myshor:
No need to sorry. If I would be QA Engineer in commercial jobs I would not catch that too... First look - the whole page is dark. 😉

PR #15500, opened 2026-05-29
Fix for zi.tools

lumynou5:
Some images being not inverted are intended.

BTW, I noticed that the global fix has CSS rule applying to `[style*="color: rgba(0, 0, 0,"]`, which is what this patch mainly fixes.  I wonder the initial intent of that because some websites (such as this) uses a color with alpha 0 to hide things, and the rule ignores the alpha channel.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -38639,13 +38639,19 @@ div.mt5
 zi.tools
 
 INVERT
-button.ant-switch-small
+button.ant-switch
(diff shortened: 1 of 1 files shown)


Myshor:
`*` section fixes are shared to all URLs and thunderbird emails. So I am almost sure it is some part of fix to emails where there was some color force made in html.

I will add exclude for (0,0,0,0) and (0, 0, 0, 0) in global section just after merging your code. 😉

PR #15511, opened 2026-05-30
Add dynamic theme fixes for www-awu.aleks.com

rteafischl:
image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -37455,6 +37455,15 @@ INVERT
 
 ================================
 
+www-awu.aleks.com
+
+CSS
+div > .ansed_root > .ansed_canvas {
+    filter: invert(70%);
+}
+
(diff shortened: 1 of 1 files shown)


Myshor (changes requested):
Request changes to block merge.

I need an example URL where I can test your fix. If you can't share the URL, I need to see screenshots for these 4 scenarios:

1. No Dark Reader applied.
2. Dark Reader **without** fix.
3. Dark Reader **after** fix.
4. Dark Reader **after** fix using **light scheme**.

My biggest suspicion is that your manual `filter: invert(70%)` will break the light scheme.

Can you check if using the native `INVERT` command would be enough instead?

```
www-awu.aleks.com

INVERT
div > .ansed_root > .ansed_canvas

```

rteafischl:
Heres an image of the applied change with INVERT.
image

Myshor (approved):

PR #15530, opened 2026-06-01
Add dynamic theme fixes for knowablemagazine.org

arturstat:
See -> https://knowablemagazine.org/

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -18738,6 +18738,44 @@ svg[class$="logo-image"]
 
 ================================
 
+knowablemagazine.org
+
+INVERT
+img[alt="Knowable Magazine | Annual Reviews"]
+img[alt="Knowable Magazine"]
+
+CSS
+html[data-darkreader-scheme="dark"] div.category-header {
+    overflow: hidden;
+    position: relative;
+}
+html[data-darkreader-scheme="dark"] div.category-header::before {
+    background: rgba(0, 0, 0, 0.65);
+    content: "";
+    inset: 0;
+    position: absolute;
+    z-index: 0;
+}
+html[data-darkreader-scheme="dark"] div.category-header h1 {
+    position: relative;
+    z-index: 1;
+}
+html[data-darkreader-scheme="dark"] div.standard-content-hero {
+    background-position: center;
+    background-size: cover;
+    overflow: hidden;
+    position: relative;
+}
+html[data-darkreader-scheme="dark"] div.standard-content-hero::before {
+    background: rgba(0, 0, 0, 0.65);
+    content: "";
+    inset: 0;
+    position: absolute;
+    z-index: 0;
+}
+
+================================
+
 knowyourmeme.com
 
 INVERT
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -18738,6 +18738,44 @@ svg[class$="logo-image"]
 
 ================================
 
+knowablemagazine.org
+
+INVERT
+img[alt="Knowable Magazine | Annual Reviews"]
```
arturstat, line 18744: Inverts the logos on top and middle of the page.
https://knowablemagazine.org/

Image

Image

src/config/dynamic-theme-fixes.config
```diff
@@ -18738,6 +18738,44 @@ svg[class$="logo-image"]
...
+img[alt="Knowable Magazine"]
```
arturstat, line 18745: Inverts the logo near the bottom left of the page.
https://knowablemagazine.org/

Image

src/config/dynamic-theme-fixes.config
```diff
@@ -18738,6 +18738,44 @@ svg[class$="logo-image"]
...
+    background: rgba(0, 0, 0, 0.65);
+    content: "";
+    inset: 0;
+    position: absolute;
+    z-index: 0;
+}
+html[data-darkreader-scheme="dark"] div.category-header h1 {
+    position: relative;
+    z-index: 1;
+}
```
arturstat: The purpsoe of code from line 18748 to line 18762 is to darken the area near the top of the following and similar pages. The objective is to increase the contrast between the background and the text, so that readability is improved.
https://knowablemagazine.org/content/health-disease

Image

src/config/dynamic-theme-fixes.config
```diff
@@ -18738,6 +18738,44 @@ svg[class$="logo-image"]
...
+    overflow: hidden;
+    position: relative;
+}
+html[data-darkreader-scheme="dark"] div.standard-content-hero::before {
+    background: rgba(0, 0, 0, 0.65);
+    content: "";
+    inset: 0;
+    position: absolute;
+    z-index: 0;
+}
```
arturstat, line 18751: The purpose of the code from line 18763 to line 18775 is to darken the area of the following and similar pages, to be consistent with the similar area addressed on the lines of code discussed in the previous comment.
https://knowablemagazine.org/about-knowable-magazine

Image

src/config/dynamic-theme-fixes.config
Myshor: In most cases like this I use text-shadow to make text more visible. Smaller fix is easier to mantain:
```
div.category-header h1 {
    text-shadow: ${rgb(225, 212, 201)} 2px 2px 2px !important;
}
```
obraz

src/config/dynamic-theme-fixes.config
Myshor: As I am not darkening previous image I would not touch this either.

When doing sitefixes we are trying to make them as short as possible and fix only things that are really needed to be fixes as not visible or hard to read.

Keep in mind making bigger CSS fixes will grow sitefixes file fast and when more and more people will use Synchronize Sitesfixes option it will generate bigger bandwidth on github.

PR #15548, opened 2026-06-04
Invert logo at momentosesabores.com

arturstat:
Invert the top center logo at:
https://www.momentosesabores.com/


Myshor:
Please avoid using wildcard selectors (like *=) when a concrete selector or attribute is available. Broad selectors are more resource-heavy for the engine. Let's stick to specific attributes (like alt) for better performance and cleaner code.

PR #15559, opened 2026-06-08
Google docs comment counter pill (dark mode)

gabriel-peracio:
Fixed indicators for comment count on document tabs pills.

The current dynamic version lacks contrast, making it impossible to see the number.

## Unmodified (Light mode, default):
image

## Default (Dynamic Theme):
image

## My proposed fix:
image


Myshor:
Nice with dark scheme... can you share Dark Reader in light scheme with your fix? Because I do not have any google documents with comments to check by myself.

PR #15571, opened 2026-06-10
Fixed readyStateCompleteListeners never flushing when initialized after DOM ready

### LINKED ISSUE #15570: SVG fill attributes are not overridden when enable() runs from a deferred script
I'm using the darkreader npm package in an Electron app. I call `DarkReader.enable()` from a script that runs while the document is still loading (readyState is "interactive"). Everything gets themed except inline SVG icons that use `fill` attributes, they stay dark on dark.

I traced it to `src/inject/utils/dom.ts`. The fill handling in `overrideInlineStyle` is deferred with `addReadyStateCompleteListener` because it needs layout to decide between the text and background color heuristics. The problem is that the `readystatechange` listener that flushes `readyStateCompleteListeners` is only registered when `!isDOMReady()`, so it only exists if the module was initialized before readyState reached "interactive". That always holds for the extension since content scripts run at document_start, but a page script using the API typically runs at "interactive", so the listener is never registered and the queued callbacks just sit there forever.

Easy repro: page with an SVG using `fill="#333"`, call `DarkReader.enable()` from a deferred script. The fill never gets a `data-darkreader-inline-fill` override. Enable after window load and it works.

The fix is a one liner, register the listener when `!isReadyStateComplete()` instead. No behavior change for the extension path. Opening a PR with the change.

KamilDev:
Fixes #15570

The readystatechange listener in dom.ts was only registered when the module initialized before readyState "interactive". When the API is enabled from a deferred page script, initialization happens at "interactive", so `readyStateCompleteListeners` (deferred SVG fill overrides, image loads) never ran. Register the listener while readyState is not yet "complete" instead. The handler already deals with both flush stages, and nothing changes for the extension, which initializes at document_start.

### DIFF
--- a/src/inject/utils/dom.ts
+++ b/src/inject/utils/dom.ts
@@ -242,7 +242,10 @@ export function cleanReadyStateCompleteListeners(): void {
     readyStateCompleteListeners.clear();
 }
 
-if (!isDOMReady()) {
+// Listen until readyState is "complete", so that readyStateCompleteListeners
+// are flushed even when this module is initialized at readyState "interactive"
+// (e.g. when the API is enabled from a deferred script)
+if (!isReadyStateComplete()) {
     const onReadyStateChange = () => {
         if (isDOMReady()) {
             readyStateListeners.forEach((listener) => listener());


alexanderby:
Unfortunately this will have negative effect on websites that load slowly (or when the internet connection is slow), as it postpones certain mechanisms until the state is complete (all style sheets and images finished loading). Particularly, this check is used to remove the fallback style, which is inserted to prevent white flashes.

Is your SVG dynamically generated? Dark Reader is supposed to watch for DOM and inline style changes, but maybe there's a bug. Also this SVG analysis is not perfect, maybe switching it off completely will work better for you. E.g. by supplying a "fix" argument like `{ignoreImageAnalysis: ['svg *'], invert: ['svg']}`. We also still have experimental theme property `{immediateModify: true}` that disables the ready state chack.

KamilDev:
Nothing gets postponed, because the change removes no case from the guard, it only adds one. The line decides whether the listener gets registered, nothing else:

```js
if (!isReadyStateComplete()) {          // was !isDOMReady()
    const onReadyStateChange = () => { /* unchanged */ };
    document.addEventListener('readystatechange', onReadyStateChange);
}
```

and it's negated, so the narrower predicate registers in more cases, not fewer:

```ts
export let isDOMReady: () => boolean = () => {
    return document.readyState === 'complete' || document.readyState === 'interactive';
};

export function isReadyStateComplete(): boolean {
    return document.readyState === 'complete';
}
```

`!isDOMReady()` is true only at `loading`. `!isReadyStateComplete()` is true at `loading` and at `interactive`. Every case that registered before still registers, plus one that never did. So no mechanism can fire later than it used to, the fallback style included.

That added case is init at `interactive`, which is what a deferred script gets. There the listener was never registered, so `readyStateCompleteListeners` never flushed at all. The SVG fill overrides weren't running late, they weren't running.

The SVG is static markup. `overrideInlineStyle` queues the `fill` handling itself because the heuristic needs `getBoundingClientRect()`, so the watchers can't cover it.

PR #15574, opened 2026-06-11
Add dynamic theme fixes for magazine-hd.com

arturstat:
Invert top log at:
https://www.magazine-hd.com/

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -20739,6 +20739,13 @@ body
 
 ================================
 
+magazine-hd.com
+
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -20742,7 +20742,7 @@ body
 magazine-hd.com
 
 INVERT
-img.custom-logo
+div#header-logo-image img.custom-logo
```
arturstat, line 20745: Update to invert only the desktop version logo, as the mobile version does not need any inversion.

PR #15581, opened 2026-06-12
Add dark theme detection hints for additional sapo.pt websites

arturstat:
See:
https://24noticias.sapo.pt/
https://blogs.sapo.pt/

### DIFF
--- a/src/config/detector-hints.config
+++ b/src/config/detector-hints.config
@@ -77,11 +77,12 @@ MATCH
 ================================
 
 24noticias.sapo.pt
-blogs.sapo.pt
-sapo.pt
-tek.sapo.pt
 
-SYSTEM THEME
+TARGET
+body
+
+MATCH
+.black-in-black
 
 ================================
 
@@ -297,6 +298,16 @@ MATCH
 
 ================================
 
+blogs.sapo.pt
+
+TARGET
+html
+
+MATCH
+[data-theme="dark"]
+
+================================
+
 calendar.cron.com
 
 TARGET
@@ -448,6 +459,18 @@ MATCH
 dictionary.com
 
 NO DARK THEME
+
+================================
+
+digitalinside.com.br
+digitalinside.es
+digitalinside.sapo.pt
+
+TARGET
+body
+
+MATCH
+.darkmode
 
 ================================
 
@@ -1270,6 +1293,12 @@ link[href*="dark.css"]
 
 MATCH
 *
+
+================================
+
+sapo.pt
+
+SYSTEM THEME
 
 ================================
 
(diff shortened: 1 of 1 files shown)


src/config/detector-hints.config
```diff
@@ -76,6 +76,15 @@ MATCH
 
 ================================
 
+24noticias.sapo.pt
```
Myshor, line 79: Not System theme:

Image

When dark:
Image

So this will need totally different detection hint if you want to add it here.

src/config/detector-hints.config
```diff
@@ -76,6 +76,15 @@ MATCH
...
+blogs.sapo.pt
```
Myshor: Still better to check html [data-theme="dark"]

As I do not have account there I can not confirm if it can be maually set or not in some preferences but when we can check specific part of code it's still more safe than SYSTEM THEME

src/config/detector-hints.config
```diff
@@ -76,6 +76,15 @@ MATCH
...
+sapo.pt
+tek.sapo.pt
```
Myshor: Still better to check html [data-theme="dark"]
it has manual switch like 24noticias.sapo.pt

Image

Myshor (changes requested):
Only sapo.pt which was already added earlier I did not have any comments. Rest needs changes if you want to have them detected.

src/config/detector-hints.config
arturstat: Sorry for missing that one.

src/config/detector-hints.config
arturstat: For tek.sapo.pt I am not able to reliably use a detection hint.
html [data-theme="dark"] is missing when I select auto.
When in light mode it starts with .black-in-black, which disappears immediately.
So, this one cannot be used reliably either.
I am going to leave this one out, because dark reader is accurately detecting dark mode for me.

Myshor: Yes and with auto it's not fully dark. For me it opens with white header.
obraz

arturstat: For me dark reader still detects dark mode correctly in both dark and auto modes, but there is a delay of some milliseconds.

Myshor: Code in repo should work in both Edge (chromium) and Firefox. If I see not dark in auto, and additionally I see Manual switch I won't merge SYSTEM THEME. I do not see any further sense to talk about it. You make changes in code which we accept and merge or we are not merging.

arturstat:
@Myshor 
Modified according to what is working for me.
I decided to leave tek.sapo.pt out for the reasons given above.
Reverted sapo.pt back.
Also added the following websites:
https://digitalinside.com.br/
https://digitalinside.es/
https://digitalinside.sapo.pt/
Please check my last commit.
Thank you.

Myshor (approved):

PR #15587, opened 2026-06-12
Fix invisible images at portaldasfinancas.gov.pt

arturstat:
Fix invisible images at:
https://www.portaldasfinancas.gov.pt/at/html/index.html
https://sitfiscal.portaldasfinancas.gov.pt/geral/home?areaDestino=AD
https://sitfiscal.portaldasfinancas.gov.pt/geral/home?segmento=CD


arturstat:
Added an additional icon.
I think you won't be able to see it without login.

PR #15600, opened 2026-06-15
Invert third party logos at shopify.com

arturstat:
Invert logos from third parties at the shopify authentication page.
In order to reproduce open:
https://en.giesswein.com/
and proceed with the login process.
Screenshot 2026-06-15 173502

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -30777,7 +30777,7 @@ INVERT
 .header-country-select__trigger
 .lia-message-count::before
 .DateTime::before
-img[class^="_Logo_"]
+img[src*="logo-black"]
 
 ================================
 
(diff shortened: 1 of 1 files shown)


arturstat:
@Myshor 
The problem is that not all logos contain logo-black in the src.
Open https://www.hom.com/ and try to login.
However, all the logos that I have seen from all the stores are black and require inversion.

arturstat:
See also:
https://hanro.com/
https://www.skiny.com/

Myshor:
OK Now as I see more examples I will put your previous suggestion into sitefixes.

arturstat:
The background here is always white, which implies the logos must all be dark.
Probably safe to invert them all.
I use class^= because the last part looks like some random number, which might change in the future.

PR #15603, opened 2026-06-15
Add react site to global Dark List

manvithreddy07:
https://react.dev/

This Site already had Dark mode Implemented , So it will be good to have this site added into the Global Dark List.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -964,6 +964,7 @@ raycast.com
 razeenf.ca
 razer.com
 razorsecure.com
+react.dev
 redeclipse.net
 reelgood.com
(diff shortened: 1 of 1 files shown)


arturstat:
@manvithreddy07 
This website has a toggle for light and dark modes in the top right corner.
It cannot be added to the global dark list.
I created pull request https://github.com/darkreader/darkreader/pull/15605 to add a detection hint for this website.
When the hint is implemented, dark reader will detect dark mode almost instantaneously for this website.

manvithreddy07:
Thanks for the clarification. The detection hint approach makes sense since react.dev supports both light and dark modes. **I'll leave it to PR #15605.**

PR #15606, opened 2026-06-16
Add React Bits Site to Global Dark List

manvithreddy07:
**Website:**   https://reactbits.dev/
This Site already provides a native dark theme. Applying Dark Reader on top of the website is unnecessary and may lead to visual inconsistencies.

### DIFF
--- a/src/config/dark-sites.config
+++ b/src/config/dark-sites.config
@@ -964,6 +964,7 @@ raycast.com
 razeenf.ca
 razer.com
 razorsecure.com
+reactbits.dev
 redeclipse.net
 reelgood.com
 reflektclothing.co.uk


arturstat:
@manvithreddy07 
This website is similar to react.dev but without a theme button.
This is a candidate for the global dark list.
However, I see it also has the html.dark class, so I have added it to pull request https://github.com/darkreader/darkreader/pull/15605
```
react.dev
reactbits.dev

TARGET
html

MATCH
.dark
```
The detection hint is working well for both websites.
Just in case they decide to put a light/dark button there as well.

Myshor:
Did not found any system preference theme change. No theme button found. Merging proposition.

PR #15609, opened 2026-06-16
Add dark theme detection hints for geizhals

arturstat:
Dark reader detects a dark theme for these websites even when light mode is selected.
https://geizhals.at/
https://geizhals.de/
https://geizhals.eu/


arturstat:
For more information:
https://en.wikipedia.org/wiki/Geizhals

PR #15613, opened 2026-06-16
fix hover/focus select text on mozilla.org

Hamham6:
Before (hover):
Screenshot 2026-06-16 at 15 02 04

After (hover):
Screenshot 2026-06-16 at 15 02 47

Before (focus):
Screenshot 2026-06-16 at 15 02 27

After (focus):
Screenshot 2026-06-16 at 15 02 54

I checked light mode and it looks fine (no changes, since text was already black).

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -23258,7 +23258,7 @@ div.mzp-c-navigation-logo {
 }
 select:hover,
 select:focus {
-    color: black;
+    background-image: none !important;
 }
 
 IGNORE IMAGE ANALYSIS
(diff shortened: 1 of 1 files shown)


Myshor:
Can you share URL to test? I can not find easily any select dropdown on site.

Hamham6:
From [this page](https://www.mozilla.org/en-US/foundation/trademarks/list/), at the bottom of the page is a newsletter sign-up. If Javascript is on, then selecting the email address box makes two select boxs for "country" and "language" appear. If Javascript is off those select boxs will be visible from the start.
I'm sure that's not the only page this shows up on, but it happened to be the page I found.

Myshor:
I think the best solution which will make this fields acting almost the same as on the page without Dark Reader enabled is to remove totally this white background on hover and focus.

We will still have blue border around them.
obraz

I will push my fix with removed background-image and merge. 😉

PR #15625, opened 2026-06-18
Fix Vlerick Canvas logo in Dynamic mode

Xarudon:
Fixes the Vlerick Canvas logo on learning.vlerick.com in Dynamic mode.

The Canvas header logo uses `background-image: var(--ic-brand-header-image)`.
In Dynamic mode, Dark Reader currently causes the computed `background-image`
for `.ic-app-header__logomark` to become `none`, so the logo disappears.

This fix restores the Canvas branding image variable for the logo element.

Tested:
- Dynamic mode
- Light mode
- Dark mode
- Firefox

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -19830,6 +19830,15 @@ CSS
 
 ================================
 
+learning.vlerick.com
+
+CSS
+.ic-app-header__logomark {
+    background-image: var(--ic-brand-header-image) !important;
+}
+
+================================
+
 learningstylequiz.com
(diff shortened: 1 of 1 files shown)


Myshor:
Can you share some screens.

The site requires login and without credentials I can not confirm how it works.
So it would be nice to have screen before fix, after fix in dark, after fix in light.

maximedc04:
Without darkreader enabled.
image

With darkreader enabled (before fix).
image

With darkreader enabled (after fix in dark).
image

With darkreader enabled (after fix in light).
image

The fix is for the logo in the top left.

PR #15648, opened 2026-06-22
Update dynamic theme fixes for *.iefp.pt

arturstat:
Got to the following websites and search for a city on the map, for example "Braga".
You will see a new map, which this fix inverts.
https://www.iefp.pt/
https://www.iefp.pt/redecentros


arturstat:
Added classes under IGNORE IMAGE ANALYSIS in order to fix top image at:
https://crcvirtual.iefp.pt/

PR #15654, opened 2026-06-23
Update dynamic theme fixes for *.sapo.pt

arturstat:
Ignore top right SAPO logo at:
https://ajuda.sapo.pt/

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -854,6 +854,9 @@ html[data-darkreader-scheme="dark"] div#bsu-footer.bsu-footer-light a.bsu-footer
 html[data-theme="dark"][data-darkreader-scheme="dark"] div#bsu-footer div.bsu-footer-text {
     filter: none !important;
 }
+.barrita .extras .info {
+    background-color: #a9cbff !important;
+}
 
 IGNORE IMAGE ANALYSIS
 .wrapper.ink-grid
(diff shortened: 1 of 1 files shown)


arturstat:
@Myshor
Now I can barely see the text in the identified.
It was alright before. I don't know why you did that.

Screenshot 2026-06-23 185507

Myshor:
I was checking in Firefox only...

In Firefox without fixes it looks like this:
obraz

I will try to make some neutral fix to make it work fine in both browsers.

arturstat:
After removing from CSS:
```
.barrita .extras .info {
    background-color: #a9cbff !important;
}
```
I see this, which is fine:

Screenshot 2026-06-23 190951

arturstat:
OK then.
We have to force the background color and text color such that it works well on both browsers.
I will investigate.

Myshor:
I will move selector into INGORE IMAGE ANALYSIS

this will return 
obraz

in both browsers.

Myshor:
Done in https://github.com/darkreader/darkreader/pull/15659

arturstat:
By moving just this selector:
```
IGNORE IMAGE ANALYSIS
.barrita .extras
```
I see this which is the same.
Can you try in Firefox?

Screenshot 2026-06-23 192008

arturstat:
@Myshor 
Please try removing .info and see if you get the same result as in my previous message.
Thank you.

Myshor:
> @Myshor Please try removing .info and see if you get the same result as in my previous message. Thank you.

Without .info I am back to
obraz

It's just the new inverting for whole part of code instead on blob images generation for background images.

BTW... Dark Reader 4.9.128 is available in Chrome Webstore. Try to update your extension in browser. Then you will start to have the same view as I have in Firefox. 😉

PR #15655, opened 2026-06-23
Update dynamic theme fixes for *.seg-social.pt

arturstat:
Add support for additional pictograms:
https://www.seg-social.pt/
You may need to login in order to reproduce.


Myshor:
Added one more invert for some black lines image.
Screen after applying invert:
obraz

PR #15661, opened 2026-06-23
Add dynamic theme fixes for cimri.com

### LINKED ISSUE #15660: [Broken Website] cimri.com darkened images
### Prerequisites

- [x] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this website issue to avoid opening a duplicate.
- [x] I can reproduce this website issue in a new, unmodified web browser profile with Dark Reader installed as the only extension.

### Website Issue Description

images are darkened because of global rules.

### Website Address

cimri.com

### Steps To Reproduce

1. open cimri.com
2. make you sure to you are using dynamic mode for the domain

### Screenshots

Image

### Operating System

Windows 11

Brave 1.91.175

### Dark Reader version

4.9.125

### Additional Context

_No response_

karayelxyz:
fixes #15660 

image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -392,6 +392,15 @@ CSS
 
 ================================
 
+*.cimri.com
+
+CSS
+img {
+    mix-blend-mode: unset !important;
+}
+
+================================
+
 *.discogs.com
 
 INVERT
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -392,15 +392,6 @@ CSS
 
 ================================
 
-*.cimri.com
-
-CSS
-img {
-    mix-blend-mode: unset !important;
-}
-
-================================
-
 *.discogs.com
 
 INVERT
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -392,6 +392,15 @@ CSS
 
 ================================
 
+*.cimri.com
```
Myshor: Why to use *.cimri.com when we can use cimri.com instead?

Myshor:
Will add some more fixes to logos which are almost invisible and merge. ;)

src/config/dynamic-theme-fixes.config
karayelxyz: Hi Patryk, I just followed the file syntax :)

Myshor: *.page.url we use if site can have some prefixes... for example sales.cimri.com, market.cimri.com

if page uses suffix like cimri.com/market then it's better to add site without asterisk. Easier to catch and find it than going through all pages with asterisk in the beginning. 😉

However I made the change and sorting after that already.

PR #15677, opened 2026-06-27
Hide close button in dark mode at smithsonian.com

arturstat:
Close button near the top right should not be visible all the time.
https://www.smithsonian.com/

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -31621,6 +31621,7 @@ smithsonian.com
 
 INVERT
 div.logos-wrapper img
+div.nav-links:not([style="display: block;"]) button.nav-close::before
 
 ================================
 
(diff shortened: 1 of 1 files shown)


Myshor:
Do you know when it should appear? I still see it on my monitor after invert.

My proposal is to use:
```
smithsonian.com

INVERT
div.logos-wrapper img

CSS
.nav-close {
    color: var(--darkreader-neutral-background, #e8e6e3);
}
```

arturstat:
When you open the menu in the mobile version you should be able to see the close button.
Your proposal is always hiding the close button.
We need to look for an alternative solution.

arturstat:
When you click the hamburguer icon you should be able to see the close button in either the desktop or the mobile version.
But it should be invisible otherwise.

Myshor:
Not sure what hamburger does in desktop. :P But I can make fix for that close button and merge. 😉

arturstat:
Looks fine!
Thank you.

PR #15678, opened 2026-06-27
Fix for privacypack.org

pnaf:
Fixes graphical elements on `/create` page by using site-specific colors.

Before:

After:


Myshor:
Do we really want to have dark mode with big selection white areas after clicking "Create your pack" button?
obraz

pnaf:
Good point! I focused too much on the brighter backgrounds on the darkened page that I forgot about the white backgrounds on the regular page.

Updated.

PR #15688, opened 2026-06-29
Salesforce setup > Sandbox buttons fix

zmerdev:
Salesforce setup page > sandboxes, has un-inverted buttons with background images, unfortunately directly inverting them messes up the text color, which is very onerous to select correctly. so i just manually fixed them by removing the background image and recoloring.

Salesforce setup has custom tenents in the domain like ***companyname**.my.salesforce-setup.com* hence the *

Before
{2FE1F9CE-148C-484C-B46D-D8A34E7F0E46}

After (the chosen fix)
{90E8E09E-A02D-443C-8293-F06A91550EF8}

Direct Invert (i couldnt figure out how to fix the text when inverting the image)
{8420CECF-D58F-4B0F-93D1-C0AF1C1F3EFA}

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -853,6 +853,31 @@ main.about header#msi-key-visual.kv div.kv__content h1 {
 
 ================================
 
+*.my.salesforce-setup.com
+
+CSS
+body .rich-tabhdr-cell-active > table[id$="shifted"] td.rich-tabhdr-side-cell,
+body .rich-tabhdr-cell-inactive > table[id$="shifted"] td.rich-tabhdr-side-cell,
+body .rich-tabhdr-cell-active > table[id$="shifted"] td.rich-tabhdr-side-cell td.rich-tab-active,
+body .rich-tabhdr-cell-inactive > table[id$="shifted"] td.rich-tabhdr-side-cell td.rich-tab-inactive {
+    background-image: none !important;
+}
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -24152,6 +24152,31 @@ img[alt="mobile icon"]
 
 ================================
 
+*.my.salesforce-setup.com
```
Myshor: Very Important Question:
Is my.salesforce-setup.com uses prefixes that needs asterisk in the beginning?

I bet yes... in that case this entry should be between:
`*.msi.com`
and
`*.myconnectwise.net`

Correct it as right now it's not passing checks because of that alphabetical sorting which is wrong now.

Myshor (changes requested):

src/config/dynamic-theme-fixes.config
zmerdev: I didnt realize there was an asterisk section of the cofig, i will fix the PR.  and yes it requires the *.  the url for salesforce looks like example.my.salesforce-setup.com

zmerdev: Fixed

https://github.com/darkreader/darkreader/pull/15688/changes#diff-16ef5bfd1ce1c367ac55330b0229f74bbb1a9b85580afb3cf5711f09ea1dc99dR856

Myshor (approved):

PR #15711, opened 2026-07-06
fix for dreamwidth.org

pinions:
fixes for universal overrides affecting inline styles in dynamic theme, and for logo and white background in filter/filter+.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -437,24 +437,23 @@ html[data-darkreader-scheme="dimmed"] .tropo #logo {
 .ContextualPopup {
     background: var(--darkreader-neutral-background) !important;
 }
-:is(.entry-content,.comment-content) :is([style*="color: rgba(0,0,0,"]:not([style*="color: rgba(0,0,0,0)"]),
-[style*="color: rgba(0, 0, 0,"]:not([style*="color: rgba(0, 0, 0, 0)"]),
-[style*="color:#000"],
-[style*="color:black"],
-[style*="color:#282828"],
-[style*="color:#444"],
-[style*="color: #222222"],
-[style*="background:#fff"],
-[style*="background-color:#fff"],
-[style*="background: #fff"],
-[bgcolor="#F5F5F5"],
-[style*="background-color: #F5F5F5"],
-[style*="background: #e6e6e6"],
-[style*="background: rgb(255, 255, 255) !important"] *:not(a),
-[style*="background-color: rgb(254, 254, 254) !important"] *:not(a),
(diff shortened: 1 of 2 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -437,24 +437,31 @@ html[data-darkreader-scheme="dimmed"] .tropo #logo {
...
-[style*="background:#f9f9f9"]) {
-    background-color: inherit !important;
-    color: inherit !important;
+[class*='ui-widget'],
+.gradient,
+.taginfo {
+    background-image: none !important;
+}
+#tagform input[type="image"] {
+    filter: invert(90%);
```
Myshor: Why using this filter instead of INVERT section? INVERT section inverts images in dark scheme only. This fix will invert images in dark and light which can cause some ugly results in light scheming.

Myshor (changes requested):
Moving to request changes as I need toblock the merge for this for a while. 😋

It seems you force dark colors for text and light in background in your fixes.
I need some examlpe URL to check how it will look like specially in ligh and dark scheme of Dark Reader.

Same request for removing background image and setting invert filter. Some example URLs would be nice to know what you ware fixing.

src/config/dynamic-theme-fixes.config
pinions: i'll fix the INVERT section shortly. for some reason when i tried it before the css editor didn't register the changes, but it seems to be working now.

as for the background-image, here are some examples; these all have background gradients applied instead of colors. these pages and popups are only accessible if you're logged in.

{F8574763-602F-4B85-9E44-D06CA5D6108D}
{7664E8AA-472C-4B2E-A688-BC6AA704F7CB}
{E0C0649E-F7BF-4E3D-BDB4-CAF1CD38C588}

as for the fixed colors, the first has the current site fixes applied, while the others have them removed.

{D51EB12E-D241-43CC-A29F-9521F560B67A}
{FBDBA810-955B-424A-AB13-0E0DEE5359EA}
{22082B51-D999-4E4D-B5AC-CF91F41D677A}
{C272EFBE-3F38-4250-8D3C-AAE44768EC41}
{73F8A60A-92F4-40A7-BB76-6053EBC06141}

tl;dr: this site allows users to use custom inline css, which creates a ton of odd edge cases where something isn't readable, so it's just easiest to skip styling it completely. the universal overrides caused a lot of issues with this when they were first applied and previous attempts to fix with less intrusive methods just caused other problems. so while it does create white backgrounds, consistent readability >>> aesthetic.

Myshor (approved):
Ok... Thanks for the example screens.

PR #15713, opened 2026-07-06
Fix invisible digits at *.iherb.com

arturstat:
Fix invisible digits over the image near the top.
https://www.iherb.com/


arturstat:
In order to reproduce change Country/Region to United States.

arturstat:
Ignore text under "Dermatological Beauty" near the top:
https://www.iherb.com/c/dermatological-beauty

PR #15720, opened 2026-07-11
Fix invisible logo at researchgate.net

arturstat:


Myshor:
As I understand from the selectors name. This logo you invert and add white background is shown after login.

Can you share the screenshots of that logo before fix, after fix and after fix in light scheme?

arturstat:
Here it goes in the requested order.

Screenshot 2026-07-13 152502

Screenshot 2026-07-13 152554

Screenshot 2026-07-13 152625

PR #15742, opened 2026-07-16
Picrew color picker fix

deoxys3:
The colors in the color picker are often shown incorrectly, and this seems to fix it. Tested in picrew.me/en/image_maker/644129 by comparing the skin color options in both the dark and light versions of the page.

I did not test if this issue is present in the desktop website, but it is in the mobile version.


Myshor:
I will select the same but different selector to avoid some generated part of class name.

PR #15749, opened 2026-07-17
Add 'olafbenz.com' to dark sites configuration

arturstat:


src/config/dark-sites.config
```diff
@@ -857,6 +857,7 @@ obtainium.imranr.dev
 odysee.com
 offshorecorptalk.com
 okayeg.com
+olafbenz.com
```
Myshor: I opened this and ended just on start with:

Image

It's only for the first time but still it's white as... snow. 😋

Myshor (changes requested):
Check my comment.

I can understand adding benz.store to detection-hints because it opens with white elements and is detected as dark theme but olafbenz.com shows white cookies acceptation to so it shouldn't be added to dark-sites IMHO. 😉

src/config/dark-sites.config
arturstat: I assure you the website is all dark except for that.
I never see the cookie notice because adguard blocks it.
Dark Reader is handling this website very well except for one issue.
I am going to propose a few modifications shortly.

arturstat:
@Myshor please check.
Thank you.

src/config/dynamic-theme-fixes.config
```diff
@@ -26068,6 +26068,15 @@ img[class="logo"]
 
 ================================
 
+olafbenz.com
+
+CSS
+header .main-menu .brand-logo a {
```
Myshor, line 26074: I was gonna propose IGNORE IMAGE ANALYSIS but checked it's not working as it should so good workaround.

Myshor (approved):
Looks ok now. 😉

PR #15759, opened 2026-07-20
Update dynamic theme fixes for ipma.pt websites

arturstat:
https://dataclima.ipma.pt/
https://escolas.ipma.pt/
https://www.ipma.pt/

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -9997,6 +9997,15 @@ pre {
 
 ================================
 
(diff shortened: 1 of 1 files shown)


Myshor:
I was gonna ask for some example urls for the fixes but played a little bit with everyone. Compared how ipma.pt looks like in some opened articles with and without your fixes and in the end I will accept and merge so you won't need to go through the pages and search for every selector again. ;)

arturstat:
Thank you!

Issue #14603, opened 2025-08-25
[Broken Website] bloomberg.com


atiq-cs:
### Prerequisites

- [x] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this bug to avoid opening a duplicate.
- [x] I can reproduce this bug in a new, unmodified web browser profile with Dark Reader installed as the only extension.
- [x] I understand I need to use the [Broken Website Report template](https://github.com/darkreader/darkreader/issues/new?assignees=&labels=Broken+Website&template=broken-website-report.yml&title=%5BBroken+Website%5D+Replace+with+title) if this bug I am reporting occurs on a single website.

### Bug Description
bloomberg.com site and article links broken.

Repros on brave browser (probably repros on all chromium based browsers)

There's a terrible bug on "invert listed only" option.

I have to refresh pages twice after the first load for this extension to effect on each page. For example, I added the URL: https://www.bloomberg.com on "invert listed only"

Now, say I am reading this page:
https://www.bloomberg.com/news/articles/2025-08-22/meta-to-license-ai-lab-midjourney-s-tech-in-new-partnership

I have to refresh it twice for this extension to take effect or to enable dark mode on that page!! Very annoying user experience. Keeps asking for donations while major functionalities are broken, annoying.

Sometimes, it will enable dark mode on first load of the page. However, page foreground and background colors are badly broken. So, I have to refresh twice in that cases as well.

### Website Address

https://www.bloomberg.com/ai

### Steps To Reproduce

1. Enable extension on brave browser
2. Add the URL: https://www.bloomberg.com on "invert listed only"
3. Now, navigate to an article on that site i.e., https://www.bloomberg.com/news/articles/2025-08-22/meta-to-license-ai-lab-midjourney-s-tech-in-new-partnership
4. first loads are always broken, refresh to load it properly

### Expected Behavior

enable dark mode properly

### Actual Behavior

- sometimes messed up dark mode on first load on bloomberg.com   

<img width="2712" height="1932" alt="Image" src="https://github.com/user-attachments/assets/84e7f576-459e-4fd1-967e-7933835b3cb5" />

- sometimes no dark mode on first load   

<img width="2200" height="1885" alt="Image" src="https://github.com/user-attachments/assets/b57fa1a7-7f27-4024-af6e-9b70af290a88" />

### Screenshots

this extension is badly broken on bloomberg.com

### Operating System

pop_os 24.04 LTS

### Web Browser name and version

[...]

PR #15760, opened 2026-07-21
Fix delayed style injection before body exists

yuming-h:
This was causing issues with bloomberg.com articles loading dark text on dark background: https://github.com/darkreader/darkreader/issues/14603


atiq-cs:
Hi,
Lmk if you need help testing this.

PR #15769, opened 2026-07-23
handle email embedded content with inline important color text override (global fix)

fmigneault:
## Description

Certain emails embed HTML tables with fonts explicitly set with `color: rgb(0, 0, 0) !important;`.
Because of the inline `!important`, the text does not get inverted even though `--darkreader-inline-color` correctly gets applied, since the inline color takes CSS priority. 

image

To ensure the text color inversion applies as intended, this fix supplies the `-webkit-text-fill-color` which is less traditionally employed by embedded contents. It reuses the resolved `--darkreader-inline-color` to ensure consistent result. The scope is limited to the Gmail content container `.a3s` where the rule was detected for inline color inversion.

## Before Fix

image

## After Fix

image

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -21878,6 +21878,9 @@ CSS
         background-image: url(//ssl.gstatic.com/ui/v1/icons/mail/rfr/density_compact_v1_2x.png) !important;
     }
 }
+.a3s [data-darkreader-inline-color] {
+    -webkit-text-fill-color: var(--darkreader-inline-color) !important;
+}
 .buk {
     background-image: url(//ssl.gstatic.com/ui/v1/icons/mail/rfr/density_default_v1_1x.png) !important;
 }
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -228,6 +228,10 @@ body.mediawiki.skin-vector #mw-page-base {
 .overflow-hidden .solid-bg {
     background: transparent !important;
 }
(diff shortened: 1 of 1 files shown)


Myshor:
Can you forward that email to me (email can be found in my profile).

I would check how this email is presented in thunderbird. Maybe your fix will be needed globally to include it in thunderbird too. 😉

src/config/dynamic-theme-fixes.config
```diff
@@ -21878,6 +21878,9 @@ CSS
         background-image: url(//ssl.gstatic.com/ui/v1/icons/mail/rfr/density_compact_v1_2x.png) !important;
     }
 }
+.a3s [data-darkreader-inline-color] {
```
Myshor: We try to not include specific email issues in site fix like mail.google.com

As Dark Reader can be used with Thunderbird I would prefer to check your problematic email in thunderbird.
Fixes for thunderbird goes into `*` section of sitefixes so they work in Thunderbird and all web e-mails like gmail, Outlook Web and others. 😉

Myshor (changes requested):
Request changes to block merge until problematic mail will be sent to me and checked globally. If it will be only problem in gmail I will merge if not I will prepare global fix. 😉

fmigneault:
@Myshor 
I've sent you an example email.
Happy to see if you find a more robust solution.

Myshor:
Thanks... checked in thunderbird and in browser. So it's broken no matter if in gmail or other place.

I will push fix into your branch and merge. Should work for all places. :)

Myshor (approved):

PR #15771, opened 2026-07-24
Fixes for v2.bricklink.com

StevenSchifter:
Invert: BrickLink logo alt text case changed
CSS: Removed element type specifier from banner style to make it work again
Ignore inline style: Removed obsolete entries

Example page: https://v2.bricklink.com/en-us/catalog/color-guide

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -36925,7 +36925,7 @@ INVERT
 v2.bricklink.com
 
 INVERT
-img[alt="BrickLink logo"]
+img[alt="BrickLink Logo" i]
 
 CSS
 [data-testid="feature-banner"][style] {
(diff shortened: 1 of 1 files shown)


src/config/dynamic-theme-fixes.config
```diff
@@ -36925,16 +36925,14 @@ INVERT
 v2.bricklink.com
 
 INVERT
-img[alt="BrickLink Logo"]
+img[alt="BrickLink logo"]
```
Myshor: First check and it's not fix but break up on my side:

Image

Myshor: It seems we need both inverts as logo on main page is still with big "L"

Myshor (commented):
I will push some fix for that invert selector and merge. 😉

PR #15783, opened 2026-07-27
Add background color to brand logo at expert-technomarkt.de

arturstat:
Define background color for the logos in the navigation menu.
https://www.expert-technomarkt.de/


Myshor:
As I was making invert for logos on bottom of the page I will change your background adding into invert too.

Issue #15054, opened 2026-02-24
[Broken Website] vanguard.com incorrect on first load, but is fixed by toggling DR off/on.


meeotch:
### Prerequisites

- [x] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this website issue to avoid opening a duplicate.
- [x] I can reproduce this website issue in a new, unmodified web browser profile with Dark Reader installed as the only extension.

### Website Issue Description

Per request, from the other issue I submitted about nytimes.com (https://github.com/darkreader/darkreader/issues/15034) - I'm submitting this as a separate issue for vanguard.com.  

If I load the web page below in firefox or chrome (brave), private window so other extensions aren't running, what's displayed is not fully "dark". But if I click on the extension's icon to get the pop-up, and toggle it to OFF and then back ON again, the result is what I expect - nicely darkified.

https://investor.vanguard.com

### Website Address

https://investor.vanguard.com

### Steps To Reproduce

1. load the website
2. see that the top banner is light, the rest of the page is dark. 
3. toggling DR off/on fixes it. 

### Screenshots

<img width="1883" height="707" alt="Image" src="https://github.com/user-attachments/assets/571bab74-4d9a-4119-849c-ec6c45053087" />

### Operating System

Windows 11 25H2

### Web Browser name and version

firefox 147.0.4, brave 1.87.190

### Dark Reader version

4.9.121

### Additional Context

If you're logged into your account, there are other pages that have different, interior elements that are un-darkified. But that's probably not helpful to you for debugging, unless you have an account at Vanguard.

fyi - My Windows 11 system settings are set to Dark, and my Firefox settings set to "Automatic". The problem above persists if I set Firefox to "Light", and leave DR enabled.  (But that seems to break other pages.  So I'm assuming that the correct settings are Firefox=Automatic or Dark, DR=enabled, DR Detect Dark Theme also = enabled?)

From the discussion in the other github issue:  "nice example when something is loading dynamically and CSS style from Dark Reader is not applied to that part because of that".

Myshor:
Just adding comment it's bug created after some discussion in #15034

PR #15785, opened 2026-07-28
fix for vanguard.com

evanwherchek:
Fix for #15054 

Note: this is my first pull request in this project(and any open source project). I've tested this fix and it seems to work correctly. However, if anything is missing, please let me know. Thanks!


Myshor:
If it would be that easy to fix it would be added in February.

After making refresh of the page that header is back to white background.

PR #15791, opened 2026-07-28
Update dynamic theme fixes for *.naturitas.*

arturstat:
Invert account icon on the top right.
https://www.naturitas.pt/
https://www.naturitas.es/


arturstat:
Apparently, the icon in question is only available after login.
Screenshot 2026-07-29 010204

Myshor:
I was able to move it in the code to some visible part and check without login just using developer tools in firefox and confirmed in needed invert. Merging.

PR #15808, opened 2026-08-02
Fix 2 bugs in MV3 state persistence causing auto-sync with system dark mode to fail in chromium

### LINKED ISSUE #15554: [Bug] Auto sync according to system mode doesn't work - Brave
### Prerequisites

- [x] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this bug to avoid opening a duplicate.
- [x] I can reproduce this bug in a new, unmodified web browser profile with Dark Reader installed as the only extension.
- [x] I understand I need to use the [Broken Website Report template](https://github.com/darkreader/darkreader/issues/new?assignees=&labels=Broken+Website&template=broken-website-report.yml&title=%5BBroken+Website%5D+Replace+with+title) if this bug I am reporting occurs on a single website.

### Bug Description

The system dark mode detection is not working on Brave. The system is in auto (light) and the extension always picks dark mode.

`window.matchMedia("(prefers-color-scheme: dark)").matches` evaluates to false and `window.matchMedia("(prefers-color-scheme: light)").matches` evaluates to true.

- Brave 1.91.168
- macOS 26.5 (25F71)

### Website Address

https://github.com/

### Steps To Reproduce

1. Run the the extension on Brave
2. Enable automation and set it to "Use system colour scheme"
3. Observe as the extension wrongfully assumes the system is in dark mode

### Expected Behavior

The extension correctly picks the mode (light or dark) according to the active mode on the system.

### Actual Behavior

The extension always picks dark mode.

### Screenshots

_No response_

### Operating System

macOS 26.5

Brave 1.91.168

### Dark Reader version

4.9.125

### Additional Context

`https://codepen.io/ayc0/pen/MWpOmbQ` correctly detects the system version.

babymastodon:
Fixes #15554 

This fixes a regression for chromium in version 4.9.125. I had Codex narrow it down these two specific bugs. I still not entirely sure why these two long-standing bugs suddently resulted in a regression but the reasoning seems plausible.

Bug 1: Commit 8714273c (#9619) set `Extension.wasLastColorSchemeDark = isDark ` before calling `MV3syncSystemColorStateManager`  so `else if (Extension.wasLastColorSchemeDark !== isDark) ` was always guarenteed to be false, `Extension.systemColorStateManager.saveState();` was never called, so the system color state was never saved to local storage.

Bug 2:  StateManager's `collectState` incorrectly used || instead of ??, meaning that the boolean value `false` fails to be persisted and gets converted to the default value (darkMode = true). This would cause us to get stuck in dark mode, even if we fixed Bug 1.

Now I believe b2a5b4a6d59fa6d7f57b1f0d6ee0f36319894e04 (in version 4.9.125) made `MV3syncSystemColorStateManager` synchonous which triggered Bug 1. Previously, this function ran async so we bypassed both of these bugs by dumb luck. But this is just speculation at this point.

### DIFF
--- a/src/background/extension.ts
+++ b/src/background/extension.ts
@@ -129,10 +129,9 @@ export class Extension {
         if (isDark === null) {
             // Attempt to restore data from storage
             return Extension.systemColorStateManager.loadState();
-        } else if (Extension.wasLastColorSchemeDark !== isDark) {
-            Extension.wasLastColorSchemeDark = isDark;
-            return Extension.systemColorStateManager.saveState();
         }
+        Extension.wasLastColorSchemeDark = isDark;
+        return Extension.systemColorStateManager.saveState();
     }
 
     private static alarmListener = (alarm: chrome.alarms.Alarm): void => {
--- a/src/utils/state-manager-impl.ts
+++ b/src/utils/state-manager-impl.ts
@@ -132,7 +132,7 @@ export class StateManagerImpl<T extends Record<string, unknown>> {
     private collectState() {
         const state = {} as T;
         for (const key of Object.keys(this.defaults) as Array<keyof T>) {
-            state[key] = this.parent[key] || this.defaults[key];
+            state[key] = this.parent[key] ?? this.defaults[key];
         }
         return state;
     }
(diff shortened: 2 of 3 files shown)


alexanderby:
Thanks for your investigation. let me see why it keeps working in all browsers for me...

PR #15815, opened 2026-08-04
Added TFQueue (queue.tf) to dark-sites.config.

zuchr:
(Our implementation of "darkreader-lock" seems to be breaking the site for Firefox users.)


Myshor:
If you are the author (or one of the authors) then fix this white popup for updates.
How I can accept adding site to dark-sites when I see something like this for a welcome. 😋
obraz

zuchr:
Fixed! Sorry about that, I thought we pushed that change to the live site.

PR #15851, opened 2026-08-11
Update dynamic theme fixes for shop.militaryrange.eu

arturstat:
Added support for all language versions of the website.
Removed brand logos from INVERT and set white background instead.
Some flags don't look well when inverted, so it is preferable to set white background, see for example: https://shop.militaryrange.eu/brand/french-army


Myshor:
Adding IGNORE IMAGE ANALYSIS to EU icon broken in top right corner of the page:
obraz

PR #15903, opened 2026-08-20
Add dynamic theme fixes for oledtest.org

arturstat:
The purpose is to ignore several backgrounds and define backgrounds and colors where needed.
The fixes are required for the tests to work as intended.
For example, to test white uniformity on a monitor we require an actual white background.
The fixes target the tests listed in the following pages:
https://oledtest.org/
https://oledtest.org/tools/

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -26715,8 +26715,11 @@ CSS
 oledtest.org
 
 CSS
+:root {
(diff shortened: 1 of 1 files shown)


Myshor:
Added some change fixing background for these buttons:
obraz

Myshor:
BTW
Watch out when copying styles directly from DevTools - stripped that redundant , 1 fallback inside var(--tw-bg-opacity, 1). Let's keep the syntax clean next time. 😉

PR #15924, opened 2026-08-25
Fix inline Desmos calculator

TesseractPi:
Control bar was improperly inverted (white on white)

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -3309,6 +3309,10 @@ CSS
 .lrn .lrn.lrn_widget pre {
     mix-blend-mode: normal !important;
 }
+.dcg-calculator-api-container-v1_11 .dcg-container .dcg-expression-top-bar {
+    background: linear-gradient(#282a2b, #181a1b);
+}
+
 
 ================================
 


src/config/dynamic-theme-fixes.config
```diff
@@ -3309,6 +3309,10 @@ CSS
 .lrn .lrn.lrn_widget pre {
     mix-blend-mode: normal !important;
 }
+.dcg-calculator-api-container-v1_11 .dcg-container .dcg-expression-top-bar {
+    background: linear-gradient(#282a2b, #181a1b);
+}
+
```
Myshor, line 3315: Please remove this trailing empty line. It breaks the configuration parser and causes CI tests to fail.

src/config/dynamic-theme-fixes.config
```diff
@@ -3309,6 +3309,10 @@ CSS
 .lrn .lrn.lrn_widget pre {
     mix-blend-mode: normal !important;
 }
+.dcg-calculator-api-container-v1_11 .dcg-container .dcg-expression-top-bar {
+    background: linear-gradient(#282a2b, #181a1b);
```
Myshor, line 3313: If a custom background override is strictly necessary, note that missing `!important` might cause specificity issues if styles are loaded dynamically by the host app.

Also, if the issue is strictly caused by improper inversion of inline styles, consider whether adding this selector under `IGNORE INLINE STYLE` would be a cleaner approach.

Myshor (changes requested):
Thanks for the PR.

Since `apclassroom.collegeboard.org` requires an account behind an auth-wall, maintainers cannot verify this fix directly. 

Please provide screenshots demonstrating the result:
- Before fix (Dark Reader ON)
- After fix (Dark Reader ON)
- After fix but switched to light scheme in Dark Reader (to make sure there are no contrast regressions)

Please also address the inline comments regarding the CI linter failure and the CSS rule.

PR #15930, opened 2026-08-26
Detect github.com dark theme when color mode is auto

### LINKED ISSUE #15929: [Feature Request] system theme aware detector hints, for github "sync with system"
### Prerequisites

- [X] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this feature request to avoid opening a duplicate.

### Is this feature request related to a problem?

Yes

### Feature Request Description

i have github set to "sync with system" and my os is dark, so github renders its own dark theme. dark reader still applies its theme on top, with "detect dark theme" on.

the hint for github.com is:

```
TARGET
html

MATCH
[data-color-mode="dark"]
```

github only sets `data-color-mode="dark"` when you pick "single theme". with "sync with system" it stays `auto`, and the real theme comes from `data-light-theme` / `data-dark-theme` depending on `prefers-color-scheme`. so the selector never matches and dark reader keeps theming an already dark page.

i know this has history:

- #11995 added `[data-color-mode="auto"]`
- #15155 someone on a light os then got no dark mode at all, because `auto` matched while github was actually light
- #15161 removed `auto` again, correctly
- #15187 proposed `MATCH SYSTEM DARK` / `MATCH SYSTEM LIGHT`, closed

so a plain `auto` match is definitely wrong. but the case is still broken, and i dont think it can be fixed in the config alone, because `Element.matches` cant see `prefers-color-scheme`. the condition that is actually true is:

```
[data-color-mode="auto"] and (prefers-color-scheme: dark) and [data-dark-theme^="dark"]
or
[data-color-mode="auto"] and (prefers-color-scheme: light) and [data-light-theme^="dark"]
```

both branches are needed, because "sync with system" lets you pick any of the 12 themes per branch, so you can have a dark theme selected for the light branch. @Myshor confirmed that in #15187.

the existing `SYSTEM THEME` command doesnt cover it either, since it short circuits before `MATCH` and assumes the site always follows the system, which github doesnt when you use single theme.

what id like to do is add `MATCH SYSTEM DARK` and `MATCH SYSTEM LIGHT` to detector-hints, which add their selectors only when `prefers-color-scheme` matches, and use them for github. same idea as #15187, but only that, nothing else. about 25 lines across `detector.ts`, the generator, definitions, the config and the config test.

i have it written and the unit tests pass. happy to open the pr, or drop it if you'd rather not add new hint commands.

### Screenshots

_No response_

### Additional Context

dark reader 4.9.129, macos dark mode, github appearance set to "sync with system".

pvinis:
Fix #15929

github only sets `data-color-mode="dark"` for "single theme". with "sync with system" it stays `auto` and the active theme comes from `data-light-theme` / `data-dark-theme` depending on `prefers-color-scheme`, so the current hint never matches and dark reader themes an already dark page.

`Element.matches` cant see `prefers-color-scheme`, so this adds two commands to detector-hints:

- `MATCH SYSTEM DARK` — selectors that only apply when `prefers-color-scheme: dark`
- `MATCH SYSTEM LIGHT` — selectors that only apply when `prefers-color-scheme: light`

they get appended to `MATCH` at detection time, nothing else changes. both are needed for github, because "sync with system" lets you pick any of the 12 themes per branch, so a dark theme can be selected for the light branch.

a plain `[data-color-mode="auto"]` match was tried in #11995 and removed in #15161 because it broke #15155 (light os, github actually light, dark reader turned itself off). this doesnt have that problem:

| github setting | os | matches | result |
| --- | --- | --- | --- |
| single theme, dark | any | yes | dark reader off, github dark |
| single theme, light | any | no | dark reader themes it |
| sync with system, dark theme picked | dark | yes | dark reader off, github dark |
| sync with system, light theme picked | dark | no | dark reader themes it |
| sync with system, light theme picked | light | no | dark reader themes it |
| sync with system, dark theme picked | light | yes | dark reader off, github dark |

`^="dark"` covers `dark_dimmed`, `dark_colorblind` etc, and excludes every `light*` value.

same idea as #15187, but only this, nothing else.

one thing i left out on purpose: flipping the os theme with the page open doesnt re-run detection. thats already true of `SYSTEM THEME` today, and with automation set to system dark reader re-applies anyway.

`npm run test:unit` (75 passed) and `npm run lint` pass.

### DIFF
--- a/src/config/detector-hints.config
+++ b/src/config/detector-hints.config
@@ -823,6 +823,12 @@ html
 MATCH
 [data-color-mode="dark"]
 
+MATCH SYSTEM DARK
+[data-color-mode="auto"][data-dark-theme^="dark"]
+
+MATCH SYSTEM LIGHT
+[data-color-mode="auto"][data-light-theme^="dark"]
+
 ================================
 
 gotquestions.org
--- a/src/definitions.d.ts
+++ b/src/definitions.d.ts
@@ -213,6 +213,8 @@ export interface DetectorHint {
     url: string[];
     target: string;
     match: string[];
+    matchSystemDark: string[];
+    matchSystemLight: string[];
     noDarkTheme: boolean;
     systemTheme: boolean;
     iframe: boolean;
--- a/src/generators/detector-hints.ts
+++ b/src/generators/detector-hints.ts
@@ -9,6 +9,8 @@ import type {SiteFixesIndex, SitesFixesParserOptions} from './utils/parse';
 const detectorHintsCommands: { [key: string]: keyof DetectorHint } = {
     'TARGET': 'target',
     'MATCH': 'match',
+    'MATCH SYSTEM DARK': 'matchSystemDark',
+    'MATCH SYSTEM LIGHT': 'matchSystemLight',
     'NO DARK THEME': 'noDarkTheme',
     'SYSTEM THEME': 'systemTheme',
     'IFRAME': 'iframe',
--- a/src/inject/detector.ts
+++ b/src/inject/detector.ts
@@ -202,10 +202,18 @@ export function stopDarkThemeDetector(): void {
 let hintTargetObserver: MutationObserver;
 let hintMatchObserver: MutationObserver;
 
+function getHintSelectors(hint: DetectorHint): string[] {
+    const systemSpecific = isSystemDarkModeEnabled() ? hint.matchSystemDark : hint.matchSystemLight;
+    return (hint.match || []).concat(systemSpecific || []);
+}
+
 function detectUsingHint(hint: DetectorHint, success: () => void) {
     stopDetectingUsingHint();
 
-    const matchSelector = (hint.match || []).join(', ');
+    const matchSelector = getHintSelectors(hint).join(', ');
+    if (!matchSelector) {
+        return;
+    }
 
     function checkMatch(target: Element) {
         if (target.matches?.(matchSelector)) {
--- a/tests/unit/config/config.tests.ts
+++ b/tests/unit/config/config.tests.ts
@@ -107,7 +107,7 @@ test('Detector Hints config', async () => {
 
     // selectors should have no comma
     const commaSelector = /\,(?![^\(|\"]*(\)|\"))/;
-    expect(hints.every(({target, match}) => ![target].concat(match).some((s) => commaSelector.test(s)))).toBe(true);
+    expect(hints.every(({target, match, matchSystemDark, matchSystemLight}) => ![target].concat(match, matchSystemDark, matchSystemLight).some((s) => commaSelector.test(s)))).toBe(true);
 
     // only a single selector is allowed for target
     expect(hints.every(({target, noDarkTheme, systemTheme}) => noDarkTheme || systemTheme || typeof target === 'string' && !target.includes('\n'))).toBe(true);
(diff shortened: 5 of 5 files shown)


Myshor:
Thanks for opening the PR and providing a concrete implementation.

The diff is clean and concise. Extending the generator and resolving selectors dynamically via `isSystemDarkModeEnabled()` in `detector.ts` nicely covers the GitHub edge case.

Since this introduces new detector hint commands (`MATCH SYSTEM DARK` / `MATCH SYSTEM LIGHT`) and touches the core detector injection logic, I've approved the CI workflows to run the automated checks and unit tests.

I will leave the final architectural approval on extending the detector hint syntax to @alexanderby.

src/config/detector-hints.config
```diff
@@ -823,6 +823,12 @@ html
 MATCH
 [data-color-mode="dark"]
 
+MATCH SYSTEM DARK
```
Myshor, line 826: I am not sure if the changes to `detector-hints.config` should be merged in this exact PR.

Here is the backward compatibility concern:
Many users (including myself) have the **"Synchronize site fixes"** option enabled. If this config is merged into `main`, clients running the current production version of Dark Reader will fetch the updated config before the new extension build containing the updated `detector.ts` parser is published and approved on Chrome Web Store and Mozilla Add-ons.

When testing by pasting these new commands into DevTools on the current release build, the unsupported tags are stripped out. In production, fetching unsupported hint syntax on older extension versions could lead to unexpected parsing issues or silent failures until the store updates roll out.

It would be safer and more practical to:
1. Merge and release the core engine changes (`detector.ts`, generators, definitions) in the extension bundle first.
2. Submit the actual `MATCH SYSTEM DARK / LIGHT` rules for `detector-hints.config` in a follow-up PR once the new extension version is officially live in Web Stores.

Myshor (commented):
Overall the core implementation in `src/inject/detector.ts` and the generator looks clean.

I left an inline note regarding `detector-hints.config` and the potential backward compatibility / release cadence issue with the "Synchronize site fixes" feature.

Workflows approved for CI checks. Leaving final architecture decision to @alexanderby.

src/config/detector-hints.config
pvinis: my only reason for not really using the "sync site fixes" is that basically dark reader is plug and play. i install, i tap once one it to turn on the "on/off/auto" and put it to auto, and im good! except im not really, because of the fixes *or* the fix on this pr.

that "sync site fixes" is to deep in the settings, and off by default, that i def prefer a better(?) fix thats more general than sticking to the best-so-far solution 🤷. but yea, obviously, as the two of you prefer, we do it that way :) i just enjoy dark reader and want to make it better.

Myshor: Once @alexanderby approves the core changes, I can push the detector-hints.config update separately and monitor how it behaves with the sync enabled on the current production build. If it causes any issues, we can just do a quick revert. But first, let's wait for Alex's review. 😉

alexanderby:
I looked into the issue with Github. The problem is I assumed that color scheme value "light dark" acts like the system color scheme is preferred, but it's not. Along with another bug fix, the dark theme detection for Github works fine without a hint. A few more tests on other websites, and the new behavior will be published soon.

pvinis:
@alexanderby awesome. feel free to do whatever you want with the issue and the pr here. if you need anything from me, let me know. otherwise, i look forward to having this on a version soon :) thanks.

PR #15956, opened 2026-09-05
Update dynamic theme fixes for hugoboss.com

arturstat:
Invert performance icons image at:
https://www.hugoboss.com/pt/en/all-brands/men/highlights/inspiration/boss-performance/ https://www.hugoboss.com/fr/toutes-les-marques/homme/stories/inspirations/boss-performance/ https://www.hugoboss.com/es/todas-las-marcas/hombre/especiales/inspiracion/boss-performance/

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -17917,6 +17917,7 @@ div[style^="background-color: rgb(255, 255, 255) !important;"]:not(:has(a[style^
 div[style^="background-color: white !important;"]
 div[style^="color: rgb(0, 0, 0) !important;"]:not(div.dch-fullbleed-grid-teaser-item__image ~ div div[style^="color: rgb(0, 0, 0) !important;"])
 span[style^="color: rgb(0, 0, 0) !important;"]
+section[data-as-promotion^='{"id":"boss-perfomance-icons-still--promo--2091958R"'] div[style^="background-color:  !important;"] img
 
 CSS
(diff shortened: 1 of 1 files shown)


arturstat:
@Myshor 
The problem is that the alt is not working for all languages.
Screenshot 2026-09-07 132540

arturstat:
We can make the selector simpler while still working:
`section[data-as-promotion^='{"id":"boss-perfomance-icons-still--promo'] img`

arturstat:
I just proposed PR https://github.com/darkreader/darkreader/pull/15969 with this suggestion.
Please check.
Thank you.

Myshor:
OK... I am not sure how I was checking your 3 links but It's not working with english...

> We can make the selector simpler while still working: `section[data-as-promotion^='{"id":"boss-perfomance-icons-still--promo'] img`

It seems to be working... I was thinking about using beginning of src url for this image which would both work for desktop and mobile.
`img[src^="https://images.hugoboss.com/is/image/hugobosscsprod/260126_HBME_112_SU26_BOSS_PM"]`

Why it's better idea IMHO? Because if they will use the same picture in totally different place in their shop it will be still inverted. With your proposal it's strictly related to boss-performance promotion.

So I think I will still update your PR. (Or you can update it by yourself as I will have some lunch break now 😉)

arturstat:
Or they can use the same section[data-as-promotion^='{"id":"boss-perfomance-icons-still--promo'] elsewhere.
No one can guess.

PR #15959, opened 2026-09-05
pixiv.net: remove stale dynamic fix, add NO DARK THEME detector hint

k6G52m4Dz75W:
## Summary
Two related changes for pixiv.net:

1. **Remove** the stale dynamic theme fix added in 2020 (PR #4192).
2. **Add** a `NO DARK THEME` detector hint so that Dark Reader keeps transforming on this site, covering a first-paint theme race that produces a visible white flash for dark-theme users on every page load.

## Why the old fix is removed
It targets styled-components generated classes (`.jAENWx`, `.iYRMgo`) that no longer exist on the site — hash classes rotate on every deployment, and `document.querySelectorAll('.jAENWx, .iYRMgo').length` currently returns `0`. Additionally, dynamic fixes can never apply here: Dark Reader's dark theme detector sees `<html data-theme="dark">` (set by an inline head script before first paint) and stands down entirely, so no dynamic fix stylesheet is injected on this site at all.

## Root cause (why the hint is needed)
The site has a first-paint theme race, captured with a `CSSStyleSheet.insertRule` recorder (times relative to navigation start):

- An inline head script correctly sets `<html data-theme="dark">` before first paint.
- Despite that, the first styled-components batch is rendered with the **light** theme:
  - `t=1195ms` — `.kQLKe { background: #ffffff; border-radius: 8px 8px 0 0; }`
- ~157 ms later, the dark variant of the same component shape appears and components re-render with new class names:
  - `t=1352ms` — `.iiAyvy { background: #1f1f1f; border-radius: 8px 8px 0 0; }`
- All subsequently lazy-loaded chunks render dark directly (`t=1848ms`, `2538ms`, `3515ms`), confirming the theme context is correct after the initial sync.

Because the detector interprets `data-theme="dark"` as "this site handles dark on its own", Dark Reader stands down and the site's internal race is left exposed. The `NO DARK THEME` hint keeps Dark Reader transforming from the first paint, which covers the race window. On an already-dark token palette the transform is near-identity, so the result visually matches the native dark theme.

## Verification
Applied the same hint through this extension's Dev Tools → "Detect dark theme hints" editor (identical content to the `src/config/detector-hints.config` diff here): after a reload, the white flash on page load is gone for dark-theme users. The light/dark rule pair and timestamps above were captured from a single page load with all other extensions disabled.

## Maintenance note
If the site fixes its theme-initialization race (ThemeProvider synchronously initialized from `data-theme`/localStorage before first render), this hint is no longer needed and can be reverted in a follow-up PR.

The maintainer of the site has been informed of the bug, but there is no guarantee that it will be fixed (soon).

### DIFF
--- a/src/config/detector-hints.config
+++ b/src/config/detector-hints.config
@@ -1363,6 +1363,13 @@ NO DARK THEME
 
 ================================
 
+pixiv.net
+*.pixiv.net
+
+NO DARK THEME
+
+================================
+
 pkmer.cn
 
 TARGET
--- a/src/config/detector-hints.config
+++ b/src/config/detector-hints.config
@@ -1364,9 +1364,12 @@ NO DARK THEME
 ================================
 
 pixiv.net
-*.pixiv.net
 
-NO DARK THEME
+TARGET
+html
+
+MATCH
+[data-theme="dark"]
 
 ================================
 
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -28529,16 +28529,6 @@ IGNORE IMAGE ANALYSIS
 
 ================================
 
-pixiv.net
-
-CSS
-.jAENWx,
-.iYRMgo {
-    color: ${aliceblue} !important;
-}
-
-================================
-
 pixivision.net
 
 INVERT


Myshor:
Forcing `NO DARK THEME` on a website that natively provides a full dark theme (`data-theme="dark"`) is an anti-pattern. Dark Reader should not continuously transform native dark stylesheets and DOM trees just to work around a ~150ms client-side theme initialization race condition.

Updated the changes as follows:
- Kept the removal of the stale 2020 fix (`.jAENWx, .iYRmgo`), as those legacy styled-components hashes are long gone.
- Dropped the redundant `*.pixiv.net` wildcard entry. Checked subdomains like `sketch.pixiv.net`, `factory.pixiv.net`, and `sensei.pixiv.net` - none of them have a native dark mode or use `data-theme="dark"`. With the base `pixiv.net` detector hint, they simply fail the match condition and Dark Reader transforms them as intended.
- Replaced the detector hint with an explicit target and match for `html[data-theme="dark"]`. This ensures Dark Reader transforms pages in light/default state while cleanly standing down when the native dark mode is active.

k6G52m4Dz75W:
Hi @Myshor, thanks again for the review on this PR. I've spent some more time comparing the merged version against the pre-PR behavior, and I'd like to flag two things — one factual note about the merged hint, and one proposal that I think actually resolves the anti-pattern concern you raised.

**1. The merged `TARGET html` + `MATCH [data-theme="dark"]` hint is behaviorally a no-op on pixiv.net — and if anything it widens the uncovered flash window**

The generic detection path (no hints) already returns "dark" for pixiv: `runCheck()` in `src/inject/detector.ts` checks `document.documentElement.dataset.theme === 'dark'` before ever reaching the color-sampling code, and the site sets that attribute before first paint. The merged hint matches the same attribute on the same element (`detectUsingHint()`), so the verdict is identical. Since `detectDarkTheme` defaults to `true` (`src/defaults.ts`), this is the out-of-the-box path.

The only real difference is timing. Without a hint, `runDarkThemeDetector()` waits for `canCheckForStyle()` (body rendered, some style present) before checking; with a hint, the match fires as soon as the content script receives the theme — so the stand-down happens *earlier*, and Dark Reader covers *less* of pixiv's hydration race (the styled-components window I measured in the PR description: light values at ~1195 ms, dark variant at ~1352 ms). The hint path also drops the color-sampling fallback, which is the only branch that can keep Dark Reader covering when the attribute check is inconclusive.

Net effect: as merged, the entry neither changes the detection verdict nor fixes the flash — it just stands down a bit sooner. One repro note: on an MV3 cold start, service-worker wake-up can push the whole inject→detect cycle past the race window, which is probably why the flash didn't reproduce during review; on a warm start it does reproduce reliably. Given all this, would you consider dropping the pixiv.net entry? Removing it should, if anything, shorten the flash.

**2. Proposal: when the detector stands down, keep URL-matched dynamic fixes ("theme off, fixes on")**

Right now detection is strictly binary: `ADD_DYNAMIC_THEME` applies theme + fixes together, and on detection success the content script calls `removeDynamicTheme()`, which tears the fixes down with it, followed by `CLEAN_UP` on the background side. That means for a natively dark site with a known first-paint bug — exactly pixiv's case — the fix database has no tool that can help: any dynamic fix we write dies together with the theme at the moment of stand-down.

What if detection success kept URL-matched dynamic fixes instead of doing a full cleanup? For example, a new opt-in command in `detector-hints.config` (say `KEEP FIXES`, as a companion to `TARGET/MATCH`): on a dark detection the content script drops the generated theme but re-applies the site's dynamic fix entries. That directly addresses your anti-pattern concern — Dark Reader would no longer transform native dark stylesheets, it would only apply explicitly documented per-site patches — and it gives the fix database an actual role on native-dark sites. Without something like this, there is regrettably no point in writing a per-site fix for pixiv at all, since no such fix can survive the stand-down. Pixiv is unlikely to fix this on their side (we have no realistic channel to their maintainers), so client-side mitigation is the only realistic path for users.

If (2) is not acceptable upstream, the pragmatic fallback for pixiv would be restoring the `NO DARK THEME` hint from the first revision of this PR — semantically a lie, as you said, but the transform is near-identity on pixiv's palette and it's the only current mechanism that covers the window.

Happy to open a separate issue for the feature discussion if this PR isn't the right place, and I can attach the insertRule recording of the hydration race if that would help.

Myshor:
Thanks for the detailed follow-up and breakdown.

Since this PR is already merged and closed, continuing the discussion here gets buried and keeps it strictly as a two-person thread.

Please go ahead and open a dedicated issue for this. Proposing a new engine behavior or config directive (like keeping dynamic fixes active after detector stand-down) affects core Dark Reader architecture beyond just pixiv.net. An open issue will be much better tracked, and we can loop in @alexanderby there to weigh in on architectural direction regarding native dark theme hydration races.

Feel free to link this PR and attach your insertRule recording in the new issue.

PR #15962, opened 2026-09-06
Add dynamic theme fixes for epey.com

canklot:
Dark reader is trying to invert an image but its inverting the whole element. This cause the already dark background becoming light.

filter: invert(100%) hue-rotate(180deg) contrast(90%) !important;

Added some elements under IGNORE IMAGE ANALYSIS to prevent problematic inversion


canklot:
before fix
after fix
Here is the before and after the fix

PR #15969, opened 2026-09-07
Update dynamic theme fixes for hugoboss.com

arturstat:
Update selector in order to work across all language versions of the website:
https://www.hugoboss.com/pt/en/all-brands/men/highlights/inspiration/boss-performance/ https://www.hugoboss.com/fr/toutes-les-marques/homme/stories/inspirations/boss-performance/ https://www.hugoboss.com/es/todas-las-marcas/hombre/especiales/inspiracion/boss-performance/

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -17917,7 +17917,7 @@ div[style^="background-color: rgb(255, 255, 255) !important;"]:not(:has(a[style^
 div[style^="background-color: white !important;"]
 div[style^="color: rgb(0, 0, 0) !important;"]:not(div.dch-fullbleed-grid-teaser-item__image ~ div div[style^="color: rgb(0, 0, 0) !important;"])
 span[style^="color: rgb(0, 0, 0) !important;"]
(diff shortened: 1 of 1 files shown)


arturstat:
I propose shortening the selector to:
`section[data-as-promotion^='{"id":"boss-perfomance-icons'] img`
As I don't see any conflict elsewhere.

Myshor:
Overriding with the `src`-based selector.

Matching generic `img` tags inside a CMS promotion container (`section[data-as-promotion...] img`) is an anti-pattern. CMS promo modules frequently get reused or updated with product shots, banners, or model photography. If any photo asset ends up inside that section, a descendant `img` selector will invert it into a negative. Shortening the promotion ID only increases collision probability across the site.

Targeting the specific asset key from the CDN (`img[src^="https://images.hugoboss.com/is/image/hugobosscsprod/260126_HBME_112_SU26_BOSS_PM"]`) is strictly scoped to this exact iconography strip, works across all language locales, and carries zero risk of inverting unrelated imagery elsewhere on the domain.

Issue #14790, opened 2025-11-15
[Bug] Unresponsive page with 100% CPU


ocavue:
### Prerequisites

- [x] I [searched for any existing report](https://github.com/darkreader/darkreader/issues?q=is%3Aissue) about this bug to avoid opening a duplicate.
- [x] I can reproduce this bug in a new, unmodified web browser profile with Dark Reader installed as the only extension.
- [x] I understand I need to use the [Broken Website Report template](https://github.com/darkreader/darkreader/issues/new?assignees=&labels=Broken+Website&template=broken-website-report.yml&title=%5BBroken+Website%5D+Replace+with+title) if this bug I am reporting occurs on a single website.

### Bug Description

My webpage becomes unresponsive with 100% CPU usage when dark mode is enabled.

### Website Address

https://prosekit.dev/playground/-/react/full/

### Steps To Reproduce

1. Open a Chrome Incognito window  
2. Make sure Dark Reader is off  
3. Go to https://prosekit.dev/playground/-/react/full/ or https://1afb35f4.prosekit-dev.pages.dev/playground/-/react/full/  
4. The CPU usage is normal, and the webpage is interactive  
5. Turn on Dark Reader  
6. The CPU usage reaches 100%, and the webpage is frozen  

### Expected Behavior

The webpage is not frozen

### Actual Behavior

The webpage is frozen

### Screenshots

https://github.com/user-attachments/assets/658a1099-efa3-4a72-8f6f-41cf0c18b0d1

### Operating System

macOS 

### Web Browser name and version

Chrome 142.0.7444.135

### Dark Reader version

Dark Reader 4.9.113

alexanderby:
Nice example, thank you.

alexanderby:
I've lost access unfortunately: Internal server error.

ocavue:
> I've lost access unfortunately: Internal server error.


Cloudflare is down: https://www.cloudflarestatus.com/incidents/8gmgl950y3h7

alexanderby:
The problem is, this library rebuilds the code blocks' HTML entirely after Dark Reader inserts its inline style. We are working on stopping injecting inline styles. But in the meantime we will ignore the containers with `[data-node-view-content]` attribute.

The other question is, why dark theme is not being detected.

ocavue:
> this library rebuilds the code blocks' HTML entirely after Dark Reader inserts its inline style.


[...]

PR #15970, opened 2026-09-07
Add subtree opt-out for inline style mutations

nperez0111:
Hi, long-time darkreader user here.

I make a rich text editing library called [BlockNote](https://github.com/TypeCellOS/BlockNote) and our editor does not play well with having attributes being stamped on it's elements.

It looks like this has already been an issue before, https://github.com/ueberdosis/tiptap/issues/3818#issuecomment-2247047938 https://github.com/darkreader/darkreader/issues/14790 , since I see that there is a ProseMirror carve out (our base rich text editing framework). But, BlockNote uses elements that are deeper in the tree, so the carve-out for ProseMirror is not enough for our needs.

I thought the most reasonable way to go about this was to contribute something to darkreader, rather than having to add extension-specific detection to our editing library (like [what was proposed here](https://github.com/TypeCellOS/BlockNote/pull/3062)). Rich-text editors based on ProseMirror have this issue because ProseMirror heavily relies on the MutationObserver API to watch for changes to the rich text content, but darkreader stamping attributes everywhere both causes a lot of noise or in some cases can cause infinite loops like what is happening in BlockNote.

So, my suggestion here is to not just try to ignore anything that is ProseMirror/Tiptap/BlockNote based, and instead offer an attribute that marks this specific behavior as being disabled on this sub-tree so that we don't get elements being stamped causing infinite loops in our library.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,6 +34,21 @@ lock.name = 'darkreader-lock';
 document.head.appendChild(lock);
 ```
 
+### Preventing inline style changes within a component
+
+Add `data-darkreader-ignore-inline` to prevent Dynamic mode from rewriting inline
+styles and color attributes throughout a subtree, including future descendants:
+
+```html
+<div data-darkreader-ignore-inline>
+    <span style="color: red">Editor content</span>
+</div>
+```
+
+Set the attribute before Dark Reader processes the component; it does not undo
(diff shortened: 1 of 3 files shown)


Myshor:
@alexanderby Core logic and dynamic theme performance impact (element.closest on inline mutations). Passing to you for review.

PR #15981, opened 2026-09-09
Add brightness filter for grey placeholder images

ehanoj:
FIX for bbc.com: When loading an article, white boxes appear where images will be displayed (which are lazily loaded later). Buxfix resolves this issue.

### DIFF
--- a/src/config/dynamic-theme-fixes.config
+++ b/src/config/dynamic-theme-fixes.config
@@ -5633,6 +5633,7 @@ INVERT
 .orb-icon .orb-icon-arrow
 [class*="NavigationLink-LogoLink"] > span > svg
 span[class*="LogoIconWrapper"]
+img[src$="grey-placeholder.png"]
 
 CSS
 .wr-icon-weather-type__svg-background,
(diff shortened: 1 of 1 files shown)


ehanoj:
testing:
https://www.bbc.com/news/articles/c99dm1000p7o
before
after

Myshor:
Thanks for catching this issue with the lazy-loaded image placeholders and opening the PR!

I reviewed the changes and pushed an update directly to your branch to refine the approach. Here is why the adjustments were needed:

1. **Dark Reader supports both Dark and Light schemes:** Dark Reader is not only for dark mode; users can also customize or run light schemes. Hardcoding `filter: brightness(0) !important;` in the raw `CSS` block applies the rule unconditionally. In a light scheme, this forces a pitch-black box onto a light background instead of keeping BBC's native light-grey placeholder.
2. **Visual blending (`brightness(0)` vs `INVERT`):** Crushing the placeholder to `#000000` via `brightness(0)` creates an unnatural, harsh black cutout against BBC's dark background. Moving the selector to the `INVERT` section ties it directly to the theme engine: the light-grey placeholder naturally inverts to a matching dark-grey tone in dark mode, while leaving it completely untouched in light mode or when toggled off.
3. **Selector precision:** Using `img[src*="grey-placeholder"]` is broader than necessary. Since BBC consistently serves this asset as a PNG, anchoring to the file ending via `img[src$="grey-placeholder.png"]` is cleaner and prevents accidental substring matches elsewhere.

The commit is now on your branch so your contribution will be included in the release. Thanks again for helping improve the BBC experience!

